PR #13, opened 2015-07-12
Refactoring changes and fix for VS

Flamefire:


Flamefire:
This is now a working set with the most "complicated" renaming and stuff done. When using this, you got to fix submodules to (they point to my repos) once they are merged. Check .gitmodules file

Flow86:
merged manually. closing

PR #151, opened 2015-08-03
Fixes TerrainRenderer test and improves performance. Fixes #148

### LINKED ISSUE #148: Crash in terrainrenderer
There is a crash in the terrainrenderer when scrolling to point 0,0 (or very close) in one of Maqs savegames (can you attach that for reference?)
Nothing to do but scroll to 0,0 (even in pause modus)
I assume wrong offsets/counts in the texture

Flamefire:


MarcusSt:
Would be great to have more pull requests dealing with a single overall idea instead of mixing a lot of different commits into one. That way it is possible to fix bugs (like crashes) quickly without checking a lot of other commits.

Flamefire:
Most of that stuff is related to the Terrainrenderer. And you guys are committing directly into the master. So whats wrong here? Every change in its own commit and the PR to have it checked before it is put in the master. Using multiple PRs would mean multiple branches but I need all changes in one branch, which would incur quite some overhead.

MarcusSt:
1. Feature branches should be used for bigger, _related_ changes. You can then still merge them to a local branch containing all your changes/branches. They can be merged with one pull request.
2. Big pull requests with _unrelated_ commits cause problems:
  * Most of the bugs that got fixed in the recent days and took a lot of time were the results of big pull requests that contained partly unrelated commits. It gets extremely difficult to spot the one change that caused the bug. Take the recent problem with std::set and nofPassiveSoldiers.
  * There is no way to test those easily or at least get an overview over what changed. A pull request only containing changes to one building for example is easier to debug than one that changes about every file there is.
  * Combining structural and logical changes is even more problematic if blame or diff doesn't work any more, as it prevents looking for logical changes that could cause crashes.
  * Even blame and diff are no real help for debugging after thousands of changed lines.
3. Most of our commits have been fixes to new bugs introduced at a single point in time by merging a pull request. All of those have been tested and are somewhat time-critical. Then there's a little difference in commit size. Most of my recent commits were bug fixes of some kind or just really small changes. In my opinion it does not make sense to create a feature branch and make a pull request for only changing a few lines that have been carefully tested. This also has the advantage of having a nightly build containing those changes that gets tested by others as well.

**Yours** (this pull request):
> 37 additions and 40 deletions.
> 289 additions and 312 deletions. 
> 186 additions and 246 deletions. 
> 25 additions and 27 deletions
> 80 additions and 81 deletions. 
> 20 additions and 45 deletions. 
> 104 additions and 104 deletions.
> 6 additions and 6 deletions.
> 45 additions and 36 deletions.
> 15 additions and 15 deletions.
> 13 additions and 16 deletions.
> 163 additions and 186 deletions. 
> Total: 983 additions and 1114 deletions (one pull request).

**Mine:**
> 2 additions and 1 deletion.
> 4 additions and 0 deletions.
> 5 additions and 5 deletions.
> 6 additions and 5 deletions.
> 2 additions and 2 deletions.
> 31 additions and 11 deletions. 
> 11 additions and 1 deletion.
> 7 additions and 5 deletions.
> ...

Flamefire:
But what is the difference between having multiple branches merged into 1 branch and make a PR for that one to having multiple small commits? Or did I got you wrong?
Here are basically 2 branches: TerrainRenderer fixes (first 7 commits + 0a302a4) and random improvements (all others, where each one is a single fix for a single problem without much effect) The latter partly depend on the previous.
I do agree that smaller PRs are better, but this also adds quite some overhead especially if they depend on each other. I try to use small, self-containing commits where possible to improve on testing and stuff.
Most of the changes that you summed up, change a single file, well, REWRITE a single file essentially. If I'd squash them, it would be much smaller...

MarcusSt:
The idea is that if you work on a lot of different things at a time to have feature branches for each of those features/bugs you are working on so they can be merged separately. For testing all of them together, you can still merge them into a single branch.
For me to merge a pull request, I have to go through the changes, check for mistakes, etc. This becomes more difficult the bigger the total number of changes per pull request gets. Therefore, a separate pull request for TerrainRender would have been great. Bug fixes should be checked/merged with a higher priority than changes like 52c6155, which are very prone to slip-ups, as you change a lot of very similar stuff and easily forget a "!", but do not change anything but looks.

Flamefire:
Ok, I'll try to keep them simpler next time

PR #158, opened 2015-08-04
Use generic Push/PopObjectContainer

Flamefire:
Requires boost 1.54


Flow86:
will merge it as soon as I added a proper boost-version for all architectures. I won't forget it.

Flow86:
perhaps we could even switch to boost::serialization and/or to protobuf (so we could get rid of the ridiculous messages-stuff)

MarcusSt:
We should get rid of recursion in serialization... ;-) But that's another story.

Flamefire:
Boost is included in the contrib folder so this actually works. Although some overly greedy gitignores removed some boost files which made the build fail. Trying to build again. Just a sec.

MarcusSt:
@Flow86 Can it be merged like this? Should work, as boost is in contrib.

Flamefire:
And builds on the test server

Flow86:
nein weil du mir das cmake gelöscht hast - es hatte schon einen grund warum das drin ist.

Flamefire:
Hatte es für einen Fehler in der gitignore gehalten, sorry.
Warum ist das überhaupt drin? Finde nicht, dass es rein sollte. Libraries ja, dann muss man die sich nicht zusammensuchen, aber Build-Umgebungszeug? Warum dann nicht gleich ne ganze IDE oder so was includen? Ist eh selten, dass binaries im Git sind und gerade bei CMake würde ein Link in der Readme reichen, statt es direkt zu includen. So müsste man es auch noch aktuell halten...

Flamefire:
I'll investigate a bit into boost serialization. It would help quite a lot: Only 1 function for serialization AND deserialization instead of 1 each. Not sure how boost handles recursion... We may need to either increase the stack size or reduce our hierarchy depth.

Flow86:
removing recursion at all would be much better. perhaps we could simply serialize all items (from x=0..width and y=0..height) and then save the relations between them?

Flamefire:
And what about e.g. a military building with soldiers in it? It would also make the code a mess and a hell to maintain. SO question related: http://stackoverflow.com/questions/31827802/boost-serialization-with-recursion?noredirect=1#comment51578780_31827802

Flamefire:
@Flow86 Can you comment on my reasoning for removing cmake?

Flow86:
@Flamefire da wir es für die windows nutzer so einfach wie möglich halten sollten, aber hast schon recht, eigentlich gehört es da nicht rein

Flamefire:
@Flow86 Heißt halt, dass wir das auch aktuell halten müssten und für die meisten Entwickler ist es eh unnötig weil schon vorhanden.
BTW: u.U. müssen wir mal die anderen libraries aktualisieren oder zumindest prüfen...

Wenn das so ok ist bitte mergen.

Flow86:
I have built boost now for most architectures (32bit linux boost is the only one left, which I will built I think this evening)

Flamefire:
Alright, perfect. Then I'd add another commit to this PR which removes boost?
And could you update Cmake while you are on it. The version on the server seems to be buggy e.g. for finding boost.

Flow86:
thats the standard ubuntu cmake version, but I'll have a look then

Flow86:
merged manually

PR #172, opened 2015-08-08
Two small compile fixes for FreeBSD/clang

uqs:


Flow86:
shouldnt be the locale header inside mygettext.h?

uqs:
Depends on what sort of IWYU (include what you use) paradigm you subscribe to, I guess. This cpp file uses LC_ALL, and thus needs to have a declaration for it.

Either via header pollution from some other include, or by just declaring that it needs this header directly.

I think the latter makes more sense. I have used http://include-what-you-use.org/ at work before, not on RTTR though (yet?)

Flow86:
its okay

PR #245, opened 2015-09-02
Make build/cmake.sh a plain sh(1) script for better portability.

uqs:
While here, remove 8bit chars to make it UTF-8, remove some unneeded
subshells and allow all possible spellings of 'yes'.

### DIFF
--- a/build/cmake.sh
+++ b/build/cmake.sh
@@ -1,12 +1,12 @@
-#!/bin/bash
+#!/bin/sh
 
-# Editable Variables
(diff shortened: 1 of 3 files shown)


Flow86:
its not possible to merge it currently, can you fix the merge please?

uqs:
Merged with the svn keyword removal. Also added start.sh and build/start.sh

Though for me a simple make that build/start.sh does is not sufficient, as the video drivers are only linked during make install. Will investigate this some more ...

Please take another look.

PR #248, opened 2015-09-03
Clean up some warnings about unused fields and constants

uqs:
Also squelch the annoying virtual overload warnings


Flamefire:
The 2nd commit is fine with me. But please remove the first. Instead you can cherry pick this one: https://github.com/Flamefire/s25client/commit/35fe247f69fd7c3967fa42360dd8a5f1741757b4

Keep the warning about overloaded-virtual because it reminds us of fixing this. It can and should! Besides; Boost uses long longs so that warning supression is required.

uqs:
I feared as much, dropped the first commit.

Flamefire:
@Flow86 The cleanup is ok?

Issue #212, opened 2015-08-16
Compiling on OS X is impossible


ThanosSiopoudis:
Hi,
I've tried compiling s25client on Mac and it seems rather impossible to get a bundle.
I've tried on OS X 10.9 and 10.11, Xcode and tools installed, cmake installed (from Homebrew), as well as required libraries (SDL, SDL_MIXER, gettext).

Error:
```
MacBook-Pro:build Thanos$ ./cmake.sh --prefix=.
Setting Path-Prefix to "."
Setting Architecture to "apple.local"
Setting Binary Dir to "./bin"
Setting Data Dir to "./share/s25rttr"
Setting Library Dir to "./share/s25rttr"
Disabling build of ""
Activating release build
Generating files for "Xcode"
Running "cmake -G 'Xcode' -DCMAKE_INSTALL_PREFIX=  -DPREFIX=. -DCOMPILEFOR_PLATFORM=apple.local -DBINDIR=./bin -DDATADIR=./share/s25rttr -DLIBDIR=./share/s25rttr -DCMAKE_BUILD_TYPE=Release './..'"
-- Checking for apple.local
-- Set CMAKE_C_COMPILER to gcc
-- Set CMAKE_CXX_COMPILER to g++
CMake Error at cmake/apple.common.cmake:15 (MESSAGE):
  woooops: this should not happen! your system has no known architecture!
Call Stack (most recent call first):
  cmake/apple.local.cmake:12 (INCLUDE)
  cmake/crosscompile.cmake:13 (INCLUDE)
  CMakeLists.txt:33 (INCLUDE)


-- Configuring incomplete, errors occurred!
An error occured - please check above!
```

Flamefire:
It seems like the apple cmake file file is included before COMPILE_ARCHS is set but it assumes it is.

@Flow86 Can you check this?

Flamefire:
Should be fixed. Otherwise open a new issue please

PR #250, opened 2015-09-03
Rework CMake system to make it easier to handle

Flamefire:
Uses the preferred way to do cross compiling with CMake by using toolchains. The local compilers are used when none is specified (no *.local.cmake anymore) which now also allows to use CMake directly on rttr (without cmake.sh) cmake.sh now only takes care of cross-builds and symlink creation although symlinks might be better to put to CMake too.

Later CMAKE_INSTALL_PREFIX should be used to replace the current PREFIX usage.

This should fix build on OSX --> Helps #212 

@Flow86 : Please check if packaging still works. Build on all 5 targets was tested on the build server as well as on a linux system and MSVC2012 on windows.
@uqs : Your input might also be valuable if you see anything that can be done better.

Requires:

- [ ] Check packaging
- [ ] https://github.com/Return-To-The-Roots/libutil/pull/10
- [ ] https://github.com/Return-To-The-Roots/s-c/pull/8
- [ ] https://github.com/Return-To-The-Roots/s25update/pull/6
- [ ] https://github.com/Return-To-The-Roots/version/pull/3


cmake/Modules/CMakeDetectOsXArchs.cmake
```diff
@@ -0,0 +1,35 @@
+# Detects all supported OSX architectures by reading libSystem.B.dylib
+# Sets the possible values in DETECTED_OSX_ARCHS
+# If a NO{ARCH} variable is set, then {ARCH} will be ignored
+
+MACRO(DetectOsXArchs)
+    # find lipo
+	FIND_PROGRAM(LIPO NAMES apple-lipo lipo)
```
Flow86: thats not correct, it could be named *-apple-darwin*-lipo too

cmake/Modules/CMakeDetectOsXArchs.cmake
```diff
@@ -0,0 +1,35 @@
...
+	SET(CMAKE_LIPO "${LIPO}" CACHE PATH "" FORCE)
```
Flow86: you should overwrite this only if not set yet by the toolchain?

cmake/Modules/CMakeDetectOsXArchs.cmake
```diff
@@ -0,0 +1,35 @@
...
+
+	MESSAGE(STATUS "Checking ${CMAKE_PREFIX_PATH}/usr/lib/libSystem.B.dylib for possible architectures")
+
+	# read supported platforms	
+	EXECUTE_PROCESS(
+		COMMAND ${LIPO} "-info" "${CMAKE_PREFIX_PATH}/usr/lib/libSystem.B.dylib"
```
Flow86: here we should use CMAKE_LIPO then, since the other one is only a temporary var

build/postinstall.sh.cmake
```diff
@@ -201,9 +128,9 @@ extract_debug_symbols()
 mecho --blue "## Extracting debug info from files and saving them into dbg"
 
 # strip out debug symbols into external file
-case "$COMPILEFOR" in
-	apple)
-		echo "extraction of debug symbols for $COMPILEFOR currently not supported" >&2
+case "$SYSTEM_NAME" in
+	Apple)
+		echo "extraction of debug symbols for Apple currently not supported" >&2
 		i686-apple-darwin10-strip -S ${DESTDIR}bin/s25client
```
Flow86: perhaps we should use cmake to find the corresponding strip too and use it here?

build/preinstall.sh.cmake
```diff
@@ -29,7 +29,7 @@ mecho()
 
 ###############################################################################
 
-COMPILEFOR=@COMPILEFOR@
+SYSTEM_NAME=@CMAKE_SYSTEM_NAME@
```
Flow86: why not name the var identically? its not a problem adding CMAKE_ here to the name, you would at least know what was meant?

build_paths.h.cmake
```diff
@@ -24,32 +24,4 @@
 ///////////////////////////////////////////////////////////////////////////////
 #endif
 
-///////////////////////////////////////////////////////////////////////////////
-//	Other Configuration
-
-//	set global prefix (normal /usr/local)
-//#define PREFIX "/usr/local"
-#cmakedefine PREFIX "${PREFIX}"
```
Flow86: so we remove the pathing here at all now?

cmake/toolchains/c.windows.x86_64.cmake
```diff
@@ -1,13 +1,11 @@
-
 # this one is important
 SET(CMAKE_SYSTEM_NAME Windows)
 SET(CMAKE_SYSTEM_PROCESSOR x86_64)
 
 # specify the cross compiler
-SET_CCACHE(CMAKE_C_COMPILER x86_64-pc-mingw32-gcc)
-SET_CCACHE(CMAKE_CXX_COMPILER x86_64-pc-mingw32-g++)
-SET_CCACHE(CMAKE_RC_COMPILER x86_64-pc-mingw32-windres)
-
```
Flow86: why do you remove the ccache stuff?

build/cmake.sh
```diff
@@ -122,9 +122,9 @@ while test $# != 0 ; do
 			$ac_shift
 			LIBDIR=$ac_optarg
 		;;
-		-arch | --arch | -target | --target)
+		-arch | --arch | -target | --target | -toolchain | --toolchain)
 			$ac_shift
-			ARCH=$ac_optarg
+			ToolChain=$ac_optarg
```
Flow86: var names should be all upper case

Flamefire: Done

cmake/Modules/CMakeDetectOsXArchs.cmake
Flow86: okay but standard name is okay - if we use the lipo from toolchain first (see comment below)

build/postinstall.sh.cmake
Flamefire: Not sure. According to this: http://cmake.3232098.n2.nabble.com/Cross-compiling-a-static-library-CMake-2-8-3-td5907410.html CMake should automaticly detect ranlib and some other tools by the used compiliers. Best, if you'd check that on the build server first (carefull, ranlib is set manually in most cases) But for the sake of this PR leave it as is (no functional changes, using CMake objcopy is a new feature/bugfix independant of this, so let's keep it simple and get this one done first)

cmake/apple.common.cmake
```diff
@@ -1,79 +1,55 @@
```
Flow86: this whole file is now the same as apple.cmake ?

build/preinstall.sh.cmake
Flamefire: Because I wanted to avoid CMAKE_ prefixes in shell scripts. 1) Because they can cause problems when INSTALL is not used carefully, and 2) because it has nothing to do with CMake anymore and COULD come from other sources. "SYSTEM_NAME" makes it still clear, what it is and where it came from

build_paths.h.cmake
Flamefire: I actually wanted to include that in a seperate commit. Gonna revert this, although it can be removed later

build/preinstall.sh.cmake
Flow86: okay, only wanted to now if you thought about the naming ;-)

cmake/Modules/CMakeDetectOsXArchs.cmake
Flamefire: added

cmake/Modules/CMakeDetectOsXArchs.cmake
Flamefire: check added

cmake/Modules/CMakeDetectOsXArchs.cmake
Flamefire: Fixed

cmake/toolchains/c.windows.x86_64.cmake
Flamefire: Because it is added later. These are toolchains, they have nothing to do with caching. Check CMakeLists.txt

cmake/apple.common.cmake
Flamefire: apple.common.cmake should not be there. Removed.

cmake/toolchains/c.windows.x86_64.cmake
Flow86: well your thought is correct, but it wont work in this case here, since ccache for example does not work for apple directly (thats why that macro was used to use the symlinks from /usr/*/ccache/ or ccache-lipo (which is the ccache wrapper for apple to allow universal builds which will fail with normal ccache)

Flamefire: Ok so I added the CCache stuff in CmakeLists.txt now. The macros cannot be used at this point!

Flow86: okay perfect!

uqs:
This broke the build on FreeBSD again. And I can't find the obvious thing to fix. It breaks when trying to link sound-convert.

[ 30%] Linking CXX executable sound-convert
cd /data/src/s25client/build/s-c/src && /usr/local/bin/cmake -E cmake_link_script CMakeFiles/sound-convert.dir/link.txt --verbose=1
/usr/bin/CC    -O2 -g -std=c++11 -I/data/src/s25client/build   CMakeFiles/sound-convert.dir/getopt.cpp.o CMakeFiles/sound-convert.dir/getopt_long.cpp.o CMakeFiles/sound-convert.dir/main.cpp.o  -o sound-convert  ../../libsiedler2/src/libsiedler2.a ../../libutil/src/libs25util.a /usr/local/lib/libboost_iostreams.so /usr/local/lib/libboost_system.so /usr/local/lib/libboost_filesystem.so ../../libendian/src/libendian.a -lminiupnpc -Wl,-rpath,/usr/local/lib 
/usr/bin/ld: cannot find -lminiupnpc
CC: error: linker command failed with exit code 1 (use -v to see invocation)


This is bogus, why would sound-convert link against libminiupnpnc.so?

% grep -i miniupnpc CMakeCache.txt
MINIUPNPC_INCLUDE_DIR:PATH=/usr/local/include
MINIUPNPC_LIBRARY:FILEPATH=/usr/local/lib/libminiupnpc.so
s25util_LIB_DEPENDS:STATIC=general;/usr/local/lib/libboost_system.so;general;/usr/local/lib/libboost_filesystem.so;general;miniupnpc;
//Details about finding miniupnpc
FIND_PACKAGE_MESSAGE_DETAILS_miniupnpc:INTERNAL=[/usr/local/include][/usr/local/lib/libminiupnpc.so][v()]
//ADVANCED property for variable: MINIUPNPC_INCLUDE_DIR
MINIUPNPC_INCLUDE_DIR-ADVANCED:INTERNAL=1
//ADVANCED property for variable: MINIUPNPC_LIBRARY
MINIUPNPC_LIBRARY-ADVANCED:INTERNAL=1

PS: I really which this was all in one git repo, so I can use git log -Sminiupnpnc to see where and when things changed related to that. :)

uqs:
Nevermind, found the problem, will send a fix. Also found that this is just a transitive dependency and thus sound-convert needs to link against miniupnpc when linking against s25util

Issue #272, opened 2015-09-21
Wrong shading


Flamefire:
@Flow86 @MarcusSt The shading calculation at https://github.com/Return-To-The-Roots/s25client/blob/master/src/GameWorldBase.cpp#L1221 seems to be off. Shouldn't that be 1,2,3?

Going to fix that if there is no specific reason for that.

BTW: The algorithm is different then from e.g. http://settlers2.net/documentation/world-map-file-format-wldswd/

Spikeone:
![shading](https://cloud.githubusercontent.com/assets/1276342/10051261/f5caddc0-6220-11e5-8d37-66acfbcb470d.png)

Good luck I still had this picture

Flamefire:
Which map is this?

Spikeone:
Island of Hills (OMAP00.SWD) in that case

Flamefire:
Ok. Then I'll look into shadow calculation again. RTTR seems pretty flat compared to s2

Spikeone:
Yeah, quite strange - I just know that you may check the mapeditor by xaser, he nearly had a fully working shading with SDL which looks great.
When the first rttr versions appeared, everything was lighter than its now ^^

Flamefire:
Xaser uses the original method via gouroud shading:

>The editor calculates shading (light) values for each vertex. Note that the gouraud shading in settlers 2 is precalculated.
The game trusts this data and uses it to get the right color value from one of the GOUx.DAT-files (x = 5-Greenland, 6-Wasteland, 7-Winterland).
This files contain a color value for each of the light values the editor set. If you plan to change this block to see what happens, you have to
load this map in the game to see it. The editor is not interested in this block cause it calculates the values from scratch.

>If you want to know how the original settlers 2 editor and game calculates the ShadingInformation, take a look at:
	http://bazaar.launchpad.net/~xaser/%2Bjunk/s2mapeditor/annotate/head%3A/CMap.cpp
The function called 'modifyShading(.....)' is what you need.

>In the GOUx.DAT files there are color values ( = color indices = mapped rgb data 8bpp). If you want to shade a pixel, you have to read it
from the tileset (the 8bit pixel value) and you need the shading value from this block (more specifically the interpolated value at the point you
want to draw). The GOUx.DAT files are two-dimensional arrays, for every of the 256 possible shading values (0x00 - 0xFF) every shaded palette is saved in this array.
So do the following to get the new color value for a given color value and a given shading value:

[...]

PR #273, opened 2015-09-22
Fixes according to static analysis

Flamefire:
Fix all found issues by using PVS Studio.

Mostly shadowing issues which can lead to confusion of members vs parameters but also some safety stuff (boost::array, missing copy operator, possibly throwing dtor, bool to int conversion, ...)
Renaming was done by VisualAssist so should be save (if any error, it would not be able to compile, besides that I'd just trust the refactoring function)

For review:

- [ ] Shading issue: https://github.com/Return-To-The-Roots/s25client/issues/272
- [ ] Safer use of minimap (replace void*)


Flow86:
you can work around the fPIC stuff if you disable static boost libraries and use the dynamic ones (perhaps we should add some kind of configure check if boost is build with fpic and if not, use dynamic librarys?)

Flamefire:
@Flow86 I'd rather make it an option to use static boost libs: `OPTION(Boost_USE_STATIC_LIBS ON)`
Not sure if an automatic check would work.

Flow86:
yes an automatic check would be possible since you could try to link a dynamic library (with -fpic) and boost which will fail if boost is without fpic

stefson:
I guess s25rttr is broken with system-boost-1.56.0 ; the bug I filled in #276 is not occuring when upgrading to system-boost-1.58.0 

Will do some further investigations tomorrow.

PR #302, opened 2015-10-29
Fix typo: Japaneses -> Japanese

bluss:


bluss:
I chose to fix it in both the user visible string and the code.

PR #303, opened 2015-10-29
Fix iterator error in GameClientPlayer::JobNotWanted

bluss:
Upon erasing an element (and all == true), it would increment the
iterator twice.

Also remove the check against end, which is redundant with the loop
condition.

### DIFF
--- a/src/GameClientPlayer.cpp
+++ b/src/GameClientPlayer.cpp
@@ -800,13 +800,17 @@ void GameClientPlayer::AddJobWanted(const Job job, noRoadNode* workplace)
 
 void GameClientPlayer::JobNotWanted(noRoadNode* workplace,bool all)
 {
-    for(std::list<JobNeeded>::iterator it = jobs_wanted.begin(); it != jobs_wanted.end(); ++it)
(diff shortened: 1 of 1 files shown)


bluss:
Please review, maybe I'm wrong about this code, but I think this was an error.

I didn't find any bug, just ran across this in the code, but the interpretation [from use in buildings/nobHarborBuilding.cpp line 145] is that if a Harbour would have multiple jobs and they were sequential in the jobs list, this bug would mean only every second of the jobs would be properly canceled.

Flamefire:
I'll review that later. But seems alright.

Flow86:
yes I think its correct, the old code would skip one entry (which is wrong) if the call was with "not all", otherwise both codes do the same

src/GameClientPlayer.cpp
```diff
@@ -800,13 +800,17 @@ void GameClientPlayer::AddJobWanted(const Job job, noRoadNode* workplace)
 
 void GameClientPlayer::JobNotWanted(noRoadNode* workplace,bool all)
```
Flow86: the "workplace" here should be const, since its not modified inside this function?

bluss: Yes

Issue #298, opened 2015-10-23
Crash in Ware::GoalDestroyed; void Ware::GoalDestroyed(): Assertion `location' failed.


bluss:
Savegame: http://www2.zippyshare.com/v/7VWYvUFo/file.html

Crashes soon after load, when a fortress is occupied by the blue player

Log:
```
s25client: /home/x/src/s25client/src/Ware.cpp:166: void Ware::GoalDestroyed(): Assertion `location' failed.

Program received signal SIGABRT, Aborted.
(gdb) bt full
#0  0x00007ffff62e1107 in __GI_raise (sig=sig@entry=6)
    at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
        resultvar = 0
        pid = 30460
        selftid = 30460
#1  0x00007ffff62e24e8 in __GI_abort () at abort.c:89
        save_stage = 2
        act = {__sigaction_handler = {sa_handler = 0x7fffffffe969, 
            sa_sigaction = 0x7fffffffe969}, sa_mask = {__val = {140737324831169, 
              8297184, 166, 4, 140737488345216, 47381949632, 5020826, 0, 0, 0, 0, 
              21474836480, 140737324830695, 140737488345368, 140736929570816, 
              140737324846344}}, sa_flags = 8297516, 
          sa_restorer = 0x7e9d90 <Ware::GoalDestroyed()::__PRETTY_FUNCTION__>}
        sigs = {__val = {32, 0 <repeats 15 times>}}
#2  0x00007ffff62da226 in __assert_fail_base (
    fmt=0x7ffff6410d08 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", 
    assertion=assertion@entry=0x7e9c2c "location", 
    file=file@entry=0x7e9ae0 "/home/x/src/s25client/src/Ware.cpp", 
    line=line@entry=166, 
    function=function@entry=0x7e9d90 <Ware::GoalDestroyed()::__PRETTY_FUNCTION__> "void Ware::GoalDestroyed()") at assert.c:92
        str = 0x830ecc0 ""
        total = 4096
#3  0x00007ffff62da2d2 in __GI___assert_fail (
    assertion=assertion@entry=0x7e9c2c "location", 
    file=file@entry=0x7e9ae0 "/home/x/src/s25client/src/Ware.cpp", 
    line=line@entry=166, 
[...]

PR #305, opened 2015-11-01
Fix wrong building material ordering

Flamefire:
This fixes the problem described in the discussion of #298 where a building site orders to many building materials.

This changes it in a way that those superflous materials are still carried out of the harbour but instantly carried back in. This avoids problems like deleting the Ware in AddWare and accessing it later. This can be caused when Ware->RecalcRoute is called, the ware calls WareDontWantToTravelByShip and this calls AddWare. The caller of RecalcRoute might not be aware of this


Flamefire:
Please carefully review this!
Replay: http://www96.zippyshare.com/v/h5ekrbfg/file.html
Save: http://www96.zippyshare.com/v/DVVSu1Vz/file.html

Flamefire:
New PR from own branch: #306

PR #306, opened 2015-11-02
Fix wrong building material ordering

Flamefire:
This fixes the problem described in the discussion of #298 where a building site orders to many building materials.

This changes it in a way that those superflous materials are still carried out of the harbour but instantly carried back in. This avoids problems like deleting the Ware in AddWare and accessing it later. This can be caused when Ware->RecalcRoute is called, the ware calls WareDontWantToTravelByShip and this calls AddWare. The caller of RecalcRoute might not be aware of this

Please carefully review this!
Replay: http://www96.zippyshare.com/v/h5ekrbfg/file.html
Save: http://www96.zippyshare.com/v/DVVSu1Vz/file.html

### DIFF
--- a/src/GameClientPlayer.cpp
+++ b/src/GameClientPlayer.cpp
@@ -1342,7 +1342,9 @@ unsigned GameClientPlayer::GetBuidingSitePriority(const noBuildingSite* building
     }
 
     LOG.lprintf("GameClientPlayer::GetBuidingSitePriority: ERROR: BuildingSite or type of it not found in the list!\n");
-    return 0xFFFFFFFF;
+    assert(false);
+    // We may want to multiply this value so don't return the absolute max value
+    return std::numeric_limits<unsigned>::max() / 1000;
 }
 
 void GameClientPlayer::ConvertTransportData(const TransportOrders& transport_data)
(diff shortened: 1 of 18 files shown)


Flamefire:
@jhkl So can this be merged?

Flow86:
one question, since we have a lot pieces where we do some kind of "if state = x do xyz else if state = y ..." and since we have access to boost now, perhaps a look at boost::msm/fsm would appropriate and result to better code?

Flamefire:
I would not do that. I just had a look at it (good find though, didn't know that) and it looks pretty complicated so beginners in coding RTTR will have much more problems understanding, what is going on. Besides that I'm afraid we will run into a lot of issues regarding synchronisation, saving and loading as those state machines have also states which must be saved/loaded and I'm not sure that everything will be (cross-platform) as deterministic as we need it to be...
The current code is ok and there are other problems that should be tackled first. E.g. using smart-pointers to detect dangling pointers.

jhkl:
> @jhkl So can this be merged?

Yep. Sorry, I thought I wrote that here, but I think I only told you so in IRC. :)

Flamefire:
Ok. I just thought you would have merged that already if there was nothing left to change

PR #310, opened 2015-11-08
Ware fix

### LINKED ISSUE #309: Crash of programm
Hi!
I have another Crashing-savegame. This time it is not related to where the screen is. I attached the save. Just load it and wait 30s and you will see the crash…
Version 20151107 (Latest nightly)
 
Downloadlink (7days)
https://fil.email/1YpL3o?&showconfirmation=true

Regards
David

Flamefire:
Closes #309 and supersedes https://github.com/Return-To-The-Roots/s25client/commit/cc15d43a09fc7ab7f962cd3d71290ada8bc3d73a
Actually there should not be any ware with goal = destroyed harbour. If there is, then it is a bug. Currently the Handling of wares that were waiting in warehouses was not correct. This is fixed now by calling `WareNotNeeded` instead of `GoalDestroyed`


Flamefire:
Should be done. Can you cross-check again please? If you don't find anything else please merge.

PR #314, opened 2015-11-16
Fix crash when ordering wares for destroyed harbor (again)

### LINKED ISSUE #313: Crash - sea map
Hi!
I loaded with the latest nightly version my savegame from:
https://github.com/Return-To-The-Roots/s25client/issues/309
and continued playing. Now the first harbour attack was without a crash possible. I think the new crash is from the 2nd harbour attack (Left island). At the same time of the crash the ship quintia arrives on my island, but just delivering a scout and meat. So I don’t think that there is a relation… 
I attached again the savegame.
https://fil.email/18aHYM

Regards
David 
PS: I wanted to say a big thank you to the developing team! The support and fix time is amazing, with the last issues I reported!

jhkl:
This fixes #313 (I hope)

@Flamefire Would you like to merge this (or find a better solution ;) )

### DIFF
--- a/src/buildings/nobHarborBuilding.cpp
+++ b/src/buildings/nobHarborBuilding.cpp
@@ -110,12 +110,14 @@ nobHarborBuilding::nobHarborBuilding(const MapPoint pos, const unsigned char pla
 void nobHarborBuilding::Destroy()
 {
     em->RemoveEvent(orderware_ev);
-    players->getElement(player)->HarborDestroyed(this);
 
     // Der Wirtschaftsverwaltung Bescheid sagen
     GameClientPlayer& owner = gwg->GetPlayer(player);
     owner.RemoveWarehouse(this);
     owner.RemoveHarbor(this);
+
+    // Den Schiffen Bescheid sagen
+    players->getElement(player)->HarborDestroyed(this);
 
     // Baumaterialien in der Inventur verbuchen
     if(expedition.active)
@@ -561,7 +563,7 @@ void nobHarborBuilding::OrderExpeditionWares()
 void nobHarborBuilding::WareLost(Ware* ware)
 {
     // ggf. neue Waren für Expedition bestellen
-    if(expedition.active && (ware->type == GD_BOARDS || ware->type == GD_STONES))
+    if(expedition.active && (ware->type == GD_BOARDS || ware->type == GD_STONES) && !IsBeingDestroyedNow())
         OrderExpeditionWares();
     RemoveDependentWare(ware);
 }


src/buildings/nobHarborBuilding.cpp
```diff
@@ -110,13 +110,15 @@ nobHarborBuilding::nobHarborBuilding(const MapPoint pos, const unsigned char pla
...
     em->RemoveEvent(orderware_ev);
-    players->getElement(player)->HarborDestroyed(this);
 
     // Der Wirtschaftsverwaltung Bescheid sagen
     GameClientPlayer& owner = gwg->GetPlayer(player);
     owner.RemoveWarehouse(this);
     owner.RemoveHarbor(this);
 
+    // Den Schiffen Bescheid sagen
+    players->getElement(player)->HarborDestroyed(this);
```
Flamefire: Better use `owner.HarborDestroyed()`. I'd avoid the kinda global `player` variable (bad style and unsafe... it is the same as owner but that is not obvious)

Flamefire:
The fix is working (the part in `WareLost` is the essential one) But I'm thinking if there is a better way. The problem is similar as in #310: The beeing-destroyed harbor calls a function that eventually calls `ware.NotifyGoalAboutLostWare` which tells that same harbor that the ware is lost. But actually it should be the other way around: The harbor should tell the wares that it doesn't exist anymore... This will (or would) be done by the warehouse in this case (this is where the wares are registered)

Part of the way how to do this would be to remove the the calls to `ware.NotifyGoalAboutLostWare` in noShip.cpp#1035-1045 (same for figures) as we know, that the wares will be notified by the harbor. The problem then is the 2nd case there: `state == STATE_TRANSPORT_UNLOADING`. There it is a bit complicated.

Oh yeah... We definitely should go this way and fix a huge bug there too. Imagine the following: Ship is unloading and harbour is destroyed. Ship gets notified -> `state == STATE_TRANSPORT_UNLOADING` -> `HandleState_TransportDriving` is called -> `DriveToHarbour` actually finds the harbour as it was not yet removed (will be done by `noBaseBuilding`)
So some parts from the case HARBOR_DOESNT_EXIST:` need to be put in `HarborDestroyed` instead of the call to `HandleState_TransportDriving`. Figures and wares do not need to be handled (will be done by the harbour itself), the sea attackers look strange (only goal is set to `NULL`??? is this maybe also done by the harbor?) which leaves only lines #858-869 which calculates a new target. This is where the first part of your fix is important (remove harbour before notifying ships)

I would also put some assertions:

1. In WareLost: IsBeingDestroyedNow==false
2. `DriveToHarbour` should always find the harbour (so the state `HARBOR_DOESNT_EXIST` is useless) because the harbour would tell the ship when it vanishes.

Please do some carefull review of all this, I think we can prevent/fix quite some common sea-faring crashes with this.

Flamefire:
Superseded by #323

PR #317, opened 2015-11-27
Improved hitpoint text

Ribosom:
Removed max hitpoints. Colour of hitpoints will be red if a soldier has not the maximal amount of hitpoints.


Flamefire:
Just an Idea as you used colors here: What about a color scale? Like green=full HP, orange= more than half HP, red = max half HP

Flamefire:
Can you tell me if you like the color scale idea? If not, I'll merge that as it is right now.

Ribosom:
Sorry had not much time this weekend. I will try the colour scale later. :-)

Flamefire:
Looks good. Thank you!

PR #323, opened 2015-12-05
Seafaring fixes

Flamefire:
This refactores the seafaring system to solve a couple of bugs:

- #313 
- Loosing of wares because of calling idle for lost ships
- Wrong states for sea attacks

Tested with a (smallish) replay that includes transports, expeditions and attacks and did not async

However we should bump the savegame version before merging this as the changed use of states might lead to invalid situations with old saves (e.g. attack considered transport)

- [x] Roughly validated by @jhkl If anyone wants to double check please leave a message
- [x] Test for mem-error (see below)

### DIFF
--- a/src/EventManager.cpp
+++ b/src/EventManager.cpp
@@ -85,6 +85,7 @@ EventManager::EventPointer EventManager::AddEvent(EventPointer event)
     // Should be in the future!
     // At the current GF is allowed as long as a std::list is used for the events of one GF because it retains iterator validy even on modify
     assert(event->gf_next >= GAMECLIENT.GetGFNumber());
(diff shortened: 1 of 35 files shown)


Flamefire:
During tests I found a memory corruption potentially caused by seafaring. Not sure if this is due to the changes or existed before but it's hard to find and reproduce. I have a replay where the CRT checks can detect that (mostly) but it's really slow and does not help much. Maybe memgrind or so is better... https://www.sendspace.com/file/sduxjt

Created with the last commit from this PR: 24d30f1

Flamefire:
Test for that memory error so far shows, that probably a soldier that is an attacker of a harbor is deleted but not removed from its aggressors.

PR #325, opened 2015-12-06
Refactor Eventmanager

Flamefire:
This makes the `EventManager` much cleaner through typedefs and better names.

Also fixes 2 issues:

- Memory leaks when calling `Clear` -> List contents must be freed!
- It could happen, that the currently active event gets deleted which destroys the list -> Detect it

It also reduces the includes

- [x] Waiting for last commit (adds check if an objects events are still active)


Flamefire:
Ready to merge

Flow86:
I think it looks good

PR #329, opened 2015-12-10
Add LAN option

### LINKED ISSUE #327: LAN discovery / LAN lobby
Add either a LAN discovery (without a lobby server) so one can see open games in the same LAN or make add a server that creates a lobby similar to the Internet Lobby but for LAN mode (dedicated lobby server)

Easy implementation: Use internet lobby server (and ship it with RTTR) but make it usable in LAN. This means:
- Switchable IRC connection (no Lobbybot in LAN)
- Add option to change Lobby server in RTTR (e.g. use same `Desktop` as for internet but if it is reached via the LAN button, then one has to give the servers IP, which is also stored for next usage)

Flamefire:
Closes #327 

- [x] Requires rebase after https://github.com/Return-To-The-Roots/libutil/pull/16 is merged!

Adds a LAN mode with automatic server discovery through UDP. New desktop is added for the long existing LAN button!

To catch call issues the `ServerType` enum was factored out to a new file, transformed to a boost scoped enum (type safety) and all usages fixed accordingly.

2 Small fixes slipped in: Datatype for lobby list was wrong and passing the host now as a string instead of a char*

Ports are constant and choose right after the regular RTTR ports. Maybe other ports are better?

### DIFF
--- a/src/GameClient.cpp
+++ b/src/GameClient.cpp
@@ -1599,10 +1599,6 @@ void GameClient::ExecuteGameFrame(const bool skipping)
 /// Führt notwendige Dinge für nächsten GF aus
 void GameClient::NextGF()
 {
-    if(framesinfo.nr == 83831)
-    {
-        _CrtSetDbgFlag(_CRTDBG_ALLOC_MEM_DF | _CRTDBG_LEAK_CHECK_DF | _CRTDBG_CHECK_ALWAYS_DF);
-    }
     // Statistiken aktualisieren
     StatisticStep();
     //  EventManager Bescheid sagen
--- a/src/GameClient.cpp
+++ b/src/GameClient.cpp
@@ -78,7 +78,6 @@ void GameClient::ClientConfig::Clear()
     password.clear();
     mapfile.clear();
     mapfilepath.clear();
-    servertyp = 0;
     port = 0;
     host = false;
 }
@@ -189,7 +188,7 @@ GameClient::~GameClient(void)
  *  @author OLiver
  *  @author FloSoft
  */
-bool GameClient::Connect(const std::string& server, const std::string& password, unsigned char servertyp, unsigned short port, bool host, bool use_ipv6)
+bool GameClient::Connect(const std::string& server, const std::string& password, ServerType servertyp, unsigned short port, bool host, bool use_ipv6)
 {
     Stop();
     ggs.LoadSettings();
@@ -1600,6 +1599,10 @@ void GameClient::ExecuteGameFrame(const bool skipping)
 /// Führt notwendige Dinge für nächsten GF aus
 void GameClient::NextGF()
 {
+    if(framesinfo.nr == 83831)
+    {
+        _CrtSetDbgFlag(_CRTDBG_ALLOC_MEM_DF | _CRTDBG_LEAK_CHECK_DF | _CRTDBG_CHECK_ALWAYS_DF);
+    }
     // Statistiken aktualisieren
     StatisticStep();
     //  EventManager Bescheid sagen
--- a/src/desktops/dskHostGame.cpp
+++ b/src/desktops/dskHostGame.cpp
@@ -55,8 +55,8 @@ static char THIS_FILE[] = __FILE__;
  *  @author Devil
  *  @author FloSoft
  */
-dskHostGame::dskHostGame(ServerType serverType) :
-    Desktop(LOADER.GetImageN("setup015", 0)), temppunkte_(0), hasCountdown_(false), serverType(serverType)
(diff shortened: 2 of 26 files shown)


src/GameClient.cpp
```diff
@@ -1601,6 +1600,10 @@ void GameClient::ExecuteGameFrame(const bool skipping)
 /// Führt notwendige Dinge für nächsten GF aus
 void GameClient::NextGF()
 {
+    if(framesinfo.nr == 83831)
+    {
+        _CrtSetDbgFlag(_CRTDBG_ALLOC_MEM_DF | _CRTDBG_LEAK_CHECK_DF | _CRTDBG_CHECK_ALWAYS_DF);
+    }
     // Statistiken aktualisieren
```
Flow86: debug stuff???

Flamefire: I knew I forgot that... thanks

src/desktops/dskHostGame.cpp
```diff
@@ -673,10 +675,8 @@ void dskHostGame::CI_Countdown(int countdown)
  */
 void dskHostGame::CI_CancelCountdown()
 {
-    if (isSinglePlayer_)
-    {
+    if (serverType == ServerType::LOCAL)
         return;
```
Flow86: why not introduce some "isX"-methods?

Flamefire: Only reasonable method would be `IsSinglePlayer` to show the intention better. But not really required IMO because the comparison of `serverType` is already clear enough.

Compare: `if(serverType == ServerType::LOBBY)` vs `if(IsServerTypeLobby())` --> Not much gained imo

src/desktops/dskHostGame.cpp
```diff
@@ -524,6 +525,19 @@ void dskHostGame::Msg_Group_ComboSelectItem(const unsigned int group_id, const u
...
+{
+    if (serverType == ServerType::LOCAL)
+        WINDOWMANAGER.Switch(new dskSinglePlayer);
+    else if (serverType == ServerType::LAN)
+        WINDOWMANAGER.Switch(new dskLAN);
+    else if (serverType == ServerType::LOBBY && LOBBYCLIENT.LoggedIn())
+        WINDOWMANAGER.Switch(new dskLobby);
+    else
+        WINDOWMANAGER.Switch(new dskDirectIP);
+}
```
Flow86: why not use a switch-case?

src/desktops/dskHostGame.cpp
Flow86: yea perhaps

src/desktops/dskHostGame.h
```diff
@@ -37,7 +35,7 @@ class dskHostGame :
     public:
 
         /// Map übergeben, damit die Kartenvorschau erstellt werden kann
-        dskHostGame(bool single_player = false);
+        dskHostGame(ServerType serverType);
```
Flow86: shouldnt the server type be fixed? (const&)

src/desktops/dskHostGame.h
```diff
@@ -90,11 +88,12 @@ class dskHostGame :
 
         void LC_Status_Error(const std::string& error);
 
+        void GoBack();
     private:
         int temppunkte_; // TODO - wegmachen und durch korrekte punkte ersetzen!
         GlobalGameSettings ggs_;
         bool hasCountdown_;
-        bool isSinglePlayer_;
+        ServerType serverType;
```
Flow86: shouldnt the server type be fixed? (const&)

src/desktops/dskHostGame.h
Flamefire: Yes it should be const, but not const& (its an enum -> pass by value)

Flamefire:
Incorporated your suggestions

Flow86:
has some branch conflicts now, cannot automerge it

Flamefire:
It also needs a rebase when the libutil changes were merged. If you approve this I'll rebase, include the libutil and merge afterwards

Flow86:
:+1:

Flamefire:
Ok. Then only the libutil is missing. Can you recheck and merge if ok? Answered your comments there

Flow86:
libutil is merged

Flamefire:
Alright. I'll merge this tomorrow together with the rest of the PRs except the server fixes. There WILL be bugs, but I can't distinguish bugs introduced by the PR and existing bugs or bugs discovered by additional checking (asserts) anymore, so fixes will follow. Because otherwise the PR will just get bigger and bigger or I'd need to fix a bug but commit to a separate branch which depends on that branch... Or do you have any objections?

Flow86:
no its okay

PR #419, opened 2016-03-14
Speed indicator

### LINKED ISSUE #406: Improve speed changes
From #404: Pressing 'v'  multiple times will ignore some of them. So it should be improved, so that the client sends the speed he wants, not only a speed change.
Also the icon is only shown on high speed settings but we have 6 levels of speed which make it seem as speed is not changed.

Flamefire:
This allows multiple speed changes during 1 NWF (pressing 'v' multiple times by now only accepts the first one)

It also adds a little number indicating the current speed stepping. Scheme is simple: Show difference to 30 ms in steps of 10 ms with negative (slower) values red and faster ones yellow. Nothing for 30ms and no number for 20ms.

This closes #406

speed

### DIFF
--- a/src/FramesInfo.h
+++ b/src/FramesInfo.h
@@ -24,7 +24,7 @@ struct FramesInfo
 public:
     FramesInfo();
     void Clear();
-    /// Changes the GF length to GFLengthNEw and adapts the NWF length accordingly
+    /// Changes the GF length to GFLengthNew and adapts the NWF length accordingly
     void ApplyNewGFLength();
 
     /// Current GameFrame (GF) (from start of the game)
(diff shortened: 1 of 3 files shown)


Spikeone:
Faster might displayed in green I guess.

But that's maybe just a taste.

Flamefire:
I was also thinking about green. But most of the text is yellow, this is why I choose yellow. But I don't mind any color... Any other opinions?

Spikeone:
Green = faster then set at the beginning
Yellow = speed as set
Red = slower than set

Flamefire:
I wanted to make it relative to the settings (normal, fast, very fast...) and only show, what is necessary. So normal is nothing shown and fast is missing the number

Spikeone:
Okay then, I don't really care so, it's the way it is :

Flamefire:
Relative to setting at begin is a good idea though... But that requires storing that setting somewhere.

Spikeone:
But number size may be increased (font size) looks a bit tiny I guess.

Also I'm not sure if we should really save speed, maybe while a game is running but not in a savegame as you may change the speed after loading.

jhkl:
Looks hard to read, but I guess you don't have to read it very often -> Ok

Issue #422, opened 2016-03-15
Transmit lua script like map packets


Flamefire:
Currently the lua script is sent as a string in the map info packet. This easily exceeds the maximum packet size. So the lua script should be compressed (probably even with the map data it self) and sent like the map data.

This allows bigger scripts (requested in #36) and fixes bugs due to to large TCP packets.

PR #423, opened 2016-03-18
Lua interface rework

### LINKED ISSUE #421: Add LUA save/load
Out from #36:

Saving and loading of lua states should be possible.

Research shows, that automatic save/load is really problematic. So I'd add Save/Load callins to lua and make the script save its own state as required. Instead of OnStart, OnContinue can be called or a param be added to OnStart.

Details: Pass lua an object to Save/Load that can be used to Write/Read integers and strings. This is serialized via serializer and appended to the savegame. Should be pretty easy and keeps the savegame small (LUA VM state might be rather large...)

Flamefire:
This switches the lua interface to a C++ wrapper based OO-Design

It also adds a simple test case that runs all Lua functions and does also some minor error checking (I'd love to have unit tests...)

It changes the current interface completely so current lua scripts won't work anymore.
However the transition should be straight forward, see https://github.com/Flamefire/s25client/blob/lua/tests/maps/LuaFunctions.lua

@Spikeone: Have a look at the Lua file: https://github.com/Flamefire/s25client/blob/lua/tests/maps/LuaFunctions.lua and tell me if there is anything to improve for you

Will close #421, #422

Todo: 
- [x] test player join event (I think one cannot set nation for open spots...)
- [x] Run onSettingsReady only on host(?) (setting addons etc should only be done by the host)


.gitmodules
```diff
@@ -25,3 +25,6 @@
 [submodule "RTTR/languages"]
 	url = https://github.com/Return-To-The-Roots/languages.git
 	path = RTTR/languages
+[submodule "contrib/kaguya"]
+	path = contrib/kaguya
+	url = https://github.com/satoren/kaguya.git
```
Flow86: we should fork the kaguya repo to the rttr scope if we use it at the end

Flamefire: I thought about that. But I did not see an advantage (for now). The guy maintaining it is pretty active and this way we are better linked to the upstream. We maintain the other repos, hence they are in our scope, but I don't want/need us to maintain kaguya ;)

Flamefire:
In the works: Change game settings via Lua

.gitmodules
Flow86: yes but I dont like the idea depending on "some stuff somewhere on the internet" ;-)

Flow86:
one side note: I've authorized travis and appveyor now, appveyor build is running ;-)

.gitmodules
Flamefire: We don't really depend on it (in a bad way). This is what submodules are for: Pull in some *specific* source code. So we just pull in another library at a version we choose. Why should we have extra work for maintaining our copy of that repo?

Flamefire:
Ok nice :) Can I have the URL to the badge or can you add the 2 badges (travis, appveyor for master branch) to the readme? Always nice to see a green icon ;)

Flamefire:
@Spikeone It is now possible to change game settings including addons while even allowing the user to change some addons. Have a look at the now extended example script.

Todo: 
- [ ] add player join event (I think one cannot set nation for open spots...)
- [ ] Run onSettingsReady only on host (setting addons etc should only be done by the host)

Spikeone:
Just <3

Flamefire:
Almost complete now. Still need to test the player joining and do some refactoring for names of commands (many are called toggle but are actually set commands)

Spikeone:
So I can close playerslots and add AI players then? Anyway, great work at all :)

Flamefire:
Yes :)

Spikeone:
Uh uh, now we need some input boxes like vbs does, simple yes no selection, pause game, simple input box so we can actually use puzzles or riddles where you have to find like a password or solve something and then... uhhh those possibilitys. Maybe I wont convert my old scripts but rewrite them from scratch

Flamefire:
Currently adding message boxes. But I would not add decisions in the near future as this might cause asyncs...

Spikeone:
Oh and can you also add like pictures in messages somehow? so we can specify any image by x and y offset from the upper left of that message box to be displayed? That might not be a very handy way but... at least it works and one could do nearly everything with that (like tutorial pictures etc.)

Flamefire:
Ok :)

Spikeone:
<3 love it

Flamefire:
Tested and I'd considers this ready to go. Wiki is already updated

PR #428, opened 2016-03-19
Fix template deduction error (for GCC >= 5.3)

gattschardo:
without specifying `::Type` default template deduction fails in newer versions of GCC (5.3 in my case)

### DIFF
--- a/src/world/MapGeometry.cpp
+++ b/src/world/MapGeometry.cpp
@@ -22,7 +22,7 @@
 // Include last!
 #include "DebugNew.h" // IWYU pragma: keep
 
-Point<int> GetNeighbour(const Point<int>& p, const Direction dir)
+Point<int> GetNeighbour(const Point<int>& p, const Direction::Type dir)
 {
     /*  Note that every 2nd row is shifted by half a triangle to the left, therefore:
     Modifications for the dirs:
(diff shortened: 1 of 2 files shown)


Flamefire:
Thanks for the PR. However I fail to see the error here. What we have is basically a kind of scoped enum (emulated due to C++03) and passing a value of type `Direction` should be totally fine. Also I don't see the template involved in this.
Can you post the exact errors you are getting and, if you can find any, some resources on why it has to be solved like that? Just want to make sure, this is the right way and not just a compiler issue

gattschardo:
Hi Flamefire, thanks for taking a look so quickly!

You are actually right, this PR is now unecessary, since you fixed the issue yourself with b20376c5abc9f4ccc802966a320d758cf5b0b9d6 -- I had this lying around a couple of days before I got around to submitting it and didn't realize it was obsolete after the rebase :). My bad, sorry for the noise!

Flamefire:
No worries :) Always good to see someone is contributing. Even just compiling (with other compilers/OS) sometimes helps to find issues
Issue in this case is, I'm using VS2015 for developing most of the time and VS got some mixture of C++03 and C++11 and is also pretty forgiving about includes and stuff. I need to push it to the build server to see if it works on all platforms. But now we got Travis CI integration for PRs which should make this kind of bugs quite rare :)

PR #432, opened 2016-03-23
Add add-on to disable automatic mouse positioning to allow split-screen multiplayer

jonathan-reisdorf:
This PR solves an issue I had while getting local split-screen multiplayer to work.
I'm using Linux and use 2 USB mice and xinput like this to create two mouse cursors:
```
$ xinput create-master player2
$ xinput list
$ xinput reattach <mouse id> <player2 pointer id>
```
Then I run the s25client twice in windowed mode, each window has half of the screen.
The first window starts a multiplayer lan server, the second window joins it.
So far, so good, but after some seconds of playing, reproducably the mouse cursor of the left window was caught in the right window, so that both mouses were controlling the window.
The reason for this is the SetMousePos calls that are done if a dialog like for the building is opened. That convenience feature does not move the "right" cursor in the right window, but the "left" cursor because of course it does not know something about two system cursors being present.
Therefore I added this add-on which can be enabled under the misc section, that allows disabling this feature.


jonathan-reisdorf:
Oh, by the way, this is my first PR to this project and also my first C++ code, so feel free to double check everything :)

src/addons/AddonDisableAutomaticMousePos.h
```diff
@@ -0,0 +1,46 @@
...
+ *  Addon that allows disabling the automatic mouse cursor positioning on modals, so that e.g. local split-screen multiplayer with xinput is possible
+ *
+ *  @author Jonathan Reisdorf
+ */
+class AddonDisableAutomaticMousePos : public AddonBool
+{
+    public:
+        AddonDisableAutomaticMousePos() : AddonBool(ADDON_DISABLE_AUTOMATIC_MOUSE_POS,
+                                             ADDONGROUP_OTHER,
+                                             gettext_noop("Disable Automatic Mouse Positioning"),
```
Flamefire: Using `_("FooBar")` instead of `gettext_noop("FooBar")` enables translations

Flamefire:
If this actually works, I'd accept this as a workaround. Can remove the re-formating of the files? There are some spaces added/removed, that make this change appear bigger than it should (check "Files Changed" for lines modified that are not added by you) Don't worry about the addon alignment, as there will be a bigger change to that very soon
Can you also check, how this affects the main menu? I'm not sure, if GAMECLIENT.getGGS() is already available and valid there but there are some dialogs and mouse movement too.

More general: Are you sure, that there is no other way? This addon would make placing flags much more cumbersome :/

jonathan-reisdorf:
Thank you @Flamefire, I have applied the suggested changes. Regarding your general question: Nothing I can think of. In my view it's correct that the application doesn't know something about the presence of a second cursor, and, if achievable at all, it sounds imo like a quite hard thing to implement. Of course once could, instead of shifting the cursor, shift the modals, so that the first button is under the cursor, in case the add-on is enabled. Not so sure if that's a good user experience, though.

jonathan-reisdorf:
As for the main menu: As you said, GAMECLIENT.getGGS() isn't available for instance the first time it's called in the main menu, e.g. when you join a LAN game, but instead it's instantiated automatically the first time it's needed (at least it seems so?). Therefore, I can reproduce that the the add-on flag returns false for that first time even if it's enabled, which makes the cursor jump in the main menu, but just for the first time, for subsequent times it correctly returns true. Is that a regression that needs to be fixed?

Flamefire:
I think this is due to the fact, that addons are meant for playing, not for the menu.
I'd suggest you remove the addon and make it a regular setting instead (look at `SETTINGS` and `dskConfig`) It is actually quite easy to add another checkbox, setting, and save/load. Also make sure, that in `LoadDefaults` the value is set to false

jonathan-reisdorf:
:+1: sounds good, even though the setting is pretty prominent then. I'll add it to the general section then.

jonathan-reisdorf:
I've done the changes now and it now also works in the menu. (not sure why the file diff shows changes in  src/GlobalGameSettings.cpp, looks the same to me, maybe my editor removed some invisible chars?)

Flamefire:
:+1: 
You can reset a file with `git checkout <commit> -- <file>` Looks like a newline issue in this case

Currently I don't really like the name. New users will be confused. I understand it need to be short. Maybe "Intelligent mouse cursor"? (or just "Intelligent cursor") Probably someone else has an idea? ( @Spikeone @jhkl @MarcusSt @Flow86 )

At least I'd add tooltips to the buttons: "Automatically moves mouse cursor (e.g. to default window buttons)" and "Disables automatic cursor movement" or something like this. But I'm pretty bad with names and UI stuff ;-)

jonathan-reisdorf:
Hm, maybe contextual mouse shifting? Well, I guess I'm also not the best when it comes to naming...

jhkl:
Why not something like "Two mice support"? In the tooltip it would be nice to somehow describe the xinput stuff. 

Maybe even show that option only when running Linux?

Flamefire:
Would not mention the 2 mice support as it would be kinda misleading IMO (2 mice are not really supported by RTTR)
Showing the option only in linux (better: non-windows) is a good idea :) Then during load it should be set to false on windows.

Spikeone:
I also dislike those smartmoves sometimes - so why we just don't  call it that way?

jonathan-reisdorf:
What about Smart Mouse Cursor Shifting? Haven't tried yet if it fits into the available space, but if not, maybe Smart Cursor Shifting? Of course that's not too self explanatory, so I guess adding tooltips is required. About showing it only on Linux: I'm not that sure if multiple cursors are only achievable in Linux, afaik there's 3rd party software for Windows (EitherMouse, Pluralinput, and TeamPlayer), not sure about Mac. I can test later if one of these tools work the way they should on Windows.

Flamefire:
"Shifting" is strange... Should be rather "Placement" or "Moving". Or just "Smart mouse cursor" and the tooltips we need either way.
And as @Spikeone said, he would disable this sometimes even for single mouse, I'd leave the setting for all OSes

jonathan-reisdorf:
OK, I named it "Smart Mouse Placement" now and added tooltips. Hope they are descriptive enough.

Flamefire:
If others like the name I'm ok with it. Otherwise I'd suggest the shorter "Smart cursor", in German "Intelligenter Cursor" as its shorter. But thats just minor.

Could you rename the property and make sure the meaning is consistent? Because we now have "Smart Mouse Placement" which goes into "DisableCursorJumping" -> Off is On and On is Off ;)

jonathan-reisdorf:
"Smart Cursor" isn't enough imo as it's pretty generalistic. It could even mean sth like that the appearance of the cursor changes vs a hardware cursor that's always the same. As for the translations: Ich finde, die Übersetzungen müssen nicht zwangsläufig immer 1:1 mit dem Englischen übereinstimmen. Intelligenter Cursor fänd ich zB ok für DE.
About the off = on thing: I did that on purpose because there is an issue with the existing system: If I didn't reverse the logic, smart cursors would stop working on existing applications that are just updated, because the current structure method that gets values from the config file is type casting to integer type, therefore on configs that don't contain the property at all, always 0 (false) is returned, not the default that is specified and only applied if you don't have a local config file present. Of course I could check if the actual config entry is just an empty string and then apply the default value, but it's done nowhere else like this so I thought it'd be easier to just switch the logic so that false is always ok for existing systems.

Flamefire:
But if "Intelligenter Cursor" is ok, why shouldn't "Smart cursor" be? I know its pretty general but its also nicely short and catchy ;) And we have the tooltips anyways. But if it fits in the space, its ok.

Oh yes, you are right... I still think it will be quite confusing. I'd rather fix the default value issue instead...

jonathan-reisdorf:
I'm okay with that. Have done the supposed changes and just test played in split screen, it all works well now.

Flamefire:
Ok looks good so far. Still one thing to polish: The tooltips are to long. E.g. No need to mention the setting name for Off. I'd say: "Don't move mouse cursor automatically.\nUseful for multi-mouse mode"
And for On, it could be simply: "When opening a dialog window the cursor is placed at the default button automatically" Maybe even omit "window" ("The first" is wrong, as e.g. for safety dialogs the default is "No") Or "Place cursor on default button for new dialogs"
Reasoning is that nobody likes reading that much text for settings.
@Spikeone you got more experience with design. Opinion?

jonathan-reisdorf:
I wouldn't say "multi-mouse mode" as it suggests that we have such a mode built into the application. Shortening the texts sounds good, but the reason for them to be added in the first place is to increase understandability of the whole feature - I wouldn't sacrifice that just because it looks nicer

Flamefire:
Good point.
Yes, but they should complement each other. I think tooltips for 1 option of the kind On->"Makes it do `<this>`", Off->"Does not make it do `<this>`" are not that much helpful, than 2 different ones.
I think "Don't move mouse cursor automatically.\n`<Something about why this would be useful in very few words>` is ok for Off, as the first line says all that needs to be said.
What about "Place cursor on default button for new dialogs / action windows"?

jonathan-reisdorf:
OK, then what about
"Don't move the mouse cursor automatically.\nUseful e.g. for split-screen / dual-mice multiplayer on Linux (see wiki)" for off
and
"Place mouse cursor on default button for new dialogs / action windows (S2 default)" for on?

Would you add the hint with the S2 default / wiki or is that too much for tooltips?

Flamefire:
I'd probably drop "mouse" as I find "cursor" already specific enough, so this word is just noise to me, or is there a reason for relating to "mouse cursor"?
I like the S2 default (maybe even just default) and the wiki hint. Especially the last one for multiplayer as this is often requested. I'd drop "Linux" as the same might work for OS X and windows(?)

jonathan-reisdorf:
I added it because there's also the text cursor, but of course it gets more clear with the context (window/button). Regarding Mac / Windows compatibility: I tested it on Mac, no success so far. Mac simply doesn't have the possibility to create more than one pointer, even though it makes use of Linux' xinput with some modifications. Other searches clearly state that it's not possible for Mac.
For Windows I tried out EitherMouse (no success at all) and dicolab's TeamPlayer (quite buggy because the rttr window didn't catch the 2nd mouse automatically but just when clicked with it, but then, at the same time, the first window lost the track, because it's technically just one pointer that is switching between the positions). Wasn't able to try out PluralInput, because they currently switched their download offline and I couldn't find another source.
So I guess we can consider this being a Linux-only feature (at least yet). But you're right, it doesn't need to be mentioned if that can also be done in the wiki...

Flamefire:
That is unfortunate... Even SDL2 does not support it. I once created a branch using ManyMouse to support (kind of) split screen (there is a youtube demo somewhere around) but it wasn't accepted and I deleted the sources. :(
About the tooltips: I like them. So PR looks good to me. If you like, you can also add the German translations before I merge this.

jonathan-reisdorf:
Oh, the solution you had sounds much better than this workaround (just saw the video), too bad that you don't have the source anymore :/
About the translation: How does it work? The last time I worked with .po files is.. ages ago. I have a po editor here and it only gives me the option to update existing entries, not to add new ones, and the changes I did are not in there - is there a sort of "migration" command or do I just add them in there manually?

Flamefire:
Damn... We need some good way to update the pot files I guess. Ok, so I'll just accept this for now and we will translate this later

Flow86:
there is a way to update them, "Catalog" -> "Refresh from sources". But one should do this only for "rttr.pot" (the template).

We can also download the current template from launchpad (since code gets mirrored there, the rttr.pot is up to date ...)

Flamefire:
-> See #435. Please describe the preferred way. I guess just download everything from LP and commit it?

PR #439, opened 2016-04-01
Addon text improvements

Flamefire:
This makes the addon texts translatable and also shortens them a bit. No need for "Allows you to change..." in the tooltips. Lengthy tooltips are hard to read especially as we got many settings.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -53,27 +53,10 @@ AddDirectory(world)
 
 ################################################################################
 
+FILE(GLOB RTTR_PO_FILES ../RTTR/languages/*.po)
+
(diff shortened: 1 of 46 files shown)


Flow86:
you know that they were translated in the addon-table, thats why the gettext_noop was there (so the text parser will find the strings, but that the widgets do not translate them "twice")

Flamefire:
So these strings are translated when they are displayed? Why not translate them here? It is shorter and clearer. We just need to remove the 2nd translation...

Flow86:
I dont know anymore why, but I think there was a reason for that

Flamefire:
Well then if it works like this PR (with the removal of the 2nd translation) I would use this. Cleaner, more consistent and hence easier for new devs

Flow86:
okay

PR #464, opened 2016-04-28
Update .travis.yml

Flamefire:
CodeCov.io added support for settings via `codecov.yml`
This also adds coveralls statistics as codecov is a bit buggy ATM

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 sudo: false
 language: cpp
+branches:
+    only:
+        - master
 
 os:
     - linux
@@ -80,6 +83,8 @@ script:
     # Execute tests
     - export CTEST_OUTPUT_ON_FAILURE=1
     - make test
+    # Reset environment
+    - cd ..
     - set +e
 
 after_success:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,5 +1,8 @@
 sudo: false
 language: cpp
+branches:
+    only:
+        - master
```
Flow86: that will break pull request checks

Flamefire: It doesn't. PRs are handled additionally. Have a look at the first commit in this PR, that did not have this. There are 2 builds from travis, which should not be.

Flamefire:
CodeCov seems to have some issues at the moment. I'd like to add support for coveralls as well, to compare those 2 services and make sure, there is at least 1 working ;)

Flow86:
coveralls is activated now, you only have to write the .coveralls.yml

Flamefire:
The yml file for coveralls is only needed for private repos. Works now and looks good. Although I liked the old design of codecov better. Let's see how it looks like once their issues are resolved.

stevepeak:
@Flamefire you can remove the "changes" above the diff if you desire

```
comment:
  layout: "header, diff"  # old-style
```

Flamefire:
@stevepeak I don't mind having them. They might be helpful (if they only were correct -> https://github.com/codecov/support/issues/164)

PR #471, opened 2016-05-02
Implement MVC pattern for GameWorld* classes

### LINKED ISSUE #472: s25 wont start "missing video driver"
When I try to open the client, I am getting the following Terminal output:

```
Last login: Sat Apr 23 11:41:19 on ttys000
localhost:~ Alex$ /tmp/rttr.command
Updating "put your S2-Installation in here":                       nan% - decompression failed: download failure?
lade "/Users/Alex/Library/Application Support/Return To The Roots/CONFIG.INI": fertig (0ms)
searching for drivers in ./share/s25rttr/driver/video/*.dylib
0 video drivers found!
Video driver couldn't be loaded!


ERROR: Das Spiel konnte nicht gestartet werden
Verbinde mit 78.46.212.230:4123
Verbindung erfolgreich hergestellt mit debug.rttr.info:4123
Sending replay...
Will now send 26 stack frames
/Applications/s25client.app/Contents/MacOS/rttr.command: line 63:  6960 Abort trap: 6           ./bin/s25client
localhost:~ Alex$
```

In the folder: /Applications/s25client.app/Contents/MacOS/share/s25rttr/driver/video/
is a "*.dylib" driver. So I dont get, why its not working out to load it?

Flamefire:
Next step for decoupling: Implement MVC concept

- World = Contains only data and a few basic setters/getters (Model)
- GameWorldBase: World = Contains game related data (players, settings, pathfinding) and more complex methods used in game context (Enhanced Model)
- GameWorldGame: GameWorldBase = Can modify the world, used by the GameObjects to interact with the world (Controller)
- GameWorld: GameWorldGame = Top level class used by the GameManager (Serializing/Deserializing Stuff) [Could probably go, Nothing really useful in here. Save/load can be done by the MapLoader]
- GameWorldViewer *<- GameWorldBase: A players view on the World. Stuff specific to a player belongs here (Renderer, in the future this can contain all "visual" settings (BQ, roads...) (This is the View obviously) There can be multiple viewers for a World (1 for each local player)
- GameWorldView *<- GameWorldViewer: A View on the players view. Used to actually show something to the player. There can be multiple Views for each Viewer (observation windows)

Notes:
- The View/Viewer stuff is confusing. We need better names...
- The GameWorldGame should actually contain a GameWorldBase(referenz) as this is the idea behind a controller. We should also check, which methods we can move up (allow to modify stuff only in controller, player specific requests in the Viewer) and expose only a `const GameWorldBase` to everything but the GameWorldGame

This also adds a `GameLobby` class, splits up `PlayerInfo` and fixes some bugs. The aim is to create real "States" that contain only the information of that state. No need to keep "isReady" after the GameLobby.
Also lot's of stuff got even more decoupled.
Finally it also fixes rendering as some things there needed to change during this PR. This might fix #472 amd #475


src/TerrainRenderer.cpp
```diff
@@ -1071,30 +1073,29 @@ void TerrainRenderer::AltitudeChanged(const MapPoint pt, const GameWorldViewer&
...
 
     for(unsigned i = 0; i < 6; ++i)
-        UpdateBorderVertex(gwv.GetNeighbour(pt, i), gwv);
+        UpdateBorderVertex(gwv.GetNeighbour(pt, Direction::fromInt(i)), gwv);
 
     // den selbst sowieso die Punkte darum updaten, da sich bei letzteren die Schattierung geändert haben könnte
     UpdateTrianglePos(pt, gwv, true);
     UpdateTriangleColor(pt, gwv, true);
 
     for(unsigned i = 0; i < 6; ++i)
```
Flow86: why not run over Directions?

src/world/GameWorld.cpp
```diff
@@ -85,9 +84,6 @@ bool GameWorld::LoadMap(const std::string& mapFilePath, const std::string& luaFi
     }
 
     CreateTradeGraphs();
-
-    GetTerrainRenderer().GenerateOpenGL(*this);
-
```
Flow86: wo wird das nun anstelle aufgerufen?

src/TerrainRenderer.cpp
Flamefire: Direction is a scoped enum. It is (currently) not possible to iterate over it directly (But should be implemented later)
The fromInt method is a zero-overhead funtion for converting

src/world/GameWorld.cpp
Flamefire: During creation of `GameWorldViewer`: https://github.com/Return-To-The-Roots/s25client/pull/471/files#diff-e660a654c4f718d425de16c0f5d5ececR34
And that one is constructed  in `dskGameInterface`.

Checking error codes showed an GL_OUT_OF_MEMORY in the first call of `glDrawArrays` when drawing the textures (only when using VBO)

src/TerrainRenderer.cpp
```diff
@@ -763,16 +736,21 @@ void TerrainRenderer::Draw(const GameWorldView& gwv, unsigned int* water)
 
     lastOffset = PointI(0, 0);
 
+    // Arrays aktivieren
+    glEnableClientState(GL_VERTEX_ARRAY);
+    glEnableClientState(GL_TEXTURE_COORD_ARRAY);
+    glEnableClientState(GL_COLOR_ARRAY);
```
Flow86: shouldnt we disable them (after we have drawn) if we have enabled them?

Flamefire: Is done. Check line 826ff

Flow86: ah youre right, I only have seen that you removed some of those disable lines ;-)

Flamefire: I just made sure all those enabling is done where it belongs and is also disabled after it was enabled. The glInterleaved stuff enabled those in the background which made it non-obvious where things are enabled... There were also some bindings that did not belong there.

PR #473, opened 2016-05-03
Add more info about failing drivers #472

Flamefire:
This should help in debugging issues like #472 

It also shows the current directory when rttr is started (we use relative paths internally)

@Flow86 Please merge as soon as this completes building in travis to help that case

### DIFF
--- a/release/bin/macos/rttr.command
+++ b/release/bin/macos/rttr.command
@@ -1,11 +1,14 @@
 #!/bin/sh
 
-cd $(dirname $0)
+# Get absolute path to scripts folder
+scriptFolder="$(cd "$(dirname "$0")" && pwd)"
+# Get absolute path to script
+scriptPath="$scriptFolder/$(basename "$0")"
+cd "$scriptFolder" || exit 1
 
 if [ ! "$1" = "interminal" ] ; then
-	scriptPath=$(pwd)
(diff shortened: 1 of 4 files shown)


release/bin/macos/rttr.command
```diff
@@ -3,16 +3,17 @@
 cd $(dirname $0)
 
 if [ ! "$1" = "interminal" ] ; then
+	scriptPath=$(pwd)
 	echo "#!/bin/sh" > /tmp/rttr.command
-	echo "$0 interminal" >> /tmp/rttr.command
+	echo "$scriptPath/$0 interminal" >> /tmp/rttr.command
```
Flow86: thats wrong, since $0 already contains full path

Flamefire: No. Only the invocation which can be relative. Manually calling it is impossible like that. So need basename $0

Flow86: you call it manually, thats what is done "doubleclicking" the app. since there wont be \$0 written inside the command, its the expanded var

Flamefire: I meant when you call it via command line. Then `$0` gets only the relative path

Flamefire:
@Flow86 Launching should work now. Tested on OSX with spaces in path and launching from the terminal

PR #502, opened 2016-07-17
Improve lobby and HostGame-Window

Flamefire:
This adds the option to talk in the lobby when you are hosting/have joined a game. However this is currently disabled as this requires server support. Coming soon.


Flamefire:
@Flow86 Does the server run a different version then in the repo? I get disconnected when I try to send a message from within the game, but this does not happen with a local lobby server.

Flow86:
server should always run the latest version (rebuilds+restarts once per day)

Flow86:
ah! but it does not use the old liblobby_s anymore

Flamefire:
Ouch... This is unfortunate. I just got the lobby to build and work on all systems. :(
Why is it a different codebase? This makes changes quite hard (e.g. adding something to lobbyPlayer) Can't we have everything on Github and use the same repos? I've seen e.g. you've implemented another Logger and Serializer. I don't think this does any good as improvements should benefit all usages.
An idea: Add the lobbyServer as another repo here (due to libTCP dependency which looks good), remove the server stuff from libLobby, create a static lib `libLobby` that contains common parts (especially stuff that is transferred)  and a `libLobbyClient` (used as currently with the LobbyClient class) and then make the lobbyServer use libLobby. I'd volunteer to try and clean up the dependencies and get it to run with CMake on all build servers, unless you'd like to do this. After this is done, you could switch to using this :)

Flow86:
yea since the new lobby server uses a different factory for messages, and logging and tcp-server (multithreaded!) etc (which I wanted to move the client to too) it is not as easy.
You should have got an email to access my private gitlab where the lobby server code resides

Flamefire:
Yes got it. I looked through the code and so far:

- I think it is possible to use the current libLobby messages also with the server (minor modifications in the messages)
- libLog looks good, but IMO has to many features for this case. If we make libUtils logger threadsafe (very easy) and enhance the interface a bit (kinda easy) we can use that one and get rid of 1 library.
- libTCP is very nice. I think merging libLobby and lobbyServer step by step will make it easier to use it also in the client without a complete rewrite.

Why not move those repos to github and set up CI tests?

Flamefire:
Closing this with the ingame-lobbyChat disabled for now as the lobby server needs to be adjusted.
Best would be if the lobbyPlayer has a flag, if he is ingame or not. However for this we need to find a way to do this without abandoning older clients that do not have this flag. Probably version based serialization? Also other questions need answers. @Flow86

PR #507, opened 2016-07-19
Improve partials detection in coverage data

Flamefire:
Tested some options for generating coverage data. Results:

- gcc has a lot of unrelated branches that I can't get rid off
- Clang seems to produce better branch-coverage data (often perfect) but I've seen some false-positives which I consider even worse (see https://llvm.org/bugs/show_bug.cgi?id=28769)

So for now we can either disable branch detection by omitting the `-b` flag to gcov or live with the fact that we get some false partials. I choose the 2nd option, so one can check the partials and validate manually if they are covered.


Flamefire:
Doesn't help

Flow86:
do you have added -O0 ?

Flamefire:
Yes: https://github.com/Return-To-The-Roots/s25client/pull/507/files#diff-af3b638bc2a3e6c650974192a53c7291L243

See also the following SO question: http://stackoverflow.com/questions/38478916/unexpected-branch-count-in-coverage-data/38479806

PR #523, opened 2016-07-26
Zoom by mouse wheel

jhkl:
This adds zooming by mouse wheel.

Existing fix states for zoom changed into 50%, 75%, 100%(default), 125% and 150%
These can switched by pressing Z (or Shift-Z in reverse order).
Ctrl+Z resets to default
Magnifying glass in top right now shows zoom percentage (except 100%).

### DIFF
--- a/src/gameData/GameConsts.h
+++ b/src/gameData/GameConsts.h
@@ -24,6 +24,13 @@
 /// Geschwindigkeitsabstufungen - Längen der GFs in ms
 const boost::array<unsigned, 6> SUPPRESS_UNUSED SPEED_GF_LENGTHS = {{80, 60, 50, 40, 30, 1}};
 
+const boost::array<float, 5> ZOOM_FACTORS = { 0.5f, 0.75f, 1.f, 1.25f, 1.5f};
+const size_t ZOOM_DEFAULT_INDEX = 2;
+const float ZOOM_ACCELERATION = 0.001f;
+const float ZOOM_WHEEL_INCREMENT = 0.03f;
+const float ZOOM_MIN = 0.1f;
+const float ZOOM_MAX = 3.f;
+
 /// Konvertierungstabelle von Rohstoff-Indizes von den Bergwerken --> Map
 const boost::array<unsigned char, 5> SUPPRESS_UNUSED RESOURCES_MINE_TO_MAP = {{3, 0, 1, 2, 4}};
 
--- a/src/gameData/GameConsts.h
+++ b/src/gameData/GameConsts.h
@@ -24,13 +24,6 @@
 /// Geschwindigkeitsabstufungen - Längen der GFs in ms
 const boost::array<unsigned, 6> SUPPRESS_UNUSED SPEED_GF_LENGTHS = {{80, 60, 50, 40, 30, 1}};
 
-const boost::array<float, 5> ZOOM_FACTORS = { 0.5f, 0.75f, 1.f, 1.25f, 1.5f};
-const size_t ZOOM_DEFAULT_INDEX = 2;
-const float ZOOM_ACCELERATION = 0.001f;
-const float ZOOM_WHEEL_INCREMENT = 0.03f;
-const float ZOOM_MIN = 0.1f;
-const float ZOOM_MAX = 3.f;
-
 /// Konvertierungstabelle von Rohstoff-Indizes von den Bergwerken --> Map
 const boost::array<unsigned char, 5> SUPPRESS_UNUSED RESOURCES_MINE_TO_MAP = {{3, 0, 1, 2, 4}};
 
--- a/src/gameData/GuiConsts.h
+++ b/src/gameData/GuiConsts.h
@@ -20,7 +20,7 @@
 
 #include <boost/array.hpp>
 
-const boost::array<float, 5> ZOOM_FACTORS = { 0.5f, 0.75f, 1.f, 1.25f, 1.5f};
+const boost::array<float, 5> SUPPRESS_UNUSED ZOOM_FACTORS = {{ 0.5f, 0.75f, 1.f, 1.25f, 1.5f}};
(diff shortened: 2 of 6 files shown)


MarcusSt:
There are still some problems :-(

- At 300%, it seems I cannot change the zoom factor in any direction.
- Zooming out once using the wheel will lead to non-stop zooming out until either a crash or I move the wheel in the opposite direction.
- At some zoom levels, it seems to begin shaking: https://youtu.be/hIYhZtJkgqE

src/world/GameWorldView.cpp
```diff
@@ -66,10 +69,54 @@ const GameWorldBase& GameWorldView::GetWorld() const
     return gwv.GetWorld();
 }
 
-void GameWorldView::SetZoomFactor(float zoomFactor)
+void GameWorldView::SetNextZoomFactor()
 {
-    this->zoomFactor = zoomFactor;
+    if (zoomFactor == targetZoomFactor)
```
Flamefire: Don't do this. This are floats. Floats are NEVER equal.

src/world/GameWorldView.cpp
```diff
@@ -66,10 +69,54 @@ const GameWorldBase& GameWorldView::GetWorld() const
...
+        return;
+
+    float zoomDirection = targetZoomFactor - zoomFactor;
+
+    // last step
```
Flamefire: Its cleaner to put this comments into the if blocks

src/world/GameWorldView.cpp
```diff
@@ -66,10 +69,54 @@ const GameWorldBase& GameWorldView::GetWorld() const
...
+        if (zoomDirection > 0)
+            zoomSpeed += ZOOM_ACCELERATION;
+        else
+            zoomSpeed -= ZOOM_ACCELERATION;
+    }
+
+    zoomFactor = zoomFactor + zoomSpeed;
     CalcFxLx();
+
+    // allow overshoot, but swing back
```
Flamefire: I'd disallow this and add this check in SetZoomFactor. Don't think this "swing back" looks nice and it might even annoy users ("doesn't work!!!")

src/world/GameWorldView.cpp
```diff
@@ -66,10 +69,54 @@ const GameWorldBase& GameWorldView::GetWorld() const
...
 }
 
-void GameWorldView::SetZoomFactor(float zoomFactor)
+void GameWorldView::SetNextZoomFactor()
 {
-    this->zoomFactor = zoomFactor;
+    if (zoomFactor == targetZoomFactor)
+        return;
+
+    float zoomDirection = targetZoomFactor - zoomFactor;
```
Flamefire: Maybe better name? e.g. `remainingZoomDiff`

src/world/GameWorldView.cpp
```diff
@@ -66,10 +69,54 @@ const GameWorldBase& GameWorldView::GetWorld() const
...
+    // acceleration to unlimited speed
+    else
+    {
+        if (zoomDirection > 0)
+            zoomSpeed += ZOOM_ACCELERATION;
+        else
+            zoomSpeed -= ZOOM_ACCELERATION;
+    }
+
+    zoomFactor = zoomFactor + zoomSpeed;
```
Flamefire: Maybe move the first if here. Example: targetZoom = zoom + ZOOM_ACCELERATION/2 -> Overswinging. Probably even back and forth.

src/gameData/GameConsts.h
```diff
@@ -24,6 +24,13 @@
 /// Geschwindigkeitsabstufungen - Längen der GFs in ms
 const boost::array<unsigned, 6> SUPPRESS_UNUSED SPEED_GF_LENGTHS = {{80, 60, 50, 40, 30, 1}};
 
+const boost::array<float, 5> ZOOM_FACTORS = { 0.5f, 0.75f, 1.f, 1.25f, 1.5f};
```
Flamefire: These are not game constants. I'd suggest a new file either GuiConsts or ZoomConsts, so GUI and Game are not mixed

src/gameData/GuiConsts.h
```diff
@@ -0,0 +1,31 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with Return To The Roots. If not, see <http://www.gnu.org/licenses/>.
+
+#ifndef GUICONSTS_H_
+#define GUICONSTS_H_
+
+#include <boost/array.hpp>
+
+const boost::array<float, 5> ZOOM_FACTORS = { 0.5f, 0.75f, 1.f, 1.25f, 1.5f};
```
Flamefire: Please add `SUPPRESS_UNUSED` just to be sure...

jhkl: Why?

Flamefire: If you include that in another cpp file that does not use this, you get an unused warning. This does not get generated for C-arrays though.
Oh and you need double-braces  ` ... SUPPRESS_UNUSED ZOOM_FACTORS = {{ ... }};` as the array is a subobject :/ -> https://travis-ci.org/Return-To-The-Roots/s25client/jobs/147873685

Flamefire:
Found a bug: zoomFactor=2, target=1.
It accelerates 1 times to 1 (negative), but then (next update) it brakes back at line  
`if (abs(remainingZoomDiff) <= 0.5 * zoomSpeed * zoomSpeed / ZOOM_ACCELERATION)`
Values: 
zoomSpeed = -0.00100000028
remainingZoomDiff =-0.999853492
ZOOM_ACCELERATION = 0.00100000005, 
0.5 * zoomSpeed*zoomSpeed/ZOOM_ACCELERATION = 0.00050000025657939641
abs(remainingZoomDiff) = 0 (???)

Then it even accelerates and decelerates again and again never really leaving 2

Issue #550, opened 2016-08-04
Sea Exploration Path and Attack problem 


shadowvzs:
Hi, i played on King of The Sea map, there many island, i played vs computer and conquered every island except 3 (bottom-left, bottom-right corner and nearly right-middle).

settings was: minimum scout for exploration, harbor block.

bottom-left island harbour is exploreable with ships but non attackable (even my fortress was full on nearby islands.)

right islands kinda never discovered after hours (and i used 4x speed and around 10 harbour for exploration multiple times...)

.sav 
http://s000.tinyupload.com/?file_id=08932683748380210788

last replay
http://s000.tinyupload.com/?file_id=45776228506641921039

Spikeone:
Tested it an can confirm both bugs.

Flamefire:
Can you tell me a point in the replay, where the bottom-left harbor should be attackable (is explored) and/or attach a savegame from that point?
In the savegame I also see, that the harbor of right-middle was once explored...

Spikeone:
Well, I never thought about that seems like a good idea. I can only guess that the random exploration is a result of the basically random and indirect seafaring. We can not set a target for exploration because navigation was bad => It's also possible to rediscover your own harbors.

But - as I know the seafaring is quite bad if it is the only possible way to attack I agree that we may change it to "ignore own or friendly harbors".

Just asking: random map generator? King of the sea is not a randomly generated map.

Flamefire:
Well random exploration is ok. The scout itself works like that: Get all points in reach that are not fully explored and pick one of them randomly. This makes very much sense as e.g. a geologist would not plant a sign, where one already is. Currently the scout ignores, if the point is visible through a team member, so I'd consider that a bug to be fixed.
Sea exploration works (currently) like that: Get all harbors on same sea and pick one randomly to go to (except current harbor). This is a big bug, as exploration ships will drive to any (even own) harbors which is completely pointless except at the very beginning. This makes exploring unexplored harbors very hard (chance of 1/number of harbors), so this should be fixed.
Seafaring itself is not really random: When you send an expedition into one direction, it will pick the closest, usable harbor in that direction. That is quite good.
Note that neither of this is about attacking. Although you need an exploration expedition (or something else) to reveal the harbor.

[...]

PR #567, opened 2016-08-11
Seafaring fixes

Flamefire:
This fixes the non-attackable harbor issue in #550 by starting the coast-search after the critical NW point, which should be ok for almost all maps (except one really tries to cause this bug)

It also makes explorations (ship and scout) correctly honor visible points so they are not explored again.


Spikeone:
Just asking to make sure: this also works with fow and pacts?

Flamefire:
Yes. Exploration targets are points that are not visible. They are visible when they are visibile for allies and fow is not visible.

PR #577, opened 2016-08-15
More lua functions

Flamefire:
Adds SetMissionGoal and some tests. For the GUI tests we need dummy graphics and a dummy driver. This is also implemented


src/drivers/VideoDriverWrapper.cpp
```diff
@@ -211,7 +213,7 @@ bool VideoDriverWrapper::DestroyScreen()
 bool VideoDriverWrapper::hasExtension(const std::string& extension)
 {
     // Extension mit Leerzeichen gibts nich
-    if( extension.empty() || extension.find(' ') != std::string::npos)
+    if(!isOglEnabled_ || extension.empty() || extension.find(' ') != std::string::npos)
```
Flow86: wouldnt it be better to create some kind of "OpenGLVideoDriver" and wrap up all gl*-Calls to that base class?

Flamefire: Yes I thought about that. I'd also want to move more openGL calls (e.g. drawing rectancles etc) into the videodriverwrapper.
But in the end I decided against a wrapper class for openGL because a) it is a lot of work with all those calls in the whole codebase and b) it creates quite some overhead due to the virtual call per ogl call which I wanted to avoid. Hence: As long as we can run the tests with this solution I'd stay with that. There is more important work to do ;)

Issue #590, opened 2016-09-08
binaries are misplaced


hasufell:
For a regular installation, there are 3 binaries that end up in `RTTR_DATADIR`:
* s25update
* s-c_resample
* sound-convert

This blocks packaging on most distros that have basic QA, since platform dependent files must not be placed in `/usr/share` and similar (which is the default of `RTTR_DATADIR` and rightly so).

At least for `sound-convert` I was able to fix it by just installing it into `RTTR_DRIVERDIR` (had to change `libutil/src/files.h` and `s-c/src/CmakeLists.txt`). I'm not completely sure how the rest are invoked/found, but they should either be placed in `RTTR_DRIVERDIR`, `RTTR_LIBDIR` (within a subdir) or `RTTR_BINDIR`.

Flamefire:
According to the CMakelists they should be placed in `RTTR_LIBDIR` ("Directory for shared binaries")
That defaults to `RTTR_DATADIR` but is never really used except for setting `RTTR_DRIVERDIR`

I'm currently to busy to fix that, so if anyone got the time to implement and test that...?

hasufell:
Well, the question is if there's anything else that need to be changed except the installation destination (like [libutil/src/files.h](https://github.com/Return-To-The-Roots/libutil/blob/0b536749933c002164f960985d583bc5da12e847/src/files.h#L131) for `sound-convert`).

Flamefire:
Hm. IMO s-c_resample is used by sound-convert only so I think it is enough if they are place together.
s25update is referenced from the start script and from within itself. Probably do a quick grep to see if I missed anything.

Mailaender:
```
[  563s] -- Installing: /home/abuild/rpmbuild/BUILDROOT/s25rttr-0.8.1.git+0a8edac1-0.x86_64/usr/lib64/RTTR/s-c_resample
[  563s] -- Installing: /home/abuild/rpmbuild/BUILDROOT/s25rttr-0.8.1.git+0a8edac1-0.x86_64/usr/lib64/RTTR/sound-convert
[  563s] -- Installing: /home/abuild/rpmbuild/BUILDROOT/s25rttr-0.8.1.git+0a8edac1-0.x86_64/usr/lib64/RTTR/s25update
```

Shouldn't those be be placed in `/usr/bin` instead?

Flamefire:
[...]

PR #591, opened 2016-09-09
Use RTTR_LIBDIR for non-core binaries

hasufell:
Thanks to the submodule mess, this depends on:
* https://github.com/Return-To-The-Roots/s-c/pull/10
* https://github.com/Return-To-The-Roots/libutil/pull/19
* https://github.com/Return-To-The-Roots/s25update/pull/8

I'm not 100% sure that behavior is correct when a non-default `RTTR_LIBDIR` is used. The only in-code change I made wrt that is https://github.com/Return-To-The-Roots/libutil/pull/19/files

For `s25update` all the shell scripts (and main.cpp too) have hardcoded paths, so I just left them untouched (as in: they might already be broken).


hasufell:
fixes #590

Flamefire:
Could you test that it works? For the shell scripts, seems like that is a huge problem. Probably we should just turn it into a launcher or make it query the install configuration somehow? How do others do that? Or shall we simply say that autoupdate is only for one-folder installations? Would make it pretty hard for maintainers as old version users can't play with newer version users...

Flow86:
those shell scripts get updated by the old updater correctly, after that, the new updater location would be used - so the shell scripts have to be modified

hasufell:
> those shell scripts get updated by the old updater correctly, after that, the new updater location would be used - so the shell scripts have to be modified

For default installations, the location is unchanged since `RTTR_LIBDIR` defaults to `RTTR_DATADIR`.

> Would make it pretty hard for maintainers as old version users can't play with newer version users...

If you mean package maintainers, then probably not, because `s25update` is not used there. It would circumvent the package manager and most distros don't install it.

hasufell:
> Could you test that it works?

How exactly do I test that sound-convert and s-c_resample work?

Flamefire:
@Flow86 The shell scripts use relative paths assuming a "one-directory-installation". Solution I see: Configure the shell scripts with cmake where the paths are known. But one needs to be carefull with absolute vs relative paths then... However I think for package installs this will not be used at all.

> If you mean package maintainers, then probably not, because s25update is not used there. It would circumvent the package manager and most distros don't install it.

Exactly. So users will be stuck with whatever version the maintainer used and can't play with others using recent versions. My firefox e.g. is always quite some versions behind the latest and that is a popular program. So what would happen to RTTR in the distros? Updated once per year? :-/

For testing: The converter is executed to create the `SOUNDS.LST` in the settings dir (`/* 55 */RTTR_SETTINGSDIR "/LSTS/SOUND.LST", // Die konvertierte sound.lst` from files.h) on startup. Delete it and run the installed program and check if it is recreated and sounds are working.

Flow86:
@Flamefire ah okay now I know what youre talking about. yes those shell scripts should get there paths by cmake, but we have to look to get relative paths correctly

hasufell:
> Exactly. So users will be stuck with whatever version the maintainer used and can't play with others using recent versions. My firefox e.g. is always quite some versions behind the latest and that is a popular program. So what would happen to RTTR in the distros? Updated once per year? :-/

I don't want to start a debate of the pros and cons of distros, packaging and so on, but distros don't and will not package "updaters". Those are for local installs.

If your distro updates only once per year (which is not the general case), you should switch to another distro.

> For testing: The converter is executed to create the SOUNDS.LST in the settings dir (/* 55 */RTTR_SETTINGSDIR "/LSTS/SOUND.LST", // Die konvertierte sound.lst from files.h) on startup. Delete it and run the installed program and check if it is recreated and sounds are working.

Yes, I tested that.

Flamefire:
Merged the submodules. Can you please update this with the correct submodules now so it passes the build?
After merging this I'll pull the files.h into the main repo so we have less dependencies with the paths.

PR #604, opened 2016-10-03
Better support for UTF-8 paths

Flamefire:
Allow foreign chars in filenames


src/ProgramInitHelpers.cpp
```diff
@@ -0,0 +1,43 @@
...
+#endif // _WIN32
+        // Use also the encoding (mostly UTF8) for bfs paths: http://stackoverflow.com/questions/23393870
+        bfs::path::imbue(std::locale());
+    } catch(std::exception& e){
+        std::cerr << "Error initializing your locale setting. ";
+#ifdef _WIN32
+        std::cerr << "Check your system language configuration!";
+#else
+        char* lcAll = getenv("LC_ALL");
+        char* lcLang = getenv("LC_LANG");
```
Flow86: LC_LANG is wrong - standard var is "LANG" only ( I know it was wrong before, already)

src/ProgramInitHelpers.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include <stdexcept>
+
+bool InitLocale()
+{
+    // Check and set locale (avoids errors caused by invalid locales later like #420)
+    try{
+        // Check for errors and use classic locale to avoid e.g. thousand separator in int2string conversions via streams
+#ifdef _WIN32
+        // On windows we want to enforce the encoding (mostly UTF8). Also using "" would use the default which uses "wrong" separators
+        std::locale::global(boost::locale::generator().generate("C"));
```
Flow86: can't we use "C" for both systems, then we dont need a difference?

Flamefire: There was a problem with Linux/OSX which I don't want to trigger again because I don't know if "C" will cause any change compared to "classic". In windows it seems to be the same. But "Never change a running system" applies here ;)

src/ProgramInitHelpers.cpp
Flamefire: Good catch.

Flamefire: Fixed now. And could you disable the notifications from coveralls? They seem to be buggy, although the coverage itself is ok. So only no status report as comments, the commit status can be kept.

Issue #592, opened 2016-09-10
installation fails


hasufell:
This is because the installation is not proper cmake install rules, but gets piped to some shell scripts (I don't even know why). These shell scripts make assumptions about where the build directory is and when used with distro packaging, it can lead to things like:

``` sh
jule@localhost ~/git/s25client-build $ make DESTDIR="/home/jule/git/s25client/install" install
[  0%] Built target version
       version: started
                source directory: "/home/jule/git/s25client/"
                build  directory: "/home/jule/git/s25client-build/"
                build_version_defines.h is unchanged
       version: finished
[  0%] Built target updateversion
[  1%] Built target videodrv
[  1%] Built target audiodrv
[  2%] Built target videoSDL
[  7%] Built target s25util
[  9%] Built target endian
[ 11%] Built target audioSDL
[ 12%] Built target mygettext
[ 22%] Built target siedler2
[ 23%] Built target lobby_c
[ 24%] Built target resample-sndlib-20
[ 25%] Built target s-c_resample
[ 26%] Built target sound-convert
[ 27%] Built target s25update
[ 32%] Built target translations
[ 96%] Built target s25Main
[ 96%] Built target s25client
[100%] Built target Test
Install the project...
-- Install configuration: ""
-- Running preinstall script
## Installing for "Linux"
## Using Path-Prefix "/usr/local"
## Using Binary Dir "/usr/local/bin"
## Using Data Dir "/usr/local/share/s25rttr"
## Using Library Dir "/usr/local/share/s25rttr"
## Using Destination Dir "/home/jule/git/s25client/install/"
## Creating directories
mkdir: created directory '/home/jule/git/s25client/install'
mkdir: created directory '/home/jule/git/s25client/install//usr'
mkdir: created directory '/home/jule/git/s25client/install//usr/local'
mkdir: created directory '/home/jule/git/s25client/install//usr/local/bin'
mkdir: created directory '/home/jule/git/s25client/install//usr/local/share'
mkdir: created directory '/home/jule/git/s25client/install//usr/local/share/s25rttr'
mkdir: created directory '/home/jule/git/s25client/install//usr/local/share/s25rttr/S2'
mkdir: created directory '/home/jule/git/s25client/install//usr/local/share/s25rttr/RTTR'
mkdir: created directory '/home/jule/git/s25client/install//usr/local/share/s25rttr/driver'
[...]

PR #605, opened 2016-10-08
Rework install system

### LINKED ISSUE #590: binaries are misplaced

Flamefire:
This moves as much as possible from shell scripts into CMake and uses the correct CMake install commands and destination directories.

Will fix #590 and #592 

This is a WIP. So please provide some input ( @hasufell @Flow86 )

Main changes:

- Store only relative paths in RTTR_*DIR variables and find the prefix at runtime from RTTR_BINDIR variable and the executables path. This is based on the ideas of http://doc.aldebaran.com/libqi/guide/cxx-path.html
- Remove RTTR_PREFIX due to above change
- Use GNU default dirs as the default for the directories
- Move preinstall.sh commands into cmake -> no more preinstall.sh
- Rename postinstall.sh to prepareRelease.sh and call it manually in create_release.sh
- Move bugle and start.sh out of the build dir and "configure" them by cmake. Goal is to make it possible to build rttr in any other folder


CMakeLists.txt
```diff
@@ -393,55 +384,65 @@ ADD_CUSTOM_TARGET(updateversion ALL
...
 ################################################################################
+# Install
+
+# Start script
+IF ("${CMAKE_SYSTEM_NAME}" STREQUAL "Windows")
+	install(PROGRAMS "release/bin/rttr.bat" DESTINATION "${RTTR_BINDIR}")
+ELSEIF("${CMAKE_SYSTEM_NAME}" STREQUAL "Linux" OR "${CMAKE_SYSTEM_NAME}" STREQUAL "FreeBSD")
+	install(PROGRAMS "release/bin/rttr.sh" DESTINATION "${RTTR_BINDIR}")
+ELSEIF(NOT "${CMAKE_SYSTEM_NAME}" STREQUAL "Darwin")
+	message(FATAL_ERROR "${CMAKE_SYSTEM_NAME} not supported")
```
Flow86: thats wrong. darwin IS supported

Flamefire: Right. Tis is why there is a `NOT`.

Flamefire:
Ok, I think I'm done so this can be merged. Please double check or tell me, if this needs more changes or time.
I tested: MSVC compilation, test, execution, linux x64 compilation test, execution, and `make create_nightly` on linux x64

hasufell:
Doesn't work as expected:
```
[ 26%] Linking CXX executable s25update
'/home/hasufell/git/s25client/build/s25update/src/s25update' -> '/home/hasufell/git/s25client/build/lib64/RTTR/'
cp: cannot create regular file '/home/hasufell/git/s25client/build/lib64/RTTR/': No such file or directory
make[2]: *** [s25update/src/CMakeFiles/s25update.dir/build.make:152: s25update/src/s25update] Error 1
make[2]: *** Deleting file 's25update/src/s25update'
make[1]: *** [CMakeFiles/Makefile2:1268: s25update/src/CMakeFiles/s25update.dir/all] Error 2
make: *** [Makefile:139: all] Error 2
```

This is after doing it the standard cmake way:
```
mkdir <build-dir>
cd <build-dir>
cmake <source-dir>
make
```

Flamefire:
This gets now created.

CMakeLists.txt
```diff
@@ -159,74 +161,60 @@ ELSE()
...
+	# Windows uses one directory only
+	SET(RTTR_BINDIR "." CACHE INTERNAL "")
+	SET(RTTR_DATADIR "." CACHE INTERNAL "")
+	SET(RTTR_GAMEDIR "." CACHE INTERNAL "")
+	SET(RTTR_LIBDIR "." CACHE INTERNAL "")
+	SET(RTTR_DOCDIR "." CACHE INTERNAL "")
+	SET(RTTR_DRIVERDIR "./driver" CACHE INTERNAL "")
+ELSE()
+	include(GNUInstallDirs)
+	SET(RTTR_BINDIR "${CMAKE_INSTALL_BINDIR}" CACHE STRING "Directory for the binaries")
```
hasufell: These should not be cachable. Users are now able to set `CMAKE_INSTALL_BINDIR` and `RTTR_BINDIR` independently, which doesn't really make sense. Only the GNUInstallDirs variables should be exposed, everything else not, imo. So if you use RTTR_DATADIR for the preprocesser, you don't want to cache and reverse-overwrite it, instead you want something like:

```
SET(RTTR_DATADIR "${CMAKE_INSTALL_FULL_DATADIR}/s25rttr/S2")
```
which ensures it's a proper full path.

Flamefire: True. But the full story is quite complicated and this was the best solution I could come up with.
What we want: Install (and run) Rttr either with CMake-Install, from a zip, from a package or inside the build directory. Hence the description of this PR: The RTTR_*DIR variables contain relative paths. Given the path of the executable one can then (using RTTR_BINDIR) find the install dir and from there the data dir. If we would store e.g. `RTTR_DATADIR` as an absolute path then portable installations and runs from inside the build dir won't be possible. It is still not perfect (there is a case where the executable path can't be found yet, see #616, but I'm working on that and will have a fix soon) but I can't think of a better solution that is able to cope with all use cases. I'm also missing experience with how other programs handle such cases.
So all in all I opted for using the RTTR_*DIR variables as THE gold values and only initialize them from the GNUInstallDirs as default values to use sensible default values.

hasufell:
> I'm also missing experience with how other programs handle such cases.

You'd add a cmake variable to allow running the program in-place (portable mode).

Flamefire:
So basically store absolute paths in the program itself unless it is compiled in portable mode? 
Then it won't run from the build dir or when using DESTDIR
Imo the current RTTR variables make everything possible already. You (probably) could set them to absolute paths if you want and you have uniform variable names over all OS (Gnu dirs don't make much sense in Windows) 
I'd really like to see another solution that meets all requirements but it got to be complete (CMake and in the program)

PR #606, opened 2016-11-15
Random map generation

vader1986:
Hey guys, can you do a rough review to point out major problems, please? That way I get a better understanding of the scope of adaptions required to contribute to the project. 

Following things I've got in mind:
- i haven't been programming c++ for a while, so please let me know where i can write stuff easier
- i copy pasted a little code from map editor (& doing bit cleanup). Affected classes/structs are "Map" and "MapWriter". Do I need to add a notes to the file? 
- do i meet the coding standards? do you do unit-testing?
- also I'm not sure "master" is the correct target branch?

Also this is work in progress. Right now, it only adds a "Random" button to the map selection. I'm not sure if this is sufficient as a first contribution or must be extended (by configuration possiblilities for the user, such as map-size, number of players, ...)?

I know a lot of stuff looks rather messy, so feel free to complain :-)


vader1986:
Ah nice, i'll check the CI stuff first :-)

src/mapGenerator/Defines.h
```diff
@@ -0,0 +1,59 @@
...
+#define TEXTURE_MEADOW3_HARBOUR        0x4A
+#define TEXTURE_STEPPE_MEADOW2         0x0E
+#define TEXTURE_STEPPE_MEADOW2_HARBOUR 0x4E
+#define TEXTURE_FLOWER                 0x0F
+#define TEXTURE_FLOWER_HARBOUR         0x4F
+#define TEXTURE_LAVA                   0x10
+#define TEXTURE_COLOR                  0x11
+#define TEXTURE_WATER__                0x13
+#define TEXTURE_STEPPE__               0x80
+#define TEXTURE_STEPPE___              0x84
```
Flow86: why not using enums? I think there could already be such a list in the rest of the code?

src/mapGenerator/GreenlandGenerator.cpp
```diff
@@ -0,0 +1,415 @@
...
+#include <cstdlib>
+#include <boost/range/algorithm.hpp>
+
+#include "mapGenerator/Defines.h"
+#include "mapGenerator/GreenlandGenerator.h"
+#include "mapGenerator/VertexUtility.h"
+#include "mapGenerator/ObjectGenerator.h"
+
+#ifndef MIN
+#define MIN(x, y) x < y ? x : y
```
Flow86: why not using std::min / std::max - macros are mostly dangerous and should not be used

src/mapGenerator/ObjectGenerator.h
```diff
@@ -0,0 +1,137 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Return To The Roots. If not, see <http://www.gnu.org/licenses/>.
+
+#ifndef ObjectGenerator_h__
+#define ObjectGenerator_h__
+
+#ifndef IntPair
+#define IntPair std::pair<uint8_t, uint8_t>
```
Flow86: please use a typedef not defines

src/mapGenerator/MapWriter.cpp
```diff
@@ -0,0 +1,284 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Return To The Roots. If not, see <http://www.gnu.org/licenses/>.
+
+#include <iostream>
+#include "mapGenerator/MapWriter.h"
+
+bool MapWriter::Write(const std::string& filePath, Map* map)
```
Flow86: there's already the libsiedler2 library for loading/writing files - if the writeMap is missing, you should implement it there. avoids duplicate code

Flow86:
Hi,

yes we've (exactly @Flamefire ) added unittesting, so please try to create some tests for your code.

Code style looks good (some annotations above)

Perhaps you could even split some commits into different PRs ( i.e 70b17db )

before merge, you should also squash the PR (so its only one large commit, not 1000 single ones)

src/mapGenerator/Generator.cpp
```diff
@@ -0,0 +1,219 @@
...
+#include <cstdlib>
+#include <stdexcept>
+#include "mapGenerator/Generator.h"
+#include "mapGenerator/MapWriter.h"
+#include "mapGenerator/Defines.h"
+#include "mapGenerator/VertexUtility.h"
+#include "mapGenerator/ObjectGenerator.h"
+
+#ifndef PI
+#define PI 3.14159265
```
vader1986: thx

Flow86:
oh also we have a portable random number generator which supports "seeding" - so the map file should be at least consists of the seed, so others (or you) can recreate the map too - that would be nice

src/mapGenerator/GreenlandGenerator.cpp
vader1986: I used it before, then saw some error message about "min"/"max" not being in the std namespace. Just noticed that probably only an include was missing. I'll change it.

src/mapGenerator/MapWriter.cpp
vader1986: Yeah, I've been looking through the code of the "libsiedler2/src" folder but couldn't figure out how to use it. That's why I had a look into the map editor code which was easier to understand. Initially, I actually wanted to use the save game format to make full use the existing code, but couldn't figure out a good way to do it. Is there any specific place (or person) for code related questions?

vader1986:
OK, a portable random number generator would be useful. I'll check if I can find it in the code. 

For the PR, the commit 70b17db was a mistake (i'm using git-flow and actually merged it to my develop branch). I can squash the commits once i'm done, right? Or is it annoying when I keep doing single commits for now (the CI seems to get triggered each time)?

Flamefire:
I actually would not mind single commits. The Merge is already a kind of Squash and this way one can later see what went in and possibly why.
Don't worry about the CI and many commits. The CI is triggered for every push you do and this is fine. This is what it is for.
The Random generator is the RANDOM singleton. Just look for it.

src/mapGenerator/MapWriter.cpp
Flamefire: Basically: The map consists of layers. The header is one of them. So you'd want to add all the layers. The map itself is (IRC) an archive of instances of those layers. So you'd need to add the layers in the correct order (IMO simply: map header and some raw data entries) and fill them. Take a look at the map loading from that archive. Full map creating might not be implemented yet, this means it should work, but could be a bit more explicit work than it should

src/mapGenerator/Defines.h
Flamefire: Probably is. Look for TerrainTypes. Maybe you'd need a converter though from RTTR types to S2 types (the other way round definitely exists)

vader1986: Yeah, I found the TerrainTypes now, too, so I'm going to use them :-) 

Must really look like i only copy pasted code together. Give me some time and I'll try to get more into the code base as it is. My vacation start at 9th of December, then I get lots of time to fix this. I've been more focused on getting nice maps generated, cause my brother has been pushing me since he knows this project. Finally I could show something hehe.

vader1986:
I got one question about the random number generator. 
What's the "obj_id" in the "int Rand(__FILE__, __LINE__, obj_id, max)" method? 

Is it used when you call the Rand-function multiple times from the same line of code? I tried to integrate the method properly and replace "rand()" but somehow get weird results. So I think I misunderstand how it is actually working.

vader1986:
Another questions about unit tests:
1. can/how can i use subfolders for unit tests (subfolders or src/test/ i mean)? Probably needs adaptions of the CMakeLists-file, right? 
2. could be give me an example how to use RANDOM number generator with seed? Would be useful for unit-testing

vader1986:
@Flamefire @Flow86 i updated the map generation to use the classes from libsiedler2 now (and did some refactoring and UI to enable map configuration). Does anyone of you have a good example how to use the RANDOM class? Right now, I'm using rand() everywhere, because RANDOM was giving strange results. I could also use rand() in combination with srand() to enable seeding. 

Apart from the random class, I could probably write some more unit tests (for that I've been splitting classes up a bit more).

Flamefire:
`rand` might not result in the same values over all systems. For examples of `RANDOM` just search the code. There are direct usages (it registers source line for async debugging, so it looks a bit different) and also functor usages for e.g. `random_shuffle`

Can't really help ATM as I'm overseas

vader1986:
Well, thanks for the reply. The class works ok for very small maps (64x64) - but for large maps the random number sequences are repeating which leads to repeating map-features (which are supposed to be "random"). Not sure how to fix this.

vader1986:
Hey guys ( @Flow86 ),

I think I'm more or less through with the unit tests right now. Some tests are missing, because I couldn't get the random class work properly (with seeding). For many calls the Random-class seems to return the same series of numbers, therefore, for small maps (64x64) it seems to be ok, but for large maps it's absolutely not working well. That's why I still use "rand()" in two places. Once I understand how I Random-class works properly I can replace those two lines, if that's ok for you.  Then I can also cover the functions with unit tests which are currently relying on random values (which is not a good idea for testing I think). 

For further improvements I'd like to create separate pull-requests to avoid huge reviews in the future. For now, are there any major concerns left? The changes shouldn't affect the gameplay at all, I hope. It's just writing out a "Random.swd" and selecting it when pressing the "Random" button, but the game itself is not affected.

src/test/MapUtilityTest.cpp
```diff
@@ -0,0 +1,392 @@
...
+ */
+BOOST_FIXTURE_TEST_CASE(SetHarbor_HarborPlaceAvailable, MapUtility)
+{
+    Map* map = new Map(16, 16, "map", "author");
+    
+    for (int i = 0; i < map->width * map->height; i++)
+    {
+        if (i % map->width < map->width / 2)
+        {
+            map->textureLsd[i] = 0x08; // half of the map meadow
```
Flow86: why not use(/create) constants for the values somewhere (libsiedler2?)

vader1986:
Cool, totally forgot about this :-)
Added the gold/iron/coal/granite distribution to UI!

vader1986:
Ok, should work now (only there seems to be a problem executing the last CI job).

vader1986:
Looks like the CI jobs are running through now

src/mapGenerator/MapUtility.cpp
```diff
@@ -0,0 +1,247 @@
...
+        map->z[*it] = std::max((unsigned char)(z - d), (unsigned char)map->z[*it]);
+    }
+}
+
+unsigned int MapUtility::ComputeWaterSize(Map* map, const Vec2& position, const unsigned int max)
+{
+    const int width = map->width;
+    const int height = map->height;
+    
+    std::queue<Vec2> searchRoom;
```
vader1986: Naming is stupid. Should be searchSpace rather than room.

src/mapGenerator/MapUtility.cpp
```diff
@@ -0,0 +1,247 @@
...
+    std::vector<int> neighbors = VertexUtility::GetNeighbors(center.x, center.y, map->width, map->height, z);
+    for (std::vector<int>::iterator it = neighbors.begin(); it != neighbors.end(); ++it)
+    {
+        const int x2 = *it % map->width, y2 = *it / map->width;
+        const double d = VertexUtility::Distance(center.x, center.y, x2, y2, map->width, map->height);
+        map->z[*it] = std::max((unsigned char)(z - d), (unsigned char)map->z[*it]);
+    }
+}
+
+unsigned int MapUtility::ComputeWaterSize(Map* map, const Vec2& position, const unsigned int max)
```
vader1986: Method could be more general, like "GetBodySize(Map* map, TerrainType terrain, Vec2 pos, int max)".

vader1986:
@Flow86 Can you please check again?

vader1986:
@Flamefire @Flow86 can you give me a quick hint which changes you still consider required? I'll be relocating soon and won't have much time to do further adaptions anytime soon. If there's already a base implementation someone else can join with further pull requests to improve the first implementation.

Flamefire:
I'm overseas atm so can't check code. Could you please post 1-2 Screenshots of the ui? 
Besides that a note: I'd like to  switch to typed enums (boost/c++11) instead of prefixed ones. I've seen you changed it the other way round in one commit. 
If the ui looks good than I'd merge that :)

vader1986:
Hi @Flamefire ,
Thanks for the reply. I'll change back to typed enums (values all upper case, right?) next weekend. I attached two screenshots of the new UI. When pressing the "Random" button, a new map is generated with the current settings (see second screenshot) and directly selected in the table. The settings for map generation can be opened using the "..." button next to the "Random" button (see first screenshot).

screen shot 2017-02-07 at 6 18 53 pm
screen shot 2017-02-07 at 6 18 58 pm

src/desktops/dskSelectMap.cpp
```diff
@@ -91,6 +109,10 @@ dskSelectMap::dskSelectMap(const CreateServerInfo& csi)
     AddTextButton(4, 590, 530, 200, 22, TC_GREEN2, _("Load game..."), NormalFont);
     // "Weiter"
     AddTextButton(5, 590, 560, 200, 22, TC_GREEN2, _("Continue"), NormalFont);
+    // random map generation
+    AddTextButton(6, 380, 530, 150, 22, TC_GREEN2, _("Random"), NormalFont);
+    // random map players
+    AddTextButton(7, 540, 530, 40, 22, TC_GREEN2, _("..."), NormalFont);
```
Flamefire: You don't really need this. Show it when generating the map (Click on "Random")

src/desktops/dskSelectMap.cpp
```diff
@@ -91,6 +109,10 @@ dskSelectMap::dskSelectMap(const CreateServerInfo& csi)
     AddTextButton(4, 590, 530, 200, 22, TC_GREEN2, _("Load game..."), NormalFont);
     // "Weiter"
     AddTextButton(5, 590, 560, 200, 22, TC_GREEN2, _("Continue"), NormalFont);
+    // random map generation
+    AddTextButton(6, 380, 530, 150, 22, TC_GREEN2, _("Random"), NormalFont);
```
Flamefire: Name it more explicit "Create Random Map" or so. Even better (if you got enough time) Put "Random" on the left menu and generate like 5 random maps or so, so the user can choose and make a "Regenerate" button on the bottom that is only shown when "Random" is selected. Might be some more work and could also be done later. Just an idea...

src/gameData/const_gui_ids.h
```diff
@@ -63,7 +63,8 @@ enum GUI_ID
     CGI_MISSION_STATEMENT,
     CGI_MAP_DEBUG,
     CGI_AI_DEBUG,
-    CGI_NEXT = CGI_MAP_DEBUG + 40
+    CGI_NEXT = CGI_MAP_DEBUG + 40,
+    CGI_MAP_GENERATOR
```
Flamefire: This needs to be before NEXT and NEXT needs to be updated (someone forgot it already)

src/ingameWindows/iwMapGenerator.cpp
```diff
@@ -0,0 +1,313 @@
...
+iwMapGenerator::iwMapGenerator(MapSettings& settings) : IngameWindow(CGI_MAP_GENERATOR,
+                                                IngameWindow::posLastOrCenter,
+                                                250, 400, _("Map Generator"),
+                                                LOADER.GetImageN("resource", 41), true, false), mapSettings(settings), tmpSettings(settings)
+{
+    AddTextButton(0,  20, 360, 100, 20, TC_RED2, _("Back"), NormalFont);
+    AddTextButton(1, 130, 360, 100, 20, TC_GREEN2, _("Apply"), NormalFont);
+
+    ctrlComboBox* combo = AddComboBox(2, 20, 30, 210, 20, TC_GREY, NormalFont, 100);
+    combo->AddString("2 Players");
```
Flamefire: This can be a loop and also internationalized. Look for `_(...)` and boost::format (already used)

src/ingameWindows/iwMapGenerator.cpp
```diff
@@ -0,0 +1,313 @@
...
+    combo->AddString("3 Players");
+    combo->AddString("4 Players");
+    combo->AddString("5 Players");
+    combo->AddString("6 Players");
+    combo->AddString("7 Players");
+    
+    combo = AddComboBox(3, 20, 60, 210, 20, TC_GREY, NormalFont, 100);
+    combo->AddString("Islands");
```
Flamefire: Localize those too (add `_(...)`)

src/ingameWindows/iwMapGenerator.cpp
```diff
@@ -0,0 +1,313 @@
...
+    combo = AddComboBox(4, 20, 90, 210, 20, TC_GREY, NormalFont, 100);
+    combo->AddString("64 x 64");
+    combo->AddString("128 x 128");
+    combo->AddString("256 x 256");
+    combo->AddString("512 x 512");
+    combo->AddString("1024 x 1024");
+
+    AddText(5, 20, 120, _("Player Distribution"), COLOR_YELLOW, 0, NormalFont);
+    combo = AddComboBox(6, 20, 140, 210, 20, TC_GREY, NormalFont, 100);
+    combo->AddString("Very Close");
```
Flamefire: Localize

src/ingameWindows/iwMapGenerator.cpp
```diff
@@ -0,0 +1,313 @@
...
+    combo->AddString("Close");
+    combo->AddString("Medium");
+    combo->AddString("Far");
+    combo->AddString("Very Far");
+
+    AddText(7, 20, 170, _("Landscape"), COLOR_YELLOW, 0, NormalFont);
+    combo = AddComboBox(8, 20, 190, 210, 20, TC_GREY, NormalFont, 100);
+    combo->AddString("Greenland");
```
Flamefire: Localize

src/ingameWindows/iwMapGenerator.cpp
```diff
@@ -0,0 +1,313 @@
...
+    combo->AddString("Winterworld");
+    combo->AddString("Wasteland");
+    
+    AddText(9, 20, 225, _("Gold:"), COLOR_YELLOW, 0, NormalFont);
+    ctrlProgress* progressBar = AddProgress(10, 100, 220, 130, 20,
+                                            TC_GREY, 139, 138, 100);
+    progressBar->SetPosition(settings.ratioGold);
```
Flamefire: Move to `Reset` (see below)

src/ingameWindows/iwMapGenerator.cpp
```diff
@@ -0,0 +1,313 @@
...
+            break;
+
+        case 0: // back
+        {
+            Close();
+        } break;
+            
+        case 1: // apply
+        {
+            tmpSettings.ratioGold = GetCtrl<ctrlProgress>(10)->GetPosition();
```
Flamefire: Why not get rid of tmpSettings and query all the values here? No need to react on the change events.

src/ingameWindows/iwMapGenerator.cpp
```diff
@@ -0,0 +1,313 @@
...
+            tmpSettings.ratioGranite = GetCtrl<ctrlProgress>(16)->GetPosition();
+            mapSettings = tmpSettings;
+            Close();
+        } break;
+    }
+}
+
+void iwMapGenerator::Msg_ComboSelectItem(const unsigned int ctrl_id, const int selection)
+{
+    // combo-box: number of players
```
Flamefire: General Note: I started using file-local enums for the ctrl-ids which makes it much more easier to read and does not require comments: `if(ctrl_id == CTRL_PLAYER_NUM)`
Comments should inside the if for the else-ifs so no space between the closing brace and the `else if`

src/ingameWindows/iwMapGenerator.cpp
```diff
@@ -0,0 +1,313 @@
...
+    if (ctrl_id == 2)
+    {
+        switch (selection)
+        {
+            case 0: tmpSettings.players = 2; break;
```
Flamefire: New line after every semicolon. Can use newline and intendation after the case

src/ingameWindows/iwMapGenerator.cpp
```diff
@@ -0,0 +1,313 @@
...
+                break;
+            case 2:
+                tmpSettings.type = LT_WASTELAND;
+                break;
+            default: break;
+        }
+    }
+}
+
+void iwMapGenerator::Reset()
```
Flamefire: Also reset the Coal ratios here instead of ctor

src/ingameWindows/iwMapGenerator.cpp
```diff
@@ -0,0 +1,313 @@
...
+    if (tmpSettings.minPlayerRadius <= 0.2)
+    {
+        combo->SetSelection(0);
+    }
+    else
+    if (tmpSettings.minPlayerRadius <= 0.3)
+    {
+        combo->SetSelection(1);
+    }
+    else
```
Flamefire: `else if` and no braces required (but possible)

src/test/AreaDescTest.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+BOOST_AUTO_TEST_SUITE(MapTest)
+
+/**
+ * Tests the AreaDesc::IsInArea method for a point inside of the area independent of 
+ * the player positions.
+ */
+BOOST_FIXTURE_TEST_CASE(IsInArea_PointInsideIgnorePlayer, AreaDesc)
+{
+    AreaDesc* area = new AreaDesc(0.5, 0.5, 0.0, 0.2, 100.0, 0, 0, 10, 10, 0);
```
Flamefire: Why pointers? Just use plain instances please

src/test/AreaDescTest.cpp
```diff
@@ -0,0 +1,94 @@
...
+    const bool isInArea = area->IsInArea(8, 8, 0.0, 16, 16);
+    
+    BOOST_REQUIRE(isInArea);
```
Flamefire: Does not need a variable (but is ok here)

src/test/MapTest.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * Tests the default constructor of the Map class. Width and height of the map
+ * are expected to be zero.
+ */
+BOOST_FIXTURE_TEST_CASE(Constructor_DefaultZeroSize, Map)
+{
+    Map* map = new Map();
+    
+    BOOST_REQUIRE_EQUAL(map->width,  (unsigned)0);
```
Flamefire: Don't use `(unsigned)0`. Use `0u`
Same below

src/test/MapTest.cpp
```diff
@@ -0,0 +1,111 @@
...
+}
+
+/**
+ * Tests the constructor of the Map class. The name must match the name specified
+ * in the constructor arguments.
+ */
+BOOST_FIXTURE_TEST_CASE(Constructor_CorrectName, Map)
+{
+    std::string name("name");
+    Map* map = new Map(64, 64, name, "author");
```
Flamefire: Use instances, no pointers. And use at least 2 different authors. Like: MapA Author1, MapB Author2 After that require equal on each instance' author

src/test/VertexUtilityTest.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+/**
+ * Tests the VertexUtility::GetIndexOf method for a position inside of the bounds given by
+ * width and height. The resulting index must match the scheme: index = y * width + x.
+ */
+BOOST_FIXTURE_TEST_CASE(GetIndexOf_InsideOfBounds, VertexUtility)
+{
+    const int x = 4;
+    const int y = 2;
+    const int width = x * x;
```
Flamefire: Pretty special case for the sizes... Might trigger an unlucky coincidence (especially as `y*y==x`)

src/test/VertexUtilityTest.cpp
```diff
@@ -0,0 +1,146 @@
...
+    const int height = y * y;
+    const int index = VertexUtility::GetIndexOf(x, y, width, height);
+    
+    BOOST_REQUIRE_EQUAL(index, y * width + x);
```
Flamefire: Check at least 2 indices. You can combine the index calculation and IndexOf into the check if you like. Would be more explicit

src/desktops/dskSelectMap.cpp
vader1986: My thought behind this was that I want to click only once for regenerating the map. If I always open the settings it takes longer to find the map I like to play. Therefore, I split up the generation and the settings into two separate buttons. Just give it a try, you'll see it is more comfortable this way ;)

src/desktops/dskSelectMap.cpp
vader1986: Actually, that's also a nice idea.

vader1986:
Thanks for the review. I think I implemented most of the changes.
What's left: 
- change the structure with the "Random" and "..." buttons
- switch to typed enums (boost/c++11) instead of prefixed ones

The algorithm is implemented in a pretty messy way, but good enough for a starting point. I made some effort on improvement for that in a separate branch, but I don't think I'll complete it anytime soon. But might be an interesting starting point for someone else. After some refactoring I also noticed there might be a lot of stuff potentially reusable for an in-game editor tool. 

Also I changed the unit test structure a bit on a separate branch to be able to create a sub-folder structure.

vader1986:
Hi, thanks for the review, got no problems with that. Just requires some time as I'm currently working about 60h/w while searching for new job in different country. Therefore, my plan is having some free time in April, hope I can implement the requested changes then.

Of course I appreciate the review, you're spending your free time here in the same way I do. Only difference, I learn something ;)

src/desktops/dskSelectMap.cpp
```diff
@@ -91,6 +96,10 @@ dskSelectMap::dskSelectMap(const CreateServerInfo& csi)
     AddTextButton(4, 590, 530, 200, 22, TC_GREEN2, _("Load game..."), NormalFont);
     // "Weiter"
     AddTextButton(5, 590, 560, 200, 22, TC_GREEN2, _("Continue"), NormalFont);
+    // random map generation
+    AddTextButton(6, 380, 530, 150, 22, TC_GREEN2, _("Create Random"), NormalFont);
```
vader1986: "Create random map" is a bit long for the size of the button, therefore, i put in "Create Random". That it creates a map is obvious for the user I think.

src/desktops/dskSelectMap.cpp
```diff
@@ -91,6 +96,10 @@ dskSelectMap::dskSelectMap(const CreateServerInfo& csi)
...
+    // random map settings
+    AddTextButton(7, 540, 530, 40, 22, TC_GREEN2, _("..."), NormalFont);
```
vader1986: This one I would like to keep because it allows the user to configure the map independently of the time-consuming generation itself.

src/desktops/dskSelectMap.cpp
Flow86: perhaps simply "Random Map"?

src/desktops/dskSelectMap.cpp
Flow86: Yes but thats kind of confusing for the user - this should perhaps be in "Settings" (like the default addon stuff)

src/gameData/TerrainData.cpp
```diff
@@ -19,6 +19,36 @@
...
+        case TT_WATER_NOSHIP:       return 0x06;
+        case TT_BUILDABLE_MOUNTAIN: return 0x22;
+        case TT_BUILDABLE_WATER:    return 0x13;
+        case TT_LAVA:               return 0x10;
+        case TT_LAVA2:              return 0x14;
+        case TT_LAVA3:              return 0x15;
+        case TT_LAVA4:              return 0x16;
+    }
+    throw std::logic_error("Invalid parameters given");
+}
```
Flow86: aren't those TerrainType-enum exactly defined as these values? then you should use 

```return uint8_t(t)```

File: src/mapGenerator/ObjectGenerator.h [hunk 31]
```diff
@@ -0,0 +1,175 @@
...
+    private:
+    
+    /**
+     * Generates a random number.
+     * @param max maximum value
+     * @return a new random number
+     */
+    static int Rand(const int max)
+    {
+        return rand() % max;
```
Flow86: this is still not portable - so with the same seed, you will get different results on different platforms

src/mapGenerator/RandomConfig.h
```diff
@@ -0,0 +1,120 @@
...
+    
+    /**
+     * Generates a random number between min and max.
+     * @param min minimum value
+     * @param max maximum value
+     * @return a new random number
+     */
+    static int Rand(const int min, const int max)
+    {
+        return min + rand() % (max - min);
```
Flow86: another rand-function?

src/desktops/dskSelectMap.cpp
Spikeone: But you can set basic addon settings in the options menu and still change it when hosting. 
Just think of the following: you found some players and instead of just hosting and changing some settings for the random map you first have to leave the host menu, change options, rehost, damn forgot something... Etc. 
But thats just something I mind

vader1986: Ah ok, now I get it. I'll check how the things with the add-ons work as soon as I get some more time to spend here. Since there're quite a bunch of settings, I'd probably have to include another category for random map generation.

vader1986: Long-term it would be nice to have the map generated once the game starts (so people don't see the mini-map before). That was my initial thought, but I couldn't figure out which pieces of code to touch for this (seem to be quite time consuming to get that far), so I started with this "light-weight" approach which can be implemented most independently existing code.

src/gameData/TerrainData.cpp
vader1986: i'll double-check this, probably true

src/mapGenerator/ObjectGenerator.h
vader1986: The problem with the existing portable generator is the bad "randomness" of generated numbers for a large amount (which is required for map generation). Also, the way the map generation is implemented, it doesn't really matter if it is portable or not (apart from testability), because only the host is generating the map before the game starts.

src/mapGenerator/RandomConfig.h
vader1986: indeed, the one before is useless, i'll remove it

src/gameData/TerrainData.cpp
vader1986: Looks like they're not

vader1986:
> For tests use different values for e.g. x/y coordinates or width/height to detect parameter swapping.
> If you make a param a pointer, than you allow NULL and may also pass ownership to the object. In
> your case you never want that. Use references (const where possible) instead.

I deleted all param pointers. Don't really know why I put them there in the first place. 

> const on integrals like int is not required. There would be to many of them if we put that everywhere.

I deleted the const in all places now (I hope). 

> Try to make everything as local as possible. So implementations to cpp files (exceptions are simple 
> getters/setters, but they could be in cpps too), no using namespace in headers, types only used in 
> the cpp (like the ctrl-enum) can be put in the cpp and even in an anonymous namespace: 
> namespace{ enum{CTRL_1, CTRL_2}; } and define variables in the functions that need them, if they 
> are not used outside.

I restructured some stuff in order to get things more local (e.g. moving anonymous enum to cpp, removed some members, ...).

> Last: I tried to switch to compound types for e.g. points throughout the code. This makes calls much
> easier to read: calcDistance(pointA, pointB, maxDist) instead of calcDistance(xA, xB, yA, yB,
> maxDist) or the like. Even worse for values: calcDistance(5, 2, 7, 4, 7) -> "What is what???"

Yeah, I changed that now. Initially I had a lot of compiling-related issues due to my lack of knowledge on C++. I tried to use more of the existing code now.

ToDos:

- use different x/y/width/height parameters for unit tests
- move the random map settings to addOns

General notes from my side: 

- There's still one thing missing, that's the random-function. Not sure there's any way to make the portable RANDOM-class more "random", rather than repeating its values after a while. Just replacing the native rand() function by the portable RANDOM class leads to very weird maps.
- I think there are some conceptional problems in the random map implementation which are a bit UI related. In general, it would be nice it the random map is generated once all players are ready and the game starts, so that no player knows the map before. However, this is really tough to achieve right now. For example, in the multiplayer menu, there could be a check box "Random Map" which disables the map selection and displays a blank image for the mini-map preview.

Flamefire:
The random function is flawed. There are better alternatives for prngs. I wanted to change that but that breakes replays and possibly save games so I didn't do it yet. 
I wouldn't move the settings to addons. The map selection screen looks like the right place.

src/ingameWindows/iwMapGenerator.cpp
```diff
@@ -0,0 +1,320 @@
...
+#include "gameData/MaxPlayers.h"
+#include "helpers/containerUtils.h"
+#include "libutil/src/colors.h"
+
+#include <boost/format.hpp>
+#include <string>
+
+enum
+{
+    CTRL_PLAYER_NUMBER = 2,
```
Flamefire: Tipp: don't set values and use the enum for all controls. The compiler can choose values :) no need to change this here though just for the future

src/ingameWindows/iwMapGenerator.h
```diff
@@ -0,0 +1,67 @@
...
+// You should have received a copy of the GNU General Public License
+// along with Siedler II.5 RTTR. If not, see <http://www.gnu.org/licenses/>.
+#ifndef iwMAPGENERATOR_H_INCLUDED
+#define iwMAPGENERATOR_H_INCLUDED
+
+#pragma once
+
+#include "mapGenerator/MapSettings.h"
+#include "mapGenerator/MapStyle.h"
+#include "IngameWindow.h"
```
Flamefire: This is the only include required

Flamefire:
I'd say: do a last check for: test values, superfluous includes (especially in headers) and C-arrays (use boost::array or vector and double check sizes, seen one for players where you used "7" instead of the max player constant) 
And when you say that you are done then merge it so it can be tested by users

src/ingameWindows/iwMapGenerator.cpp
vader1986: I assigned the values to be able to use the enums for the controls in all places, including the constructor where the controls are added (and require an ID). If I want to avoid the values, then I need to add an enum-value for all controls (even labels, which i never access again).

Flamefire: You can use a placeholder. Like start with `CTRL_LAST_ID = 10` and then all your named ctrls afterwards. Then you can use id 1-10 for the unnamed ones and the others will be named with ids starting at 11. This way you only need to specify one value and if you add more unnamed values you can simply change the placeholder

src/mapGenerator/Map.cpp
```diff
@@ -26,11 +26,11 @@ Map::Map() : width(0), height(0)
 Map::Map(unsigned int width,
          unsigned int height,
          const std::string& name,
-         const std::string& author) : width(width), height(height), name(name), author(author), players(MAX_PLAYERS)
+         const std::string& author) : width(width), height(height), name(name), author(author), positions(MAX_PLAYERS)
 {
     const unsigned int size = (unsigned int)width * height;
 
-    for (int i = 0; i < MAX_PLAYERS; i++)
+    for (unsigned int i = 0; i < positions.size(); i++)
```
Flamefire: You can even do this in the ctor: `positions(MAX_PLAYERS, Point<uint16_t>::Invalid())` or use an algo like `std::fill(positions.begin(), positions.end(), Point<uint16_t>::Invalid())`

vader1986:
@Flamefire I think I'm finished so far. 

I made a couple of branches (in my fork) for refactoring & prototyping different approaches for random map generation which are giving better results for particular settings. Also there are probably some ideas for improvement for user experience (e.g. save random map settings, avoid revealing the map before playing, etc.). However, I think the feature provides at least a bit value for the user, even in the current stage.

src/ingameWindows/iwMapGenerator.cpp
```diff
@@ -34,14 +34,15 @@
...
-    CTRL_RATIO_GRANITE = 16
+    CTRL_MAP_STYLE,
+    CTRL_MAP_SIZE,
+    CTRL_PLAYER_RADIUS,
+    CTRL_MAP_TYPE,
+    CTRL_RATIO_GOLD,
+    CTRL_RATIO_IRON,
+    CTRL_RATIO_COAL,
+    CTRL_RATIO_GRANITE,
+    CTRL_LAST_ID // last enum value
```
Flamefire: Ok this does also work but I meant it the other way round. So just for documentation: Use `CTRL_LAST_ID = 10` as the **first** value in the enum and all the named entries come afterwards. This way you can use id `1` to `10` for all unnamed controls (labels etc.). It is better than this solution as it is shorter as instead of `CTRL_LAST_ID + x` you just use `x`

src/mapGenerator/Map.cpp
```diff
@@ -23,18 +23,12 @@ Map::Map() : width(0), height(0)
...
+         const std::string& author) : width(width), height(height), name(name), author(author), positions(MAX_PLAYERS, Point<uint16_t>::Invalid())
 {
-    const unsigned int size = (unsigned int)width * height;
-
-    for (unsigned int i = 0; i < positions.size(); i++)
-    {
-        positions[i].x = 0xFF;
-        positions[i].y = 0xFF;
-    }
+    const unsigned size = (unsigned)width * height;
```
Flamefire: Note: Unnecessary cast. And try to use C++ casts ( `static_cast` and family) Can be done in a follow up though

Flamefire:
Good to go from my side. I'd merge this and improve everything left in follow-ups.

vader1986:
I changed the two issues you pointed out. Feel free to merge now. Other improvements I'll add in small pull requests.

Flamefire:
Done. Thanks for your work.
I did not squash the commits to keep them for documentation purposes as this might help later

stefson:
@vader1986 

I have tested this for a while now and ran into a few problems/bugs with the random generation. One is that, if you generate a map in the mode of Migration with one island per player around a big one, there are rarely any bays to land an expediton in. Means, you build a harbour and a shipyard on the starting island and can't send it anywhere but only cancel the expedition. 

If there is a possibility to build in a check for at least one bay per island, or even better at least as many bays as players, then I'd be happy to test it! 👍

Flamefire:
@vader1986 If you are going to fix that: Make a testcase first that is supposed to fail. Then fix it. Thats how I do the bugfixing now to make sure once fixed bugs don't come back at some point.

PR #620, opened 2017-01-29
Fix CMake

SoapGentoo:
* Detect if at least C++11 is active from the `__cplusplus` macro
  and not by running regex on the `CXXFLAGS`. This is more portable
  and is forward compatible with compilers such as GCC 6, that
  run in C++14 by default. Currently, when not specifying any `-std=`
  setting with GCC 6, the build system sets C++98 by default, which
  is undesirable.
* Make adding optimizations flags to `CXXFLAGS` optional. Some
  distributions (such as Gentoo), have a strict policy that the
  user is in full control of all optimization flags.
* Do not build with `-Werror` by default. While this is supposed to
  make the code more robust, you cannot forsee new warnings
  produced by compilers appearing after the release of your code.
  See also:

   https://blog.flameeyes.eu/2009/02/future-proof-your-code-dont-use-werror/

  For instance, the current code produces warnings of the sort:

    `AIPlayerJH.cpp:1421:5: warning: this ‘if’ clause does not guard... [-Wmisleading-indentation]`

  This is a new GCC 6 warning, which the authors could not have forseen 2 years ago.


CMakeLists.txt
```diff
@@ -237,15 +223,20 @@ MACRO(CheckAndAddFlag flag)
 	unset(checkedFlag)
 ENDMACRO()
 
+OPTION(ENABLE_WERROR "Build with warnings turned into errors" OFF)
```
Flow86: please change Jenkinsfile, .travis.yml  and appveyor.yml so the ci builds are run with WError

driver/src/CMakeLists.txt
```diff
@@ -15,6 +8,8 @@ ADD_FILE_DEPENDENCIES(${PROJECT_SOURCE_DIR}/build_version.cpp ${CMAKE_BINARY_DIR
 
 ADD_LIBRARY(videodrv STATIC VideoDriver.cpp VideoDriver.h VideoInterface.h ${PROJECT_SOURCE_DIR}/build_version.cpp)
 ADD_LIBRARY(audiodrv STATIC AudioDriver.cpp AudioDriver.h AudioInterface.h Sound.cpp Sound.h Interface.h ${PROJECT_SOURCE_DIR}/build_version.cpp)
+SET_PROPERTY(TARGET videodrv PROPERTY POSITION_INDEPENDENT_CODE ON)
+SET_PROPERTY(TARGET audiodrv PROPERTY POSITION_INDEPENDENT_CODE ON)
```
Flow86: that SET_PROPERTY line are missing for videoSDL etc also (see travis ci builds)

CMakeLists.txt
Flow86: (perhaps -Wmisleading-indentation should be added too ...)

SoapGentoo:
@Flow86 I've fixed the issues, sorry for the delay. I still think all the optimization flags ought to be removed. CMake in `RELEASE` mode adds `CFLAGS="-O3 -DNDEBUG"`, which pretty much enables all of these. While I now know that there is a flag to disable them, many users might want to chaneg them anyhow, and having to dig for a CMake flag to disable optimizing flags is counter-intuitive.

SoapGentoo:
@Flow86 ping

driver/video/SDL/src/CMakeLists.txt
```diff
@@ -37,7 +30,8 @@ IF (SDL_FOUND)
 		LIBRARY_OUTPUT_DIRECTORY_DEBUG "${_OUTPUT_DIR}"
 		LIBRARY_OUTPUT_DIRECTORY_RELWITHDEBINFO "${_OUTPUT_DIR}"
 	)
-	
+	SET_PROPERTY(TARGET videoSDL PROPERTY POSITION_INDEPENDENT_CODE ON)
```
Flow86: thats missing for video/WinAPI too

SoapGentoo: @Flow86 fixed

PR #653, opened 2017-07-03
Remove or rename duplicate file names

Flamefire:
This avoids problems with code coverage detection and possible double/non includes


Flow86:
glfw könnten wir gleich ganz entfernen, benutzt eh keiner

Flamefire:
As mentioned I'd also remove the WinAPI and keep SDL only. I think the apple build relies on using SDL anyway (the SDL_main include) so we could skip on the extra dlls/shared binaries and link that directly. Saves the incompatibilities with the editor and is less maintenance as we have only one implementation to update and test with.

However that got to be a separate PR as I don't want to mix to many things in one.

PR #659, opened 2017-07-08
Human pathfinding fixes

### LINKED ISSUE #626: PathConditionHuman Is incorrect, allows walking across water
~~This is an incorrect check~~
~~https://github.com/Return-To-The-Roots/s25client/blob/master/src/pathfinding/PathConditions.h#L42~~

~~the comment above states~~
~~`// If no terrain around this is usable, we can't go here`~~
~~in reality what it is checking is if any point is usable~~

People can walk across V's of water which shouldn't be allowed. You can repro this easily with shipbuilders walking along beaches.

See below comment

Flamefire:
This adds tests to detect e.g. humans walking on water (#626)
and fixes #626 

I made some new test cases an an illustration here: 
testpathfindingillustration

Testcase 1: Walk along the shore in every possible direction
Testcase 2: Walk "over" water in every direction which should only be possible, if there is a road
Testcase 3: Walk along "dangerous" terrain like lava which should not be possible -> green path taken
Question: Should it be possible to "touch" the dangerous terrain (straight from yellow to yellow)? I think not, all paths in Testcases 2 and 3 should be 4 steps long.
Can you confirm this @Spikeone ?

In the current version the test case fails because they can walk straight in Testcase 2

Update

- Testcase 2 is tested with `TT_WATER, TT_WATER_NOSHIP, TT_SWAMPLAND`
- Testcase 3 is tested with `TT_SNOW, TT_LAVA, TT_LAVA2, TT_LAVA3, TT_LAVA4`
- Walking straight is tested with `TT_DESERT, TT_MEADOW_FLOWERS, TT_MOUNTAIN1, TT_MOUNTAIN2, TT_MOUNTAIN3, TT_MOUNTAIN4, TT_SAVANNAH,
        TT_MEADOW1, TT_MEADOW2, TT_MEADOW3, TT_STEPPE, TT_MOUNTAINMEADOW, TT_BUILDABLE_WATER, TT_BUILDABLE_MOUNTAIN`


Flamefire:
@Spikeone What shall we do with buildable water? I'd say if it is buildable, then people can also walk on that. Correct?

Spikeone:
I totally agree, although I have not tested it in the original.

Flamefire:
Alright. I updated the original post with the tested landscape types. Can you check that they are all correct?

PR #660, opened 2017-07-13
Drawing improvements

### LINKED ISSUE #585: Lobby chat notification color
When hosting a server there is the lobby chat and the game chat.
When you get a message, the buttons are blinking - sadly they use the wrong texture, currently IO id 8 is used but should be IO id 3.

Just have a look at the "Lobby Chat" button and "Return"

lobbychat

Flamefire:
This refactors lots of drawing related code and adds some tests.
Main points:

- Test screen added for visual tests (double click version text on main screen to enter)
- button up/down state improved (no more artefacts on pressed button)
- illuminate also borderless buttons
- deepening control now has a base class and textdeepening does not inherit from ctrlText (as it isn't a ctrlText)
- add function to get text bounds and use it to implement getting element bounds
- Will add button blinking to fix #585


src/Window.h
```diff
@@ -216,7 +216,7 @@ class Window
         // GUI-Notify-Messages
 
         // Nachrichten, die von oben (WindowManager) nach unten (zu Controls) gereicht werden
-        virtual void Msg_PaintBefore(){}
+        virtual void Msg_PaintBefore();
```
Flow86: override?

src/desktops/dskHostGame.cpp
```diff
@@ -431,6 +431,7 @@ void dskHostGame::UpdatePlayerRow(const unsigned row)
  */
 void dskHostGame::Msg_PaintBefore()
 {
+    Desktop::Msg_PaintBefore();
```
Flow86: evtl sollten wir allgemein eine art typedef base in den klassen machen, dann kann man immer "base::..." schreiben?

src/Window.h
Flamefire: This is the base class ;-)

src/desktops/dskHostGame.cpp
Flamefire: Yes, I've been thinking about this. Especially when you change the inheritance to introduce another level this would prove very helpful. However it would be a lot of work.

src/Window.h
Flow86: oh ;-)

src/desktops/dskHostGame.cpp
Flow86: > However it would be a lot of work.
one line per class?

Flamefire: Yep. For 40 classes or so. Its just boring ;)
I see if I can make a regex or so to do this automatically. We need it for the gameobjects too to get rid of all the `Destroy_xxx` functions and the like

PR #670, opened 2017-08-05
Avoid unnecessary libminiupnp remove

Mailaender:
Looks like it is no longer necessary as it isn't bundled and installed anymore.


Mailaender:
Removed it from the RPM without problems: https://build.opensuse.org/request/show/514714

PR #677, opened 2017-08-21
Added missing tooltips, a quick hack for some international characters

frin:
Note: I added a small hack for text drawing method that converts some international characters to their ASCII counterparts (for example Slovenian), they look better than drawn question marks (?) in their place. Best solution would of course be added glyphs for all missing characters, but until then this does the job.


Flow86:
@frin I would merge the missing tooltips, but I have to agree with @Flamefire about the glyphs. perhaps you can add these to the font? Thank you very much!

frin:
Ah, didn't realize you had the ability of modifying the font, definitely the right way to go, I will look into it.

frin:
I've added glyphs and removed the replace hack. I merged with the latest version of the repo and rebased my stuff, but not sure if that's proper way to do it, looks like a mess with 26 commits even though there's just 2. Is that okay, or should I just close this PR and create one from scratch?

Flamefire:
Just a rebase would have been right. Try an interactive rebase skipping the merge and force push to clean this.

frin:
There we go 👍

Flow86:
okay I'll merge that now, since its only the tooltips. 👍

Flamefire:
And the glyphs are also there. Thank you! 
Could you fix the pr title?

frin:
Changed it to what the commit message is. Don't forget to merge the languages repo as well, with the added two new translations.

PR #681, opened 2017-08-23
Show direction glyph toward flag halfway out, save and Mac delete enhancements

frin:
This is a similar behavior as seen in original S2. (Drawing while in road building mode)

### DIFF
--- a/src/GameClient.cpp
+++ b/src/GameClient.cpp
@@ -1684,9 +1684,16 @@ unsigned GameClient::SaveToFile(const std::string& filename)
 
     // Und alles speichern
     if(!save.Save(filename))
+    {
+        GameMessage_System_Chat saveAnnouncement = GameMessage_System_Chat(playerId_, "Error occurred while saving game.");
+        send_queue.sendMessage(socket, saveAnnouncement);
         return 1;
-    else
+    }
+    else {
+        GameMessage_System_Chat saveAnnouncement = GameMessage_System_Chat(playerId_, "Game saved.");
+        send_queue.sendMessage(socket, saveAnnouncement);
         return 0;
+    }
 }
 
 void GameClient::ResetVisualSettings()
--- a/src/ingameWindows/iwSave.cpp
+++ b/src/ingameWindows/iwSave.cpp
@@ -127,7 +127,13 @@ void iwSave::SaveLoad()
     tmp += ".sav";
 
     // Speichern
-    GAMECLIENT.SaveToFile(tmp);
+    int error = GAMECLIENT.SaveToFile(tmp);
+    if (error == 0)
+    {
+        // Just close the dialog if save was successful, since player usually wants to continue playing or quit the game
+        Close();
+        return;
+    }
 
     // Aktualisieren
     RefreshTable();
--- a/src/world/GameWorldView.cpp
+++ b/src/world/GameWorldView.cpp
@@ -290,7 +290,7 @@ void GameWorldView::DrawGUI(const RoadBuildState& rb, const TerrainRenderer& ter
                     continue;
 
                 if((gwv.IsRoadAvailable(rb.mode == RM_BOAT, curPt) && gwv.IsOwner(curPt) && GetWorld().IsPlayerTerritory(curPt))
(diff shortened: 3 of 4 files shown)


Flow86:
can you please provide a screenshot? I cant imagine it ;-)

frin:
Yes of course, marked with magenta arrows http://i.imgur.com/QdQWpUe.png

Previously you wouldn't know the slope of the road for the last road segment.

frin:
Oh I guess this gets added automatically to existing PR. Well, there's more enhancements. I updated the PR title

Spikeone:
Do you have a screenshot with extreme heights? Showing what that looks like? Your current screenshot shows quite moderate height differences

frin:
Yeah here is the biggest height difference, it is slightly overlapping, but so does the original S2, screenshot for comparison. Still it's not that bad.
s2_height

src/GameClient.cpp
```diff
@@ -1684,9 +1684,16 @@ unsigned GameClient::SaveToFile(const std::string& filename)
 
     // Und alles speichern
     if(!save.Save(filename))
+    {
+        GameMessage_System_Chat saveAnnouncement = GameMessage_System_Chat(playerId_, "Error occurred while saving game.");
```
Flamefire: Not a good idea: Saving is done per client. You would most likely flood the game with messages from every player. Rather just show it to the local player maybe using the notification system.
Also use the localization feature: `_("Translated text")`

src/ingameWindows/iwSave.cpp
```diff
@@ -127,7 +127,13 @@ void iwSave::SaveLoad()
     tmp += ".sav";
 
     // Speichern
-    GAMECLIENT.SaveToFile(tmp);
+    int error = GAMECLIENT.SaveToFile(tmp);
+    if (error == 0)
```
Flamefire: Maybe this would be the place to show the message?

src/world/GameWorldView.cpp
```diff
@@ -308,7 +308,21 @@ void GameWorldView::DrawGUI(const RoadBuildState& rb, const TerrainRenderer& ter
...
                     }
 
-                    LOADER.GetMapImageN(id)->DrawFull(curPos);
+                    // If item at direction is a flag and not the start, draw the direction glyph halfway out
+                    if (GetWorld().GetNO(curPt)->GetType() == NOP_FLAG && curPt != rb.start)
+                    {
+                        // Last clicked position (coordinate) in the road route
+                        Point<int> lastPos = GetWorld().GetNodePos(rb.point) - offset + curOffset;
+                        Point<int> halfwayPos;
+                        halfwayPos.x = (curPos.x + lastPos.x) / 2;
```
Flamefire: Easier: Replace `Point<int>` by `DrawPoint` and use `halfwayPos = (curPos + lastPos) / 2`

src/world/GameWorldView.cpp
```diff
@@ -290,7 +290,7 @@ void GameWorldView::DrawGUI(const RoadBuildState& rb, const TerrainRenderer& ter
                     continue;
 
                 if((gwv.IsRoadAvailable(rb.mode == RM_BOAT, curPt) && gwv.IsOwner(curPt) && GetWorld().IsPlayerTerritory(curPt))
-                   || (gwv.GetBQ(curPt) == BQ_FLAG))
+                   || (gwv.GetBQ(curPt) == BQ_FLAG) || (GetWorld().GetNO(curPt)->GetType() == NOP_FLAG && curPt != rb.start))
```
Flamefire: Can the last condition become a problem if the flag belongs to an other player? Maybe the last 2 conditions (Owner and player territory) should always apply. But would need to think carefully about this.

frin:
Interesting suggestions (didn't know saves are separate for each client). I will look into these and update code accordingly.

Flamefire:
@frin Are you still working on that? Would love to see this changes merged :)

Shawn8901:
@Flamefire  I think this PR can be closed, as every change was merged via #821

Issue #674, opened 2017-08-12
AI stuck


luk3ZPL:
AI stopped build any building (stuck) when there is no suitable place for it (probably from AI priority list).

1) Which version is affected?
	20170811

2) What did you do/shall be done to reproduce the bug?
	Look at AI red player from save (again this AI have really bad luck ;) :
	http://www.mediafire.com/file/1iq0agqmc9tw1mb/AI_bug.7z

3) What is expected? (Unless it is a crash)
	AI should expand borders a little bit (at the begining of the game when is isolated) to find a suitable place for building.
	
4) What happens instead?
	AI stopped build any building (stuck).

Flamefire:
Some analysis: This is caused by an unfair map created by the mapgenerator from @vader1986 (you may want to look into this)
The red player has only hut building qualities so he cannot build a sawmill. This is currently the condition for expansion for the AI written by @jhkl (any ideas?) I might have an idea how to fix this by simply removing the condition that checks for planned or built sawmills. Reasoning: The AI will try to build a sawmill first. So if it isn't planned or in the construction queue then it is impossible and the AI should expand to try somewhere else. However the AI currently tries to build only around warehouses. Maybe it should try somewhere else if this fails to find a space?

luk3ZPL:
Any workaround will be good.

jhkl:
Expansion sounds like a valid fallback if nothing else works ;)
I think it tries to build around random military buildings as well (if there are any).

vader1986:
@flamefire: I'll probably go for some bugfixes/map generation improvements around Christmas. Really sorry, but I'm quite busy and have no internet connection at home.

frin:
I managed to fix the AI problem, tested on this case in particular, PR https://github.com/Return-To-The-Roots/s25client/pull/682. I've detailed the change in the PR comment. Once the AI builds the first military building it spreads well, building sawmill and other buildings.

The problem was that there were no build tiles suitable to build a sawmill within initial border.
![screen shot 2017-08-25 at 19 25 55](https://user-images.githubusercontent.com/176282/29727065-ab5b4bf4-89d2-11e7-99b0-b8feedc580b3.png)

PR #682, opened 2017-08-25
Fix for AI issue #674

frin:
Changes made: if AI generates a build job with a military building and sawmill but can't build a sawmill (cause no middle build quality tile is available), it will build a military building instead (in this specific case a military building was not wanted, so build is forced when there is no sawmill built yet). It also downgrades fortress/watchtower to guardhouse in this case. Only applies if no sawmill is built yet.

Also fixes a minor build job sorting bug/already built check, where fortress wasn't included with other military buildings.

### DIFF
--- a/src/ai/AIConstruction.cpp
+++ b/src/ai/AIConstruction.cpp
@@ -66,7 +66,7 @@ void AIConstruction::AddBuildJob(AIJH::BuildJob* job, bool front)
         delete job;
         return;
     }
-    if(job->GetType() < BLD_FORTRESS) // non military buildings can only be added once to the contruction que for every location
+    if(job->GetType() <= BLD_FORTRESS) // non military buildings can only be added once to the contruction que for every location
     {
         if(front)
             buildJobs.push_front(job);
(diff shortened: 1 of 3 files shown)


src/ai/AIConstruction.cpp
```diff
@@ -593,6 +593,16 @@ bool AIConstruction::MilitaryBuildingSitesLimit()
     return complete + 3 > inconstruction;
 }
 
+bool AIConstruction::IncrementWanted(BuildingType type)
+{
+    if (GetBuildingCount(type) <= buildingsWanted[type])
```
Flamefire: Shouldn't this be >=? 
And I think putting the increment and the check in the same function is misleading for the reader of the call. It is also not clear why this leads to forceBuild so checking the count there would be better. So this function can actually be removed

src/ai/AIJHHelper.cpp
```diff
@@ -101,7 +101,23 @@ void AIJH::BuildJob::TryToBuild()
...
+            {
+                // We can't build watchtower or fortress because no BQ_HOUSE or higher is available, build a
+                // guardhouse instead
+                type = BLD_GUARDHOUSE;
+            }
+            forceBuild = aiConstruction.IncrementWanted(type);
+        }
+    }
+
+    if(!aiConstruction.Wanted(type) && !forceBuild)
```
Flamefire: Better swap conditions

src/ai/AIConstruction.cpp
frin: That check's there so that AI doesn't keep building and building new military buildings if first one isn't complete yet so the "<=" is correct. I guess the function can be removed and check moved to the AIJHHelper.cpp, but I thought it would be more proper to actually update the wanted list.

Flamefire: But why would you want to increase the wanted buildings if there are more wanted than built? 
And this implementation won't avoid the case you are talking about: built=0 wanted=1 so increase wanted to 2. Then to 3 etc. 
Maybe rather a) build military bld even if no more wanted in case of space problems and b) don't do a) of there are military blds in construction.
Maybe even make sure we don't add military buildings to the wanted buildings unless we have a sawmill (at least as a building site) or can't build it (so add it to wanted buildings if no position is found, similar to your solution)

Flamefire:
Superseded by #686

PR #684, opened 2017-08-29
Optimization and fixes

Flamefire:
Collection of various improvements, see commits.

Most notable:

- RTTR_Assert now available in drivers.
- Some critical bugfixes (ingame window placement, lobby crash, gentoo compile fixes)
- Font fixes and performance improvements (width calculation was off by 1 which resulted in overdrawing of buttons, performance for the debug output (printing of map coords etc) is very bad, now slightly better)


CMakeLists.txt
```diff
@@ -235,6 +238,10 @@ IF (MSVC)
 	IF(RTTR_ENABLE_WERROR)
 		ADD_DEFINITIONS(/WX) # warning = error
 	ENDIF()
+	# Link time code generation (Required for bzlib) which also disables incremental linking
+	ADD_FLAGS(CMAKE_EXE_LINKER_FLAGS /LTCG)
+	FORCE_ADD_FLAGS(CMAKE_EXE_LINKER_FLAGS_DEBUG /INCREMENTAL:NO)
```
Flow86: don't do that for debug! do not set LTCG if not necessary. incremental linking is extremly important if you do  live-debugging (with disabling incremental, no edit+continue works anymore)

CMakeLists.txt
```diff
@@ -393,6 +400,8 @@ ADD_CUSTOM_TARGET(updateversion ALL
 ################################################################################
 # Then add all other dependencies
 
+FILE(GLOB COMMON_SRC ${PROJECT_SOURCE_DIR}/commonSrc/*.cpp ${PROJECT_SOURCE_DIR}/commonSrc/*.h)
```
Flow86: don't use glob - if a file's missing the buildsystem does not recognize that and you get strange errors

driver/audio/SDL/src/CMakeLists.txt
```diff
@@ -22,9 +22,9 @@ ELSE()
 		SET(SL_PREFIX "")
 	ENDIF ()
 		
-	ADD_LIBRARY(audioSDL SHARED AudioSDL.cpp AudioSDL.h SoundSDL_Music.cpp SoundSDL_Music.h SoundSDL_Effect.cpp SoundSDL_Effect.h ${PROJECT_SOURCE_DIR}/driver/src/DriverInterface.cpp)
-	ADD_DEPENDENCIES(audioSDL updateversion)
-	TARGET_LINK_LIBRARIES(audioSDL audiodrv endian s25util ${SDL_LIBRARY} ${SDLMIXER_LIBRARY} ${Boost_LIBRARIES})
+    FILE(GLOB LIB_SRC *.cpp *.h)
```
Flow86: same here

src/TerrainRenderer.cpp
```diff
@@ -349,18 +347,15 @@ void TerrainRenderer::UpdateTriangleTerrain(const MapPoint pt, bool updateVBO)
...
         RTTR_Assert(w > 0 && h > 0);
-        unsigned texW = helpers::roundToNextPowerOfTwo(w);
-        unsigned texH = helpers::roundToNextPowerOfTwo(h);
+        Point<float> texSize(VIDEODRIVER.calcPreferredTextureSize(Extent(w, h)));
 
-        float texScaleW = 1.f / texW;
-        float texScaleH = 1.f / texH;
         // Tip of the triangle is in the middle in x
-        texCoord[1].x = (w + 1) / 2.f * texScaleW;
+        texCoord[1].x = (w + 1) / 2.f / texSize.x;
```
Flow86: use multiplications these are much faster

CMakeLists.txt
Flamefire: The bzip2 in the contribution dir forces this. But maybe we should just update that one instead

src/TerrainRenderer.cpp
Flamefire: Not if you do the division before. I now leave it to the compiler to figure that out. Most likely the asm will be the same but the code is more clear

CMakeLists.txt
Flamefire: Yeah but we are using this already. It's just so cumbersome to add a new class without a glob. But yes maybe we should change it back. But at least I'd put it in an extra cmake file to keep it clean (we have 100s files or so)

PR #685, opened 2017-08-29
Sound refactoring

Flamefire:
Refactors sound related code and removes the virtual inheritance used before. Makes it much cleaner.

Also uses SoundHandles that are held by the driver but can be closed from either side. So switching the driver would be possible as that invalidates all held handles which then get reloaded.

Note: It might be better to just store the `SoundItem`s in a list. They could contain an `ArchivItemSound` instead of inheriting from it.   
Similar for the bitmaps: Contain instead of inherit which would remove the virtual inheritance. However this requires some changes of the loader and/or "post-load-processing" where the classes are actually loaded into the other classes. Would still be easily possible, as we know, where which filetypes are. So we could detect wrong filetypes on load time not on use time.


Flow86:
something lost in space?

```
/workdir/driver/audio/SDL/src/SoundSDL_Effect.cpp:22: error: prototype for 'SoundSDL_Effect::SoundSDL_Effect()' does not match any in class 'SoundSDL_Effect'
/workdir/driver/audio/SDL/src/SoundSDL_Effect.h:26: error: candidates are: SoundSDL_Effect::SoundSDL_Effect(const SoundSDL_Effect&)
/workdir/driver/audio/SDL/src/SoundSDL_Effect.h:28: error:                 SoundSDL_Effect::SoundSDL_Effect(Mix_Chunk*)
/workdir/driver/audio/SDL/src/SoundSDL_Effect.cpp:26: error: definition of implicitly-declared 'virtual SoundSDL_Effect::~SoundSDL_Effect()'
``

Flow86:
AH! Thats one reason why you should not use FILE(GLOB ...)!

Flamefire:
I partially agree. However this is a bug in the buildsystem. This PR deletes the file in question. It should not exist anymore. So while the GLOB might lead to bugs it also unveils unused files which avoids other bugs/mistakes.
So what shall we do? Don't use globs and are forced to manually add each file to the CMakeLists which will result in huge file lists or stick to how it is?

PR #687, opened 2017-09-02
Fix warnings and edit-and-continue

Flamefire:
Enable /ZI
Dynamic link of lua
Rebuild and add bzip2d.lib/pdb
Fix MSVC resource compile
Make files to copy a list (required till CMake 3.5)
Disable SafeSEH for debug


cmake/Modules/CMakeMacroAddFlags.cmake
```diff
@@ -7,8 +7,8 @@ MACRO(ADD_FLAGS parameter)
 	# Make a copy of the current arguments in ${parameter}
 	SET(new_parameter ${${parameter}})
 	
-        # fix arch-argument bug
-        STRING(REPLACE "-arch;" "-arch=" new_parameter "${new_parameter}")
```
Flow86: is the indentation change intentional?

cmake/Modules/CMakeMacroAddFlags.cmake
```diff
@@ -37,8 +37,8 @@ MACRO(ADD_FLAGS parameter)
 		ENDFOREACH(option ${TMP})
 	ENDFOREACH(required_arg ${ARGN})
 	
-        # unfix arch-argument bug
-        STRING(REPLACE "-arch=" "-arch " new_parameter "${new_parameter}")
```
Flow86: is the indentation change intentional?

cmake/Modules/CMakeMacroForceAddFlags.cmake
```diff
@@ -37,8 +37,8 @@ MACRO(FORCE_ADD_FLAGS parameter)
 		ENDFOREACH(option ${TMP})
 	ENDFOREACH(required_arg ${ARGN})
 	
-        # unfix arch-argument bug
-        STRING(REPLACE "-arch=" "-arch " new_parameter "${new_parameter}")
+    # unfix arch-argument bug
+    STRING(REPLACE "-arch=" "-arch " new_parameter "${new_parameter}")
```
Flow86: is the indentation change intentional?

cmake/Modules/CMakeMacroAddFlags.cmake
Flamefire: Yes. It was one level to much (same for the others)

PR #725, opened 2017-10-25
Refactoring

Flamefire:
Make some function usages more uniform.

Add/RemoveFigure takes the position first which is more intuitive
A figure/movable is now always at its position. Before it was moved before it moved to account for drawing, but that is confusing and may lead to subtle bugs. This is now handled in the drawing code.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -41,7 +41,7 @@ def transformIntoStep(arch, wspwd) {
                                                          \$VOLUMES \
                                                          --name "${env.BUILD_TAG}-${arch}" \
                                                          git.ra-doersch.de:5005/rttr/docker-precise:master -c \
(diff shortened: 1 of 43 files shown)


Jenkinsfile
```diff
@@ -75,9 +75,10 @@ catchError() {
                 checkout scm
                 
                 sh """set -x
-                      git submodule foreach "git reset --hard || true" || true
-                      git reset --hard || true
-                      git submodule update --init || true
+                      git reset --hard
+					  git submodule foreach git clean --force
```
Flow86: bitte einrückung noch korrigieren

Flamefire: Done. Also added VERBOSE=1 to make to see what is failing there (osx)

PR #726, opened 2017-10-27
Exploration fixes and faster replay loading

Flamefire:
Fixes some issues with exploration and reworks the file format for replays and savegames.

The listing of replays is now much faster as it doesn't load the whole file anymore.

This invalidates all current replays/savegames which is ok, as the position change in the last PR already did that w/o bumping the format version.

Also:

- Use a resource struct instead of magic numbers
- WINTER cheat for human AI
- Productivity is now calculated in nobUsual (fix inconsistent calculation values as the event was fired in nobUsual but the calculation was in nofWorkman)
- Pause param for MissionStatement
- Language was set incorrectly and MyGetText was missing functionality for querying
- Lua localization
- Better RNG (XorShift) and refactoring of the old one (state transition should not depend on max value
- Add icons to SDL windows
- Add lua localization
- Add editor


Flamefire:
Merged before this gets any bigger. Failed appveyor build was due to unrar download failure

PR #737, opened 2017-11-07
WIP: ENH: new darwin15 toolchain and do not use hardcoded strip

Flow86:
Use new darwin15 toolchain


Flamefire:
Merged my big PR and updated all gitattributes and line endings. So please rebase this.
Module conflicts can be ignored (just skip your 1st commit) and the src conflicts get resolved automatically at least in tortoiseGit (line endings only)

Flow86:
interesting that the PR here is completly broken O.o

Flamefire:
The conflicts were due to the submodules. The many commits due to a wrong rebase. Did you pull your master before you rebased?

Flow86:
sure, thats whats strange here, local log is clean

unbenannt

Flamefire:
Had the same in #741 for some reason. I pulled master again, rebased and force-pushed which fixed it there. I think this is a github bug but don't know how it happened.

Flow86:
yep that worked. added another commit and rebased again, now its correct again

CMakeLists.txt
```diff
@@ -53,6 +53,23 @@ INCLUDE(CheckCXXCompilerFlag)
 INCLUDE(CheckCCompilerFlag)
 INCLUDE(CheckIncludeFile)
 
+SET (RTTR_RPATH "")
```
Flamefire: Shouldn't this rather go into the `cmake/<platform>.cmake` files?
And what is `$ORIGIN`? Haven't heard of that.

CMakeLists.txt
```diff
@@ -187,6 +204,14 @@ IF (WIN32)
 	SET(RTTR_LIBDIR "." CACHE INTERNAL "")
 	SET(RTTR_DOCDIR "." CACHE INTERNAL "")
 	SET(RTTR_DRIVERDIR "./driver" CACHE INTERNAL "")
+ELSEIF(APPLE)
+	# Apple wants an Apple Bundle
+	SET(RTTR_BINDIR "./s25client.app/Contents/MacOS/bin" CACHE INTERNAL "")
```
Flamefire: I think `./s25client.app/Contents/MacOS` should rather be the Install prefix passed during cmake configure as not every apple user would want to create the `.app`-structure

CMakeLists.txt
```diff
@@ -465,9 +494,35 @@ endif()
 # Start script
 IF (WIN32)
 	install(PROGRAMS "release/bin/rttr.bat" DESTINATION "${RTTR_BINDIR}")
+ELSEIF(APPLE)
+	install(PROGRAMS "release/bin/macos/rttr.command" DESTINATION "${RTTR_LIBDIR}")
```
Flamefire: Although this might technically be the correct path, don't use `RTTR_LIBDIR` as that is the wrong path (name variable)
If we do want to hard-code the whole .app-structure, define variables `SET(RTTR_MACOS_DIR "./s25client.app/Contents/MacOS")` or even the parent dir: `SET(RTTR_MAC_CONTENT_DIR "./s25client.app/Contents")`. Because looking at this lines make me want to change them as they *look* wrong

cmake/darwin.cmake
```diff
@@ -56,7 +56,7 @@ IF ( NOT "${CMAKE_OSX_ARCHITECTURES}" MATCHES "ppc" AND NOT "${CMAKE_OSX_ARCHITE
     ADD_FLAGS(APPLE_CFLAGS -mtune=core2)
 ENDIF ( NOT "${CMAKE_OSX_ARCHITECTURES}" MATCHES "ppc" AND NOT "${CMAKE_OSX_ARCHITECTURES}" MATCHES "i386" )
 
-ADD_FLAGS(APPLE_CFLAGS -fexceptions -ffast-math -msse -fomit-frame-pointer)
```
Flamefire: Please try `-msse2` instead of removing this if you have the time. If not: Have you found or filed a bug report in GCC? Could you link the bug-id/url?

CMakeLists.txt
Flow86: yea thats kind of interesting:
> $ORIGIN is a special variable that means ‘this executable’, and it means the actual executable  filename, as readlink would see it, so symlinks are followed.  In other words, $ORIGIN is special and resolves to wherever the binary is at runtime.

on work, we have for example `$ORIGIN/../lib` set, so a binary in `x/bin` will search for its libraries in `x/lib`

CMakeLists.txt
Flow86: yes that nees to be adapted

cmake/darwin.cmake
Flow86: nice try, but does not work

existing  bug reports are here: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78380

release/bin/macos/rttr.command
```diff
@@ -23,6 +23,8 @@ chmod 0755 ../rttr.command ../bin/RTTR/s25update ../bin/s25client ../bin/RTTR/so
 RTTR_TEST_FILEA="share/s25rttr/S2/DATA/CREDITS.LST"
 RTTR_TEST_FILEB="share/s25rttr/S2/GFX/PALETTE/PAL5.BBM"
 
+unset LANG
```
Flamefire: Why? I was rather thinking about supporting that additionally in mygettext

Flow86: well, i tried everything to remove that error caused by the entry (with and without UTF-8) it always generated the error "locale::facet::_S_create_c_locale name not valid"

Flamefire: IMO this could only be caused by wrong locale settings (e.g. var set wrong, lang not installed)
See https://stackoverflow.com/questions/19100708/mongodb-mongorestore-failure-localefacet-s-create-c-locale-name-not-valid/32762296#32762296 and https://stackoverflow.com/questions/19100708/mongodb-mongorestore-failure-localefacet-s-create-c-locale-name-not-valid

We should not work-around user bugs... Unsetting `LANG` would influence the whole system-language setting and the user might end up with a wrong language in rttr

Flow86: yea but that only happens on osx, where you dont have the possibility to install "locales" or something like that.

Flamefire: I found a couple other options that I like much better:

1. Use libc++ instead of libstdc++ https://stackoverflow.com/questions/11190072/setting-locales-on-os-x-crashes
2. Build gcc with `--enable-clocale` https://www.thecodingforums.com/threads/std-locale-problem-on-mac-osx.752939/
3. Don't use `std::locale("")` which boost also does for OSX and related. (Based on https://svn.boost.org/trac10/ticket/1458 and current boost src)

While it would be great to know if 1. or 2. works I'd go for 3 and remove the `unset LANG` as I'm sure keeping it would make our "System-default" language setting fail.

Flow86: I'll test the "--enable-clocale" first - for libc++ I have to rebuild the boost library too

Flow86: `--enable-clocale` seems to have no effect

Flamefire: Added a new commit which should fix this -.-

Flamefire:
Superseded by compileEnv branch

Issue #740, opened 2017-11-08
Segfault on showing military aid


stefson:
I just tested a randoom map (I think Pangea?) with default settings, turns out the game segfaults straight away if using the space bar, which is used to display the 'little yellow markers' for flags and houses.

Flamefire:
Seems like the new_map.LST folder was not loaded. Can you post the log of this crash?

Flamefire:
Can't reproduce this. It really looks like `RTTR/LSTS/GAME/map_new.LST` is not loaded. AND you activated the military aid addon which is non-default.

stefson:
I tried to start a new game with S2 standards, which disables the military aids, and it works just fine without any issues regarding the space bar. 

About the `map_new.LST` file, there is one folder called `RTTR/LSTS/GAME/map_new.LST/` , but no file with this name. Do you meant this folder or is there an issue regarding the filename? 

these are the absolute paths of the files which are to be found in this folder

```~ $ equery f s25rttr | grep map_new.LST
/usr/share/s25rttr/RTTR/LSTS/GAME/map_new.LST
/usr/share/s25rttr/RTTR/LSTS/GAME/map_new.LST/20000.rle.nx0.ny0.bmp
/usr/share/s25rttr/RTTR/LSTS/GAME/map_new.LST/3150.player.nx140.ny80.bmp
/usr/share/s25rttr/RTTR/LSTS/GAME/map_new.LST/3151.player.nx140.ny80.bmp
/usr/share/s25rttr/RTTR/LSTS/GAME/map_new.LST/3152.player.nx140.ny80.bmp
/usr/share/s25rttr/RTTR/LSTS/GAME/map_new.LST/3153.player.nx140.ny80.bmp```

Flamefire:
Yes the folder. Please enable military aid, press space and attach the log if that crashed

stefson:
here you go, full log from console including output from gdb/backtrace

[log.zip](https://github.com/Return-To-The-Roots/s25client/files/1459000/full.log.zip)

Flamefire:
I may have found it. The problem lies a bit deeper and is due to locales. The "20000.rle.nx0.ny0.bmp" is converted to an int via a standard stream which is supposed to return `20000`. However I think that it may not read it like this can return something else.
[...]

PR #741, opened 2017-11-09
Fix non-selectable addons and better locale support

Flamefire:
Fix addons not selectable in addons window if they are not visible on the first page
Fix locale dependent number conversions leading to not-loaded assets


stefson:
I had a bit of trouble with the addon selection menu while testing to disable the military aid option in #740 , does this PR deal with this issue?

Flamefire:
Yes. Needs a bit more work though

PR #758, opened 2017-11-17
Changed ExhausibleWellsAddon to ExhausibleWaterAddon

Shawn8901:
Added new option with "Inexhausible Water, with water everywhere" which puts water on every free resource tile and prevents water notifications


src/addons/AddonExhaustibleWater.h
```diff
@@ -19,19 +19,23 @@
...
+class AddonExhaustibleWater : public AddonList
 {
 public:
-    AddonExhaustibleWells()
-        : AddonBool(AddonId::EXHAUSTIBLE_WELLS, ADDONGROUP_ECONOMY, _("Exhaustible Wells"),
-                    _("Wells will now dry out, limiting the available water"), 0)
-    {}
+    AddonExhaustibleWater()
+        : AddonList(AddonId::EXHAUSTIBLE_WATER, ADDONGROUP_ECONOMY, _("Exhaustible Water"),
+                    _("If Water is exhaustible wells will now dry out."), 0)
```
Flamefire: Maybe say also that "water everywhere" disables water found notifications

src/world/MapLoader.cpp
```diff
@@ -136,6 +124,30 @@ void MapLoader::InitNodes(const glArchivItem_Map& map, Exploration exploration)
             resource = Resource(Resource::Granite, mapResource - 0x58);
         else if(mapResource > 0x80 && mapResource < 0x90) // fish
             resource = Resource(Resource::Fish, 4);       // Use 4 fish
+        else
```
Flamefire: What about simply `else if (mapResource == 0x20 || mapResource == 0x21 || waterEverywhere)` and then the old code?
Then we have no duplication and there can be different amounts of water (why not?)

src/world/MapLoader.cpp
```diff
@@ -45,11 +45,11 @@ class nobBaseWarehouse;
 
 MapLoader::MapLoader(World& world, const std::vector<Nation>& playerNations) : world_(world), playerNations_(playerNations) {}
 
-bool MapLoader::Load(const glArchivItem_Map& map, Exploration exploration)
+bool MapLoader::Load(const glArchivItem_Map& map, Exploration exploration, bool waterEveryWhere)
```
Flamefire: casing: "waterEvery**w**here"

src/world/MapLoader.cpp
Flamefire: Thinking about it: Maybe don't change the MapLoader as this is the loader from the S2 maps. We might change this or add another map format.
Maybe add a method similar to `ConvertMineResourceTypes` to `GameWorldGame` which does the whole water adjustment (including the terrain stuff). That would be more flexible, what do you think?

src/world/MapLoader.cpp
Shawn8901: Code was removed.

src/world/MapLoader.cpp
Shawn8901: Code is moved to GameWorldGame.

src/addons/AddonExhaustibleWater.h
Shawn8901: Added the hint to the Addon-Description.

src/world/GameWorldGame.cpp
```diff
@@ -1315,6 +1315,31 @@ void GameWorldGame::ConvertMineResourceTypes(Resource::Type from, Resource::Type
     }
 }
 
+void GameWorldGame::FillWaterEverywhere()
```
Flamefire: I would move ALL the water code here. So name it e.g. `PlaceAndFixWater`, check if there is water on the node or we want water everywhere and then replace it with the appropriate amount of water (as in the MapLoader).
Then you can remove the current code from MapLoader where it changes the amount of water depending on terrain and simply place water there.
And while you are on it: Why not check ALL terrains around the point for unsuitable terrain or e.g. start with 7 water and reduce (to 0 if required) if any terrain around can't hold that much water. You can use `GetRightTerrain` and iterate over the 6 directions (see e.g. `IsSeaPoint`)

src/world/GameWorld.cpp
```diff
@@ -21,6 +21,7 @@
 #include "GlobalGameSettings.h"
 #include "Loader.h"
 #include "SerializedGameData.h"
+#include "addons/const_addons.h"
```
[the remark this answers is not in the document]
Shawn8901: Is there any tool or setting which hints this information or to auto organize those imports?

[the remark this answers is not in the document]
Flamefire: There is a clang tool "iwyu" but it's a bit difficult to use. I made a script for that I think.
In this case I just checked the diff
For the order we use clang-format (custom build target created if it is found)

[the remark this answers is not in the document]
Shawn8901: Yeah this one was easy, the question was for a more general intent.
There exists a nice VS addon for iwyu, thanks for the hint.

src/world/GameWorldGame.cpp
Shawn8901: I moved the code as discussed. 
I hope my refactoring for IsSeaPoint and IsWaterPoint is acceptable. i didnt want to copy & paste the evaluation so i refactored to a more flexible way to evaluate the surrounding terrain.

src/world/GameWorldGame.cpp
```diff
@@ -1315,6 +1315,54 @@ void GameWorldGame::ConvertMineResourceTypes(Resource::Type from, Resource::Type
...
+                amount = 4;
+            else if (World::IsOfTerrain(pt, TT_STEPPE))
+                amount = 2;
+            else
+                amount = 7;
+        }
+    
+        if (amount != 0)
+        {
+            curNodeResource.setType(Resource::Water);
```
Flamefire: `curNodeResource = Resource(Resource::Water, amount)`

src/world/World.h
```diff
@@ -148,6 +148,10 @@ class World : public MapBase
     /// Create the FOW-objects, -streets, etc for a point and player
     void SaveFOWNode(const MapPoint pt, const unsigned player, unsigned curTime);
     unsigned GetNumSeas() const { return seas.size(); }
+    // returns if a map point is surrounded by a given TerrainType. Result is evaluated by the given predicate.
```
Flamefire: * Return (use command like description, shorter and a good convention)

Better for this: `Return if all surrounding terrains match the given predicate`

src/world/World.h
```diff
@@ -148,6 +148,10 @@ class World : public MapBase
...
+    bool IsOfTerrain(const MapPoint pt, bool(*terrainPredicate)(TerrainType)) const;
+    // returns if a map point is surrounded by a given TerrainType
```
Flamefire: `Return`
And maybe add `fully surrounded` to distinguish this better.

Flamefire:
I like the refactoring if `IsOfTerrain`. However your code has an issue: If the point is partially lava and `waterEverywhere = true` then you would place water there. Also if the point has `TT_STEPPE` and `TT_SAVANNAH` you would place 7 water.
So you need another function `HasTerrain` for this.

src/world/GameWorldGame.cpp
Shawn8901: Yeah true, while coding i did not realize that Resource is a struct and i was somehow confused that there is no constructor....

Flamefire: Hm? There is a Ctor

Shawn8901:
Yeah, your right for that. I was unsure about this code and tried a unit test, but i am not very familiar with boost, this test suite and also with the rest of the rttr code, that i have given up on this.

src/world/GameWorldGame.cpp
Shawn8901: you may see that i am not programming mainly in c++. i meant there is no new operator... :-)

Flamefire:
If you are unsure just ask. I'm happy to help and glad that there is someone else helping out :)
Last thing: Is it intended that savannah+steppe fields have 4 water? I don't really care so I leave it up to you.

And a tip: You can also rebase on the current master. This avoids the merge commits in the PR. If you are using TortoiseGit this is very easy. Only thing is that you need to "force push" to your remote repo (tick the box), so only do that when it is not yet merged.

Shawn8901:
If have now switched STEPPE and SAVANNAH, so the lower amount is prefered.

I also tried to rebase. Please see the changes.

Im not this familiar with git, i hope i did everything right.

Flamefire:
Looks perfect :)
Minor note: Your first commit shows "Jörg" as the committer. If you want to change that, rebase again (force) and amend with changing the name. Otherwise I'd merge that once CI runs through. Just give me a note.

Shawn8901:
I have changed the commiter. So from my point of view, you can merge the pr.

Flamefire:
@Shawn8901 and @Spikeone I'm currently working on the terrain data. Why does water terrain prohibit water for wells? Do we want to keep that? This means on the coast won't be any water for wells...

Issue #762, opened 2017-11-20
Czech font chars


FELiXik:
Added czech font chars in size "1", size "2" will follow.

Can I test created font at any place? Eg. in "Read me"? I tested there size "1", but can I use any "tags" to display there text in two other sizes???

And can I ask in which case is used size 0? I didnt find it nowhere...

(continue of bug tracking on old place https://bugs.launchpad.net/bugs/371755 )
-----------------------------------------------------------------------------------------------------------
LST containing this folowing czech chars:

AaÁá
Bb
CcČč
DdĎď
EeÉéĚě
Ff
Gg
Hh
IiÍí
Jj
Kk
Ll
Mm
NnŇň
OoÓó
Pp
Qq
RrŘř
SsŠš
TtŤť
UuÚúŮů
Vv
Ww
Xx
YyÝý
ZzŽž

test text: ĚŠČŘŽÝÁÍÉÚŮ

[OUTLINE_FONTS.zip](https://github.com/Return-To-The-Roots/s25client/files/1487661/OUTLINE_FONTS.zip)

Flamefire:
Finally I can reply to this with the good news: You can test the fonts in the developer test screen. Simply double click the version number on the main screen to access it. There you have an edit box for typing in letters that are displayed in a text control next to it. With the drop down you can change the size so you can test all 3.

And yes size 0 is also used. E.g. the FPS and status display at the top use it. Just search for `SmallFont` to find more instances.

FELiXik:
Hello,
thank you a very much!

May I ask you for latest build? I tryied to download repository, VisualStudion and install all necessary in manual, but I failed on "booststrap" install (install ends with error). I´m just no SW programmer :)

Flamefire:
So your boost build failed? If only `b2 install` failed it should still work. Building a version for you now and will put it in this post.

Shawn8901:
@FELiXik there is a newer build (28.11.2017) available at rttr.info, so everything which was available on 30th Nov should be included in this version.

xatupal:
Hey! This one can probably be closed. Slovakian, Czech and Polish special characters were [recently added](https://github.com/Return-To-The-Roots/s25client/pull/1635).

Spikeone:
hey @xatupal - care to test and verify, that it works?

Flow86:
closed via #1635

PR #763, opened 2017-11-21
Fixes

Flamefire:
Fix pasting of special chars in WinAPI and add Windows pasting to SDL
Change Toggle GCs to Set GCs to avoid accidental double setting
Fix issues related to async detection and handling
Add font test part to test screen for #762

Also fix border gasps (#81):
Old: borderbloggs2
New: borderbloggs3


Flamefire:
@Spikeone Fixed the border gaps. What do you say?
While I'm on it: What shall we do with ties (2 military blds have the same distance to a point)? I don't really like that the last built building gets it. I would say: Current owner keeps the point. If there is none, the older building gets it (first built).

PR #770, opened 2017-12-01
Another try on compile fixes

Flamefire:
- Enables editor
- Removes SDL_opengl on editor
- Use typename on clang

@Flow86 Could you tell me, what happens with clang when you don't use `typename`? Haven't seen a failure yet.


Flamefire:
The apple universal is a little endian?

PR #771, opened 2017-12-01
Compile fixes

Flamefire:
Fix the endian dependency of the editor and remove the `typename` for clang


Flamefire:
@Flow86 I removed the typename and it works. What was the problem? Or did I miss anything?

Flamefire:
Well ok. Then this is a proper fix and I won't complain. Maybe it failed on something else?

I'll watch Jenkins for further failures. If any, expect my PR ;-)

Flamefire:
Succeeded. What do you think about another announcement? Like happy Christmas time, new nightly available which includes an editor.

Issue #517, opened 2016-07-25
Cannot break up with team with option `lockedTeams=false`


lvanderwall:
**Version:** 20160725 - 1ac0870

If `lockedTeams=false`, I expect that I could break up with my team members.

But players who start in the same team (in the settings dialogue) still cannot change their diplomacy settings.

Example settings:
![settings](https://cloud.githubusercontent.com/assets/16540971/17098228/91905e34-5261-11e6-97aa-574b735e4890.png)

Diplomacy: Can change alliances with AI2, but not with AI1:
![diplomacy](https://cloud.githubusercontent.com/assets/16540971/17098264/baab28ee-5261-11e6-948b-d7485b5aea34.png)

MarcusSt:
The other party needs to agree to cancel a pact prematurely. At the moment, the AI does handle any diplomacy, so it won't agree to cancel the pact. Human players will get message "The player '%s' want to cancel the '%s' between you both prematurely. Do you agree?". Only if both parties agree, the pact is cancelled.

lvanderwall:
Ok, I'll bear that in mind. Perhaps an easy AI diplomacy model might be that the AI will always agree to cancel a pact, but never to sign new treaties. That should resemble the original game:

- the AI respects pacts predefined by the settings
- the AI becomes hostile as soon as the human player attacks

Flamefire:
Maybe a Lua event would be good so you can decide that in missions: onContractCancel: Accept/Deny

lvanderwall:
An interesting idea! I would still like to have the AI model above as default diplomacy behavior (to stay as close to S2 as possible), but you could override that with onContractCancel() if you want your own settings.

E.g. you could set a time frame in which players are allowed to attack the AI...

Shawn8901:
@Flamefire @lvanderwall can some one please check if the pushed code (#773) fixes this issue?

Flamefire:
Fixed

PR #773, opened 2017-12-02
AI will accept pact cancel & basic lua interface

Shawn8901:
This commit adds the logic discussed in issue #517, that the ai will accept cancel requests.
In addition i have added basic checks for the pact state and the possibility to suggest a pact to the player.

I skipped the implementation for accepting or starting a cancel request by AI, because i had some problems with finding a GameCommandFactory (i guess im missing this), which is usable for this intent.

But despite that, i am creating this pull request, cause it resolves the discussed logic from issue #517.

### DIFF
--- a/src/SerializedGameData.cpp
+++ b/src/SerializedGameData.cpp
@@ -247,7 +247,7 @@ void SerializedGameData::MakeSnapshot(const GameWorld& gw)
     writtenEventIds.clear();
 }
 
-void SerializedGameData::ReadSnapshot(GameWorld& gw)
+void SerializedGameData::ReadSnapshot(boost::weak_ptr<Game> game, GameWorld& gw)
 {
     Prepare(true);
 
@@ -255,7 +255,7 @@ void SerializedGameData::ReadSnapshot(GameWorld& gw)
 
     expectedNumObjects = PopUnsignedInt();
 
-    gw.Deserialize(*this);
+    gw.Deserialize(game, *this);
     em->Deserialize(*this);
     for(unsigned i = 0; i < gw.GetNumPlayers(); ++i)
         gw.GetPlayer(i).Deserialize(*this);
--- a/src/SerializedGameData.cpp
+++ b/src/SerializedGameData.cpp
@@ -20,6 +20,7 @@
 #include "CatapultStone.h"
 #include "EventManager.h"
 #include "FOWObjects.h"
+#include "Game.h"
 #include "GameEvent.h"
 #include "GameObject.h"
 #include "GamePlayer.h"
@@ -209,10 +210,11 @@ void SerializedGameData::Prepare(bool reading)
     isReading = reading;
 }
 
-void SerializedGameData::MakeSnapshot(const GameWorld& gw)
+void SerializedGameData::MakeSnapshot(boost::shared_ptr<Game> game)
 {
     Prepare(false);
-
+    
+    GameWorld& gw = game->world;
     writeEm = &gw.GetEvMgr();
 
     // Anzahl Objekte reinschreiben (used for safety checks only)
@@ -247,10 +249,11 @@ void SerializedGameData::MakeSnapshot(const GameWorld& gw)
     writtenEventIds.clear();
 }
 
-void SerializedGameData::ReadSnapshot(boost::weak_ptr<Game> game, GameWorld& gw)
+void SerializedGameData::ReadSnapshot(boost::shared_ptr<Game> game)
 {
     Prepare(true);
 
+    GameWorld& gw = game->world;
     em = &gw.GetEvMgr();
 
     expectedNumObjects = PopUnsignedInt();
--- a/src/ai/AIPlayer.h
+++ b/src/ai/AIPlayer.h
@@ -50,6 +50,9 @@ public:
         return tmp;
     }
 
+    // access to ais CommandFactory
+    const AIInterface& getAIInterface() const { return aii; }
+
     /// Eigene PlayerId, die der KI-Spieler wissen sollte, z.B. wenn er die Karte untersucht
     const unsigned char playerId;
     /// Verweis auf den eigenen GameClientPlayer, d.h. die Wirtschaft, um daraus entsprechend Informationen zu gewinnen
--- a/src/lua/LuaInterfaceGame.cpp
+++ b/src/lua/LuaInterfaceGame.cpp
@@ -34,7 +34,7 @@
 #include "libutil/Serializer.h"
 #include <boost/nowide/fstream.hpp>
 
-LuaInterfaceGame::LuaInterfaceGame(boost::weak_ptr<Game> game) : game(game), gw(game.lock()->world)
+LuaInterfaceGame::LuaInterfaceGame(boost::weak_ptr<Game> gameInstance) : gw(gameInstance.lock()->world), game(gameInstance)
 {
 #pragma region ConstDefs
 #define ADD_LUA_CONST(name) lua[#name] = name
--- a/src/lua/LuaInterfaceGame.cpp
+++ b/src/lua/LuaInterfaceGame.cpp
@@ -327,7 +327,7 @@ LuaPlayer LuaInterfaceGame::GetPlayer(unsigned playerIdx)
 {
     if(playerIdx >= gw.GetNumPlayers())
         throw std::runtime_error("Invalid player idx");
-    return LuaPlayer(gw.GetPlayer(playerIdx));
+    return LuaPlayer(game, gw.GetPlayer(playerIdx));
 }
 
 LuaWorld LuaInterfaceGame::GetWorld()
@@ -390,7 +390,8 @@ bool LuaInterfaceGame::EventCancelPactRequest(PactType pt, unsigned char cancele
 
 void LuaInterfaceGame::EventSuggestPact(const PactType pt, unsigned char suggestedByPlayerId, unsigned char targetPlayerId, const unsigned duration)
 {
-    const AIPlayer* ai = GAMECLIENT.GetAIPlayer(targetPlayerId);
+    Game& gameInst = *game.lock();
+    AIPlayer* ai =  gameInst.GetAIPlayer(targetPlayerId);
     if (ai != NULL)
     {
         kaguya::LuaRef onPactCancel = lua["onSuggestPact"];
--- a/src/lua/LuaInterfaceGame.cpp
+++ b/src/lua/LuaInterfaceGame.cpp
@@ -23,7 +23,10 @@
 #include "ingameWindows/iwMissionStatement.h"
 #include "lua/LuaPlayer.h"
 #include "lua/LuaWorld.h"
+#include "gameTypes/PactTypes.h"
 #include "network/GameClient.h"
+#include "ai/AIPlayer.h"
+#include "ai/AIInterface.h"
 #include "postSystem/PostMsg.h"
 #include "world/GameWorldGame.h"
 #include "gameTypes/Resource.h"
@@ -139,6 +142,11 @@ LuaInterfaceGame::LuaInterfaceGame(GameWorldGame& gw) : gw(gw)
     lua["RES_COAL"] = Resource::Coal;
     lua["RES_GRANITE"] = Resource::Granite;
     lua["RES_WATER"] = Resource::Water;
+
+    lua["NON_AGGRESSION_PACT"] = PactType::NON_AGGRESSION_PACT;
+    lua["TREATY_OF_ALLIANCE"] = PactType::TREATY_OF_ALLIANCE;
+    // infinite pact duration, see GamePlayer::GetRemainingPactTime
+    lua["DURATION_INFINITE"] = 0xFFFFFFFF;
 
 #undef ADD_LUA_CONST
 #define ADD_LUA_CONST(name) lua[#name] = iwMissionStatement::name
@@ -370,4 +378,48 @@ void LuaInterfaceGame::EventResourceFound(unsigned char player, const MapPoint p
     kaguya::LuaRef onResourceFound = lua["onResourceFound"];
     if(onResourceFound.type() == LUA_TFUNCTION)
         onResourceFound.call<void>(player, pt.x, pt.y, type, quantity);
+}
+
+bool LuaInterfaceGame::EventCancelPactRequest(PactType pt, unsigned char canceledByPlayerId, unsigned char targetPlayerId)
+{
+    kaguya::LuaRef onPactCancel = lua["onCancelPactRequest"];
+    if (onPactCancel.type() == LUA_TFUNCTION)
+        return onPactCancel.call<bool>(pt, canceledByPlayerId, targetPlayerId);
+    return true; // always accept pact cancel if there is no handler
+}
+
+void LuaInterfaceGame::EventSuggestPact(const PactType pt, unsigned char suggestedByPlayerId, unsigned char targetPlayerId, const unsigned duration)
+{
+    const AIPlayer* ai = GAMECLIENT.GetAIPlayer(targetPlayerId);
+    if (ai != NULL)
+    {
+        kaguya::LuaRef onPactCancel = lua["onSuggestPact"];
+        if (onPactCancel.type() == LUA_TFUNCTION)
+        {
+            AIInterface aii = ai->getAIInterface();
+            bool luaResult = onPactCancel.call<bool>(pt, suggestedByPlayerId, targetPlayerId, duration);
+            if (luaResult)
+                aii.AcceptPact(gw.GetEvMgr().GetCurrentGF(), pt, suggestedByPlayerId);
+            else
+                aii.CancelPact(pt, suggestedByPlayerId);
+        }
+    }
+}
+
+void LuaInterfaceGame::EventPactCanceled(const PactType pt, unsigned char canceledByPlayerId, unsigned char targetPlayerId)
+{
+    kaguya::LuaRef onPactCanceled = lua["onPactCanceled"];
+    if (onPactCanceled.type() == LUA_TFUNCTION)
+    {
+        onPactCanceled.call<void>(pt, canceledByPlayerId, targetPlayerId);
+    }
+}
+
+void LuaInterfaceGame::EventPactCreated(const PactType pt, unsigned char suggestedByPlayerId, unsigned char targetPlayerId, const unsigned duration)
+{
+    kaguya::LuaRef onPactCreated = lua["onPactCreated"];
+    if (onPactCreated.type() == LUA_TFUNCTION)
+    {
+        onPactCreated.call<void>(pt, suggestedByPlayerId, targetPlayerId, duration);
+    }
 }
--- a/src/lua/LuaInterfaceGame.cpp
+++ b/src/lua/LuaInterfaceGame.cpp
@@ -23,7 +23,6 @@
 #include "ingameWindows/iwMissionStatement.h"
 #include "lua/LuaPlayer.h"
 #include "lua/LuaWorld.h"
-#include "gameTypes/PactTypes.h"
 #include "network/GameClient.h"
 #include "ai/AIPlayer.h"
 #include "ai/AIInterface.h"
@@ -143,8 +142,8 @@ LuaInterfaceGame::LuaInterfaceGame(GameWorldGame& gw) : gw(gw)
     lua["RES_GRANITE"] = Resource::Granite;
     lua["RES_WATER"] = Resource::Water;
 
-    lua["NON_AGGRESSION_PACT"] = PactType::NON_AGGRESSION_PACT;
-    lua["TREATY_OF_ALLIANCE"] = PactType::TREATY_OF_ALLIANCE;
+    lua["NON_AGGRESSION_PACT"] = NON_AGGRESSION_PACT;
+    lua["TREATY_OF_ALLIANCE"] = TREATY_OF_ALLIANCE;
     // infinite pact duration, see GamePlayer::GetRemainingPactTime
     lua["DURATION_INFINITE"] = 0xFFFFFFFF;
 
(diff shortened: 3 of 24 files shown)


src/lua/LuaInterfaceGame.cpp
```diff
@@ -140,6 +140,9 @@ LuaInterfaceGame::LuaInterfaceGame(GameWorldGame& gw) : gw(gw)
     lua["RES_GRANITE"] = Resource::Granite;
     lua["RES_WATER"] = Resource::Water;
 
+    lua["PACTTYPE_NON_AGGRESSION"] = PactType::NON_AGGRESSION_PACT;
```
Flamefire: What about simply `NON_AGGRESSION_PACT`?

src/lua/LuaInterfaceGame.cpp
```diff
@@ -371,3 +374,12 @@ void LuaInterfaceGame::EventResourceFound(unsigned char player, const MapPoint p
     if(onResourceFound.type() == LUA_TFUNCTION)
         onResourceFound.call<void>(player, pt.x, pt.y, type, quantity);
 }
+
+
+bool LuaInterfaceGame::EventCancelPact(PactType pt, unsigned char canceledByPlayerId, unsigned char targetPlayerId)
```
Flamefire: I'd be more verbose: `onPactCancelRequest`. Otherwise it might interfere if we add a real `onCancelPact`. Or we name that other one `onPactCanceled`. But I'd go with both to avoid confusion.

src/lua/LuaInterfaceGame.h
```diff
@@ -43,6 +44,10 @@ class LuaInterfaceGame : public LuaInterfaceBase
     void EventStart(bool isFirstStart);
     void EventGameFrame(unsigned number);
     void EventResourceFound(unsigned char player, const MapPoint pt, unsigned char type, unsigned char quantity);
+    // Called if player wants to cancel a pact
+    bool EventCancelPact(PactType pt, unsigned char canceledByPlayerId, unsigned char targetPlayerId);
+    // Called if pact is suggested
+    void EventSuggestPact(unsigned int startFrame, PactType pt, unsigned char suggestedByPlayerId, unsigned char targetPlayerId, unsigned int duration);
```
Flamefire: Implementation is missing (was removed). Check the iwPostWindow for correct handling. And don't pass the GF to lua. It is only required by the response handling.

src/lua/LuaPlayer.cpp
```diff
@@ -294,3 +297,18 @@ kaguya::standard::tuple<unsigned, unsigned> LuaPlayer::GetHQPos() const
...
+}
+
+bool LuaPlayer::IsAttackable(unsigned char otherPlayerId) 
+{
+    return player.IsAttackable(otherPlayerId);
+}
+
+void LuaPlayer::SuggestPact(unsigned char otherPlayerId, const PactType pt, const unsigned duration)
+{
+    player.SuggestPact(otherPlayerId, pt, duration);
```
Flamefire: we need a constant `INFINITE` or `DURATION_INFINITE`/`DURATION_FOREVER` in lua.

src/lua/LuaPlayer.cpp
```diff
@@ -59,6 +59,9 @@ void LuaPlayer::Register(kaguya::State& state)
                                .addFunction("GetHQPos", &LuaPlayer::GetHQPos)
                                .addFunction("IsDefeated", &LuaPlayer::IsDefeated)
                                .addFunction("Surrender", &LuaPlayer::Surrender)
+                               .addFunction("IsAlly", &LuaPlayer::IsAlly)
```
Flamefire: Great ideas!

src/lua/LuaInterfaceGame.cpp
Shawn8901: I have changed both, so ALLIANCE is also TREATY_OF_ALLIANCE now.

src/lua/LuaInterfaceGame.cpp
Shawn8901: Im ok with changing it to onPactCancelRequest.

src/lua/LuaInterfaceGame.h
Shawn8901: Yeah is had a not working implementation for that, and forgot to remove it from header file.
iwPostWindow calls GAMECLIENT as GameCommandFactory. There is the problem, that if i import GAMECLIENT as singleton instance (from header file), the GameCommandFactory builds correct instances for the host and not for the AI. 
So this game commands are not valid for the AI (tries to suggest a pact between player 0 and player 0).

is there any way to load a GameCommandFactory for another player?

About passing the GF:
While accepting the pact, the script needs to know the PactID (which is also the GF when the pact was suggested, if im right). So where should get script get knowledge of this id instead of this event handle? PactID would be a better naming. On the otherhand i could generate a "custom" id and remap this to the GF?

Flamefire: Damn, you are right. One would need the actual AI player for that (currently in GAMESERVER)

Not really: You call the lua function and that returns true or false. Then you call the accept with the GF you still have.

Shawn8901:
I have reworked the way the lua interface handles actions performed by the script.
It now redirects the game commands to the AIInterface, to build a valid Gamecommand. i assume that only AIs should be scripted. If the host player should be scriptable also, i have to add the "normal" way to build a game command via GAMECLIENT.

I hope that this implementation fits more the needs of the project.

For the testcase:
I was trying to build a test case, but i did not find a way to extends/combine the LuaTestWorld with GCs. For me it seems to be that the LuaTestWorld, does not have the needed structures for the AIPlayers initialized.

src/ai/AIPlayer.h
```diff
@@ -44,6 +44,9 @@ class AIPlayer
 
     /// Zugriff auf die GameCommands, um diese abarbeiten zu können
     const std::vector<gc::GameCommandPtr>& GetGameCommands() const { return gcs; }
+    // access to ais CommandFactory
+    const AIInterface getAIInterface() const { return aii; }
```
Flamefire: You want a reference here not a copy

src/lua/LuaInterfaceGame.cpp
```diff
@@ -385,3 +388,18 @@ bool LuaInterfaceGame::EventCancelPactRequest(PactType pt, unsigned char cancele
...
     return true; // always accept pact cancel if there is no handler
 }
+
+void LuaInterfaceGame::EventSuggestPact(const PactType pt, unsigned char suggestedByPlayerId, unsigned char targetPlayerId, const unsigned duration)
+{
+    kaguya::LuaRef onPactCancel = lua["onSuggestPact"];
+    if (onPactCancel.type() == LUA_TFUNCTION)
+    {
+        bool luaResult = onPactCancel.call<bool>(pt, suggestedByPlayerId, targetPlayerId, duration);
+        if (luaResult)
```
Flamefire: Why not take both results? True = Accept, False = Cancel/Decline

src/lua/LuaPlayer.cpp
```diff
@@ -310,5 +313,32 @@ bool LuaPlayer::IsAttackable(unsigned char otherPlayerId)
...
-    player.SuggestPact(otherPlayerId, pt, duration);
+    if (!player.isHuman())
+    {
+        AIPlayer* ai = GAMESERVER.GetAIPlayer(player.GetPlayerId());
+        AIInterface aii = ai->getAIInterface();
+        aii.SuggestPact(otherPlayerId, pt, duration);
+    }
 }
+
+void LuaPlayer::AcceptPact(const PactType pt, unsigned char otherPlayerId)
```
Flamefire: This won't work. The GF passed to `AcceptPact` must be the GF where the pact was suggested. So you'd need to get it from the pact structs.
What is this method required for anyway? You have `EventSuggestPact` already.

src/lua/LuaPlayer.cpp
```diff
@@ -310,5 +313,32 @@ bool LuaPlayer::IsAttackable(unsigned char otherPlayerId)
...
+        AIPlayer* ai = GAMESERVER.GetAIPlayer(player.GetPlayerId());
+        AIInterface aii = ai->getAIInterface();
+        GameWorldGame& world = player.GetGameWorld();
+        aii.AcceptPact(world.GetEvMgr().GetCurrentGF(), pt, otherPlayerId);
+    }
+}
+
+void LuaPlayer::CancelPact(const PactType pt, unsigned char otherPlayerId) 
+{
+    if (!player.isHuman())
```
Flamefire: Good idea to have this as an extra function. Then accompany this with `onPactStateChanged(bool true=Created, false=canceled, ...)` (or `onPactCreated`,  `onPactCanceled`)

Flamefire:
@Shawn8901 I have merged the commit I was talking about. The AI is now in the GameClient not the GameServer. You can rebase on current master and fix the remaining issues :)

src/ai/AIPlayer.h
Shawn8901: Done, somehow its exhausting to care of that, as daily java developer...

src/lua/LuaInterfaceGame.cpp
Shawn8901: Done, aii now calls cancel pact if lua methods returns false.

src/lua/LuaPlayer.cpp
Shawn8901: I am a bit unsure, do you mean/prefer a method on LuaPlayer (which Lua may subscribe if its possbile(??)) or a Callback-Handler in LuaInterfaceGame (like EventSuggestPact)?

Flamefire: I meant 2 extra events for lua like EventSuggestPact. Then we have pact suggested, approved, denied and canceled covered with events.

src/lua/LuaPlayer.cpp
Shawn8901: you are right, i removed this.

Shawn8901:
I have added the requested OnPactCreated/OnPactCanceled. 
While Pull/Rebease from Upstream to master i managed it somehow to add your commits.
Am i right that this will result in trouble when merging?
So i would guess i have to remove them again...?

Flamefire:
You have a merge somewhere which looks odd. Do a git fetch origin and then rebase on origin/master (force if necessary) If you use tortoiseGit it is very easy. Otherwise you might want an interactive rebase (-i) and skip the merge

Shawn8901:
I think i have managed it.
I also combined all commits to one, to no mess up the commit history.

Flamefire:
Looks good to me. Just some compile failures. Did you check the changes manually? just write a simple lua script, start a game and check if it works as intended.

Shawn8901:
I have a basic script for that testing purpose.
player 1 needs to be a ai.
If you offer a non aggresion pact, the ai offers an alliance.
it also tries to cancel all pacts.
Its spamming cancel requests (because i dont check if there was a request sent), but for testing its okay.

[Random.zip](https://github.com/Return-To-The-Roots/s25client/files/1579868/Random.zip)

for automated testing i am still unsure how to get a "working" gameclient. for the existing lua test cases the player/ai structure seems not to be initialized, so executing the lua script will not result in the expected pact-states.

i think i have fixed all errors.

Flamefire:
Yeah I know the GAMECLIENT issue for testing. I was trying to mitigate this. The AIs are stored in the `Game` class now. We could store a weak pointer in `LuaGameInterface` and `LuaPlayer`. Then you can access the AI w/o `GAMECLIENT` and do automated tests.

If you want to do that as additional commits on this PR it would be great and improve code coverage. If not I'm fine with the current changes and would merge them.

Shawn8901:
I am fine with this idea. i would take a look into it after Christmas.

Shawn8901:
hi, 
i think i have a few questions left.
With your comment, did you mean to store the game instance or the AI instances? For me i don't find a way to an accessible game instance w/o invoking anything from the GameClient. 

So i would guess, that i am storing an pointer to the AIs in LuaInterface, and the GameClient itself could store a pointer from the AI (if the game starts) to LuaInterface. 
Maybe i would then store a direct AI-Pointer in LuaPlayer (if the LuaPlayer is an AI).

Due this i would be able to set the AI from GameWorldWithLuaAccess, which should result in an accessible AI-Instance for the test cases.
I just want to know if this the way your suggested.
Thank you very much.

Flamefire:
I was thinking about changing `LuaInterfaceGame` to hold a weak ptr to `Game` instead of a reference to `GameWorldGame`. This means changing `GameWorld::LoadMap` and `GameWorld::Deserialize` to accept said pointer. Then you have access to the game which means the world (as before) and the AI too. Check the call tree for e.g. `GameWorld::LoadMap`. It is called from GameClient which has access to the `Game`

You should not store pointers to AI as they might change. Also the AIs are not stored according to their playerIDs so you have to search for the right one. Check `GameClient::GetAIPlayer` for that.

Shawn8901:
Hi,
ok i think i have somehow managed the most of the requested refactoring.
One point i dont really like is, that i had to change the visibility of GameWordBase::lua from protected to public to access it in GameWorldWithLuaAccess. 

Now i am at the point to refactor the `GameWorldWithLuaAccess` (which will result later on in a `GameWithLuaAccess`).
There I don't know how to go further. `LuaInterfaceGame` needs a weak_ptr, i can serve a weak_ptr or shared_ptr, but no raw pointer. If i am trying to convert the raw pointer with boost::make_shared im getting a C2280 
`d:\workspace\boost\include\boost-1_65_1\boost\smart_ptr\make_shared_object.hpp(256): error C2280: 'Game::Game(const Game &)': attempting to reference a deleted function (compiling source file D:\Workspace\s25client\src\test\testLuaGUI.cpp)`
After googling around it seems to be, that make_shared tries to copy the object, which is not possbile, because of the missing copy function. 
at this point i don't really know what to do.
Am i right with this error? 
i have commited my current state, which compiles but will fail in testing.

src/SerializedGameData.cpp
```diff
@@ -247,15 +247,15 @@ void SerializedGameData::MakeSnapshot(const GameWorld& gw)
     writtenEventIds.clear();
 }
 
-void SerializedGameData::ReadSnapshot(GameWorld& gw)
+void SerializedGameData::ReadSnapshot(boost::weak_ptr<Game> game, GameWorld& gw)
```
Flamefire: If you pass the game here, you don't need the gw anymore as it is part of game. Also use a shared_ptr (see other comment on LoadMap)

src/test/testSerialization.cpp
```diff
@@ -262,10 +263,13 @@ BOOST_FIXTURE_TEST_CASE(BaseSaveLoad, RandWorldFixture)
             std::vector<PlayerInfo> players;
             for(unsigned j = 0; j < 4; j++)
                 players.push_back(PlayerInfo(loadSave.GetPlayer(j)));
-            GlobalGameSettings newGGS = save.ggs;
+            GlobalGameSettings& newGGS = save.ggs;
             TestEventManager newEm(loadSave.start_gf);
```
Flamefire: these 2 lines are not required anymore (those are in `Game`)

src/test/testSerialization.cpp
```diff
@@ -262,10 +263,13 @@ BOOST_FIXTURE_TEST_CASE(BaseSaveLoad, RandWorldFixture)
...
             GameWorld newWorld(players, newGGS, newEm);
-            save.sgd.ReadSnapshot(newWorld);
+            Game newGame(newGGS, loadSave.start_gf, players);
+            boost::shared_ptr<Game> sharedGame;
+            sharedGame.reset(&newGame);
```
Flamefire: This is not how a shared ptr works. Most importantly you need an actual pointer, so `new Game(...)` or make the `shared_ptr` not delete it's pointer.
`make_shared` is simply a wrapper around `new`. So what you want is: `boost::shared_ptr<Game> sharedGame = boost::make_shared<Game>(newGGS, loadSave.start_gf, players)` or `boost::shared_ptr<Game> sharedGame(new Game(newGGS, loadSave.start_gf, players))`

src/world/GameWorld.cpp
```diff
@@ -37,7 +37,7 @@ GameWorld::GameWorld(const std::vector<PlayerInfo>& playerInfos, const GlobalGam
 {}
 
 /// Lädt eine Karte
-bool GameWorld::LoadMap(const std::string& mapFilePath, const std::string& luaFilePath)
+bool GameWorld::LoadMap(boost::weak_ptr<Game> game, const std::string& mapFilePath, const std::string& luaFilePath)
```
Flamefire: Use a shared ptr here. Weak ptr semantic is 'can still contain a pointer but may not'. But here we actually want to ensure that we have a valid pointer.

src/test/GameWorldWithLuaAccess.h
```diff
@@ -33,18 +33,23 @@
...
 #include "libutil/colors.h"
+#include "Game.h"
 #include <boost/test/unit_test.hpp>
+#include <boost/weak_ptr.hpp>
+#include <boost/shared_ptr.hpp>
+#include <boost/make_shared.hpp>
 #include <vector>
 
-class GameWorldWithLuaAccess : public GameWorldGame
+class GameWithLuaAccess : public Game
```
Flamefire: Idea: Replace this by a simple function that creates a `shared_ptr<Game>`, creates the lua in it and return it.
The fixtures need to store a shared_ptr then and call this function to init it.

src/world/GameWorldBase.h
```diff
@@ -62,6 +61,8 @@ class GameWorldBase : public World
     GameWorldBase(const std::vector<GamePlayer>& players, const GlobalGameSettings& gameSettings, EventManager& em);
     ~GameWorldBase() override;
 
+
+    boost::interprocess::unique_ptr<LuaInterfaceGame, Deleter<LuaInterfaceGame> > lua;
```
Flamefire: Instead of making it public create a function `SetLua(LuaInterfaceGame*)` and then you could even make it private as we have `SetLua`, `HasLua` and `GetLua`

src/SerializedGameData.cpp
Shawn8901: Yeah, before i had problems with compiling, thats the reason i keept GameWorld as parameter.
I also moved MakeSnapshop to Game (to keep this API equal), i hope its okay!

src/test/testSerialization.cpp
Shawn8901: Yeah i was just playing around with compiler errors. im not an c/c++ expert, so boost magic confuses me more and i was happy that it compiles at this point.

src/SerializedGameData.cpp
Flamefire: Just tell me what problems you have and we can resolve them together. I'm happy to help :)

src/test/testSerialization.cpp
Flamefire: no worries. `shared_ptr` is part of C++ now, so it's worth learning. All those `smart_ptr`s are just wrappers around a raw pointer to automatically delete the pointer when no one references it anymore.
But you can just do what you think should work and amend your last commit and we go from there.

Shawn8901:
i think i have somehow managed to resolve most of the requested changes.
Now the game itself compiles fine for me.
I am now at the point to work on the test cases, but before i want to start i am trying to ensure, that everything is working.
Currently i am at the point that the test case GameFunctions of testLua.cpp is failing because the HQs are not filled with Wares after my refactoring. In nobHQ the GlobalGameSettings are read via gwg->GetGGS(), this struct is in my case null and i have absolute no idea why. if i am inspecting the gwg-instance while debugging it is not null.
The GlobalGameSettings instance which is handed over from GameWithLua-Access seems also fine.
Do you have any suggestions for me about the problem?

src/world/GameWorldBase.cpp
```diff
@@ -697,3 +697,11 @@ void GameWorldBase::RecalcBQ(const MapPoint pt)
     if(SetBQ(pt, calcBQ(pt, boost::bind(&GameWorldBase::IsOnRoad, this, _1))))
         GetNotifications().publish(NodeNote(NodeNote::BQ, pt));
 }
+
+void GameWorldBase::SetLua(LuaInterfaceGame* newLua)
+{
+    if (newLua != NULL)
```
Flamefire: No need for the if. `lua.reset(NULL)` is the same as `lua.reset()`

src/test/WorldFixture.h
```diff
@@ -98,14 +100,14 @@ template<class T_WorldCreator, unsigned T_numPlayers = 0, unsigned T_width = Wor
...
-    Game game;
+    boost::shared_ptr<Game> game;
     TestEventManager& em;
     GlobalGameSettings& ggs;
     GameWorld& world;
     T_WorldCreator worldCreator;
     WorldFixture()
-        : game(GlobalGameSettings(), new TestEventManager, std::vector<PlayerInfo>(T_numPlayers, GetPlayer())),
-          em(static_cast<TestEventManager&>(*game.em)), ggs(const_cast<GlobalGameSettings&>(game.ggs)), world(game.world),
+        : game(boost::shared_ptr<Game>(new Game(GlobalGameSettings(), new TestEventManager, std::vector<PlayerInfo>(T_numPlayers, GetPlayer())))),
```
Flamefire: You can remove the explicit ctor of `boost::shared_ptr<Game>`. `new Game(...)` already works.

src/test/GameWithLuaAccess.h
```diff
@@ -44,12 +44,7 @@ class GameWithLuaAccess : public Game
 {
 public:
     GlobalGameSettings ggs;
```
Flamefire: Remove this. Passing it to Game() will copy it but one might wrongly assume this instance is still used. Just do `Game(GlobalGameSettings(), ...)` in Line 47

Flamefire: This is actually causing your issue as it shadows Game::ggs. Remove and it works

Flamefire:
Looks good to me. But I'm missing the test coverage from coveralls. Could you run `./checkUTF8.sh` from within the repo dir? There might be some non-utf8 file encoding.

Shawn8901:
I converted src/network/CreateServerInfo.h to UTF-8, maybe it helps.

Flamefire:
Looks great. Are you done so this can be merged?

Shawn8901:
Yes sir! i'm Done.

Flamefire:
Awesome, thank you!

PR #774, opened 2017-12-04
Load terrain data from lua

### LINKED ISSUE #784: Mountain meadow has no water
The texture Type mountain meadow seems to not contain any water anymore but there should be water.

Flamefire:
Tackling the long-term project to move the game data out of the source code so new stuff can be added more easily. E.g. once the description for buildings and units are out of the source adding new nations is a matter of copying the files and adjusting the paths in those files.

The base is a variation of an enum: `DescIdx`. It is templated on the type it describes to provide type-safety but the template param may be incomplete (aka forward declared). The data is modelled after ideas from @MarcusSt to use a string as an identifier and replace that with a simple number.
For example: We have a struct `TerrainDesc` which describes a terrain type (water, mountain, ...). It has a name with which it can be looked up in a `DescriptionContainer<TerrainDesc>`. From there you get an index `DescIdx<TerrainDesc>` which provides `O(1)` access to the data (lookup by name is `O(log n)` ). That index is a wrapped `uint8_t` so we have a very small memory footprint.
The data is loaded from a lua file. I was thinking about what file format to use. XML is to verbose, YAML would be best (very compact representation) but also requires yet another support library. We already have lua and some helpers for that. It is more verbose than YAML but way less than XML. And it is more flexible: E.g. I added a function `include` to include another lua file. So we can put each landscape, nation, ... in another file which makes this way more organized.

I started with the terrain. Currently we have the problem, that the terrain is not the same over all landscape types. So we only handle greenland properly. The others are partially faulty (e.g. no/other snow types) Putting that in data files allows us to easily add description on terrain per type. So greenland snow and wasteland snow are completely distinct. Even better: We could mix terrains from different landscapes or add new ones. Additionally we can implement the animation for the swamp and the icy water (we don't have it ATM). And finally it is a test for the externalization of the data. For the new nation it would be good if we had all data in files. Adding another nation and changing arrays scattered all over the code just doesn't do it...

The lua file contains a description of the data contained so please have a look (especially @Spikeone ) if this suffices. I opted for a combination of `kind` and `properties` to describe the terrain. So we can have water that is used by e.g. dugs, and snow, used by polar bears and separately make it shippable, buildable or whatever. All greenland terrains are correctly described in the contained file.
Only difference so far: Buildable water is not shippable. Although the data format would allow this (bitset used) I did not add it to the lua file format. Reasoning: Was it shippable in the original? Do we even want it? If we can build buildings it must be very shallow water, so it would be to shallow for ships. And finally I could not think of a good way to represent this in a short-enough form.

@Flow86 @Spikeone Please review this carefully as it is a prototype for future things to come. Tell me what you think, are missing or is unclear. When all is discussed and resolved I'll use this in the game and remove the hard-coded values.

Fixes #784


Spikeone:
@Flamefire okay main question for me at this point is: what should this support and what should it not support. What do I mean by asking such a question? Simple, should this be used for adding complete new sets without changing too much and having more default textures or should this be used by modders as well? Also I'm thinking about adding a terrain for some special mission where I could then set the maximum building quality of some terrain. An other question is, how to we make sure that each client uses the same information? Does the server distribute them? Do we want to add addon dependant terrain changes (e.g. wells dry out surrounding areas)?

Flamefire:
Mainly this is a prototype to allow modding. So adding new terrain sets, nations etc. It should also fix current bugs by having the ability to define properties per terrain and landscape.
Maximum building quality is something I would not allow. This would change to much in the existing code. You can set totally nothing, nothing, flags, mines and houses. That has to be enough.

We have the possibility to distribute it from the server. For that I added a safety feature that disallows including anything outside the main script directory. Besides that I won't check for identical files. If someone modifies his and plays with non-modified version the game will async eventually. If a group of player decide to use modded files they still can.
Addons are possible. But I'd keep them to a minimum and as generic as possible.

Flow86:
hmm, what if we dont use lua for the description files - the whole stuff could be defined via json files? that would remove the security problem which could be exist when using lua?

Flamefire:
That was what I explained above: I found a good [yaml parser](https://github.com/jbeder/yaml-cpp) and I already have an [xml implementation](https://github.com/Flamefire/s25client/tree/DynamicNations) but I opted for lua. First because we don't need another library that might fail and needs to be kept up to date and second because coders and users need to learn something in addition. We already have tools and experience to implement lua stuff and modders might know lua already, so that's easy.
Security is ok. IMO the only issue might be stack overflows due to include cycles and including files from the system, although I doubt the 2nd point. We don't add the libraries that can do any harm (network, file IO, ...)

Flamefire:
Any more comments? Otherwise I'd finish this and continue with the nations

Flamefire:
Finally done. I changed a lot to actually use the same texture are as S2 and validate this with the debug texture of http://settlers2.net/2015/09/tex5-lbm-greenland-debug-texture/

However we do have a problem: We are not actually drawing pixels from the texture but "cut" a triangle out of the texture (sharp edge, not pixelated like in the original) and stretch it to the destination. I'm not aware how we could possible fix this as this is how OGL works. So I'll keep it as-is unless anyone has a clever idea.

Issue #226, opened 2015-08-18
Caravan limit


Spikeone:
Would be great to add a caravan limit, that way you have to think twice before you start to trade with anybody as you have to wait till the caravan arrived, even longer if donkeys have to return.

Spikeone:
This addon depends on general trade design. Maybe it's useless after the rework.

Issue #218, opened 2015-08-17
Use other trade leader


Flamefire:
There was a request to change the trade leader: https://bugs.launchpad.net/s25rttr/+bug/366861

E.g. a scout might be better. But I'd say the trading caravan is not allowed to scout. E.g. with shared team view one could reveal the whole map with trading which would be a loophole in the fog of war design. So maybe a scout image only (no scouting abilities)

PR #792, opened 2018-01-03
Use Scout as TradeLeader

Shawn8901:
I have implemented the request from #218.

Additionally i have changed two things:
* i did not understand, why Helpers limits the trade for Figures, since they are not reduced if caravan leaves, i removed this limitation (maybe i am missing something?)
* removed the military recruitment "simulation" from test case, since there is no dependency left to Helpers

edit: coverage from coverall is missing because of 0664b9b i will rebase, if #773 will be merged before this. i am not sure what happens if i am cherry picking this commit to this branch, will git be able to handle this?

edit2: rebased.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,45 +28,6 @@ cache:
   - directories:
     - ${HOME}/.cache
 
-jobs:
-  include:
-    - &linux-gcc6
-      os: linux
-      compiler: gcc
-      env: MATRIX_EVAL="CXX=g++-6 && CC=gcc-6" BUILD_TYPE=Release
-      addons:
-        apt:
-          <<: *defApt
-          packages: [*defPackages, g++-6]
-    - <<: *linux-gcc6
-      env: MATRIX_EVAL="CXX=g++-6 && CC=gcc-6" GCOV=gcov-6 BUILD_TYPE=Debug ADDITIONAL_CMAKE_FLAGS="-DRTTR_ENABLE_BUILD_TESTING=ON -DRTTR_ENABLE_COVERAGE=ON"
-      after_success: tools/ci/uploadCoverageData.sh
-    - &linux-clang
-      # GCC6 headers required
-      <<: *linux-gcc6
-      compiler: clang
-      env: BUILD_TYPE=Debug ADDITIONAL_CMAKE_FLAGS=-DRTTR_EXTERNAL_BUILD_TESTING=ON
-    - <<: *linux-clang
-      env: BUILD_TYPE=Release
-    - os: osx
-      compiler: clang
-      env: BUILD_TYPE=Debug
-    # Latest versions of Clang and GCC
-    - os: linux
-      compiler: gcc
-      env: MATRIX_EVAL="CXX=g++-8 && CC=gcc-8" BUILD_TYPE=Debug
-      addons:
-        apt:
-          <<: *defApt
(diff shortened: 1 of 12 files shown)


src/buildings/nobBaseWarehouse.cpp
```diff
@@ -1408,24 +1408,28 @@ bool nobBaseWarehouse::IsDependentFigure(noFigure* fig) const
 /// Available goods of a specific type that can be used for trading
 unsigned nobBaseWarehouse::GetAvailableWaresForTrading(const GoodType gt) const
 {
-    // We need a helper as leader
-    if(!inventory[JOB_HELPER])
+    // We need a scout as leader or a bow to recruit
+    bool noScoutAvailable = !inventory[JOB_SCOUT];
+    if (noScoutAvailable && !inventory[GD_BOW])
```
Flamefire: Actually a bow and a helper and you should use `CanRecruit` instead

Shawn8901: actually you are right! i will add this.

Other question: 
Do you have access to the lobby server? For me it seems that the lobby is down.

Flow86: > Do you have access to the lobby server? For me it seems that the lobby is down.

thats fixed, gitlab upgrade changed some network config stuff

Flamefire:
General question: Do we actually want a scout? What about just using scout graphics? Because he does not need a bow as he just leads the caravan. What would he use the bow for?
ccing @Spikeone

Shawn8901:
in reference to #226 i thought it would be okay to use the scout/bow, because the caravan leader has a higher economic value.

Flamefire:
True. I was just thinking about the implications of sending a (fully equipped) scout.

IMO the PR is ok, but I'd like 1 more opinion on whether to actually use the scout. If yes we can close #226 too

Shawn8901:
i am okay with that, if the intention was only to use the model, please close this pr.
If it should be merged, i have modified the test case for the trade messages to scouts as leader, just in case it will be merged.

Spikeone:
@Flamefire @Shawn8901 As far as I remember the idea was to actually use a scout so caravans cost more. Currently you are able to send soldier after soldier to an ally, boosting him quite a lot. This idea was born when we played 3vs3 and 2 players just sent all soldiers to one player so he could defeat his enemy fast. 

Overall I'm not really sure if this is the best idea as it makes trade even more complicated as it already is for new players. I'd vote for just using the graphic for now and wait till the trade system is reworked - if we still intend to do so. If we will not rework the system, one should have an addon (enable / disable use "real scouts" for trade)

Shawn8901:
Then i would suggest to close/decline the PR, as it was not supposed to be a scout.

Flamefire:
If I understand this correctly the only change required is that the "scout" is taken from the normal people and not from the scouts, right? Could you add this as a separate commit (so we can revert it if we decide FOR using scouts)?

Shawn8901:
i can dig into it and try to find that code passage as soon, as i have reinstalled my dev environment for s25 (had switched from HDD to SSD)

Shawn8901:
I have updated everything to the newest master.
I have commited a single revert commit for the unit change to scout and done a commit to just change the skin.
I hope that its now like expected.

Shawn8901:
checked, also LGTM

PR #793, opened 2018-01-04
Added Trade-Message if caravan leader arrives.

Shawn8901:
I have added a simple PostMsg when a trade leader arrives in a ware house.

If this PR will be merged before #792, named PR has to add a patch to change the figure calculation for the trade leader (helpers -> scout).
If it will be merged after named PR, this test case has to be adjusted.

### DIFF
--- a/src/figures/nofTradeLeader.cpp
+++ b/src/figures/nofTradeLeader.cpp
@@ -54,8 +54,7 @@ void nofTradeLeader::Serialize(SerializedGameData& sgd) const
 void nofTradeLeader::GoalReached()
 {
     noBase* nob = gwg->GetNO(goalPos);
-    RTTR_Assert(dynamic_cast<nobBaseWarehouse*>(nob));
-    nobBaseWarehouse* targetWarehouse = static_cast<nobBaseWarehouse*>(nob);
+    nobBaseWarehouse* targetWarehouse = checkedCast<nobBaseWarehouse*>(nob);
     if(successor)
     {
         unsigned amountWares = 0;
--- a/src/figures/nofTradeLeader.cpp
+++ b/src/figures/nofTradeLeader.cpp
@@ -66,7 +66,7 @@ void nofTradeLeader::GoalReached()
             successorDonkey = successorDonkey->GetSuccessor();
         }
         GamePlayer& owner = gwg->GetPlayer(player);
-        std::string waresName = goodType == GD_NOTHING ? JOB_NAMES[jobType] : WARE_NAMES[goodType];
+        std::string waresName = _(goodType == GD_NOTHING ? JOB_NAMES[jobType] : WARE_NAMES[goodType]);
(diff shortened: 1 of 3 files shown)


src/test/testTrading.cpp
```diff
@@ -269,4 +272,40 @@ BOOST_FIXTURE_TEST_CASE(TradeFail, TradeFixture)
...
+    unsigned distance = world.CalcDistance(curWh->GetPos(), players[0]->GetHQPos()) + 2;
+    RTTR_SKIP_GFS(20 * distance);
+    // warriors are recruited
+    numHelpers -= numSwords;
+
+    const PostMsg* post = postbox->GetMsg(0);
+    BOOST_REQUIRE(post);
+    const PostMsgWithBuilding* msg = dynamic_cast<const PostMsgWithBuilding*>(post);
+    std::string text = boost::str(boost::format(_("Trade caravan with %s %s arrives from player '%s'.")) % 2 % JOB_NAMES[JOB_WOODCUTTER] % players[1]->name);
+    BOOST_REQUIRE_EQUAL(msg->GetText(), text);
```
Flamefire: I would just check, if the text contains the number, the wares and the player instead of the exact text so it is not duplicated in 3 locations.

src/figures/nofTradeLeader.cpp
```diff
@@ -48,17 +53,30 @@ void nofTradeLeader::Serialize(SerializedGameData& sgd) const
 
 void nofTradeLeader::GoalReached()
 {
+    noBase* nob = gwg->GetNO(goalPos);
+    RTTR_Assert(dynamic_cast<nobBaseWarehouse*>(nob));
+    nobBaseWarehouse* targetWarehouse = static_cast<nobBaseWarehouse*>(nob);
```
Flamefire: You can use `checkedCast` instead of the assert above and the static_cast. I added that because we have this pattern quite often. There is also `boost::polymorphic_downcast` which is very similar.

Shawn8901: good point, i switched to checkedCast

src/test/testTrading.cpp
Shawn8901: i hope you suggested what i have done now.

src/figures/nofTradeLeader.cpp
```diff
@@ -48,17 +53,29 @@ void nofTradeLeader::Serialize(SerializedGameData& sgd) const
...
+        unsigned amountWares = 0;
+        Job jobType = successor->GetJobType();
+        GoodType goodType = successor->GetCarriedWare();
+        nofTradeDonkey* successorDonkey = successor;
+        while (successorDonkey != NULL) {
+            amountWares++;
+            successorDonkey = successorDonkey->GetSuccessor();
+        }
+        GamePlayer& owner = gwg->GetPlayer(player);
+        std::string waresName = goodType == GD_NOTHING ? JOB_NAMES[jobType] : WARE_NAMES[goodType];
```
Flamefire: Just seen: You need to use `_` function to translate the name. Same for tests.

Flamefire:
👍

PR #795, opened 2018-01-06
Use libexec/s25rttr

### LINKED ISSUE #791: Consider moving RTTR_BINDIR/RTTR binaries to RTTR_DATADIR
Hi --

I've created a package of rttr for OpenBSD and I am about to import it into our official package repository. There was a concern brought up during the package creation process from other developers: that two binaries (s-c_resample and sound-convert) are placed into RTTR_BINDIR/RTTR (aka /usr/local/bin/RTTR on OpenBSD), which is a subdirectory of the usual package binary directory. It was observed that in nearly 10,000 packages in the OpenBSD package repository, none of them do this. In our experience, binaries that are not reachable from a default path would be put in something like share/s25rttr (CorsixTH does this, for example: the binary is installed as /usr/local/share/CorsixTH/CorsixTH and then reached through a wrapper script).

I would be wondering if you would consider moving s-c_resample and sound-convert to the share/s25rttr directory. This would make packaging easier for more than just OpenBSD, I believe.

Thanks!

Flamefire:
Closes #791 

@Flow86 I also simplified the debug symbol extraction so we don't need to change that when changing the paths again. Could just have a quick look if it looks alright?


release/prepareRelease.sh.cmake
```diff
@@ -123,45 +128,46 @@ extract_debug_symbols()
...
 	;;
 	*)
 		echo "$SYSTEM_NAME not supported" >&2
 		exit 1
 	;;
 esac
 
+${extract_symbols_func} ${RTTR_BINDIR}/s25client.exe
+${extract_symbols_func} ${RTTR_BINDIR}/s25edit.exe
+${extract_symbols_func} ${RTTR_DRIVERDIR}/video/libvideoWinAPI.dll
```
Flow86: this will fail on linux

Flow86: perhaps we should add a list per architecture which will be processed - we also have `set -e` so each failing command will interrupt the script with error

Flow86: something like

```

symbol_files=${RTTR_BINDIR}/s25client.exe ${RTTR_BINDIR}/s25edit.exe

...

symbol_files=${RTTR_BINDIR}/s25client ${RTTR_BINDIR}/s25edit

...

for file in symbol_files ; do
    ${extract_symbols_func} $file
done

Comment (Flamefire) [reply release/prepareRelease.sh.cmake]: I added a check if the file exists. So it doesn't fail

Comment (Flow86) [reply release/prepareRelease.sh.cmake]: yea but if its missing by accident, the build will not fail

Comment (Flamefire) [reply release/prepareRelease.sh.cmake]: Alright. Refactored this and the macOS packaging below. What do you think?

File: release/prepareRelease.sh.cmake [hunk 2]
```diff
@@ -61,6 +62,12 @@ rm -vf ${DESTDIR}${RTTR_DRIVERDIR}/audio/libaudio*.{a,lib}
 extract_debug_symbols()
 {
 	local FILE=$1
+	
+	if [ "$SYSTEM_NAME" == "Darwin" ]; then
+		# Can't extract symbols for apple, so jsut strip them
```
Flow86: typo `jsut` ;-)

release/prepareRelease.sh.cmake
```diff
@@ -195,7 +198,7 @@ case "$SYSTEM_NAME" in
 		for LIBSUFFIX in miniupnpc.5 boost_system boost_filesystem boost_iostreams boost_thread boost_locale boost_program_options ; do
```
Flow86: here `I` called the variable wrong (its LIBNAME or something like that, its not a suffix ;-) perhaps you could change that for me. thx

release/prepareRelease.sh.cmake
```diff
@@ -206,22 +209,24 @@ case "$SYSTEM_NAME" in
...
+		cp -v ${RTTR_SRCDIR}/release/bin/macos/icon.icns ${contentsPath}/Resources/ || exit 1
+		cp -v ${RTTR_SRCDIR}/release/bin/macos/PkgInfo ${contentsPath}/ || exit 1
+		cp -v ${RTTR_SRCDIR}/release/bin/macos/Info.plist ${contentsPath}/ || exit 1
+		mv -v ${DESTDIR}bin/* ${macOSPath}/bin/ || exit 1
+		mv -v ${DESTDIR}lib/* ${macOSPath}/lib/ || exit 1
 		
-		chmod +x ${DESTDIR}s25client.app/Contents/MacOS/* || exit 1
+		chmod +x ${macOSPath}/rttr.command || exit 1
+		chmod +x ${macOSPath}/bin/* || exit 1
+		chmod +x ${macOSPath}/lib/* || exit 1
```
Flow86: libraries should not be executable, shouldnt it be libexec?

Flow86: oh yes, the copying libexec's missing at all

Flamefire:
Ok, next changes in

Issue #206, opened 2015-08-14
Soldiers will be called into buildings even if the border is divided by the sea


kyromaster:
Imagine the following: You have a settlement, then comes the sea and then the settlement of another player. I had the situation that my soldiers got called into the military buildings beneath the sea. They were full of soldiers because the border to the enemy was close, even if the sea was between him and me.
IMHO this shouldn't be like this. If there's (a rather large) sea between two players settlements then there should be no soldiers called to the border.

Spikeone:
Yeah, annoying bug, also since endless maps were added because you often start in each corner on original maps.
Still for upgrading this is usefull, also mappers should know that fact and think about their map design.
Do you have a screenshot of your actual game situation? I'm quite interested in the map.

Flow86:
yes there are many maps with problems like that (i.e Lost Worlds in the top left corner)

Spikeone:
Would need some testing - maybe that's original. But a switch (as for endless/non endless maps) could be great. Sadly a map can be endless and still include that "bug". 
I'll test rage tomorrow

kyromaster:
It's the "Europe" map from Settlers2 Gold. Attached is a screenshot
![bildschirmfoto vom 2015-08-15 14 11 46](https://cloud.githubusercontent.com/assets/13797213/9288935/6534631a-435f-11e5-87be-8b8a28d76f7f.png)

Spikeone:
Working as intended:

![dosbox 2015-08-17 09-58-58-41](https://cloud.githubusercontent.com/assets/1276342/9300497/e323445e-44c6-11e5-81ec-f044085ba804.png)

kyromaster:
So isn't this a bug but intended?

PR #796, opened 2018-01-06
Added Addon FrontierDistanceReachable

Shawn8901:
This addon enables a check for frontier distance calculation. If this addon is enabled, a human path as to be within the triple distance length between both military buildings.

If the distance between both buildings is less then 10, the calculation base will a distance of 30 (maybe this value is a bit high?)

I hope that this "calculation" some how fits the request (#206).

@Flamefire: is there any test case for those frontier logic?

### DIFF
--- a/src/addons/AddonFrontierDistanceReachable.h
+++ b/src/addons/AddonFrontierDistanceReachable.h
@@ -1,0 +1,37 @@
+// Copyright (c) 2005 - 2018 Settlers Freaks (sf-team at siedler25.org)
+//
+// This file is part of Return To The Roots.
+//
+// Return To The Roots is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 2 of the License, or
+// (at your option) any later version.
+//
+// Return To The Roots is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Return To The Roots. If not, see <http://www.gnu.org/licenses/>.
+
+#ifndef ADDONFRONTIERDISTANCEREACHABLE_H_INCLUDED
+#define ADDONFRONTIERDISTANCEREACHABLE_H_INCLUDED
+
+#pragma once
+
+#include "AddonBool.h"
+
+class AddonFrontierDistanceReachable : public AddonBool
+{
+public:
+    AddonFrontierDistanceReachable()
+        : AddonBool(AddonId::FRONTIER_DISTANCE_REACHABLE, ADDONGROUP_GAMEPLAY | ADDONGROUP_MILITARY,
+                    _("Frontier Distance checks Reachability"),
+                    _("Checks if the military building is reachable by triple distance of both military buildings (minimum distance is "
+                      "10). If its not reachable within this distance a military building counts as far away."),
+                    0)
+    {}
+};
+
+#endif
--- a/src/buildings/nobMilitary.cpp
+++ b/src/buildings/nobMilitary.cpp
@@ -321,7 +321,7 @@ void nobMilitary::LookForEnemyBuildings(const nobBaseMilitary* const exception)
 {
     // Umgebung nach Militärgebäuden absuchen
     sortedMilitaryBlds buildings = gwg->LookForMilitaryBuildings(pos, 3);
-    frontier_distance = FAR_BORDER;
+    frontier_distance = DIST_FAR;
 
     for(sortedMilitaryBlds::iterator it = buildings.begin(); it != buildings.end(); ++it)
     {
@@ -334,13 +334,13 @@ void nobMilitary::LookForEnemyBuildings(const nobBaseMilitary* const exception)
             if(gwg->GetGGS().isEnabled(AddonId::FRONTIER_DISTANCE_REACHABLE)
                && !DoesReachablePathExist(*gwg, (*it)->GetPos(), pos, MAX_ATTACKING_RUN_DISTANCE))
             {
-                frontier_distance = FAR_BORDER;
+                frontier_distance = DIST_FAR;
             }
             // in nahem Umkreis, also Grenzen berühren sich
             else if(distance <= MILITARY_RADIUS[size] + (*it)->GetMilitaryRadius()) // warum erzeugtn das ne warning in vs2008?
             {
                 // Grenznähe entsprechend setzen
-                frontier_distance = NEAR_BORDER;
+                frontier_distance = DIST_NEAR;
 
                 // Wenns ein richtiges Militärgebäude ist, dann dort auch entsprechend setzen
                 if(BuildingProperties::IsMilitary((*it)->GetBuildingType()))
@@ -351,7 +351,7 @@ void nobMilitary::LookForEnemyBuildings(const nobBaseMilitary* const exception)
             {
                 // Grenznähe entsprechend setzen
                 if(!frontier_distance)
-                    frontier_distance = MID_BORDER;
+                    frontier_distance = DIST_MID;
 
                 // Wenns ein richtiges Militärgebäude ist, dann dort auch entsprechend setzen
                 if(BuildingProperties::IsMilitary((*it)->GetBuildingType()))
@@ -365,7 +365,7 @@ void nobMilitary::LookForEnemyBuildings(const nobBaseMilitary* const exception)
                 {
                     // Grenznähe entsprechend setzen
                     if(!frontier_distance)
-                        frontier_distance = MID_BORDER;
+                        frontier_distance = DIST_MID;
 
                     // dort auch entsprechend setzen
                     mil->NewEnemyMilitaryBuilding(1);
@@ -375,11 +375,11 @@ void nobMilitary::LookForEnemyBuildings(const nobBaseMilitary* const exception)
     }
 
     // Evtl. Hafenpunkte in der N? mit ber?htigen
-    if(frontier_distance <= MID_BORDER)
+    if(frontier_distance <= DIST_MID)
         if(gwg->CalcDistanceToNearestHarbor(pos) < SEAATTACK_DISTANCE + 2)
         {
             // if(gwg->IsAHarborInSeaAttackDistance(MapPoint(x,y)))
-            frontier_distance = HARBOR;
+            frontier_distance = DIST_HARBOR;
         }
 
     // Truppen schicken
@@ -389,19 +389,18 @@ void nobMilitary::LookForEnemyBuildings(const nobBaseMilitary* const exception)
 void nobMilitary::NewEnemyMilitaryBuilding(const unsigned short distance)
 {
     // Neues Grenzgebäude in der Nähe --> Distanz entsprechend setzen
-    if(distance == NEAR_BORDER)
+    if(distance == DIST_NEAR)
     {
         // Nah
-        frontier_distance = NEAR_BORDER;
+        frontier_distance = DIST_NEAR;
     }
     // in mittlerem Umkreis?
-    else if(distance == MID_BORDER)
+    else if(distance == DIST_MID)
     {
         // Mittel (nur wenns vorher auf weit weg war)
         if(!frontier_distance)
-            frontier_distance = MID_BORDER;
+            frontier_distance = DIST_MID;
     }
-
     RegulateTroops();
 }
 
--- a/src/buildings/nobMilitary.cpp
+++ b/src/buildings/nobMilitary.cpp
@@ -152,8 +152,9 @@ void nobMilitary::Serialize_nobMilitary(SerializedGameData& sgd) const
 
 nobMilitary::nobMilitary(SerializedGameData& sgd, const unsigned obj_id)
     : nobBaseMilitary(sgd, obj_id), new_built(sgd.PopBool()), numCoins(sgd.PopUnsignedChar()), coinsDisabled(sgd.PopBool()),
-      coinsDisabledVirtual(coinsDisabled), frontier_distance(sgd.PopUnsignedChar()), size(sgd.PopUnsignedChar()), capturing(sgd.PopBool()),
-      capturing_soldiers(sgd.PopUnsignedInt()), goldorder_event(sgd.PopEvent()), upgrade_event(sgd.PopEvent()), is_regulating_troops(false)
+      coinsDisabledVirtual(coinsDisabled), frontier_distance(FrontierDistance(sgd.PopUnsignedChar())), size(sgd.PopUnsignedChar()),
+      capturing(sgd.PopBool()), capturing_soldiers(sgd.PopUnsignedInt()), goldorder_event(sgd.PopEvent()), upgrade_event(sgd.PopEvent()),
+      is_regulating_troops(false)
 {
     sgd.PopObjectContainer(ordered_troops, GOT_NOF_PASSIVESOLDIER);
     sgd.PopObjectContainer(ordered_coins, GOT_WARE);
@@ -320,7 +321,7 @@ void nobMilitary::LookForEnemyBuildings(const nobBaseMilitary* const exception)
 {
     // Umgebung nach Militärgebäuden absuchen
     sortedMilitaryBlds buildings = gwg->LookForMilitaryBuildings(pos, 3);
-    frontier_distance = 0;
+    frontier_distance = FAR_BORDER;
 
     for(sortedMilitaryBlds::iterator it = buildings.begin(); it != buildings.end(); ++it)
     {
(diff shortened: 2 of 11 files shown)


Flamefire:
I don't think there is a test case yet. You might write one by using a world with 2 players and placing a couple military buildings in varying distances. Then place some water between them and call the border functions again.

BTW: This won't work: If you have some stones or a building blocking the path to the other building then it would be considered "far". But the calculation is not retriggered when the stones are removed leaving that building exposed. You can use the free-path-finder directly and use a custom functor which only considers terrain instead of the human-path.

2 style issues: Use `std::max` instead of the `(x > y) ? x : y` pattern (here: `distance > 10`) as it is more expressive. And 2nd: When you are doing more work here, consider downloading llvm 6 or at least clang-format. CMake will pick it up from your path and add a clang-format target which you can use before committing (it only formats changed files, so it is fast). Otherwise try to stay close to the code-style around. This helps for avoiding merge conflicts and to many changes where only spaces were inserted/removed. What I have seen is the space after the `if`s

Shawn8901:
Does the project use the default settings of clang-format?

Flamefire:
No. We have a custom clangformat file which is automatically picked up by clang-format with the `file` option (all encapsulated in the make target)
Our style is based on one of the clang-format styles and the most used style in our old codebase

Shawn8901:
Thanks, i will give a try on that. About the issue with the pr, i will think about the resolution. And try to write a test case.

Shawn8901:
Hi, 
sorry for the long time without update. I had/have trouble wit clang/LLVM and my SSD.... But i have now at least clang-format running. 
Just as question before i will work on a test case. 
do you mean a path finder like i have implemented in the last commit?

Flamefire:
First: Yes exactly. I think checking for water is enough. Just make sure, the water check actually checks for "unwalkable water" not just for water. If required change it to something that just requires walkable. Better might be to constrain the transition so that either left or right terrain must be walkable and neither must be dangerous.

Few issues: 

1. You need the `FreePathFinderImpl` in the cpp to instantiate the template correctly. Otherwise you'll get a linker error. (Your previous linker error was probably due to the anonymous namespace which excludes the symbol from the symbol table)
2. Then you don't need the `GetWorld` trait as you don't use it
3. Use World or GameWorldBase as the param for your checker class, no need for the template
4. Style: `namespace detail` is for something very specific to 1 task. Your pathfinder class does not fit there, as it might be used for something else, just like the human pathcheckers are used not solely for pathfinding but also for some direct node transition checks. A good way to use namespaces is to copy the directory structure (each directory=1 namespace, possibly nested). It is not done yet, except for small parts because when I started everything was in 1 folder.
5. Style: `FindXXX` as the name makes me assume it returns the found thing. Better use `HasXXX` or `[Does]XXXExists` or similar if all you return is a bool.
6. The pointer params to the path finders denote optional params. In your case there is no need for the road-vector as all you want to know is **IF** there is a way, not which.

Shawn8901:
So i think i have now added everything requested.

src/addons/AddonFrontierDistanceReachable.h
```diff
@@ -0,0 +1,41 @@
...
+// along with Return To The Roots. If not, see <http://www.gnu.org/licenses/>.
+#ifndef ADDONFRONTIERDISTANCEREACHABLE_H_INCLUDED
+#define ADDONFRONTIERDISTANCEREACHABLE_H_INCLUDED
+
+#pragma once
+
+#include "AddonBool.h"
+
+/**
+ *  Addon adds a 'order home' command button to military buildings. Pressing this button will send all soldiers of the highest rank
```
Flamefire: Change description or just remove

src/pathfinding/FindPathReachable.h
```diff
@@ -0,0 +1,26 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Return To The Roots. If not,  see <http://www.gnu.org/licenses/>.
+
+#ifndef FindPathReachable_h__
+#define FindPathReachable_h__
+
+#include "world/GameWorldBase.h"
```
Flamefire: Move include to cpp and simply fwd declare it here.

src/pathfinding/PathConditionReachable.h
```diff
@@ -0,0 +1,59 @@
...
+#ifndef PathConditionReachable_h__
+#define PathConditionReachable_h__
+
+#include "world/GameWorldBase.h"
+
+struct PathConditionReachable
+{
+    const GameWorldBase& world;
+
+	PathConditionReachable(const GameWorldBase& world) : world(world) {}
```
Flamefire: Please clang-format. Tabs and new lines are wrong. Maybe you have to rerun cmake or better add an empty new line to CMakeLists.txt to trigger it. (We are using `file(GLOB` so new files don't trigger a re-run like it should)

src/test/testFrontierDistance.cpp
```diff
@@ -0,0 +1,163 @@
...
+BOOST_FIXTURE_TEST_CASE(FrontierDistanceNearFrontierAddonActive, FrontierWorldSmall)
+{
+	DescIdx<TerrainDesc> tWater(0);
+	for (; tWater.value < world.GetDescription().terrain.size(); tWater.value++)
+	{
+		if (world.GetDescription().get(tWater).kind == TerrainKind::WATER && !world.GetDescription().get(tWater).Is(ETerrain::Walkable))
+			break;
+	}
+	this->ggs.setSelection(AddonId::FRONTIER_DISTANCE_REACHABLE, 1);
+	for (int i = 1; i < 20; i++) {
```
Flamefire: better use `world.getHeight()` for clarity

src/test/testFrontierDistance.cpp
```diff
@@ -0,0 +1,163 @@
...
+{
+	DescIdx<TerrainDesc> tLava(0);
+	for (; tLava.value < world.GetDescription().terrain.size(); tLava.value++)
+	{
+		if (world.GetDescription().get(tLava).kind == TerrainKind::LAVA && !world.GetDescription().get(tLava).Is(ETerrain::Walkable))
+			break;
+	}
+
+	this->ggs.setSelection(AddonId::FRONTIER_DISTANCE_REACHABLE,1 );
+	for (int i = 1; i < 20; i++) {
```
Flamefire: This is not `world.getHeight()` (50). Is that on purpose?

src/test/testFrontierDistance.cpp
```diff
@@ -0,0 +1,163 @@
...
+{
+	DescIdx<TerrainDesc> tWater(0);
+	for (; tWater.value < world.GetDescription().terrain.size(); tWater.value++)
+	{
+		if (world.GetDescription().get(tWater).kind == TerrainKind::WATER && !world.GetDescription().get(tWater).Is(ETerrain::Walkable))
+			break;
+	}
+
+	this->ggs.setSelection(AddonId::FRONTIER_DISTANCE_REACHABLE, 0);
+	for (int i = 1; i < 20; i++) {
```
Flamefire: `world.getHeight()`

src/pathfinding/PathConditionReachable.h
Shawn8901: Clang format just died before reaching this file. The reason was that i forgot to re-run cmake after a renaming.

src/test/testFrontierDistance.cpp
Shawn8901: no, just a copy & paste mistake.

src/pathfinding/FindPathReachable.h
Shawn8901: i hope did mean what i did on the second commit.

src/buildings/nobMilitary.cpp
```diff
@@ -328,8 +329,14 @@ void nobMilitary::LookForEnemyBuildings(const nobBaseMilitary* const exception)
         {
             unsigned distance = gwg->CalcDistance(pos, (*it)->GetPos());
 
+            // check if military building is reachable
+            if(gwg->GetGGS().isEnabled(AddonId::FRONTIER_DISTANCE_REACHABLE)
+               && DoesReachablePathExist(*gwg, (*it)->GetPos(), pos, std::max(distance * 3, MILITARY_RADIUS[3])))
+            {
+                frontier_distance = 0;
```
Flamefire: Err, wait. `frontier_distance = 0` means "far". So I think it should have been `!DoesReachablePathExist`, or did I miss anything?

src/test/testFrontierDistance.cpp
```diff
@@ -0,0 +1,180 @@
...
+      dynamic_cast<nobMilitary*>(BuildingFactory::CreateBuilding(world, BLD_WATCHTOWER, milBld1Pos, 1, NAT_BABYLONIANS));
+
+    unsigned distance0 = milBld0->GetFrontierDistance();
+    unsigned distance1 = milBld1->GetFrontierDistance();
+
+    BOOST_REQUIRE_EQUAL(distance0, 0u);
+    BOOST_REQUIRE_EQUAL(distance0, distance1);
+}
+
+BOOST_FIXTURE_TEST_CASE(FrontierDistanceMiddleAddonActive, FrontierWorldMiddle)
```
Flamefire: In the active test you use `FrontierWorldMiddle` but in the other you use `FrontierWorldSmall`. Maybe you could unify the 2 test cases to ensure this is the same? (Run in a loop, once with addon=1, once with addon=0 but keep the world setup the same. You could either issue a recalculation of the frontier distances or destroy and rebuild the buildings)

**Important**: Change the test case first before fixing the (possible) bug I mentioned above. IMO 1 test case MUST fail due to that bug. If not, either it is not a bug or you need another test case that triggers the bug.

src/buildings/nobMilitary.cpp
Shawn8901: i definitely with agree you!  Every testcase should have a partner with addon disabled. i dont know why i added a explicit "addoninactive". i will rename them, to match a naming scheme.
I will refactor those first. also i recognized that there is a test case doubled. 
I will take a look into it in the evening.

src/test/testFrontierDistance.cpp
Shawn8901: Okay i think i have managed it. there were multiple problems... The biggest was, that the water is set in my test case was walkable, so the whole test case was wrong... After i had fixed that i realized that the Condition check did not really  work for the test case.
I have fixed both problems and now i think it should work as expected.

Flamefire: Now you are using unreachable terrain? Why not use the unwalkable water (like in some cases before)? That would be more common than lava. (or even test both)

What exactly was wrong with the condition check? IMO it was correct: You have to have all terrains around a point "not unreachable". Otherwise you could land/walk next to lava...

BTW: For your world structs you could use typedefs. They are shorter :)

Shawn8901: i have changed the test back to the "land part" of PathConditionHuman.
I have tried to unit test with water, but i have to be honest, i have no idea why its not working if i change it to water. for me it seems that water is threaded as walkable (maybe 1-2 fields??) because in real game scenario it works for me...)

Flamefire: Found your problem. You put a single colon of water there. Just draw the triangles on a piece of paper and you will see why this fails. My blueprint (large so one can reuse it): 
grid

And with 1 column of water on t1 only (might be actually t2 but doesn't matter for showing): 
water


As you can easily see, walking is very much possible. You'd need to block it completely by making it at least 2 triangles, so one would have to walk ON water to cross. (Walking next to water is allowed)

Note: There is an edge case I just remembered: If there is a road, people can walk on water. A road is possible when every node has a) no danger terrain around and b) any terrain is (at least) walkable. This means that something like this should count as connected: 
water2
To handle this let the checker return true for any transition and only check nodes.

Shawn8901:
Thanks for your explanation! It was very useful. 
Yesterday i was thinking about the idea to change the PathCondition. I rejected the idea (like you before), because we need PathConditionHuman and my implementation to be synchronized. Otherwise it could happen, that a building assume to be not attackable but can be reached by a soldier, which would result in a disaster.  
To ensure this, i changed PathConditionHuman to inherit from PathConditionReachable. The Check is copied from there and a soldier uses PathConditionHuman to evaluate its path. So everything should be fine now on my opinion.
Also i removed my magic value calculation for the distance. I found that there is a constant for the max distance a soldier will walk. So its better to use this.

For the unit test i have added more water, so every building should now be on an island. which is also an okay'ish test case. For the Near test case i also added a second test case to check if its also working for snow and lava.

Flamefire:
Almost perfect. I found another case I already mentioned. Try to rebuild the following as a test case: 
rttrmap

All land except the dark blue (or even including the light blue but doesn't matter) which is water. I'm pretty sure your addon would classify the red military buildings as "far" because there is no path with the current pathfinder. However if one player builds a road over the water, then soldiers can walk over the water and attack the building which leads to the very case you were describing as a "disaster" ;-)

Build this test case and see if it fails. I think it can be fixed by moving the `IsEdgeOk` check to `PathConditionHuman` leaving the other one empty. That should cover everything although it might be to greedy in some cases (flags buildings as near that are separated by lava). To fix this the `IsEdgeOk` could be changed to simply check left&right for 'not unreachable' and omit the walkable check. This should completely fix this.

Shawn8901:
Hi there, sorry for the delay.
I have prepared the test case (at least i think that its okay), also i have implemented your suggested fixes.
so i assume to build an island (5 fields water for map-edge) and 2 fields water as river on a map with size of 38 (middle).
I would now assume in the test case, that on both addon settings the frontier distance should be 1.

for this point the unit test is failing. i will try to look for that the next days but i wanted to share what i understand for the new test case.

Flamefire:
Hm, a simple rebase would have avoided the squash. But ok, found the changes.
First something that helps: As you check both addon settings in a loop it is hard to see which check fails. Idea: Add `i*10` to each distance. As the distance can only be 0-3 it is easy to tell apart (or use `i*100` etc.)
About the new test: This is not exactly right. Putting water at the border is ok, although not required. The map should be large enough, that it works. You could place the military buildings closer together (p0 on right and p1 on left of HQ) so that is the shorter distance. But the water at the border will also solve it for the addon case, so thats fine.
However the river in the middle is not correct. Let me try to visualize how your solution looks like:
rttrmap
Ok, actually it does work. It is different from my idea but still works and is simpler. Good job!

So assuming the addon=1 test is the failing one (add the *100 trick) then the problem is that `PathConditionReachable::IsEdgeOk` is still there. That should go to solve that.
One minor issue: Your check for water in the test is `if(fieldDesc.kind == TerrainKind::WATER && (!fieldDesc.Is(ETerrain::Walkable) || fieldDesc.Is(ETerrain::Unreachable)))`, hence you allow unreachable water but actually you'd want unwalkable but reachable water. So: `if(fieldDesc.kind == TerrainKind::WATER && !fieldDesc.Is(ETerrain::Walkable) && !fieldDesc.Is(ETerrain::Unreachable))`
Otherwise you might get deadly water where no one can go near (-> "unreachable"). In the current config you'd find the regular water first, so you luckily don't run into this by now, but still should be change.

Shawn8901:
Hi, i just tried a simple rebase before. but there was an conflict caused by the reformatting of GlobalSettings.cpp, where the Addons are registered. 
Maybe it would have been enough to solve the conflict on the first rebased commit, but i did not try this (to be honest).

The test case which fails is with addon enabled. it says that the military building is not reachable.
I tried to remove the IsEdgeOk-Code, but there is no difference, so i kept it, because i would prefer the fix you suggested earlier.

to be honest the point with the i*100 i do not understand.

With the commit i just changed the selection of water.

Flamefire:
Just for the future: Yes the resolving of the first commit would have been enough. Generally there would not have been less conflicts for squashing than for simple rebasing. But that's not really important now.

Ah yes I've just seen you changed IsEdgeOk to work as described.

I mean: Currently you get an error like "mismatch: 0 != 3". If you add `100*i` to each distance you would get "mismatch: 0 != 3" for i=0 and "mismatch: 100 != 103" for i=1 making it possible to know from which setting the error is.

I found your bug: You are setting water everywhere except at the border and the river. So exactly the opposite you wanted. Also you are starting at x/y =1 instead of 0 which is odd although not really wrong.
So basically set water instead of each continue or make all conditions together with an "or" each:
`if(curPoint.x < 5 || curPoint.x > world.GetWidth() - 5 || curPoint.y < 5 || curPoint.y > world.GetHeight() - 5 || (curPoint.x >= middle - 1 && curPoint.x <= middle)) set water;`

2 other things: Could you rename your "tUnwalkable" to "tWater"? IMO that would be easier to understand, what do you think?
And remotely related: Would you mind introducing an enum for the frontier distance instead of the numbers? I find that very hard to read what 0 or 3 means. Make it an enum in the class so you have to reference it like `nobMilitary::FAR` or so. Just make sure you don't change the meaning of the underlying ints (0=Far, 3=Nearest) to not break savegames

Shawn8901:
God damit, i should just stay away from those loops....

about the enum: i fully agree with you and added it.

Flamefire:
Yeah I know the windows defines are annoying. You could prefix them with e.g. `DIST_`. I see you suffixed them which is less common but in this case ok. However I'd prefer to have all enum entries named similarly for consistency. Maybe `NEAR_HARBOR` is better? Just harbor seems ambigous:  `if(GetFrontierDistance() == HARBOR)` (what?) --> `if(GetFrontierDistance() == NEAR_HARBOR)` (reads better)

Last: You have `tUnreachable` in one place but check for unwalkable OR unreachable. I guess you don't want the unwalkable there (water is usually unwalkable but reachable, snow/lava is usually unreachable which includes unwalkable)

Flamefire:
Finally it works. Great!
I was thinking: maybe we should make this Addon the default? I'm not sure what s2 did for buildings that are separated by water but I think others said it was like this Addon. Could you test that? (maybe create a simple map in s2 with 2players and water in the middle then build a military bld and wait for the ai to do the same)

Shawn8901:
On my opinion this was already tested in the issue related to the pr (#206). so w/o addon its working as in the original.

Flamefire:
Ok thanks a lot

PR #816, opened 2018-03-15
added build support for armv7l (raspberry pi 3)

daviel:
Changed the cmake files and added arm lua-libraries. Also there is need for libutil(pull request for submodule is open: https://github.com/Return-To-The-Roots/libutil/pull/20) to be buildable on the rpi3.

Performance is quite good (20-40FPS) even with 6 CPU enemys at once.


Flamefire:
Thanks for the contribution. Would you mind fixing our mess with the lua contrib libraries? IMO what we want is a folder structure like: 
`contrib/lua52/(include|lib/(win32|win64|lin...))` so the same lua version on all systems, hence the same includes (I am pretty sure, they are all the same) but different binaries.

Mentioning the lua version in the main folder might be a good idea, but maybe we can stick to `contrib/lua` and have that README file stating the version as we can only update all at once.

Reason for asking: I want to avoid adding yet another folder with the same contents. For that `libUtil` (FindLua.cmake) must also be modified (your other PR there) and it is easier to do it right now than accept this and then fix that in more PRs

daviel:
I will take a look at it. :+1:

Leuchtefix75:
I asume you are using a headles linux or Wayland using EGL/GLES, right?
Because I still using Raspian with X11 and FB.
Which seams to use the X11 copositor for FB, and bring the FB to display as one layer of the DISPMANX compositor, if I'm right?
There is a little enforcement by FBturbo using DMA.

The HelloPI examples are using 'pipes' like DMA or may be stream-buffers ?

daviel:
@Leuchtefix75 I can't follow you on that one. I was using a standard Raspbian Image with the Pixel-Desktop therefore using X11. I was using the 4.14 Linux-Kernel with Full-KMS support activated in the raspi-config.

daviel:
@Leuchtefix75 It was the Pixel Desktop from the raspbian-repos. You should get the 4.14 Kernel now just by updating in the stable repos. But back then I was using the next branch like so: 
`sudo BRANCH=next rpi-update`

Leuchtefix75:
Nice hint! Will try it... in some Weeks(vacation)... and set it to full KMS before...
After I made a backup from my SD!
I assume in a resulution of 800x600 there should be 70FPS possible, in 2k at all 30FPS?

I facing two issues:
- implement code to detect chipset and OS features,
  like 32 or 64bit ARM support, NEON, or... 
- implement code to make use of the triggered features

For users like me its int'resting to compile it native inside a headed environment,
the same I browse the web yet.
That's how I started compiling it three years ago, since one and a half with success.
So I'm kacking the graphic stuff now, started like always from absolut ZERO,
no mind to waste powers to make it clean, just working.
For me its a project to learn something and entertain myself during time of waiting.

Leuchtefix75:
OK , had no nerv to wait and tried... it works with framerates as expected.

sudo raspi-config
=> Advanced Options => Memory Split => 32/128
=> Advanced Options => GL Driver => GL (Full KMS)

The first value is for the display, the second for render-buffers.
(display = x * y * pixel-byte-count * double-buffer)
This slash between/and the two memory valuse are essential that the GALLIUM driver can do his job!
The GPU can only access the first 256 MB of RAM, limit of the sum of the two values.

A little sorry for MCPI because it depends on the BCM_HOST and only one EGL-API  possible.
But hey... S25 runs smoth and playable!

The lag of the graphic-cursor is annoying, as close it is!
How about giving the system-cursor the face of the game-cursor on mouseover?
... selectable?

I know: Make a request! Don't like to... at now.

stefson:
I like this idea a lot! 

0.8.1 used to compile fine on armhf, and the slow loading at startup reminded me of playing SII in the 90's ;-) but it wasn't possible to start a game really, only the menues worked as expected. 

But I think this is only for aarch64, or however raspbian calls their 64bit flavour, right? If it isn't too much of a pest for you, could you also enable support for armhf, the 32bit arm version? Have a rpi2 and can test!

cmake/linux.cmake
```diff
@@ -11,6 +11,11 @@ endif(CCACHE_FOUND)
 
 # set compiler flags
 IF(RTTR_ENABLE_OPTIMIZATIONS)
-	FORCE_ADD_FLAGS(CMAKE_C_FLAGS -ffast-math -mmmx -msse -mfpmath=sse -ggdb)
-	FORCE_ADD_FLAGS(CMAKE_CXX_FLAGS -ffast-math -mmmx -msse -mfpmath=sse -ggdb)
+	IF (PLATFORM_ARCH STREQUAL "armv7l")
+		FORCE_ADD_FLAGS(CMAKE_C_FLAGS -mcpu=cortex-a53 -mfloat-abi=hard -mfpu=neon-fp-armv8 -mneon-for-64bits -mtune=cortex-a53)
```
Flamefire: This won't work for all armv7l systems. Besides that the current approach with the optimization flags is crappy as it assumes a given compiler.

Suggestion: Make an additional file `cmake/CompilerFlags.cmake` which you include from `CMakeLists.txt`. In that you add a cache-variable e.g. `TargetDevice` with a default value (maybe simply: "default") This should be used to specify for what RttR is compiled. You can then set the `STRINGS` property on that cache entry to specify possible values. Those should (for now) be `default RasPI3` (see https://blog.kitware.com/constraining-values-with-comboboxes-in-cmake-cmake-gui/ on how and why)

Later in that file you can use this variable to set your flags, when compiling for RasPI3. And while you are on it, you could also add appropriate flags for other RasPis. See https://gist.github.com/fm4dd/c663217935dc17f0fc73c9c81b0aa845 for those flags.

Reasoning: Most of those flags apply to a given compiler, not OS, so they should not be in OS specific files. I'll clean those up later. Having the ability to specify the target allows to pass default flags appropriate for that target.

stefson:
Given that lua is already installed locally with headers, the patch to make it compile on rpi2 should be rather trivial. 

What is the value of CMAKE_SYSTEM_PROCESSOR? rpi2 should be armv7-a, the CHOST is armv7a-unknown-linux-gnueabihf - altering the patch above with armv7-a instead of armv7l is not working but gives this error: 

Architecture arm not yet supported
failed with: die "Architecture ${ARCH} not yet supported"

daviel:
Thanks for all of your feedback. It's awesome that this project has so much interest. I will look into generalize the build process to allow building on all arm devices also I will take a look at the lua headers for deduplication. But be patient as I will need some time to do it all in my freetime. :)

Leuchtefix75:
RPI2 is! armv7l, so this should work:

FORCE_ADD_FLAGS(CMAKE_C_FLAGS -Ofast -mtune=cortex-a7 -mfpu=neon-vfpv4 -mfloat-abi=hard)
FORCE_ADD_FLAGS(CMAKE_CXX_FLAGS -Ofast -mtune=cortex-a7 -mfpu=neon-vfpv4 -mfloat-abi=hard)

check it at ./build/CMakeFiles/CMakeOutput.log => COLLECT_GCC_OPTIONS
isn't what I expected on that values? I need a new image!

stefson:
@Leuchtefix75 

sadly, that doesn't work for me.

Leuchtefix75:
...sounds like something else does? So I'm happy for you!

You still stuck on CMAKE_SYSTEM_PROCESSOR?
The fact is, the system reports ARMV7L, but the basic compiler knows it as CORTEX-A7.

I looked for a different program that uses autoconf:
build/libtool:host=armv7l-unknown-linux-gnueabihf
build/libtool:build=armv7l-unknown-linux-gnueabihf

stefson:
I get this error in configure: 

``
Architecture arm not yet supported
``

and I can't find the source of this error, certainly don't have any knowledge about cmake build scripts and autoconf stuff. So, as of today, the patches or any suggestion from this thread won't work for rpi2.

Leuchtefix75:
I did the cortex-a7 stuff two years ago on BANANA PI M1.

And i started from ZERO like you now. Don't worry!

go to:
./CMakeLists.txt
look for: 
# Set PLATFORM_ARCH
look for:
MESSAGE
and hashtag it:
 # MESSAGE
and set a line below:
SET(PLATFORM_ARCH "arm")

stefson:
Build the binary with success, but for some reason I can't start a new game. This is most likely a problem on my end. So yeah, if you build with system-lua and not the contrib one, it is not much of a headache.

Leuchtefix75:
Than play the last one... over and over! SinglePlayer, EndlesGame, right?

The arch topic wasn't hard, but I needed a year to learn about Linux,
than the linux.cmake was kind of __LUCK__ and GIYF,
and after that still needed a half year to find a working solution for the LUA issue, because it's INCLUDE subdir is neccessary for each(even if it's all the same)
so I finaly gave it what cmake.sh adored.

By the way: My wish, to game with the system cursor, works great ...for me!
...with a short view at the old LAUNCHPAD.NET stuff to get the rigth name for the DrawCursor to blank it!

I need a new OS-Image ! ! ! ...maybe tomorrow?

Leuchtefix75:
For compiling it needs sometimes at one tread around 600 MB RAM,
and to run the game, to have 1GHz CPU clock would be OK.
So it makes no sense to try it below ARMv7.
The game still uses a lot CPU power for graphics.
For native compiling a dynamic stategy like CheckAndAddFlag should be the way.
Another question is an official build. How I told you, CORTEX-A7 is a well base.

Flamefire:
I merged a commit reworking the lua cmake stuff. If you have lua installed on your system you don't need to make any changes to rttr now.

I'll look into reworking the flags once I got some time

Leuchtefix75:
@Flamefire 
I think the reworking is not a big deal.
#./CMakeLists.txt
`MESSAGE(FATAL_ERROR "Unknown processor arch: ${CMAKE_SYSTEM_PROCESSOR}")`
...makes only sense for crossbuild. Else I would prefer it in ./cmake/linux.cmake
or even don't check for it, or best all regarding ARM* is ARM. Don't do it to coplicated!
> hostnamectl
> echo `uname -s`-`uname -p`
> lscpu
> uname -a
> cat /proc/cpuinfo

`MACRO(CheckAndAddFlag flag)`
...can be used in ./cmake/linux.cmake with a hierarchical touch:

so it checks some
`CHECK_CXX_COMPILER_FLAG `
and the highest _SUPPORTED get
`set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${flag}")`
but...
`# Do this before adding -Werror!`

What do you think?

Flamefire:
The arch is actually only required for the updater. Hence it can be removed for non-official builds.

And yes the reworking is not a big deal, it just has to be done

Leuchtefix75:
Done, like daviel's suggestion with the values of RPI2 which is a armv7l.
I think we will do it this way for further OS-processor types, if there are?
Probably only a person who has a different system can tell what processor the OS reports?
CPU optimisations are meaningles whithout a working GL-driver.
I forked S25. Now I have a new playground and to learn how to use it!

**RPI full KMS openGL driver:**
 Different maps, different memory allocation errors on map start, but some smal maps work
...and the driver seems to be unable to handle viewport resize on video playback
may also have caused the game flaping on resize?

enabling VBO => crash on map start, with legacy driver no issue but low frame-rate
...found a hint that VBO's are only suported on GLES2 context

warnings on build that some stuff get changed in ARM-compiler 7.1 

and some issue with **strdup** on build, but can't name it at now
...wasn't there a discussion to the sound stuff?

Flamefire:
@Leuchtefix75 I'm working on the CMake and OpenGL part to help supporting this. Could you tell me, which OpenGL version is supported? I couldn't find a version with a quick search on google but you can probably find that out with a simple command on your RPI2

Leuchtefix75:
GL|ES2.0 all the arch (VC, MALI, Vivante), some refer 3.1

How about to do all SDL/GL(EGL/GLES) interworking in the .SO file, so the game-client just uses the SO's?

Leuchtefix75:
GL|ES2.0 all the arch (VC, MALI, Vivante), some refer 3.1 (capable for Vulkan-API)

How about to do all SDL/GL(EGL/GLES) interworking in the .SO file, so the game-client just uses the SO's?

Flamefire:
I added ARM support (Just CMake) in #849 

Simply pass `-DRTTR_TARGET_BOARD=RasPi3` or similar to CMake. That PR can be tested and will replace this one (it should have the same stuff unless I missed some)

Issue #815, opened 2018-03-06
Military Aid breaks distance calculations


McKenning:
When selecting option to use "Military Aid", military building proximity indicators fluctuate when certain events take place. I have seen this happen on both the player and AI side military buildings.

Screenshot 1: After a fresh load from saved game, everything appears normal, then when the AI attacks, the flags change to "interior" rather than "border" as indicated by the adjoining enemy fortresses. See bottom two roman fortresses.
![rttr_1](https://user-images.githubusercontent.com/36162467/37037757-39f79e06-2110-11e8-922f-42fb39cfe161.png)

Screenshot 2: After the attack finishes, the flags return to their correct state with the bottom roman fortress flag showing "border".
![rttr_2](https://user-images.githubusercontent.com/36162467/37037804-5633c018-2110-11e8-9d71-a1e86b17f145.png)

Screenshot 3: After an unknown trigger, the flags again go to "interior" within a couple of minutes.
![rttr_3](https://user-images.githubusercontent.com/36162467/37037873-8347d3f0-2110-11e8-847b-27c62d595b25.png)

I have also observed the flags changing when military buildings far away are occupied.

Using version 20180213 on Windows 10 x64.

Logfile attached.
[2018-03-06_07-05-19.log](https://github.com/Return-To-The-Roots/s25client/files/1785373/2018-03-06_07-05-19.log)

Shawn8901:
could you please attach the save file?

McKenning:
Savegame attached:
[Savegame.zip](https://github.com/Return-To-The-Roots/s25client/files/1802892/Savegame.zip)

Military settings:
![image](https://user-images.githubusercontent.com/36162467/37286543-d292e768-25c7-11e8-9cb7-6326854c9091.png)

Economy Settings:
![image](https://user-images.githubusercontent.com/36162467/37286586-fb3f35b8-25c7-11e8-8531-c744453a33f9.png)

Gameplay Settings:
![image](https://user-images.githubusercontent.com/36162467/37286603-0a893122-25c8-11e8-9b6a-916f4eeb0fc3.png)

Other settings:
![image](https://user-images.githubusercontent.com/36162467/37286629-19279a48-25c8-11e8-8460-03493322fed5.png)

Shawn8901:
As you are using the "Frontier Distance check reachability" - mod, could you check the scenario if that's disabled? i would guess that it treats the military building as not reachable.
[...]

PR #817, opened 2018-03-18
Small Rework for Frontier Distance Calculation

Shawn8901:
As the Current implementation is able to override a frontier distance (mid/near), this is should include a fix for that.
As i had a save game with quite a lot military buildings, i recognized some kind of lag after changing state of a military builiding (p.e. if it gets destroyed). 
After looking at the code, i realized that the path-calculation is done for every military building, which is quite useless. it should only override the state if its mid/near. 
So i moved to code.

Also i have added some basic refactoring, so i am setting the distance on the other building at the end of the loop, instead of in every if-clause, since its redundant code.


src/buildings/nobMilitary.cpp
```diff
@@ -329,60 +329,49 @@ void nobMilitary::LookForEnemyBuildings(const nobBaseMilitary* const exception)
...
+            {
+                frontier_distance = newFrontierDistance;
+            }
+
+            // set calculated frontier distance to checked building.
+            if(BuildingProperties::IsMilitary((*it)->GetBuildingType()))
+                static_cast<nobMilitary*>(*it)->NewEnemyMilitaryBuilding(newFrontierDistance);
         }
+    }
+    // check for habor points
```
Flow86: typo "habor" ;-)

Flamefire:
The changes look reasonable. Could you explain the fix and the bug you were talking about? Is it #815? What part of the old code was wrong? Would it be possible, that you temporary revert the code, add a test that fails due to the bug and then apply the fix again to see if this fixes the code? I could do or help you with that but I'd like to see a test for that quite subtile bug (if there was any). I also have 2 performance improvements added

src/buildings/nobMilitary.cpp
```diff
@@ -329,60 +329,47 @@ void nobMilitary::LookForEnemyBuildings(const nobBaseMilitary* const exception)
...
-                // Wenns ein richtiges Militärgebäude ist, dann dort auch entsprechend setzen
-                if(BuildingProperties::IsMilitary((*it)->GetBuildingType()))
-                    static_cast<nobMilitary*>(*it)->NewEnemyMilitaryBuilding(1);
-            }
-            // andere Richtung muss auch getestet werden, zumindest wenns eine normaler Militärgebäude ist, Bug 389843
-            else if((*it)->GetGOT() == GOT_NOB_MILITARY)
+                newFrontierDistance = DIST_MID;
+            } else if((*it)->GetGOT() == GOT_NOB_MILITARY)
             {
                 nobMilitary* mil = dynamic_cast<nobMilitary*>(*it);
```
Flamefire: Use safe_cast here to improve further (dynamic_cast in debug, static_cast in release mode)

src/buildings/nobMilitary.cpp
```diff
@@ -329,60 +329,47 @@ void nobMilitary::LookForEnemyBuildings(const nobBaseMilitary* const exception)
...
-    }
 
-    // Evtl. Hafenpunkte in der N? mit ber?htigen
-    if(frontier_distance <= DIST_MID)
-        if(gwg->CalcDistanceToNearestHarbor(pos) < SEAATTACK_DISTANCE + 2)
-        {
-            // if(gwg->IsAHarborInSeaAttackDistance(MapPoint(x,y)))
-            frontier_distance = DIST_HARBOR;
+            // if new frontier distance is in military range, check if its reachable.
+            if(gwg->GetGGS().isEnabled(AddonId::FRONTIER_DISTANCE_REACHABLE) && newFrontierDistance >= DIST_MID
```
Flamefire: Move that check (enabled) outside the loop to a `const bool`. The check is kinda costly (cache wise) due to a couple pointer indirections.

Shawn8901:
Yes its #815 . i forgot to mention it in the pr, i only added a reference in the commit message. 
About the bug itself. 
the code before i had touched anything with the addon checked if the frontier distance is higher (within the loop). 
following could happen. if you have 3 buildings in buildings-array:
check building 1, its directly at the border --> seen as near.
check building 2, its in attackable range --> MID < NEAR, do nothing.
check building 3, building 3 is not able to reach the target (due unreachablility or to far away) since its the first check it will override frontier_distance with FAR and it will result in a FAR away military building. 

so the "bugfix" is not to override NEAR/MID with FAR. 
Rest are only some performance improvements and refactoring for the code.
So on my side i forgot that, because it was checked by a "magic"
if (!frontier_distance) clause, which idea was not to  override anything above FAR.

i think i will try for a test case, but as you remember i am (still) having trouble in understanding the world generation itself (see our discussions in the pr for this addon)

src/buildings/nobMilitary.cpp
Shawn8901: as i am no c++ professional, i don't now safe_cast as cast operation. after googling around i found only a Microsoft command extension which is named safe_cast.
i could do a static_cast if you want.

Flamefire: It's something I introduced. Maybe I should rename it to something more common (some C++ helps use `down_cast` etc). The idea is the same: Use static_cast/dynamic_cast in a slim wrapper to capture the intent.

My version is `checkedCast` (https://github.com/Return-To-The-Roots/s25client/blob/master/common/include/commonDefines.h#L137) but I should enhance it... However it is to be used the same as `static_cast`: `safeCast<nobMilitary*>(src)` or similar.

Shawn8901:
Finally i found some time for a test case. 
i have tried to build a test-case, which will cause the error and is as simple as possible.
the unit test builds an island, which is divided by a river, from 1/3 in the north to end of the map. Also i have added another river, which goes from west to east and is connected to the first river.
The map setup itself is designed, to have all military buildings picked up by LookForMilitaryBuildings, but have a long path to the other military buildings. 
 
now the test case, adds some buildings for player 1 (right side of the map), one is on the north side of the river and the other one is on the south side.  
After that a building for p0 is added on the north side of the river like in the screenshot.

image

everybody will assume, that the buildings on the north side, should be NEAR.
For current master code, that will not be the case. p0 will be FAR, since the dark green building is evaluated after the light green and its distance will be overriden (its a bit to  far away), so there will be no soldier able to reach the dark green building).

Flamefire:
Perfect! Could you add a very short summary of this testcase to the testcase itself? You could use some simple ascii art to show the 2 rivers and Player 1&2. The testcases are not very readable by itself unfortunately and I don't think one can change that. A small (ASCII) map can help here. If thats to difficult add an image to the repo, name it properly and reference it in the testcase. I did that for some pathfinding stuff.

Flamefire:
:+1: Keep up the good work!

PR #821, opened 2018-04-13
Continue for #681 (Show direction glyph toward flag halfway out)

Shawn8901:
Since there seems no work on #681 i have picked up the changes on the glyph rendering and applied the requested changes.
Also i have combined the evaluation if a direction targets a flag (which was used 3 times in the code block).

The changes related to saving i have skipped, because someone else already added a similar logic.

As i found it is a good idea: i also included the change to enable backspace for post message deletion for macs.
@Flamefire  if you have any improvements left, i would be willing to apply them.

### DIFF
--- a/src/world/GameWorldView.cpp
+++ b/src/world/GameWorldView.cpp
@@ -227,13 +227,13 @@ void GameWorldView::Draw(const RoadBuildState& rb, const MapPoint selected, bool
 void GameWorldView::DrawGUI(const RoadBuildState& rb, const TerrainRenderer& terrainRenderer, const MapPoint& selectedPt, bool drawMouse)
 {
     // Falls im Straßenbaumodus: Punkte um den aktuellen Straßenbaupunkt herum ermitteln
-    std::vector<MapPoint> road_points;
+    std::array<MapPoint, 6> road_points;
 
     unsigned maxWaterWayLen = 0;
     if(rb.mode != RM_DISABLED)
     {
         for(unsigned i = 0; i < Direction::COUNT; ++i)
-            road_points.push_back(GetWorld().GetNeighbour(rb.point, Direction::fromInt(i)));
+            road_points[i] = GetWorld().GetNeighbour(rb.point, Direction::fromInt(i));
 
         const unsigned index = GetWorld().GetGGS().getSelection(AddonId::MAX_WATERWAY_LENGTH);
         RTTR_Assert(index < waterwayLengths.size());
@@ -248,14 +248,6 @@ void GameWorldView::DrawGUI(const RoadBuildState& rb, const TerrainRenderer& ter
             Position curOffset;
             MapPoint curPt = terrainRenderer.ConvertCoords(Position(x, y), &curOffset);
             Position curPos = GetWorld().GetNodePos(curPt) - offset + curOffset;
-
-            // we dont own curPt, no need for any rendering...
-            if(!gwv.IsOwner(curPt))
-                continue;
-
-            // check if curPt is on the border
-            if(!GetWorld().IsPlayerTerritory(curPt))
-                continue;
 
             /// Current point indicated by Mouse
             if(drawMouse && selPt == curPt)
@@ -286,6 +278,10 @@ void GameWorldView::DrawGUI(const RoadBuildState& rb, const TerrainRenderer& ter
             if(rb.mode == RM_DISABLED)
                 continue;
 
+            // we dont own curPt, no need for any rendering...
+            if(!gwv.IsPlayerTerritory(curPt))
+                continue;
+
             // we are in road build mode
             // highlight current route pt
             if(rb.point == curPt)
@@ -295,7 +291,8 @@ void GameWorldView::DrawGUI(const RoadBuildState& rb, const TerrainRenderer& ter
             }
 
             // ensure that curPt is a neighbour of rb.point
-            if(!helpers::contains(road_points, curPt))
+            unsigned int roadIdx = helpers::indexOf(road_points, curPt);
+            if(roadIdx == -1)
             {
                 continue;
             }
@@ -305,7 +302,7 @@ void GameWorldView::DrawGUI(const RoadBuildState& rb, const TerrainRenderer& ter
                 continue;
 
             // render special icon for route revert
-            if(!rb.route.empty() && GetWorld().GetNeighbour(rb.point, rb.route.back() + 3u) == curPt)
+            if(!rb.route.empty() && road_points[(rb.route.back().toUInt() + 3u) % Direction::COUNT] == curPt)
             {
                 LOADER.GetMapImageN(67)->DrawFull(curPos);
                 continue;
--- a/src/world/GameWorldView.cpp
+++ b/src/world/GameWorldView.cpp
@@ -291,8 +291,7 @@ void GameWorldView::DrawGUI(const RoadBuildState& rb, const TerrainRenderer& ter
             }
 
             // ensure that curPt is a neighbour of rb.point
-            unsigned int roadIdx = helpers::indexOf(road_points, curPt);
-            if(roadIdx == -1)
+            if(helpers::contains(road_points, curPt))
             {
                 continue;
             }
--- a/src/world/GameWorldView.cpp
+++ b/src/world/GameWorldView.cpp
@@ -293,8 +293,9 @@ void GameWorldView::DrawGUI(const RoadBuildState& rb, const TerrainRenderer& ter
                 if(rb.mode == RM_BOAT && maxWaterWayLen != 0 && rb.route.size() >= maxWaterWayLen)
                     continue;
 
+                const bool targetsFlag = GetWorld().GetNO(curPt)->GetType() == NOP_FLAG && curPt != rb.start;
                 if((gwv.IsRoadAvailable(rb.mode == RM_BOAT, curPt) && gwv.IsOwner(curPt) && GetWorld().IsPlayerTerritory(curPt))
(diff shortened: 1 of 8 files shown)


Flamefire:
Thanks for taking that over.
As I currently have no way of testing this: Could you attach screenshots that show the different cases? At least: Up, Down, near flags, near start point.

Shawn8901:
Find attached 3 examples how the glyph is rendered if near a flag.

2018-04-16 22_20_41-return to the roots - v20180416 - 8e7b512
2018-04-16 22_21_20-return to the roots - v20180416 - 8e7b512
2018-04-16 22_22_56-return to the roots - v20180416 - 8e7b512

What do you mean with near start flag?  see the following as example. but as the condition says `curPt != rb.start` there is nothing changed for the start flag.
2018-04-16 22_23_28-return to the roots - v20180416 - 8e7b512

src/world/GameWorldView.cpp
```diff
@@ -293,8 +293,9 @@ void GameWorldView::DrawGUI(const RoadBuildState& rb, const TerrainRenderer& ter
                 if(rb.mode == RM_BOAT && maxWaterWayLen != 0 && rb.route.size() >= maxWaterWayLen)
                     continue;
 
+                const bool targetsFlag = GetWorld().GetNO(curPt)->GetType() == NOP_FLAG && curPt != rb.start;
                 if((gwv.IsRoadAvailable(rb.mode == RM_BOAT, curPt) && gwv.IsOwner(curPt) && GetWorld().IsPlayerTerritory(curPt))
-                   || (gwv.GetBQ(curPt) == BQ_FLAG))
+                   || (gwv.GetBQ(curPt) == BQ_FLAG) || targetsFlag)
```
Flamefire: I'm thinking about this `if`. The first 3 conditions ensure, that a road can be build. OK. Then we check, if we could build a flag there. But that would mean, that a road can be build, right? So we can remove that.

Problem is: `IsRoadAvailable` returns false for the point, when there is a flag (at least I think so: https://github.com/Shawn8901/s25client/blob/8e7b512a3a30d8029fb689afcbf5b933bad5f66b/src/world/GameWorldBase.cpp#L102) So that's why to `|| targetsFlag` is required. 

IMO The BQ check was a failed attempt to do that. Or am I missing a point?

Shawn8901: About the bug: you are 100% right.
for me the old code is a bit fishy and hard to understand, so i did a little refactoring session on that (as requested), so now the code is a bit more clear and you can follow the flow a bit easier (at least on my opinion).
After refactoring it was also easy to see why/where the bug you noticed happen (if-else-if hell :-) )

about the loop in 287: i have removed it and replaced by one of your helpers. since i am only interested if curPt is in road_points i have switched to your contains helper instead of the indexOf. Maybe there is a bit more efficient way to get the back direction, of i think my solution is quite easy to understand (i don't see any other reason to check for the index).

if there is anything else open, please let me know.

Shawn8901:
About the bug: you are 100% right.
for me the old code is a bit fishy and hard to understand, so i did a little refactoring session on that (as requested), so now the code is a bit more clear and you can follow the flow a bit easier (at least on my opinion).
After refactoring it was also easy to see why/where the bug you noticed happen (if-else-if hell :-) )

about the loop in 287: i have removed it and replaced by one of your helpers. since i am only interested if curPt is in road_points i have switched to your contains helper instead of the indexOf. Maybe there is a bit more efficient way to get the back direction, of i think my solution is quite easy to understand (i don't see any other reason to check for the index).

if there is anything else open, please let me know.

src/world/GameWorldView.cpp
```diff
@@ -227,13 +227,13 @@ void GameWorldView::Draw(const RoadBuildState& rb, const MapPoint selected, bool
 void GameWorldView::DrawGUI(const RoadBuildState& rb, const TerrainRenderer& terrainRenderer, const MapPoint& selectedPt, bool drawMouse)
 {
     // Falls im Straßenbaumodus: Punkte um den aktuellen Straßenbaupunkt herum ermitteln
-    MapPoint road_points[6];
+    std::vector<MapPoint> road_points;
```
Flamefire: Use a `boost::array`. The number of entries is always known.

src/world/GameWorldView.cpp
```diff
@@ -249,6 +249,14 @@ void GameWorldView::DrawGUI(const RoadBuildState& rb, const TerrainRenderer& ter
             MapPoint curPt = terrainRenderer.ConvertCoords(Position(x, y), &curOffset);
             Position curPos = GetWorld().GetNodePos(curPt) - offset + curOffset;
 
+            // we dont own curPt, no need for any rendering...
```
Flamefire: I'd combine those 2 checks as it is 1 thing. And maybe add an overload to `IsPlayerTerritory` that takes a player and checks the owner. `gwv.IsPlayerTerritory` can then call that instead so it is similar to `isOwner`. Check all references first, if that does introduce a) unnecessary checks for isOwner and b) callers relying on `gwv.IsPlayerTerritory` checking for ANY player.

Besides that: IMO this is to early. What about a) mouse outside own terriory (mid=22), or selected points outside territory?

src/world/GameWorldView.cpp
```diff
@@ -267,67 +275,71 @@ void GameWorldView::DrawGUI(const RoadBuildState& rb, const TerrainRenderer& ter
...
+            {
                 LOADER.GetMapImageN(21)->DrawFull(curPos);
+                continue;
+            }
 
-            int altitude = GetWorld().GetNode(rb.point).altitude;
-
-            for(unsigned dir = 0; dir < Direction::COUNT; ++dir)
+            // ensure that curPt is a neighbour of rb.point
+            if(!helpers::contains(road_points, curPt))
```
Flamefire: Currently this is the only usage of `road_points`. If you are not using indexOf as a cheaper way than GetNeighbour (also the latter is more clear) then you can remove the container completely and just put the loop here.

Shawn8901: after switching back to an array i am more comfortable to use road_points as "cache" for the backward direction lookup

src/world/GameWorldView.cpp
Shawn8901: About IsPlayerTerritory: I have added an overload to World::IsPlayerTerritory, that consumes a playerId.  i think i changed it the way your requested

for the other point, you are right, what was one of the bugs i realized when i edited my last post....

src/world/World.h
```diff
@@ -121,6 +121,9 @@ class World : public MapBase
 
     /// Check if the point completely belongs to a player (if false but point itself belongs to player then it is a border)
     bool IsPlayerTerritory(const MapPoint pt) const;
+    /// Checks if the point complyetly belongs to a given player (if false but point itself belongs to player then it is a border)
```
Flamefire: typo

src/world/World.cpp
```diff
@@ -243,6 +243,23 @@ bool World::IsPlayerTerritory(const MapPoint pt) const
     return true;
 }
 
+bool World::IsPlayerTerritory(const MapPoint pt, const unsigned char owner) const
```
Flamefire: I would combine this with the function above. E.g `owner == 0` -> any owner. So just change the `if(ptOwner...` and call this function from the one above or use a default param. Don't forget the docu

src/world/GameWorldView.cpp
```diff
@@ -305,7 +302,7 @@ void GameWorldView::DrawGUI(const RoadBuildState& rb, const TerrainRenderer& ter
                 continue;
 
             // render special icon for route revert
-            if(!rb.route.empty() && GetWorld().GetNeighbour(rb.point, rb.route.back() + 3u) == curPt)
+            if(!rb.route.empty() && road_points[(rb.route.back().toUInt() + 3u) % Direction::COUNT] == curPt)
```
Flamefire: `road_points[(rb.route.back() + 3u).toUInt()]`

src/world/GameWorldView.cpp
```diff
@@ -267,60 +267,75 @@ void GameWorldView::DrawGUI(const RoadBuildState& rb, const TerrainRenderer& ter
...
+            {
                 LOADER.GetMapImageN(21)->DrawFull(curPos);
+                continue;
+            }
 
-            int altitude = GetWorld().GetNode(rb.point).altitude;
-
-            for(unsigned dir = 0; dir < Direction::COUNT; ++dir)
+            // ensure that curPt is a neighbour of rb.point
+            if(helpers::contains(road_points, curPt))
```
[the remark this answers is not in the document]
Shawn8901: Yeah after removing and adding this multiple times i simply forgot the negation.

src/world/World.cpp
```diff
@@ -229,14 +229,17 @@ void World::ChangeAltitude(const MapPoint pt, const unsigned char altitude)
     AltitudeChanged(pt);
 }
 
-bool World::IsPlayerTerritory(const MapPoint pt) const
+bool World::IsPlayerTerritory(const MapPoint pt, const unsigned char owner) const
 {
-    const unsigned char owner = GetNode(pt).owner;
+    const unsigned char ptOwner = GetNode(pt).owner;
+
+    if(owner != '0' && ptOwner != owner)
```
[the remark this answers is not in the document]
Shawn8901: you are completely right, don't ask me why i used the char value; i was somehow confused that day.

Shawn8901:
If GitHub supports a squash merge, i would support to squash  the pr to one commit ( bitbucket supports to do a squash and merge afterwards), as it keeps the log more readable.

PR #823, opened 2018-04-19
Remove unneeded jpg files.

Shawn8901:
As discussed in https://github.com/Shawn8901/s25client/commit/fb02268d94d0aa671e0ea518eac6b881ebc0126e#commitcomment-28648899 a PR to remove the unneeded jpg files.


Flamefire:
If at all I would add a tool to show previews of the maps as a standalone if the need arises. But we do have it in the client, so I don't see the jpgs or the tool as necessary.

PR #826, opened 2018-04-22
[WIP] Added new Addon ImprovedWareOrder

Shawn8901:
This addon allows to redirect a ware, whichs goal is a storehouse to a new goal. This should reduce the distance traveled by a ware. See #87 as reference.
Also i fixed a small bug (at least on my opinion its a bug). If you are in emergency program and don't have a wood industry and no boards in a store, another building was still able to catch the boards, because the emergency program was only checked if the ware was ordered from a storehouse.

@Flamefire in #87 i have said, that i will look for a test case. but as the ordered ware is not accessible from another class, and i have learned that its a bad idea to change code, for a unit test (if there is no real need for this function in the program itself), i am not really sure how to test the ware ordering itself. After the ware ordering has a test case we should be able to test for my changes.
Do you have any idea how to implement this as a test case?


src/GamePlayer.cpp
```diff
@@ -826,45 +826,50 @@ void GamePlayer::FindWarehouseForAllJobs(const Job job)
...
-    } else // no warehouse can deliver the ware -> check all our wares for lost wares that might match the order
+        if(ware == GD_STONES && goal->GetBuildingType() != BLD_SAWMILL)
+            return NULL;
+
+        if(ware == GD_BOARDS && (goal->GetBuildingType() != BLD_WOODCUTTER && goal->GetBuildingType() != BLD_SAWMILL))
+            return NULL;
+    }
+
+    // Look for available ware in a warehouse
+    unsigned int nearestWarehouse = std::numeric_limits<unsigned>::max();
```
Flamefire: This is a distance, not a warehouse. Please rename

src/GamePlayer.cpp
```diff
@@ -826,45 +826,50 @@ void GamePlayer::FindWarehouseForAllJobs(const Job job)
...
+    nobBaseWarehouse* wh = FindWarehouse(*goal, FW::HasMinWares(ware, 1), false, true, &nearestWarehouse);
+
+    // Look for ware waiting on the road or on the way to storehouse (if addon is enabled)
+    unsigned nearestWare = std::numeric_limits<unsigned>::max();
```
Flamefire: Same here: Distance

src/GamePlayer.cpp
```diff
@@ -826,45 +826,50 @@ void GamePlayer::FindWarehouseForAllJobs(const Job job)
...
+    Ware* bestWare = NULL;
+    const bool improvedWareOrder = gwg->GetGGS().isEnabled(AddonId::IMPROVED_WARE_ORDER);
+    if(!wh || improvedWareOrder)
     {
-        unsigned bestLength = std::numeric_limits<unsigned>::max();
-        Ware* bestWare = NULL;
         BOOST_FOREACH(Ware* curWare, ware_list)
         {
-            if(curWare->IsLostWare() && curWare->type == ware)
+            noBaseBuilding* wareGoal = curWare->GetGoal();
```
Flamefire: This is an extra mem-load which is not required in many cases (wrong type, lost ware, addon not enabled) and only used once. So don't preload it here.

src/GamePlayer.cpp
```diff
@@ -826,45 +826,50 @@ void GamePlayer::FindWarehouseForAllJobs(const Job job)
...
+            // Ware is redirectable if its lost (has no goal) or its goal is a storehouse
+            if(curWare->type == ware && (curWare->IsLostWare() || (improvedWareOrder && wareGoal->GetBuildingType() == BLD_STOREHOUSE)))
             {
-                // got a lost ware with a road to goal -> find best
                 unsigned curLength = curWare->CheckNewGoalForLostWare(goal);
```
Flamefire: Maybe add an extra param to pass the maximum distance as we (may) already have a distance (for other ware or warehouse)

Note that this also ONLY works for wares currently at flags. This is better than nothing but not optimal. Maybe rework this in a follow up.

src/GamePlayer.cpp
```diff
@@ -826,45 +826,50 @@ void GamePlayer::FindWarehouseForAllJobs(const Job job)
...
+    if(!wh || improvedWareOrder)
     {
-        unsigned bestLength = std::numeric_limits<unsigned>::max();
-        Ware* bestWare = NULL;
         BOOST_FOREACH(Ware* curWare, ware_list)
         {
-            if(curWare->IsLostWare() && curWare->type == ware)
+            noBaseBuilding* wareGoal = curWare->GetGoal();
+            // Ware is redirectable if its lost (has no goal) or its goal is a storehouse
+            if(curWare->type == ware && (curWare->IsLostWare() || (improvedWareOrder && wareGoal->GetBuildingType() == BLD_STOREHOUSE)))
```
Flamefire: What about HQ and harbour? There is a method `BuildingProperties::IsWarehouse`

src/GamePlayer.cpp
```diff
@@ -826,45 +826,50 @@ void GamePlayer::FindWarehouseForAllJobs(const Job job)
...
+    }
+
+    // order ware in Warehouse if is near then available ware
+    if(wh && nearestWarehouse < nearestWare)
+    {
+        return wh->OrderWare(ware, goal);
+    } else if(bestWare) // any ware was found
+    {
+        // Change Goal for Ware from NULL/Warehouse to orderer.
+        bestWare->SetNewGoalForLostWare(goal);
```
Flamefire: When the ware had a goal this might be a problem. First the warehouse has to be notified that the ware won't arrive. 2nd: What about cases, where the warehouse (e.g. harbour) ordered the ware? 3rd: that case might even lead to a bigger problem: When cancelling the ware, the building might reorder it immediately (in the cancel method) and could find the exact same ware. Then it orders that ware again but you overwrite its goal to something else.

I think I know now why this hasn't been implemented earlier ;)

src/GamePlayer.cpp
```diff
@@ -826,45 +826,50 @@ void GamePlayer::FindWarehouseForAllJobs(const Job job)
...
-        {
-            // Wenn Notfallprogramm aktiv nur an Holzfäller und Sägewerke Bretter/Steine liefern
-            if((ware != GD_BOARDS && ware != GD_STONES) || goal->GetBuildingType() == BLD_WOODCUTTER
-               || goal->GetBuildingType() == BLD_SAWMILL)
-                return wh->OrderWare(ware, goal);
-            else
-                return NULL;
-        }
-    } else // no warehouse can deliver the ware -> check all our wares for lost wares that might match the order
+        if(ware == GD_STONES && goal->GetBuildingType() != BLD_SAWMILL)
```
Flamefire: While this is correct for now, don't assume that woodcutters need no stones. I'm working on some code that makes the costs changeable. So just leave this as-is: Stones and boards ordered? Then return NULL for everything but woodcutter and sawmill.

src/GamePlayer.cpp
Shawn8901: Dont ask me, why i wrote to goal to a variable, since its not used multiple times.

src/GamePlayer.cpp
Shawn8901: Eine kleinere Frage hier auf deutsch, ich denke, dass es gerade einfacher ist für mich dies auf deutsch zu beschreiben.

Dass ich das Lagerhaus/HQ/Hafen benachrichtigten muss, verstehe ich; irgendwie habe ich ignoriert, dass der Bestellende noch auf seine Ware wartet. 
Die Kritik, dass ein CancelWare ggf. zu einer Neubestellung führen könnte verstehe ich auch, da die Ware bestellbar scheint, wenn sie kein Ziel hat (somit als "verlorene" Ware gilt).
 
Müsste es aber nicht "ausreichen", wenn ich das Ziel der Ware ändere, dem neuen Ziel zuordne  (TakeWare) und dann auf dem alten Ziel CancelWare aufrufe? Hierdurch müsste ich eigentlich sicherstellen, dass die Ware bereits "benutzt" wird, bevor das Lagerhaus sie bestellen könnte.

src/GamePlayer.cpp
Shawn8901: Good idea to add a max distance and give its current distance from warehouse. FindPathForWareOnRoads already supports max distance, so its just adding NULL and the new parameter.

About the Ware not at a flag:
Shouldnt it be enough, to use the following code:

`
    noRoadNode* startNode = location;

    if (!IsWaitingAtFlag()) {
        startNode = location->GetNeighbour(Direction::fromInt(next_dir));
    }
` 
And change the other location variables to startNode? I still need to check if i need to call a Carrier, but before this i want to get feedback on my idea.

Flamefire: If the ware is being carried, then yes this should be enough. IIRC you can even just set the new goal when the ware is carried and it will calculate the path, once it reaches the next flag.

Shawn8901:
As i am currently out of time to work on this and it was still somehow broken, i will close the pr. maybe later i will find some time to restart working on that.

PR #831, opened 2018-04-25
Prevent Crash on fast click startup if sounds not loaded

Shawn8901:
If a user is very fast and raises a mouse click event on the black screen (before the splash screen is shown), the loader is not initialized.
Due its state as a singleton its created, but is empty. GetSoundN will not find the sound to play and will cause an application crash (nullpointer).
As the disable_mouse flag is true in that state i have added a check for that before playing the sound.
This should prevent the crash.
I hope this attempt is okay, otherwise i could check if there was a sound found (null check), but this looked somehow strange for me.

### DIFF
--- a/src/WindowManager.cpp
+++ b/src/WindowManager.cpp
@@ -297,8 +297,11 @@ void WindowManager::Msg_LeftDown(MouseCoords mc)
     if(!curDesktop)
         return;
 
-    // Sound abspielen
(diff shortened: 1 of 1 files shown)


Shawn8901:
Yeah that will also fix the problem, but somehow it looked fishy for me (cause on a good world scenario this value will never be null), but  If you prefer this as solution i could do this.

PR #832, opened 2018-04-26
Fix lua include dir

### LINKED ISSUE #838: BUG: lua does not getfound in contrib correctly
on all platforms, this error occurs on a fresh checkout:

```
[windows.i386] CMake Error at /usr/share/cmake-2.8/Modules/FindPackageHandleStandardArgs.cmake:108 (message):
[windows.i386]   Could NOT find LUA (missing: LUA_INCLUDE_DIR LUA_LIBRARY) (Required is
[windows.i386]   exact version "5.2.1")
[windows.i386] Call Stack (most recent call first):
[windows.i386]   /usr/share/cmake-2.8/Modules/FindPackageHandleStandardArgs.cmake:315 (_FPHSA_FAILURE_MESSAGE)
[windows.i386]   libutil/cmake/Modules/FindLUA.cmake:139 (find_package_handle_standard_args)
[windows.i386]   src/libGamedata/CMakeLists.txt:41 (find_package)
```

I also tried to set LUA_DIR manually (see f8d1e57e09c4012820266b039df4bc91d8e33d13) but still no change

Flamefire:
closes #838

### DIFF
--- a/src/libGamedata/CMakeLists.txt
+++ b/src/libGamedata/CMakeLists.txt
@@ -15,12 +15,18 @@ SOURCE_GROUP(other FILES ${SOURCES_OTHER})
 # LUA
 ################################################################################
 
-if(NOT LUA_DIR)
+# If not set we set hints for the LUA library. Those can be overwritten by
+# setting CMAKE_PREFIX_PATH properly
+# Note: We only set those for the libraries we use for official builds. Normally
+# those should be on the build server but we make them available here for convenience.
+
+if(NOT ENV{LUA_DIR})
 	if(MSVC)
 		# Library is in contrib archive which is in prefix path.
-		# We just need to set LUA_DIR so the includes can be found
-		set(LUA_DIR "${CMAKE_SOURCE_DIR}/contrib/lua")
+		# We just need to the includes to be found
+		list(APPEND CMAKE_PREFIX_PATH "${CMAKE_SOURCE_DIR}/contrib/lua")
 	elseif(WIN32 OR CYGWIN)
+		list(APPEND CMAKE_PREFIX_PATH "${CMAKE_SOURCE_DIR}/contrib/lua")
 		if(CMAKE_SIZEOF_VOID_P EQUAL 8)
 			set(LUA_DIR "${CMAKE_SOURCE_DIR}/contrib/lua/win64")
 		elseif(CMAKE_SIZEOF_VOID_P EQUAL 4)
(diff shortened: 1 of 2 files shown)


Shawn8901:
Hint i am nit familiar with cmake so I dont know if it helps, on my develpoment environment i had to override lua_dir to the correct folder, becuase the cmake search did add win32 in the folder structure, but the include folder is not within the arch-folder.
After setting the base lua dir to the correct folder, the rest was working fine, so maybe just the lua_dir lookup has to be fixed?

Flamefire:
The problem is: LUA_DIR is only 1 variable, but we need paths to include file and library. So we have to choose to set it to either

The idea of the find module is:

1. Find lua.h using "usual cmake lookup" and hint from `LUA_DIR` and `LUA_DIR/include`
2. If found, extract version and use that to specify possible library names. Else use all possible library names
3. Search for library using cmake lookup and LUA_DIR and LUA_DIR/lib as hints
4. If library found but include not, search again for include in the parent folder of the library

The CMakeLists sets LUA_DIR to the path of the library in the contrib folder. So the library should be found in step 2 and the include in either step 1 or step 4.

This scheme is to use only 1 include folder for all libraries in the contrib dir. But it is against the usual structure which is like:

    Prefix
      ->include
      ->lib or lib<arch>

I also used the contrib files only as a hint because one might prefer the system libraries.

Maybe all of that is a bad idea and we should simply use the contrib files if possible... I remember that there were problems when using newer versions of the library.

Flamefire:
@Shawn8901 To what did you set your LUA_DIR and which folders did it find (LUA_INCLUDE_DIR, LUA_LIBRARY)?

Shawn8901:
image

This is what i have set the lua variables to.
i did only remove win32 from LUA_DIR.
i am compiling on a win10 machine with VS2015.

Flamefire:
@Flow86 This is the fix. It should work, but wait what travis says.

Flow86:
yea i reverted my tests - you have to rebase once

Flamefire:
With this PR? Travis looks good so far: https://travis-ci.org/Return-To-The-Roots/s25client

Flamefire:
There is something wrong with the build-server. I found a minor bug but that should only affect the windows-crossbuild and is now fixed. Travis finds lua, the include is, where it should be but the build server does not find it. I don't know why

Flow86:
cmake version bug again?

Flow86:
hmm perhaps some weird cmake cross compiling bug: travis does use a native compilation

Flow86:
I think I found the problem: since it can't find the include dir correctly - we could simply set it correctly since we know it already (see 3e1e00472b602c596e23b1408bf9cf9e125ec12d)
the library will be detected correctly then

Flow86:
oh and I had to use CMAKE_FIND_ROOT_PATH instead of CMAKE_PREFIX_PATH

Flamefire:
Ouch. Not a good idea. This way the user can't easily overwrite it. Appending to prefix path searches it after user provided paths.
What is the difference between prefix and root path? 
Either way: the library will be found through the environment variable. So root path can be left untouched. And better would be to set the lua_include_path from Jenkins instead. So regular lookup is not changed. 

Travis uses exactly the same cmake version, so yes it could be a cross compilation bug or maybe (and I think this is it) something left over in the cache... Jenkins does not reset its cmake cache does it? Paths have changed so some left over could influence the lookup.

Flow86:
yep the cmake version is the same.

prefix path should to be the host-arch paths, the find-root-path should be the target-arch paths
(https://cmake.org/cmake/help/v3.0/variable/CMAKE_FIND_ROOT_PATH.html#variable:CMAKE_FIND_ROOT_PATH)

the include path will only be set if the library from contrib is used, so that should not be a problem (otherwise users can easily overwrite it via LUA_LIBRARY.

I tested it in a clean directory and I also reset jenkins workspace but it didnt work out of the box.

Flamefire:
The reason is here: https://github.com/Return-To-The-Roots/s25client/blob/master/cmake/toolchains/c.linux.i386.cmake#L22

`CMAKE_FIND_ROOT_PATH` simply reroots the search paths. So `/workspace/contrib/lua` from `CMAKE_PREFIX_PATH` becomes `/usr/i686-pc-linux-gnu/workspace/contrib_lua`. Not sure if this also applies to the environment variable set value.

>the include path will only be set if the library from contrib is used, so that should not be a problem (otherwise users can easily overwrite it via LUA_LIBRARY)

That is not true. You always set it unless the environment variable is unset. My intention was: If the user did not specify the environment variable, we append the contrib-lua to the search path. Because the usual setting would be to have lua in system-paths (found by default in CMake) or add it to `CMAKE_PREFIX_PATH` so it will still be found (and **before** contrib-lua) Now (I think) this does not work, because (unless the env var is set, which one normally should not need to) the `LUA_INCLUDE_PATH` is already valid and hence taken. Even worse: The library might be taken from the system and may not match the lua-header (although unlikely due to FindLua).

Leave it for now, I have something in the pipeline to test it. Can you provide me the docker image or the cross-compiler (or at least how you build that) so I can test it locally first?

Flow86:
you should have access to it already in the private gitlab, I send you the link per pm

Issue #830, opened 2018-04-25
[wish] option to auto-demolish empty quarries and mines.


alfakr0ll:
Make an option to auto-demolish buildings that has no more access to it's resources. Quarry, mines, wells (if exhaustible).

This could also be expanded into a "hard mode"/"challenge mode". Every building that drops to 0% is demolished. This could add a challenge to already (too) easy maps.

PR #834, opened 2018-04-29
Added Addon to auto demolish a building if it runs out of resources.

Shawn8901:
As requested in #830 i have build a addon which will automatically demolish a building which runs out of resources.
As it is a big benefit for the player and changes the game play from the original game, its disabled by default.
The other option requested in #830 i have not implemented as i understand them as "fun mod" and on my opinion its not useful, but maybe someone else is willing to implement that.


Flamefire:
Just one thing: Please be more specific in the extended help text. Currently it provides just the same info as the short description/name. "the building" leaves the question "Which building?" open. Say something about buildings with non-infinite (I'm missing the English word) resources and maybe "permanently depleted" or so. Otherwise one might assume this applies to woodcutters or mints too.

Also: "Demolishs building if it runs out of resources" -> "Demolish building when out of resources" (shorter [space is a big issue for the names], command style [no 's'], "runs out" doesn't mean "is out of")

src/addons/AddonDemolishBldWORes.h
```diff
@@ -0,0 +1,36 @@
...
+
+#include "AddonBool.h"
+
+class AddonDemolishBldWORes : public AddonBool
+{
+public:
+    AddonDemolishBldWORes()
+        : AddonBool(AddonId::DEMOLISH_BLD_WO_RES, ADDONGROUP_GAMEPLAY | ADDONGROUP_ECONOMY,
+            _("Demolish building when out of resources"),
+            _("Automatically demolish the building if it runs permanently out of natural resources like mines."),
```
Flamefire: Better. But this sounds like mines are resources. ;) Maybe: "Automatically demolish a resource gathering building (e.g. mine) once it has no more resources." Or "when" instead of "once"

PR #844, opened 2018-05-21
Remove Humidity Check from placing resources on mineable terrain.

Shawn8901:
As the humidity check and mountains exclude each other, it has to be a mistake to check it here.
See my comment in #842  as reference. 
I also removed the first if check for placing fish, as it is useless from logical perspective, as it will always return true, if the the second is true, also its excluding the later else-clauses (as water is not vital or mineable).

### DIFF
--- a/src/mapGenerator/RandomMapGenerator.cpp
+++ b/src/mapGenerator/RandomMapGenerator.cpp
@@ -243,9 +243,7 @@ void RandomMapGenerator::SetResources(const MapSettings& settings, Map& map)
 
         uint8_t res = libsiedler2::R_None;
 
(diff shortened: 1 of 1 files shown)


Shawn8901:
@Flow86 Are you currently working on AppVeyor? Seems that its missing the boost_159 files (see directory list at the top).

https://ci.appveyor.com/project/Flow86/s25client/build/0.8.3.1379/job/gomx59lcgm0c3ltv

Flamefire:
Your are absolutely correct. Thanks for the fix. Boost 1.59 was removed from appveyor: https://www.appveyor.com/updates/

Flamefire:
I'll fix AppVeyor first, then merge this (as a test for AV)

PR #848, opened 2018-07-07
WIP: Modernize OpenGL

Flamefire:
Our current OpenGL usage is pretty old. We don't need to support age-old devices anymore especially after dropping WinXP support. Moving to OpenGL Core profile will make future development easier especially porting to e.g. Rasberry PIs etc.

The question is: What version of OpenGL should we target/require? OpenGL 2.0 should be the minimum (shaders) and OpenGL 3.3 the maximum (see https://support.apple.com/de-de/HT202823 which lists 3.3 as the version supported by all devices) Also https://www.reddit.com/r/opengl/comments/51htbk/what_is_the_best_opengl_version_to_start_with/ summarizes this  as "use 3.3". https://learnopengl.com/Getting-started/OpenGL also votes for 3.3

However the first and 2nd generation Intel Core processors don't seem to support this: https://www.intel.com/content/www/us/en/support/articles/000005524/graphics-drivers.html So it seems we should stick with 3.1 or even 2.1

3.x has deprecated/removed all the fixed-function stuff one should not use anymore which helps development. 3.1 has instancing, which can be used to boost rendering performance and 3.3 improved on that greatly.

Hence I would go for 3.3 but that requires a Core ix 3yyyy or a not-ancient graphics card. A 2.1 would ensure wider support at the cost of harder development. I would definitely go for shaders which allows to support new devices but drops some (very) old ones.

What shall it be? @Flow86


Flow86:
I think the only useful option is 3.1 or 3.3. What do you think we should aim for?

Flamefire:
Personally I would go for 3.3 which is kinda "state of the art", implement everything there and then maybe backport if we have users who want lower versions. Reasons are as above: 3.2 added the "Core profile" which we should target to ease support for the PIs and everything that can run 3.2 can also run 3.3. Backporting shouldn't be to hard then, if we have to.

But problem is: We don't know what our users want. Having some statistics would be great.

BTW: Any progress on C++11? Or could you at least install the newer libraries? CMake 3.4+, Boost 1.58+, SDL2 would be helpful.

baka0815:
Do you really want to support that old hardware?
Most common hardware should at least support OpenGL 3.3 and those old Intels with iGPUs can just install a low end dGPU to support 3.3.
Just my 2ct.

Flow86:
okay, I'm fine with 3.3 then.

@Flamefire no progress yet

stefson:
It seems to be the case that the build script wants to include something from contrib/glad - anything else that is mandantory at the moment? I'm building against system libs for all deps but kaguya right now. Reason I'm asking is, that if I checkout this branch from your fork, it gives me a lot of errors like this: ```conversion of object pointer to function pointer type``` from glad.c

stefson:
About opengl on the rpi2 - I have mesa and glxgears installed, if you could tell me for which expression to grep I can paste the output here.

Flamefire:
@stefson I have the same errors (see travis), working on CMake first then this is easier to fix. I want to disable warnings as errors and stuff for external libs. contrib/glad is included (replacement for gl.h)
Can you run `glxinfo | grep 'version'` on your rpi2?

stefson:
this is the output from glxinfo, using mesa vc4 drivers 

```
direct rendering: Yes
    GLX_MESA_multithread_makecurrent, GLX_MESA_query_renderer, 
    GLX_MESA_multithread_makecurrent, GLX_MESA_query_renderer, 
Extended renderer info (GLX_MESA_query_renderer):
OpenGL renderer string: VC4 V3D 2.1
    GL_OES_element_index_uint, GL_OES_fbo_render_mipmap, 
```

baka0815:
Citing https://github.com/anholt/mesa/wiki/VC4-OpenGL-support:
> VC4 at the hardware level is a GLES 2.0 renderer. However, along with GLES 2.0, the Mesa driver also exposes OpenGL 2.1, which is mostly correct but with a few caveats.

So it doesn't look like the mesa driver for VC4 supports more than OpenGL 2.1.

Flamefire:
Thanks! So OGL ES2.0 and OGL 2.1 it is. I'm still interested in the glxinfo output for verification.

stefson:
here is the full output of glxinfo
[glxinfo.zip](https://github.com/Return-To-The-Roots/s25client/files/2189446/glxinfo.zip)

I have tested with glxgears and got the usual 60fps on a 60hz Display, so it is indeed hardware accelerated. Seeing s25client compile and install doesn't mean that the binary will work flawless, by the way :-S

Flamefire:
Alright thanks! Sad that the instanced extension is not supported but at least GL_ARB_vertex_array_object is.

PR #849, opened 2018-07-07
Update CMake files

Flamefire:
- Empty build directory
  - This allows a clean with a simple rm -rf *
  - Moved util-scripts into util folder
- Move optimization and warning flags into separate files
- Clean up toolchain and OS files
- Add ARM flags (supersedes #816)


stefson:
for your information only, it doesn't break anything with my specially sensitive toolchain :-)

Flamefire:
The goal is to include most of https://github.com/Return-To-The-Roots/s25client/pull/816 in the proper places. This includes all other flags (optimization etc) as well.
@stefson Would you mind to test this PR when I'm done? It will take some time though, I'll ping you.

stefson:
Sure, I can test compile, but I have a rpi2 which is arm and not aarch64 aka rpi3

Flamefire:
@stefson Could you test this? Try a clean build tree and pass `-DRTTR_TARGET_BOARD=RasPi2` without any further options. BTW: Are you compiling natively or cross?

Flamefire:
You checked out the wrong branch. You need to test this PR or my `cmake` branch.

BTW: Your cmake options (may) need to be adjusted:

- `-G Unix Makefiles` Missing quotes around `"Unix Makefiles"`
- `-DENABLE_OPTIMIZATIONS=OFF`: I'm pretty sure you want this `ON` especially for ARM (my new changes won't be applied if you have this disabled)
- `-DRTTR_INSTALL_PREFIX, -DCOMPILEFOR=linux -DCOMPILEARCH=arm -DBUILD_GLFW=no` outdated and not used anymore. CMake should even warn you
- `-DCMAKE_SKIP_RPATH=YES` duplicate
- `-DCMAKE_BUILD_TYPE=Gentoo` Not sure if this is a good idea. We use `CMAKE_CXX_FLAGS_RELEASE` etc, but I don't know how you defined your _GENTOO variables
- `-C`, `-DCMAKE_USER_MAKE_RULES_OVERRIDE` etc: You are using a toolchain, an override and an initial cache I'm pretty sure this is overkill. What are the reasons? And can you provide gentoo_rules.cmake, gentoo_toolchain.cmake and gentoo_common_config.cmake for reference?

You might be fine with something like: `cmake -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_SKIP_RPATH=ON -DRTTR_DRIVERDIR=lib/s25rttr -DRTTR_GAMEDIR=share/s25rttr/S2/ -DRTTR_LIBDIR=lib/s25rttr -DRTTR_TARGET_BOARD=RasPi2 -DCMAKE_BUILD_TYPE=Release   /var/tmp/portage/games-strategy/s25rttr-9999-r3/work/s25rttr-9999`
You could remove the paths (`RTTR_XXXDIR`, most are defaults anyway) and you may need a toolchain file but you are compiling natively so it might not be required.

stefson:
You're right about me using the wrong branch, also I forgot to remove the dead gflw code from the ebuild file, which went unnoticed because it was turned off at compile time anyway. The whole script was taken from some outdated overlay and I got it somehow to work. So, for the rest of your proposals, I'll give it a try soon to work things out.

stefson:
Ok, so now I checked out the correct branch on the rpi2, without the depreciated cmake flags, and got a syntax error about cmake/Modules/ArmConfig.cmake: [config.log.zip](https://github.com/Return-To-The-Roots/s25client/files/2207535/config.log.zip)

stefson:
okay, now allows to pass configure but with a few warnings about cmake policies not being set: [config.log.zip](https://github.com/Return-To-The-Roots/s25client/files/2209053/config.log.zip)

Flamefire:
I'll handle `Policy CMP0054`.

About debug=release (performance wise): Not true. `-Og` optimizes for debugging. This not only disables most optimizations but will also further reduce performance in order to improve debugging.

`-march=core2`: You are not building this for RasPi2 but for your local machine, right?

stefson:
It takes around two hrs to compile it on a rpi2, with jobs reduced to `make -j2` - but even with ~440mb of available ram per thread it can be tight at times with memory. Now going for some runtime testing! 

-march=core2 is on my main system ofc

stefson:
The good news is - the arm binary works in some parts. s25client loads, the gui is rendered and it's possible to click around and change the language. But two showstoppers - it's not possible to change the screen resolution from 800x640 to anything other than that (I restarted ofc), and more severe, the client cannot establish a local network connection, hence fails to connect to localhost and charges to stop the server before the game/session is started. The log is not very verbose here: [S2.log.zip](https://github.com/Return-To-The-Roots/s25client/files/2209910/S2.log.zip)

Flamefire:
Great. So this can be merged I guess?

For the former: The implementation is restricted to the possible video modes. You can try to list the available video modes (resolutions) by system specific means (no idea for linux/RPi2) or build an example with SDL 1.2 and print out what we also use: https://github.com/Return-To-The-Roots/s25client/blob/master/driver/video/SDL/src/VideoSDL.cpp#L550

About the connection: Honestly no idea. We don't do anything special. Anything with firewalls? We have a timeout of 400ms for connecting (https://github.com/Return-To-The-Roots/libutil/blob/070d338bc01b3d849e1a487a07562157f0548606/src/Socket.cpp#L625), maybe you can try to increase this? You could set it to very high and check for open ports at the same time. What's also possible: IIRC the server and client run in the same thread, which should be fine. But maybe it is not in this case and the client (socket) needs the interaction of the server (the accept or so)? Also `connect` according to [this](https://linux.die.net/man/3/connect) can never return a `EWOULDBLOCK` (https://github.com/Return-To-The-Roots/libutil/blob/070d338bc01b3d849e1a487a07562157f0548606/src/Socket.cpp#L539), right? So maybe we just don't wait long enough for the slow raspberry.

stefson:
How can I increase the timeout for connecting? You gave the file and line number, but not really a value? 

I'd like to wait a bit, till at least the connection failure is solved and I rechecked that things work as well on my workstation.

Flamefire:
Sorry assumed you were familiar with C++. You could try replacing the line I linked with `if(timeout > 80)`. That would increase the timeout by a factor of 10 to 4 seconds (sleep of 50ms multiplied with 80 tries). If this does not help try going even higher with that number.

stefson:
Compile with clang hangs at 43% with this error: 

```
/var/tmp/portage/games-strategy/s25rttr-9999-r4/work/s25rttr-9999/src/ai/AIInterface.cpp:112:18: error: comparison of constant -1 with expression of type 'char' is always false
      [-Werror,-Wtautological-constant-out-of-range-compare]
    if(direction == -1) // calculate complete value from scratch (3n^2+3n+1)
       ~~~~~~~~~ ^  ~~
1 error generated.
```

full build log: [clang-build.log.zip](https://github.com/Return-To-The-Roots/s25client/files/2212750/clang-build.log.zip)

(The cmake warnings are now silent)

Flamefire:
Thanks, fixed that and a stupid C&P bug.

Important: You are using `-march=armv7-a -mfpu=vfpv3-d16` manually, which I want to make not required (add automatically) with this PR. But the flags I found for RasPi2 are `-mcpu=cortex-a7  -mfpu=neon-vfpv4` (see https://gist.github.com/fm4dd/c663217935dc17f0fc73c9c81b0aa845) Where did you get yours from? Which are correct? Are there different RasPi2s?

stefson:
This doesn't really work, gives me the same error as your travis builds on both arm and my workstation. 

arm is hardly tested and there are many different chips out there with many different levels of optimization, `-mfpu=vfpv3-d16` is a way of playing it safe. Don't like to end with some super rare bugs just because my cflags are too agressive. Escpecially neon optimization is known for breaking stuff and to trigger compiler bugs. That being said, I can give it a go with those cflags you suggested, so we know they're not going to break the build. But first priority would be to go with the less agressive ones, second a debug with -Og -ggdb, and last would be with `-mfpu=neon-vfpv4`

Flamefire:
Yeah, typo...

Ok, the FPU stuff is confusing... Maybe we need a `RTTR_USE_NEON` option? In [this blog](https://embeddedartistry.com/blog/2017/10/9/r1q7pksku2q3gww9rpqef0dnskphtc) the stuff is detailed. From that I'd say you should use `vfpv4-d16` rather than `vfpv3-d16` for RasPi2, right?
Some more info on `RTTR_TARGET_BOARD`: "OFF" will not add any architecture related stuff.

Note: We already add `-ggdb` to Debug and RelWithDebInfo builds for GCC.

stefson:
@Leuchtefix75 @daviel - can you please test this as well? Specially if you can start a new game?

stefson:
Regarding the runtime issues, I have tested the increased buffer for connecting to the server, but without any success. It does increase the timeout as intended, but the connection still fails. Only thing I found out is that there is some activity in the background with the server: 

on my workstation, netstat -nptl shows: 
```
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name    
tcp        0      0 0.0.0.0:3665            0.0.0.0:*               LISTEN      29010/s25client     
```

I'll edit this post in the evening with the results from the rpi, right now busy with other tasks.

Flamefire:
I have no ideas ATM for the connection issue. We should maybe switch to Boost.ASIO instead of rolling our own implementation.
If someone is willing to write some test for that: Create a simple program, where you set up a listening socket in the main thread (code example in libutil) and connect to it from another socket. Don't use threads or other libraries (just like we do), and then test what happens if the server is running in a separate thread.
Maybe it is some firewall stuff or so.

`use_upnp` would use universal Plug&Play to open ports in the router (I think). Valid is 0 and 1 (off and on)

The config menu says "fullscreen resolution". So this is why the window does not change. You can resize the window manually to whatever you want (almost)

stefson:
Well, I certainly am not capable of writing this test programm. As there is no iptables or other firewall/apparmor around on my rpi2 device, and the buffer got increased without much success, I'd suggest to wait a bit from input of the two other users with arm hardware. 

Compile with `-mcpu=cortex-a7 -mfpu=neon-vfpv4` will be through in the evening.

Flamefire:
I made one: https://gist.github.com/Flamefire/0340a8fc56506cc8a1f4c52e91c5b0f5

Compile with `g++ main.cpp -DUSE_THREAD=0 -lpthread -std=c++11` and test, then use `g++ main.cpp -DUSE_THREAD=1 -lpthread -std=c++11` and test again

stefson:
This gives me a binary named a.out, if this is executed via ./a.out there is no output to the console at all for `-DUSE_THREAD=0` and for `-DUSE_THREAD=1` there is only `Started server`; with 100% cpu load on one core for some 2 minutes. Both error out with error 110 when some 2 minutes have passed. High cpu load only for `-DUSE_THREAD=1` 

If executing via ./a.out is the correct way to test this, it succeeds on my workstation. Connected after 0ms for `-DUSE_THREAD=0` and 
``` Started serverConnected after 
0ms
Got client
```
for `-DUSE_THREAD=1` 

(in the meantime, `-mcpu=cortex-a7` -mfpu=neon-vfpv4 has passed)

Flamefire:
Ok, so the whole socket thing fails on the PI. No idea why though as I think this is the textbook example of a server/client app. If anyone wants to help out on that feel free.

stefson:
With ubuntu-16.04 on the rpi2, there is one error in cmake-gui during configure: [cmake-3.5-error.log.zip](https://github.com/Return-To-The-Roots/s25client/files/2220972/cmake-3.5-error.log.zip)

its an older cmake (3.5)

stefson:
Good news about ubuntu-16.04 - your test program works! 

Bad news - the Readme.md needs an update on how to compile with cmake, checkout to s25client && `mkdir build` && `cmake -DCMAKE_BUILD_TYPE=Release` gives this error: `The source directory "~/s25client-0d8e96a8512c8190e631eaa3e9e40461b0f37c6c/build" does not appear to contain CMakeLists.txt.`

And going via cmake-gui the one I posted in my last posting.

Flamefire:
> With ubuntu-16.04 on the rpi2, there is one error in cmake-gui during configure: cmake-3.5-error.log.zip

`git submodule update`

> I propably have to use build directory from within the source dir as the binary file, not some other randoom dir as I did.

Yes. Either follow the directions in the readme or adjust the commands

stefson:
So, got it runtime tested with ubuntu-16.04 on the rpi2. 

It's all working, starting new game and so, only thing not working is reasonable fps. Between 6 (menu) and 11 (game) isn't playable, but this is likely due to outdated mesa and kernel on 16.04; 18.04 is on the horizon and I'll test that soon, or raspbian with a more recent kernel. 

What are the dependencies of your main.cpp? Looks like some sort of boost, as it includes iostream headers? I would like to open a bug for gentoo, with your code as an example which fails.

Flamefire:
Fps might get better after my ogl rework. But it does work now without other changes? Others said that raspi does not support (old) OpenGL...

Nothing. Just c++11
How did you get it to work on the raspi w/I the connection working

stefson:
I used your forked cmake branch, which is this very PR and compiled it with cmake and make as suggested in the README.md, on the rpi2 with ubuntu-16.04. The main.cpp test did actually work here, while it doesn't work with Gentoo. So it's a good test to trigger the issue, which raises chances the initial bug can be found - but for that I have to know which deps to install for it. glibc propably, and boost(?), what else? 

I'll open a bug on the gentoo bugtracker, and turn back here only if there are questions.

stefson:
So I've got a question, apparently: there is a testsuite available, and it fails with this error on the rpi2 with Gentoo: 

```/tmp/s25rttr-9999/src/test/testLobby.cpp(100): fatal error: in "Lobby/LobbyChat": critical check LobbyClient::inst().Login("localhost", lobbyPort, lobbyServer.testUser, lobbyServer.testPw, false) has 
failed
Connection to 127.0.0.1:5664
Connection failed
Error: Connection timed out
Error connection to localhost:5664
Sending Message to server failed
``` 

Also there are 2 failed test with lua, for which I'm going to open a new issue. Would you like to continue to discuss the server issue here, or should I open a seperate issue for tracking and questions?

Flamefire:
Honestly: I don't care about that connection error. This is something wrong on your side and I don't want to spend the effort handling that. It looks like the on tested here https://gist.github.com/Flamefire/0340a8fc56506cc8a1f4c52e91c5b0f5 and need to be fixed outside of rttr.

So unless you can provide information on how this is an issue in rttr, I don't see a point discussing this here. This is for the Gentoo/RasPi folks, not us.

stefson:
I tested a few different distros in the last days. 

playable: 
armv6-hardfloat based: raspbian-9, ubuntu-16.04 - no vc4 acceleration yet, so fps is <10fps 
armv7-hardfloat based: arch linux, no full vc4 acceleration yet

vc4 is more about libdrm and mesa getting stuff done, raspbian-9 has kernel 4.14 but situation did not improve just because of newer kernel. 

not playable
Gentoo still not working, but it has full vc4 acceleration (60 fps in menu)

glxgears -info has to show something with Gallium, llvm-pipe means there is only 2d acceleration available.

Flamefire:
I'm pretty sure you need hardware acceleration for OpenGL. I've seen tutorials/blogs around how to do this on RasPIs so it should work. Without getting that to work first don't even try rttr. In #848 I'm getting the OpenGL stuff modernized. I even have an SDL2 based driver ready and working. If you have SDL2 installed, it will pick it up and allow creation of native GLES2 contexts (not yet, but I'm getting there)

Issue #856, opened 2018-07-28
ENH: Update Mingw Compiler


Flow86:
Issue to collect merge requests to that topic

- [x] https://github.com/Return-To-The-Roots/s25update/pull/9
- [x] https://github.com/Return-To-The-Roots/libutil/pull/21
- [ ] https://github.com/Return-To-The-Roots/s25client/pull/857

PR #857, opened 2018-07-28
ENH: Update Mingw Compiler

Flow86:
https://github.com/Return-To-The-Roots/s25client/issues/856

I also had to add FindSDL_mixer.cmake and fix the include path find code. it seems to be wrong in cmake 3.0.2


release/prepareRelease.sh.cmake
```diff
@@ -91,7 +91,7 @@ extract_debug_symbols()
 	objcopyTarget=""
 	case "$SYSTEM_NAME" in
 		Windows)
-			objcopyTarget="-pc-mingw32"
+			objcopyTarget="-w64-mingw32"
```
Flamefire: Couldn't we get this from the compiler? Needs double checking but that is how CMake works: Find the tools from the compiler prefix (not the variable but the actual exe). We can put the compiler in here (this is configured) and match it with simple regex. This would replace at least lines 74-129

release/prepareRelease.sh.cmake
```diff
@@ -238,39 +259,56 @@ case "$SYSTEM_NAME" in
...
 					mingw=/usr/i686-pc-mingw32
 				else
 					mingw=/usr/i686-mingw32
 				fi
 				lua=win32
 			;;
 			x86_64|*64)
-				if [ -d /usr/i686-pc-mingw32 ]; then
+				if [ -d /usr/i686-w64-mingw32 ]; then
+					mingw=/usr/x86_64-w64-mingw32
```
Flamefire: Isn't this the path to the compiler? We can get it from there I guess. (see above)

release/prepareRelease.sh.cmake
```diff
@@ -238,39 +259,56 @@ case "$SYSTEM_NAME" in
...
+		if [ -f ${mingw}/bin/libgcc_s_sjlj-1.dll ] ; then
+			cp -v ${mingw}/bin/libgcc_s_sjlj-1.dll ${DESTDIR} || exit 1
+			cp -v ${mingw}/bin/libintl-8.dll ${DESTDIR} || exit 1
+
+			cp -v ${mingw}/bin/libgcc_s_sjlj-1.dll ${DESTDIR}RTTR || exit 1
+			cp -v ${mingw}/bin/zlib1.dll ${DESTDIR}RTTR || exit 1
+			cp -v ${mingw}/bin/libminiupnpc-5.dll ${DESTDIR} || exit 1
+		else
+			cp -v ${mingw}/bin/libgcc_s_seh-1.dll ${DESTDIR} || exit 1
+			cp -v ${mingw}/bin/libstdc++-6.dll ${DESTDIR} || exit
```
Flamefire: Hu? libstdc++ instead of libintl and zlib?

cmake/Modules/FindSDL_mixer.cmake
```diff
@@ -0,0 +1,110 @@
+#.rst:
```
Flamefire: I don't think this is a good idea. Can't we just use a more recent version of CMake? 3.5 would be my minimum recommendation but 3.8 would be better as it adds `target_compile_features(... cxx_14)`

release/prepareRelease.sh.cmake
Flow86: yep perhaps simply we let it "find" via cmake and put the objcopy as a variable in there?

release/prepareRelease.sh.cmake
Flow86: yep

cmake/Modules/FindSDL_mixer.cmake
Flow86: yea 3.6 is in backports, so should be not a problem. I only wanted to fix it first before changing a lot of versions. We could leave the file out in the end.

release/prepareRelease.sh.cmake
Flamefire: Maybe it's even easier to just translate this into cmake and run it in script mode? It would keep all the logic here and makes maintenance easier. E.g. I'm much better in cmake than shell

cmake/Modules/FindSDL_mixer.cmake
Flamefire: Would be easier to do earlier than later. There might be more bugs in a x.0 version ;)

release/prepareRelease.sh.cmake
Flow86: yep, CMAKE_FIND_ROOT_PATH ;-)

release/prepareRelease.sh.cmake
```diff
@@ -234,43 +197,53 @@ case "$SYSTEM_NAME" in
...
-		cp -v ${mingw}/bin/libminiupnpc-5.dll ${DESTDIR} || exit 1
-		cp -v ${mingw}/bin/libiconv-2.dll ${DESTDIR} || exit 1
-		cp -v ${mingw}/bin/libintl-8.dll ${DESTDIR} || exit 1
-		cp -v ${mingw}/bin/libogg-0.dll ${DESTDIR} || exit 1
-		cp -v ${mingw}/bin/SDL_mixer.dll ${DESTDIR} || exit 1
-		cp -v ${mingw}/bin/SDL.dll ${DESTDIR} || exit 1
-		cp -v ${mingw}/bin/libvorbis-0.dll ${DESTDIR} || exit 1
-		cp -v ${mingw}/bin/libvorbisfile-3.dll ${DESTDIR} || exit 1
+
+		if [ -f $SYSROOT/bin/libgcc_s_sjlj-1.dll ] ; then
```
Flow86: hier ists doch etwas "komplizierter" da der 32bit mingw compiler die sjlj erzeugt, der 64bit die seh. Beide brauchen jedoch in der neuen Version zwingend die libstdc++

Flamefire:
I think everything is included already. @Flow86 can you delete the branch if this is true?

PR #865, opened 2018-08-27
Added cheats to the chat commands

gggfreak2003:
Added cheats to RTTR.

The cheats has to be activated by entering "winter". After the cheats are activated, they could be entered as chat commands (Press ENTER and then "!gold" for 10 units gold). Valid commands are: ironore, beer, tongs, hammer, axe, saw, pickaxe, shovel, crucible, rodandline, scythe, cleaver, rollingpin, bow, boat, iron, flour, fish, bread, wood, boards, stones, grain, coins, gold, coal, meat, ham, water. The command "givemeall" gives you 10 units of all goods.

### DIFF
--- a/src/desktops/dskGameInterface.cpp
+++ b/src/desktops/dskGameInterface.cpp
@@ -1020,7 +1020,7 @@ void dskGameInterface::OnChatCommand(const std::string& cmd)
             worldViewer.InitTerrainRenderer();
         }
     }
-    else if(cmd == "ironore")	cheatAdd(GD_IRONORE, 10, cmd);
+    else if(cmd == "ironore") cheatAdd(GD_IRONORE, 10, cmd);
     else if(cmd == "beer") cheatAdd(GD_BEER, 10, cmd);
     else if(cmd == "tongs") cheatAdd(GD_TONGS, 10, cmd);
     else if(cmd == "hammer") cheatAdd(GD_HAMMER, 10, cmd);
@@ -1070,7 +1070,7 @@ void dskGameInterface::cheatAdd(GoodType type, unsigned value, const std::string
 	}
 }
 
-void dskGameInterface::cheatAddAll(){
+void dskGameInterface::cheatAddAll(unsigned value=10){
 	if(isCheatModeOn) {
 		GoodType arr[] = { GD_IRONORE, GD_BEER, GD_TONGS, GD_HAMMER, GD_AXE, GD_SAW,
 				GD_PICKAXE, GD_SHOVEL, GD_CRUCIBLE, GD_RODANDLINE, GD_SCYTHE,
@@ -1083,10 +1083,10 @@ void dskGameInterface::cheatAddAll(){
 
 		// Iterate over the list and display numbers
 		for (GoodType val : listOfInts)
-			cheatAdd(val, 10, "");
+			cheatAdd(val, value, "");
 
 		char text[256];
-		snprintf(text, sizeof(text), "Cheat: Added 10 of all goods");
+		snprintf(text, sizeof(text), _("Cheat: Added %d of all goods"), value);
 		messenger.AddMessage("", 0, CD_SYSTEM, text, COLOR_RED);
 	}
 }
--- a/src/desktops/dskGameInterface.cpp
+++ b/src/desktops/dskGameInterface.cpp
@@ -690,7 +690,7 @@ bool dskGameInterface::Msg_KeyDown(const KeyEvent& ke)
                     isCheatModeOn = !isCheatModeOn;
                     curCheatTxt.clear();
                     char text[256];
-                    snprintf(text, sizeof(text), "Cheat mode is on");
+                    snprintf(text, sizeof(text), _("Cheat mode is %s"), isCheatModeOn ? _("on") : _("off"));
                     messenger.AddMessage("", 0, CD_SYSTEM, text, COLOR_RED);
                 }
             } else
@@ -1019,7 +1019,8 @@ void dskGameInterface::OnChatCommand(const std::string& cmd)
             const_cast<GameWorld&>(game_->world).GetDescriptionWriteable() = newDesc;
             worldViewer.InitTerrainRenderer();
         }
-    } else if(cmd == "ironore")	cheatAdd(GD_IRONORE, 10, cmd);
+    }
+    else if(cmd == "ironore")	cheatAdd(GD_IRONORE, 10, cmd);
     else if(cmd == "beer") cheatAdd(GD_BEER, 10, cmd);
     else if(cmd == "tongs") cheatAdd(GD_TONGS, 10, cmd);
     else if(cmd == "hammer") cheatAdd(GD_HAMMER, 10, cmd);
--- a/src/desktops/dskGameInterface.cpp
+++ b/src/desktops/dskGameInterface.cpp
@@ -689,6 +689,9 @@ bool dskGameInterface::Msg_KeyDown(const KeyEvent& ke)
                 {
                     isCheatModeOn = !isCheatModeOn;
                     curCheatTxt.clear();
+                    char text[256];
+                    snprintf(text, sizeof(text), "Cheat mode is on");
+                    messenger.AddMessage("", 0, CD_SYSTEM, text, COLOR_RED);
                 }
             } else
                 curCheatTxt.clear();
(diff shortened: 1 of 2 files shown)


src/desktops/dskGameInterface.cpp
```diff
@@ -689,6 +689,9 @@ bool dskGameInterface::Msg_KeyDown(const KeyEvent& ke)
                 {
                     isCheatModeOn = !isCheatModeOn;
                     curCheatTxt.clear();
+                    char text[256];
+                    snprintf(text, sizeof(text), "Cheat mode is on");
```
baka0815: Shouldn't this check for ``isCheatModeOn``?

src/desktops/dskGameInterface.cpp
```diff
@@ -1016,7 +1019,75 @@ void dskGameInterface::OnChatCommand(const std::string& cmd)
             const_cast<GameWorld&>(game_->world).GetDescriptionWriteable() = newDesc;
             worldViewer.InitTerrainRenderer();
         }
-    }
+    } else if(cmd == "ironore")	cheatAdd(GD_IRONORE, 10, cmd);
```
baka0815: Line break before the ``else``?

src/desktops/dskGameInterface.cpp
```diff
@@ -1016,7 +1019,75 @@ void dskGameInterface::OnChatCommand(const std::string& cmd)
...
+
+		// Create a list and initialize it with vector
+		std::list<GoodType> listOfInts(arr, arr + sizeof(arr) / sizeof(int));
+
+		// Iterate over the list and display numbers
+		for (GoodType val : listOfInts)
+			cheatAdd(val, 10, "");
+
+		char text[256];
+		snprintf(text, sizeof(text), "Cheat: Added 10 of all goods");
```
baka0815: Should be translated (see ``cheatAdd``):
````cpp
snprintf(text, sizeof(text), _("Cheat: Added 10 of all goods"));
````

File: src/desktops/dskGameInterface.cpp (hunk 1 above)
Comment (gggfreak2003) [reply src/desktops/dskGameInterface.cpp]: You are right. Thanks

File: src/desktops/dskGameInterface.cpp (hunk 3 above)
Comment (gggfreak2003) [reply src/desktops/dskGameInterface.cpp]: How I could add a translation?

Comment (Flow86) [reply src/desktops/dskGameInterface.cpp]: ```_("Text to translate")```

File: src/desktops/dskGameInterface.cpp [hunk 4]
```diff
@@ -1017,6 +1020,75 @@ void dskGameInterface::OnChatCommand(const std::string& cmd)
             worldViewer.InitTerrainRenderer();
         }
     }
+    else if(cmd == "ironore")	cheatAdd(GD_IRONORE, 10, cmd);
```
Flow86: isnt there a way to the the name from an GD_... constant? so no every good type works then. would be much better and less error prone

src/desktops/dskGameInterface.cpp
```diff
@@ -1017,6 +1020,75 @@ void dskGameInterface::OnChatCommand(const std::string& cmd)
...
+
+		// Create a list and initialize it with vector
+		std::list<GoodType> listOfInts(arr, arr + sizeof(arr) / sizeof(int));
+
+		// Iterate over the list and display numbers
+		for (GoodType val : listOfInts)
+			cheatAdd(val, value, "");
+
+		char text[256];
+		snprintf(text, sizeof(text), _("Cheat: Added %d of all goods"), value);
```
Shawn8901: I am not sure which way is better, but all other places in the codebase use `%` to replace the placeholder.

src/desktops/dskGameInterface.cpp
```diff
@@ -689,6 +689,9 @@ bool dskGameInterface::Msg_KeyDown(const KeyEvent& ke)
                 {
                     isCheatModeOn = !isCheatModeOn;
                     curCheatTxt.clear();
+                    char text[256];
+                    snprintf(text, sizeof(text), _("Cheat mode is %s"), isCheatModeOn ? _("on") : _("off"));
```
Shawn8901: i would here also prefer the `%`-operator instead the snprintf

Flamefire: Don't use snprintf at all. Use Boost format which is safer. BTW: No need for this message. There already is an icon showing whether cheat mode is on

src/desktops/dskGameInterface.cpp
```diff
@@ -1017,6 +1020,75 @@ void dskGameInterface::OnChatCommand(const std::string& cmd)
...
+    else if(cmd == "boards") cheatAdd(GD_BOARDS, 10, cmd);
+    else if(cmd == "stones") cheatAdd(GD_STONES, 10, cmd);
+    else if(cmd == "grain") cheatAdd(GD_GRAIN, 10, cmd);
+    else if(cmd == "coins") cheatAdd(GD_COINS, 10, cmd);
+    else if(cmd == "gold") cheatAdd(GD_GOLD, 10, cmd);
+    else if(cmd == "coal") cheatAdd(GD_COAL, 10, cmd);
+    else if(cmd == "meat") cheatAdd(GD_MEAT, 10, cmd);
+    else if(cmd == "ham") cheatAdd(GD_HAM, 10, cmd);
+    else if(cmd == "water") cheatAdd(GD_WATER, 10, cmd);
+    else if(cmd == "givemeall") cheatAddAll();
```
Flamefire: You could use the `WARE_NAMES` array to map a string to a good.

src/desktops/dskGameInterface.cpp
```diff
@@ -1017,6 +1020,75 @@ void dskGameInterface::OnChatCommand(const std::string& cmd)
...
+
+void dskGameInterface::cheatAdd(GoodType type, unsigned value, const std::string& type_string) {
+	if(isCheatModeOn) {
+		GameWorldBase& world = worldViewer.GetWorldNonConst();
+		GamePlayer& player = world.GetPlayer(worldViewer.GetPlayerId());
+
+		nobBaseWarehouse* wh = player.GetFirstWH();
+		Inventory goods;
+		goods.Add(type, value);
+		wh->AddGoods(goods, true);
```
Flamefire: This only works for singleplayer and will async in replays and multiplayer. Correct way would be to add a GameCommand for that, especially as the viewer should not change the game (`GetWorldNonConst` is a work-around for the current messy code base)

src/desktops/dskGameInterface.cpp
```diff
@@ -1017,6 +1020,75 @@ void dskGameInterface::OnChatCommand(const std::string& cmd)
...
+
+		if(type_string.length() != 0) {
+			char text[256];
+			snprintf(text, sizeof(text), _("Cheat: Added %d %s"), value, type_string.c_str());
```
Flamefire: Use `Boost.Format`

src/desktops/dskGameInterface.cpp
```diff
@@ -1017,6 +1020,75 @@ void dskGameInterface::OnChatCommand(const std::string& cmd)
...
+			messenger.AddMessage("", 0, CD_SYSTEM, text, COLOR_RED);
+		}
+	}
+}
+
+void dskGameInterface::cheatAddAll(unsigned value=10){
```
Flamefire: Default value in definition is not allowed (Compilation should fail)

src/desktops/dskGameInterface.cpp
```diff
@@ -1017,6 +1020,75 @@ void dskGameInterface::OnChatCommand(const std::string& cmd)
...
+				GD_PICKAXE, GD_SHOVEL, GD_CRUCIBLE, GD_RODANDLINE, GD_SCYTHE,
+				GD_CLEAVER, GD_ROLLINGPIN, GD_BOW, GD_BOAT, GD_IRON, GD_FLOUR,
+				GD_FISH, GD_BREAD, GD_WOOD, GD_BOARDS, GD_STONES, GD_GRAIN,
+				GD_COINS, GD_GOLD, GD_COAL, GD_MEAT, GD_HAM, GD_WATER };
+
+		// Create a list and initialize it with vector
+		std::list<GoodType> listOfInts(arr, arr + sizeof(arr) / sizeof(int));
+
+		// Iterate over the list and display numbers
+		for (GoodType val : listOfInts)
```
Flamefire: C++11 is not possible ATM, use BOOST_FOREACH

src/desktops/dskGameInterface.cpp
```diff
@@ -1017,6 +1020,75 @@ void dskGameInterface::OnChatCommand(const std::string& cmd)
...
+			char text[256];
+			snprintf(text, sizeof(text), _("Cheat: Added %d %s"), value, type_string.c_str());
+			messenger.AddMessage("", 0, CD_SYSTEM, text, COLOR_RED);
+		}
+	}
+}
+
+void dskGameInterface::cheatAddAll(unsigned value=10){
+	if(isCheatModeOn) {
+		GoodType arr[] = { GD_IRONORE, GD_BEER, GD_TONGS, GD_HAMMER, GD_AXE, GD_SAW,
```
Flamefire: Use a `boost::array` with a fixed size. This also saves the copy to a list below.

Flow86:
any progress here?

src/desktops/dskGameInterface.cpp
Flamefire: see https://github.com/Return-To-The-Roots/s25client/pull/865#discussion_r214508949 `WARE_NAMES`

Flow86:
stale for a long time

PR #866, opened 2018-08-27
Add statstics for raw materials

gggfreak2003:
Addon to show how many units are left in mines and wells. (See issue #106 )


src/addons/const_addons.h
```diff
@@ -78,7 +79,9 @@ ENUM_WITH_STRING(AddonId, LIMIT_CATAPULTS = 0x00000000, INEXHAUSTIBLE_MINES = 0x
 
                  NUM_SCOUTS_EXPLORATION = 0x00C00000,
 
-                 FRONTIER_DISTANCE_REACHABLE = 0x00D0000, COINS_CAPTURED_BLD = 0x00D0001, DEMOLISH_BLD_WO_RES = 0x00D0002)
+                 FRONTIER_DISTANCE_REACHABLE = 0x00D0000, COINS_CAPTURED_BLD = 0x00D0001, DEMOLISH_BLD_WO_RES = 0x00D0002,
+
+				 SHOWRESOURCES = 0x00E00000)
```
Shawn8901: You are mixing tab in here if i am right, you should definitely check your formatter. the project uses clang-format.

src/ingameWindows/iwBuilding.cpp
```diff
@@ -100,6 +103,32 @@ iwBuilding::iwBuilding(GameWorldView& gwv, GameCommandFactory& gcFactory, nobUsu
 
     AddText(10, DrawPoint(113, 50), _("(House unoccupied)"), COLOR_RED, FontStyle::CENTER, NormalFont);
 
+    if(gwv.GetWorld().GetGGS().isEnabled(AddonId::SHOWRESOURCES))
+    {
+		if(BuildingProperties::IsMine(building->GetBuildingType()))
```
Shawn8901: I dont see the reason why not share the code for mine and well? i would prefer the de-doubled code, instead the "optimized" call stack for wells.

Flow86: yep thats right, thats duplicated. why not simply add a `case BLD_WELL` ?

src/ingameWindows/iwBuilding.cpp
```diff
@@ -100,6 +103,32 @@ iwBuilding::iwBuilding(GameWorldView& gwv, GameCommandFactory& gcFactory, nobUsu
...
+		{
+			Resource::Type res;
+			switch(building->GetBuildingType())
+			{
+				case BLD_GOLDMINE: res = Resource::Type::Gold; break;
```
Shawn8901: This `Resource::Type::Gold` class/namespace mixing is a Visual Studio only addon to the c++0x stuff, while the project also builds for older versions. so this is breaking the auto build.

Shawn8901:
There is still an issue with the enum stuff on gcc, which breaks the TravisCI build. 

`/home/travis/build/Return-To-The-Roots/s25client/src/ingameWindows/iwBuilding.cpp:111:39: error: ‘Resource::Type’ is not a class or namespace
    case BLD_GOLDMINE: res = Resource::Type::Gold; break;`

Flow86:
stale for a long time

PR #901, opened 2018-11-03
improve addon description

Xermio:


Flamefire:
@Spikeone What do you say? LGTM

Spikeone:
@Flamefire Well, I'm not sure at all. Is it really true, that "blocked by stone" is taken into account? I hope this is not the case as stones are often meant as temporary barrier...
If the description is correct, I'd ask for an addon change, if the description is wronge, I'd ask to change the description ;)

On second thought, stone might not be granite as this is what quarrys mine - still this could confuse people like I was :confused:

Flamefire:
Checked the implementation: Only terrain is checked, no objects. @Xermio Could you update the description please?

Shawn8901:
Yeah Stone is not included. We discussed it already when i submitted the PR, that objects like stone/wood should not be checked.
So it should be updated to something like `(Path blocked by terrain like sea, lava, etc.)`?

Xermio:
True, stone does really not count as blocking. Updated the description, please check again.

Spikeone:
Are static object taken into account? For example ruins? Are LUA spawned objects taken into account? One should then add "static objects" or something like that, but with the current description I'm also okay :)

Flamefire:
Nope they are not.

PR #904, opened 2018-11-03
Load music from memory instead of from file

### LINKED ISSUE #893: Minilag when a new song is played
Once you start playing and a new song is loaded there is a mini lag in game. This happens only once for any song (guess the song is still loaded next time it plays).

Flamefire:
Fixes #893


driver/audio/SDL/src/AudioSDL.cpp
```diff
@@ -134,8 +134,20 @@ SoundHandle AudioSDL::LoadEffect(const std::string& filepath)
         return SoundHandle();
     }
 
-    SoundSDL_Effect* sd = new SoundSDL_Effect(sound);
-    return CreateSoundHandle(sd);
+    return CreateSoundHandle(new SoundSDL_Effect(sound));
+}
+
+SoundHandle AudioSDL::LoadEffect(const std::vector<char>& data, const std::string& ext)
```
Flow86: whats the ext for? I dont see where its used?

Flamefire: The idea was to keep the interface generic so an implementation could fall back to writing to file and loading from there. See the original source

PR #905, opened 2018-11-03
Add new map

Xermio:
Selfmade map called "Neighbourhood".

8 player, but playable with less too of course.
Also nice to play with 3 players vs 5 AIs.


Spikeone:
Damn, how can I download the map to check it? :x

Xermio:
Should be downloadable within the commit. @Spikeone 

https://github.com/Return-To-The-Roots/s25client/blob/62bed02610ea40476d8922b22c1403b28dbe35d4/RTTR/MAPS/NEW/Neighbourhood.SWD

Flow86:
@Spikeone 
use "Files changed" and then "View" on the file. There you can download it.

Xermio:
The simple map design was by intention. But I can make a "improved" public version.

I personally like flat terrains and don't need "visual objects". :)

Spikeone:
Hm - what about we put those map into "other" then? we should also clean up the other maps folder as well.

Spikeone:
https://drive.google.com/open?id=1K7WGWJRk4-jSHAXZDc1OspXT8aCxsagq this one is a slightly better version. Although I still propose adding this map to the "other" folder and cleaning up all maps so the "new" (RttR) folder contains all multiplayer maps which are good and other contains singpleplayer maps or lower quality multiplayer maps. There is nothing more frustrating than playing a map for the first time and realizing that one player just can't win at all.

Xermio:
Moved the map to the "OTHER" folder. Merge it as is or not.

PR #907, opened 2018-11-05
Improve colors for disabled buttons

### LINKED ISSUE #888: Better disabled button
When disabling buttons the left and top highlight are still shown as they were active which looks quite odd. The highlight brightness should be adjusted - or just use the buttons normal texture which would result in the following: 
image
(left current, right new)

Flamefire:
Closes #888 

It uses the font color as suggested and brings the texture color closer to the suggestion: 

The complete match is not possible as we colorize a colored texture where the original uses completely other colors (grey in this case) So we can basically just make it slightly darker or lighter

### DIFF
--- a/src/controls/ctrlButton.cpp
+++ b/src/controls/ctrlButton.cpp
@@ -104,7 +104,7 @@ void ctrlButton::Draw_()
 
     if(tc != TC_INVISIBLE)
     {
-        unsigned color = isEnabled ? COLOR_WHITE : 0xFF666666;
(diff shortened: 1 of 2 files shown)


Spikeone:
Is it possible to darken the "green button" texture a bit less? Guess that also helps to decrease the text contrast which is the major problem"

Flamefire:
I made it a bit lighter in 4 steps:



The last one is 2 steps darker than the normal button. Which one is it?

Spikeone:
3rd one in my opinion

Issue #798, opened 2018-01-17
Player avatars


Flamefire:
The original had avatars for each nation. We still display them at the statistics screen but use the same for every player of the same nation.
We could make the choice completely free by adding it to the match-making screen and not tie it to a nation at all.

Do we want that? How could the UI look like? (space is rather limited for an image)

Ccing @Spikeone (maybe you have an idea)

KaioKen-J:
Even tho this isnt necessary, it would be a neat addition. Space does look very limited sadly, maybe if you shorten each tile a bit, especialy the playername one, you could squeeze in the picture like shown here?
https://i.imgur.com/rlhBsgH.png

If not at least make the roman player play as Octavianus and not Brutus ^^

Flow86:
we could even integrate the player avatars from the website (lobby only)

Flamefire:
Possible. But that means getting data from the website. I don't think that would be wise as it greatly complicates things. Then rather read a custom avatar from an image in the players folder.

Spikeone:
also see #125 - still closing the old ticket as it is in german.

Flow86:
we can get the user avatar only for lobby players. if one is cached locally then, use it for the rest of the games too?

Spikeone:
@Flow86 @Flamefire Not really sure what you mean by "cached locally" but we could add it next to the name option in the options screen. 
We could also add a for example CUSTOM_PLAYER.png which everyone could edit for themselves, sending it to other players could be done by naming it with it's hash or something like that (maybe thats the cache you mean).

Flow86:
Yes: show it on settings:
- Use one of the predefined S2 ones
- Use avatar from website (if once logged in to lobby to get the website avatar)
(- Set one from a local file?)

Then we need to implement it to compress & transfer it to all players on a multiplayer game.
Also cache them perhaps so we don't need to retransfer them every time (i.e hash username+mod-date, etc)

theBasstian:
How about a SetAvatar() ?
Would it be that complicated ? I could really get use of that for specific characters in story-driven missions ...

Spikeone:
@theBasstian guess you are talking about this one: #1181

tehKaiN:
[...]

PR #909, opened 2018-11-05
Show instant Victory

Flamefire:


Flamefire:
@Spikeone Check Code starting at https://github.com/Return-To-The-Roots/s25client/pull/909/commits/60bb788aca604f1c90837e08ff33ef90ab43c640#diff-a828fcc69b4fadf2331e1b11fbe6a6caR1277

What do you think?

Spikeone:
As it will take ages for me to get a build running, it should look that way:
image

Flamefire:
That's why I referred to the code. I think it is simple enough to get an idea what it does. Currently it looks like:

bildschirmfoto vom 2018-11-09 09-27-20

Note how the title changes between victory and defeat and that additionally a log message is shown.

Your idea is also good but that requires #798 and an idea how to scale this to many players (make it simply as wide as the images get?) Also are the colors enough or do we want the player names too?

Flamefire:
Ok, for now I suggest: I create a window similar to the original, but with a list of player names instead of the avatars as we don't have them. The text will either be "The winners" or "The winner". There will be 1 name per line. The message text in the main window should be kept to be consistent with the "player x defeated" text.

You create another issue depending on the avatar issue with your suggestion. Maybe create a mockup how the player names shall be put into the color box. IMO this will be very hard because what color shall the text be? The player colors may be random (in the future) so the text can get unreadable. Also what happens for long(er) names? All this need to be solved but as a separate issue. First we need "something". As we don't have avatars I don't see any other option than a plain list of names

Spikeone:
Agreed, will create an issue with mockup

Flamefire:
@Spikeone *ping*

Spikeone:
Really need to check whats up with my notifications.

@Flamefire _pong_ - please change the "continue game"'s button texture to gray and translate "end game" :)

Flamefire:
Ah yes, forgot that ;)

"End game" is translated (original language is English), the rest is not as it's new.

Flamefire:
New dialogue:

bildschirmfoto vom 2018-11-20 09-36-08

baka0815:
Looks like there is a translation problem as "Continue game" is in English while "Spiel beenden" is in German.

PR #939, opened 2018-11-15
 Fix and test Click and DoubleClick behaviour

Flamefire:
TickCount is time in ms, document this and fix conversion
Add test via mocked class using turtle


stefson:
looks good to me.

Flamefire:
Requires https://github.com/mat007/turtle/pull/54

PR #948, opened 2018-11-21
Fix minimize minimap enlarging window

iceslab:
Fixes bug appearing when minimizing or maximizing minimap (title bar enlarges every time window is minimized). I am unsure if it's the best way to do so.


Flamefire:
Good catch, thanks!

PR #949, opened 2018-11-21
Restructure source tree

Flamefire:
Restructure source tree to a more modular layout

This now follows the "pitchfork" proposal:
https://api.csswg.org/bikeshed/?force=1&url=https://raw.githubusercontent.com/vector-of-bool/pitchfork/develop/data/spec.bs

Also cleans CMake files a bit


Flamefire:
@Flow86 What do you say? This really helps breaking down the project libs into smaller modules so they can be reused and easier tested.

I wanted to change release process but there are just to many root folders to see what is going on

Flamefire:
@Flow86 Everything ready now, will merge on weekend if no one opposes

PR #952, opened 2018-11-27
Fix cursor image while scrolling

Flamefire:
Minor refactoring to make Code more testable.
Keep cursor scoll image while scrolling
Test those for all cases:
- Regular
- Left click cancels it
- Opening/closing window does not cancel scrolling unless it is modal
- Scolling in road mode, at end of scroll you are still in road mode
- Ctrl+leftclick scrolls, left-up cancels

@Spikeone To keep scrolling when opening a window I wanted to stay true to the S2 behaviour but that is inconsistent:
- Opening a (ingame) window makes it "active" (visible on title bar)
- But you keep scrolling -> Background is actually active
- For this I don't make the background(desktop) inactive while scrolling, so desktop AND window is active
- This way input (key,mouse) goes to the desktop not the window and this works

But now we have 2 active "windows" which I don't like out of consistency and the ingame window does not get any key events until it is clicked again which may be surprising. I don't know if this is used anywhere but to avoid issues I'd deactivate the new window while scrolling. The only change to how it is currently implemented (with this PR) is that the new window title bar does not get the active mark until you stop scrolling and click into it.

### DIFF
--- a/driver/include/driver/MouseCoords.h
+++ b/driver/include/driver/MouseCoords.h
@@ -28,17 +28,20 @@
 class MouseCoords
 {
 public:
-    MouseCoords() : x(0), y(0), ldown(false), rdown(false), dbl_click(false) {}
-    MouseCoords(int x, int y, bool ldown, bool rdown, const bool dbl_click) : x(x), y(y), ldown(ldown), rdown(rdown), dbl_click(dbl_click)
+    MouseCoords() : pos(0, 0), ldown(false), rdown(false), dbl_click(false) {}
+    MouseCoords(int x, int y, bool ldown = false, bool rdown = false, bool dbl_click = false)
+        : pos(x, y), ldown(ldown), rdown(rdown), dbl_click(dbl_click)
+    {}
+    MouseCoords(Position pos, bool ldown = false, bool rdown = false, bool dbl_click = false)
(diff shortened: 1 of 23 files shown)


Spikeone:
Can't one activate the window after scrolling ends? Like the last active winow (which is the latest opened window if one was opened)

But not sure what your changes feel like ingame - gotta wait for the next build.

Flamefire:
> Can't one activate the window after scrolling ends? Like the last active winow (which is the latest opened window if one was opened)

This further complicates the process. I feel that opening a window with the hotkey during scrolling and that NOT stopping scrolling is a bug. Because all other actions (with mouse) DO stop scrolling.

So I would either:

- New window stops scrolling
- New window gets immediately inactive when opened during scrolling. I mean: You are still pressing your mouse onto the desktop right? A click onto the desktop also deactivates the window, so it is only consistent.

Spikeone:
Well, as said I sadly fail to think about any situation this gets used. As Settlers 2 is now ~23 years old and I'm playing it for like 22 years I can not really tell how an action should be but I can tell if it feels odd / different. 

So, can only tell if this fix changes other QOL features after testing it

Flamefire:
Ok, then test it please. Remaining change: Opening a window does NOT make it active when scrolling OR opening a window DOES stop scrolling

PR #993, opened 2019-01-01
Remove CMake download work-around

Flamefire:


Flow86:
this will only work if we rework the docker stuff (i.e different images for each architecture) we also should not call the docker manually, but use the Jenkins(file) commands (i.e for authentication, etc)

Flow86:
nope the old image was not changed at all

PR #995, opened 2019-01-02
Use new docker images

Flamefire:


Flow86:
please rebase PR from s25client/new_docker

Flow86:
there's still a problem with boost detection. The new libraries are named xxx-mt-s but it always tries to add a compiler prefix, which is "" but always gets auto created by FindBoost.cmake because an empty var is a "false". what a mess

Flow86:
see 253a590

Flamefire:
> The new libraries are named xxx-mt-s 

The -s suffix is from static runtime. Did you really build them with static runtime?

Flow86:
yea I fixed that now in the boost library build, so the toolset-tag is there (it was missing before)
they changed the library naming when doing a "tagged" install with version 1.69 ... (...)

The build works now too by adding /usr/.../include to implicit include directories ca9ba6be

Flow86:
follow up #996

PR #996, opened 2019-01-02
Use new docker images

Flamefire:
Use the new toolchains


Jenkinsfile
```diff
@@ -38,7 +38,7 @@ def transformIntoStep(arch, wspwd, container) {
                                                          -v ~/.ccache:/workdir/.ccache \
                                                          \$VOLUMES \
                                                          --name "${env.BUILD_TAG}-${arch}" \
-                                                         ${container} -c \
+                                                         ${container} bash -c \
```
Flamefire: Why is this required? Should it be able to execute the shell script directly which will open a bash anyway?

Flow86: nope, the image does not have default command set, so you have to explicitly call ist,

without you get this:
```container_linux.go:247: starting container process caused "exec: \"-c\": executable file not found in $PATH"```

**Comment by Flamefire** [comment]
> and now its boost again f**ing up the build.
>
>`/usr/lib/gcc/i686-w64-mingw32/7.3-win32/include/c++/cstdlib:75:15: fatal error: stdlib.h: No such file or directory`
>
>caused by Boost_INCLUDE_DIRS being /usr/i686-w64-mingw32/include

Wildly speculating:
- stdlib.h is in /usr/i686-w64-mingw32/include
- cstdlib is in /usr/lib/gcc/i686-w64-mingw32/7.3-win32/include/c++
- boost is in /usr/i686-w64-mingw32/include
- gcc has internal includes with the `/usr/lib/gcc/i686-w64-mingw32/7.3-win32/include/c++` first and `/usr/i686-w64-mingw32/include`
- adding boost include with `-isystem` adds that to the top and removes it from gcc internals
- include_next searches following include dirs but the "correct" one is skipped

~~Can you check how far I'm correct. How was it done in the other image?~~ Forget it. Known issue: https://gitlab.kitware.com/cmake/cmake/issues/16291

File: cmake/toolchains/c.windows.common.cmake [hunk 2]
```diff
@@ -26,4 +26,7 @@ set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)
 set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
 set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
 
+list(APPEND CMAKE_C_IMPLICIT_INCLUDE_DIRECTORIES ${CMAKE_FIND_ROOT_PATH}/include)
```
Flamefire: CMAKE_SYSROOT seems to be the better solution: https://gitlab.kitware.com/cmake/cmake/issues/16291#note_440550

Jenkinsfile
Flamefire: What is the `-c` doing exactly? Would removing this fix it? --> `exec ...foo.sh` should work as the script is executable and has a bash shebang

cmake/toolchains/c.windows.common.cmake
Flow86: nope, that caused to set --sysroot on the compiler, removing /usr/lib/gcc/xxxx/... completly

Flow86: oh and it does not work:

```
d /workdir/build/external/libutil/nowide/src && /usr/bin/ccache /usr/lib/ccache/i686-w64-mingw32-g++ --sysroot=/usr/i686-w64-mingw32  -DBOOST_NOWIDE_NO_LIB -DBOOST_NOWIDE_USE_FILESYSTEM -DBOOST_PP_VARIADICS=1 -DNOMINMAX -DUNICODE -D_UNICODE -D__USE_W32_SOCKETS -isystem /workdir/external/libutil/nowide/include -isystem /usr/i686-w64-mingw32/include  -mfpmath=sse -msse -O2 -g -DNDEBUG -ggdb   -o CMakeFiles/nowide-static.dir/iostream.cpp.obj -c /workdir/external/libutil/nowide/src/iostream.cpp
In file included from /usr/lib/gcc/i686-w64-mingw32/7.3-win32/include/c++/ext/string_conversions.h:41:0,
                 from /usr/lib/gcc/i686-w64-mingw32/7.3-win32/include/c++/bits/basic_string.h:6349,
                 from /usr/lib/gcc/i686-w64-mingw32/7.3-win32/include/c++/string:52,
                 from /usr/lib/gcc/i686-w64-mingw32/7.3-win32/include/c++/stdexcept:39,
                 from /usr/lib/gcc/i686-w64-mingw32/7.3-win32/include/c++/array:39,
                 from /usr/lib/gcc/i686-w64-mingw32/7.3-win32/include/c++/tuple:39,
                 from /usr/lib/gcc/i686-w64-mingw32/7.3-win32/include/c++/bits/unique_ptr.h:37,
                 from /usr/lib/gcc/i686-w64-mingw32/7.3-win32/include/c++/memory:80,
                 from /usr/i686-w64-mingw32/include/boost/smart_ptr/scoped_ptr.hpp:22,
                 from /usr/i686-w64-mingw32/include/boost/scoped_ptr.hpp:13,
                 from /workdir/external/libutil/nowide/include/boost/nowide/iostream.hpp:12,
                 from /workdir/external/libutil/nowide/src/iostream.cpp:9:
/usr/lib/gcc/i686-w64-mingw32/7.3-win32/include/c++/cstdlib:75:15: fatal error: stdlib.h: No such file or directory
 #include_next <stdlib.h>

```

Flow86: the problem is manually adding isystem paths, include_next will jump to the wrong include path then if -isystem /usr.../include is in there (would happen on linux too if cmake would add /usr/include for example)

Jenkinsfile
Flow86: yep you're right, should work too

cmake/toolchains/c.windows.common.cmake
```diff
@@ -26,4 +26,7 @@ set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)
 set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
 set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
 
-set(Boost_COMPILER "-mgw44")
+list(APPEND CMAKE_C_IMPLICIT_INCLUDE_DIRECTORIES ${CMAKE_FIND_ROOT_PATH}/include)
+list(APPEND CMAKE_CXX_IMPLICIT_INCLUDE_DIRECTORIES ${CMAKE_FIND_ROOT_PATH}/include)
+
+set(Boost_USE_STATIC_RUNTIME ON)
```
Flow86: I've retried the boost build, I could remove the static runtime flag if we want to? I don't know if that makes any difference for our binaries at all?

Flow86: oh and I also could build dynamic boost libraries, but then we have to "ship" them

cmake/toolchains/c.windows.common.cmake
Flow86: /usr/include is host system. /usr/$platform/include is correct, but cmake does not have them in the implicit rules automatically if you set a CMAKE_FIND_ROOT_PATH

Flamefire: Sorry I meant in the toolchain.
Why toolchain/include and not toolchain/usr/include?

Flow86:
yea the findsdl2 seems still not to work very well

libs/s25main/mapGenerator/RandomConfig.cpp
```diff
@@ -22,11 +22,16 @@
 #include "gameData/WorldDescription.h"
 #include <boost/bind.hpp>
 #include <boost/random/uniform_real_distribution.hpp>
+
+#ifdef _WIN32
```
Flamefire: What? Really? AFAIK pragmas were invented to implement compiler specific stuff and has to be ignored if not understood O.o

Flow86: yes, they are ignored, with a warning, but with Werror it result in a failure ;-)

Flamefire: We should disable that warning then...

Jenkinsfile
Flow86: ah I've found a reason to still include it. you get a different parameter unescaping if you do not call it "inside" the container, but let the docker assign the parameters.

(i.e if you manually try to add ">&2 | less" it doesnt work if you do not run it via "bash -c"

Flamefire: ok then

Flamefire:
new_docker2 builds now. Will check and fix the rpath issue now

cmake/toolchains/c.apple.common.cmake
```diff
@@ -35,6 +36,36 @@ ENDIF()
 SET(CMAKE_OSX_SYSROOT ${CMAKE_OSX_SYSROOT} CACHE PATH "Path to OSX SDK")
 MESSAGE(STATUS "Using OSX SDK at ${CMAKE_OSX_SYSROOT}")
 
+set(CMAKE_OSX_DEPLOYMENT_TARGET "$ENV{MACOSX_DEPLOYMENT_TARGET}" CACHE STRING "Minimum OS X version to target for deployment")
+if(NOT CMAKE_OSX_DEPLOYMENT_TARGET)
+  set(CMAKE_OSX_DEPLOYMENT_TARGET "10.11" CACHE STRING "" FORCE)
```
Flow86: shouldnt 10.11 match the used SDK?

PR #1012, opened 2019-01-19
Fix combo box crash

Flamefire:
Caused by wrong ID in scroll-down event handler of combobox


Flow86:
why did you remove all consts from non writeable parameters?

Flamefire:
Because putting const on copies of a value does not make sense. For this reason they are not part of the method signature. This means `void foo(int)` and `void foo(const int)` are exactly the same. This leads to problems in the mocking framework (warnings in e.g. MSVC)

So I removed the const from the declarations. The definitions (implementation) could still have them but there is not much wort for that.

libs/s25main/controls/ctrlComboBox.cpp
```diff
@@ -152,7 +152,7 @@ bool ctrlComboBox::Msg_WheelUp(const MouseCoords& mc)
 
     if(IsPointInRect(mc.GetPos(), GetDrawRect()))
     {
-        // Scrolled without list opened
+        // Don't scroll do far down
```
Flow86: do -> too?

libs/s25main/controls/ctrlComboBox.cpp
```diff
@@ -176,9 +176,8 @@ bool ctrlComboBox::Msg_WheelDown(const MouseCoords& mc)
 
     if(IsPointInRect(mc.GetPos(), GetDrawRect()))
     {
-        // Scrolled without list opened
-        if(list->GetSelection() + 1 < list->GetNumLines())
-            Msg_ListSelectItem(GetID(), list->GetSelection() + 1);
+        // Will be ignored by the list if to high
+        list->SetSelection(list->GetSelection() + 1);
```
Flow86: perhaps we should provide an "IncSelection" ? (and DecSelection?)

Flamefire: I don't think it is required. It is only done like that in the combobox

Flamefire:
@Flow86 merge this now?

PR #1016, opened 2019-01-22
Fix editor

Flamefire:
Pulls in the fix for destroyed textures and wrong BQ in the editor.

Also some tests created while refactoring Point/Rect (used by the editor)

### DIFF
--- a/libs/common/include/PointOutput.h
+++ b/libs/common/include/PointOutput.h
@@ -24,7 +24,7 @@
 template<typename T>
 std::ostream& operator<<(std::ostream& out, const Point<T>& point)
 {
-    return out << "(" << point.x << ", " << point.y << ")";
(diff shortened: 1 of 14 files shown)


libs/common/include/PointOutput.h
```diff
@@ -24,7 +24,7 @@
 template<typename T>
 std::ostream& operator<<(std::ostream& out, const Point<T>& point)
 {
-    return out << "(" << point.x << ", " << point.y << ")";
+    return out << "(" << point.x * 1 << ", " << point.y * 1 << ")"; // *1 to convert chars to int
```
Flow86: why not using int(...)?

Flamefire: because T might be float.

Flow86: hmm okay you're right

PR #1018, opened 2019-01-24
Map send timeout

### LINKED ISSUE #1017: Saves make the game completely freeze out
Howdy!
I don't know what to do or what is the problem but every time I save a game, and wanted to reload later, the game frozen out at the "please wait" and do nothing for hours. 
The DOS Prompt say:
Connection to 127.0.0.1:3665 
Successfully connected to local host:3665
Server: Reserved slots 0 freed due the timeout
And after this the game dies. Not responding...

After a few hours of waiting I get bored and shut it down manually. It happens every saves I made, and it doesn't matter if I made the map - 512x512 - with the in game random map generator or the online one. 
The version I used when the saved games were made was: 2018.11.29 but I tried with the new version - 2019.01.22 - but it still don't want to load any one of those saved games. 
Okay I know maybe it doesn't compatibly with previous versions but do I have to really restart every time  a new version of the game came out? 
I can not attach the save file because it over 63Mb. - Is that size normal for a 512 size map? I uploaded it here:  [https://mega.nz/#!xARnDShJ!jTLXUfnw0gxovBGtyI9N1LkTAdsvVyLSrjwsn3RW_1E](url)
If the log file needed I can upload that too.

And whoever suggests those things like delete and reinstall, or make a clean reinstall, or other things like that, I done that... all of them. And nothing worked! 

And if I will be ever in the mood to start a new game again - with the newest version - I will give it a try, but at the moment I'm already bored to restart over and over again because of this.
I hope only the save files get corrupted somehow, and that is the answer for why I can not load it.

Flamefire:
Avoid kicking players when sending large maps

Increase timeout till player gets disconnected by estimated map send time
Use state pattern via boost::variant to improve implementation

Fixes #1017 (map loads successfully now)


libs/common/include/variant.h
```diff
@@ -0,0 +1,66 @@
...
+bool holds_alternative(const boost::variant<Types...>& v) noexcept
+{
+    return v.which() == detail::indexOf<T, Types...>::value;
+}
+
+template<class... Fs>
+auto composeVisitor(Fs&&... fs)
+{
+    return detail::lambda_visitor<std::decay_t<Fs>...>(std::forward<Fs>(fs)...);
+};
```
Flow86: /source/libs/common/include/variant.h:65:2: error: extra ';' [-Werror=pedantic]

Flamefire: thx

Flamefire:
Hmm... I need another approval from you @Flow86 after the rebase. Do we really want to keep it that way? (as in: Do you have enough time? ;) )

PR #1019, opened 2019-01-26
WIP: Create bundle in CMake

Flamefire:
The idea is to use the CMake provided utilities to determine dependencies of all executables and copy them into the bundle folder. For OSX this is (kinda) easy, but for other platforms and especially when cross-compiling it is not.

One problem is, that CMake seems to assume, that all binaries are in 1 folder. But on e.g. Windows we have some in the root folder and some in the RTTR folder. CMake correctly copies the dlls into the root folder but then they are missing for the binaries in the RTTR folder. I guess rpath can be used to fix this but I don't know if a) this is even supported on windows and b) how to tell CMake to do this.

Additionally this breaks on linux where the main binary is in `/bin` but others are in `/libexec...`. CMake than assumes the bundle folder is `/bin` and ignores the other executables.

Maybe the best solution is to not use `fixup_bundle` but just the functions to get executables and dependencies and do the copying etc. ourselves? Maybe we can at least use `fixup_bundle` for OSX and the custom solution for Win+Lin.

### DIFF
--- a/tools/release/CMakeLists.txt
+++ b/tools/release/CMakeLists.txt
@@ -1,60 +1,14 @@
+set(bundleScript ${CMAKE_CURRENT_BINARY_DIR}/createBundle.cmale)
+configure_file(createBundle.cmake ${bundleScript} @ONLY)
+install(SCRIPT ${bundleScript})
+
 # Start script
 if(WIN32)
     install(PROGRAMS "bin/rttr.bat" DESTINATION "${RTTR_BINDIR}")
-    if(PLATFORM_ARCH)
-        include(ConfigureExecutable)
-        configure_executable("prepareRelease.sh.cmake" . prepareRelease.sh)
-
-        install(CODE "
-            set(ENV{CMAKE_INSTALL_PREFIX} \${CMAKE_INSTALL_PREFIX})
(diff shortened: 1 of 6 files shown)


Flow86:
windows does not support RPATH

tools/release/CMakeLists.txt
```diff
@@ -1,60 +1,14 @@
+set(bundleScript ${CMAKE_CURRENT_BINARY_DIR}/createBundle.cmale)
```
Flow86: cmale?

Flamefire: Cfemale better? :D
Damn typos...

Flow86:
``` 
CMake Error at /source/cmake/Modules/BundleUtilities.cmake:1131 (message):
error: verify_app failed
```

stefson:
I'm going to give this a testride. 

By the way, for linux the correct folder for userspace binaries is /usr/bin/, while /bin/ is reserved for binaries which are essential for the system to boot, even though it allows non-root users to execute them.

Flamefire:
It doesn't work and we probably need our own implementation or a serious rework, Gonna take a while

PR #1030, opened 2019-02-03
Don't use broken bfs::canonical on windows and enforce it

Flamefire:
Uses the weaker lexically_normal and adds script that check for usages of `::canonical(`


Flamefire:
@Flow86 ready to merge

PR #1033, opened 2019-02-05
Add shell.nix

BlueGone:
This PR give an easy way to get a working environnement on this project without polluting his system-side installation with uneeded packages (sdl, miniupnpc, ...).
Insteed of installing all the prerequesites, the only requirement is having `nix` installed. Then, in s25client, just open `nix-shell`. This works on both Linux and Darwin/MacOSX.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,10 +58,14 @@ Most of them can be installed with the package manager.
 - miniupnpc
 All of them can be installed via homebrew
 
+### Prerequesites with Nix
+Nix users can open a nix-shell to get a development environment with all packages ready.
+
 ### Steps:
 ```
 git clone --recursive https://github.com/Return-To-The-Roots/s25client s25client
 cd s25client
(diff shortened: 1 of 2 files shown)


Flamefire:
This is to set up a **development** environment, isn't it? (I'm not familiar with nix)

If so I suggest to add SDL2 and its mixer to the list as we will switch to SDL2 as the default soon. 

Besides that: I assume you tested this? Could you add step-by-step instructions to the Readme.MD so new users know about this? Looks like it will be very useful

BlueGone:
Yes, Nix gives a development environment. When opening a nix-shell in the project directory, a custom `PATH` env variable will be set, with all dependencies in it.

I just added SDL2 in the list of dependencies, and updated README.md to explain how to build for nix users.

The only test I made is building the whole project on a system with nothing installed (no cmake, SDL, not even curl), but nix.

Issue #1040, opened 2019-02-10
readme and keyboardlayout default to english


stefson:
there is a german version for both of them, and if system language is set to german it should be choosen over the english?

stefson:
I know you haven't merged this into the main tree yet, but still I want to leave some feedback: It works now if the language setting is set to `System language`, which means that with my settings from #1032 it gives me the german readme. But if language is set explicitly to German, it still defaults to English.

Flamefire:
Thanks. took care of that, so #1042 should now fix this.

stefson:
I can confirm it is fixed now. 

By the way, I noticed that these symbols are misrendered as a questionmark: `[` `]` `|` `~`

Issue #1039, opened 2019-02-10
Scrolling with the mouse wheel through the readme is going in the wrong way


stefson:
Scrolling with the mousewheel works sort of, but its going up when the user scrolls down with the mouswheel, und down when the user scrolls up.

stefson:
Same problem with the `Tastaturbelegung` , which is accessable from within the running game.

Flamefire:
Only affects SDL2, fix incoming

Issue #1008, opened 2019-01-13
Auto version update broken


Flamefire:
As we moved the version determination into a subfolder we had to put it into CMake cache. This breaks the auto-update once the date or revision changes.

What shall we do?

Currently the version is required as a (regular or cache) variable in rttrConfig when configuring the header, in the main CMakeLists for CPack and in the (OSX) bundle.

I see 2 options:

1. Set the version (possibly via a CMake function) as a regular variable in the main CMakelists and fatal_error in using code if it isn't set
2. Set the version as a target property of rttrConfig from which it can be queried.

I'd say 1. Any other ideas? @Flow86

Flow86:
thats why I always build with `-DRTTR_VERSION=OFF -DRTTR_REVISION=OFF` in jenkins

Flamefire:
Correct. This forces to get a recent version even if it was set in the cache previously. I'd suggest however to set a common value for `RTTR_VERSION` which is used by all builds in Jenkins to avoid a long running build get 2 versions.

However we need to solve this for local builds too. Shall we go with 1?

Flow86:
> Correct. This forces to get a recent version even if it was set in the cache previously. I'd suggest however to set a common value for `RTTR_VERSION` which is used by all builds in Jenkins to avoid a long running build get 2 versions.
> 
> However we need to solve this for local builds too. Shall we go with 1?

I dont understand that? how should a long running build get 2 versions? Perhaps we should simply disable caching these values?

Flamefire:
If you set `RTTR_VERSION=OFF` cmake will use the current system date for it. If the builds start around midnight this might lead to 2 versions.

> Perhaps we should simply disable caching these values?

The variables are set in the subdir so they are not visible to others unless:
- we set them as cache variables
- we set them in PARENT_SCOPE (might break when multiple parents are there)
- we set them in the top-level CMakeLists
- we set them as a target property

Flow86:
perhaps we should simply generate it manually in the ci build script so the midnight problem can never happen when building via ci?

or we simply set and use the standard cmake project variables CMAKE_VERSION_MAJOR/MINOR/PATCH for it?

Flow86:
(oh its PROJECT_VERSION_....)

Flow86:
oh and I think its not a problem at all, since the ci will only run between 1 and 5 am

Flamefire:
[...]

PR #1042, opened 2019-02-12
Fix window related bugs

Flamefire:
- Improve error messages for SDL on windows (https://www.siedler25.org/index.php?com=forum&mod=forum&action=thread&id=1583&start=-1#last)
- Fix auto version update #1008
- Refactor some window and window manager stuff
- Fix SDL2 scroll direction #1039 
- Fix localization of text files and improve its performance (lag on open) #1040


extras/videoDrivers/SDL/VideoSDL.cpp
```diff
@@ -45,7 +47,29 @@ struct DeleterDeleteRC
...
+{
+    std::unique_ptr<HGLRC, DeleterDeleteRC> context{wglCreateContext(deviceContext)};
+    if(!context)
+        throw makeLastSystemError("Failed to create OGL context");
+    return context;
+}
+void ShareOGLResources(HGLRC srcContext, HGLRC dstContext)
+{
+    if(!wglShareLists(srcContext, dstContext))
+        throw makeLastSystemError("Failed to create OGL context");
```
Flow86: copy and paste. I think here there should be a "Failed to share OGL lists".

Flamefire: Yes thx. We do have a problem though: We need Boost build with fPIC if we want to use it in the drivers. For the purpose of `makeLastSystemError` I could implement that myself, but if we need more of boost we need the fPIC or shared boost libs.

Flamefire: Fixing this by using std::error_code

Flamefire:
@Flow86 Ready to merge

extras/videoDrivers/SDL/VideoSDL.cpp
```diff
@@ -20,10 +20,12 @@
 #include "driver/VideoDriverLoaderInterface.h"
 #include "driver/VideoInterface.h"
 #include "helpers/containerUtils.h"
-#include <memory>
+#include "makeException.h"
 #include <boost/nowide/iostream.hpp>
+#include <boost/system/windows_error.hpp>
```
Flow86: why is the windows_error.hpp included here? I can't see anything related with it?

libs/common/src/makeException.cpp
```diff
@@ -0,0 +1,35 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with Return To The Roots. If not, see <http://www.gnu.org/licenses/>.
+
+#include "makeException.h"
+#include <boost/system/config.hpp>
+#ifdef BOOST_WINDOWS_API
+#include <windows.h>
+#elif defined(BOOST_POSIX_API)
+#include <errno.h>
```
Flow86: shouldn't it be 'cerrno'? and why not removing boost here completly and use `_WIN32` instead?

libs/rttrConfig/CMakeLists.txt
```diff
@@ -17,20 +17,9 @@ SET_PROPERTY(TARGET rttrConfig PROPERTY POSITION_INDEPENDENT_CODE ON)
...
-        message(FATAL_ERROR "Could not get git revision. The source has to be in a git repo or you have to define RTTR_REVISION")
-    endif()
-    set(RTTR_REVISION ${RTTR_REVISION} CACHE STRING "Git revision" FORCE)
-endif()
-if(NOT RTTR_VERSION)
-    STRING(TIMESTAMP RTTR_VERSION "%Y%m%d")
-    set(RTTR_VERSION ${RTTR_VERSION} CACHE STRING "Version (yyyymmdd)" FORCE)
+# This should happen in the top-level CMakeLists
+if(NOT RTTR_REVISION OR NOT RTTR_VERSION)
+    message(FATAL_ERROR "Internal error: RTTR_REVISION or RTTR_VERSION not set")
```
Flow86: we should add a "in the top level CMakeList.txt" to the message

Flamefire: The comment above already says that. Should be enough

extras/videoDrivers/SDL/VideoSDL.cpp
Flamefire: Removed

libs/common/src/makeException.cpp
Flamefire: BEcause this should also be used for cygwin. The boost define has both although we could do it manually.

Flow86: ah yes I don't know if cygwin defines _WIN32 and since we're using boost anyway ...

PR #1049, opened 2019-03-08
Fix Undefined Behaviour / RTTR_Assert triggering in nofShipWright

### LINKED ISSUE #1048: nofShipWright::nofShipWright SIGTRAP on sea map
I compiled s25client myself from git, commit ID 2f33ff0075597ff1310e0ba3954262fc1476a3d8 (should be current master). This is on debian buster/testing.

A few minutes after loading the [attached savegame](https://github.com/Return-To-The-Roots/s25client/files/2942214/Wasserwege.Auto.save.sav.gz), the game enters a debug trap. I attached gdb and obtained the following backtrace (no debug symbols unfortunately):

```
(gdb) bt
#0  0x00005555558c0b62 in nofShipWright::nofShipWright(Point<unsigned short>, unsigned char, nobUsual*) ()
#1  0x000055555587498d in JobFactory::CreateJob(Job, Point<unsigned short>, unsigned char, noRoadNode*) ()
#2  0x00005555557ed751 in nobBaseWarehouse::OrderJob(Job, noRoadNode*, bool) ()
#3  0x0000555555767543 in GamePlayer::FindWarehouseForJob(Job, noRoadNode*) ()
#4  0x0000555555766f66 in GamePlayer::AddJobWanted(Job, noRoadNode*) ()
#5  0x0000555555813953 in nobUsual::nobUsual(BuildingType, Point<unsigned short>, unsigned char, Nation) ()
#6  0x00005555558118f2 in nobShipYard::nobShipYard(Point<unsigned short>, unsigned char, Nation) ()
#7  0x000055555586e9fd in BuildingFactory::CreateBuilding(GameWorldBase&, BuildingType, Point<unsigned short>, unsigned c
#8  0x000055555588cab2 in nofBuilder::HandleDerivedEvent(unsigned int) ()
#9  0x0000555555877816 in noFigure::HandleEvent(unsigned int) ()
#10 0x000055555574c14c in EventManager::ExecuteEvents(std::_Rb_tree_iterator<std::pair<unsigned int const, std::__cxx11::
#11 0x000055555574c014 in EventManager::ExecuteCurrentEvents() ()
#12 0x000055555574bd7b in EventManager::ExecuteNextGF() ()
#13 0x0000555555755ed4 in Game::RunGF() ()
#14 0x000055555569a38e in GameClient::NextGF(bool) ()
#15 0x0000555555699ca4 in GameClient::ExecuteGameFrame() ()
#16 0x0000555555694294 in GameClient::Run() ()
#17 0x00005555555e8daf in GameManager::Run() ()
#18 0x00005555555d0ceb in RunProgram(boost::program_options::variables_map&) ()
#19 0x00005555555d117f in main ()
```

If there’s anything useful I could add, let me know.

horazont:
The assertion is unsafe at this point, even though it might
appear logical. The reason it is unsafe is that it is called from
nobShipYard::nobShipYard (the consturctor) indirectly through the
constructor nobUsual::nobUsual (constructor of the base class).

The dynamic_cast<nobShipYard*> then becomes undefined behaviour,
because the constructor of nobShipYard has not started to run on
the workplace object; it is still running the nobUsual
constructor.

See also https://stackoverflow.com/a/6299463/1248008

Fixes #1048.

### DIFF
--- a/libs/s25main/buildings/nobUsual.cpp
+++ b/libs/s25main/buildings/nobUsual.cpp
@@ -47,13 +47,10 @@ nobUsual::nobUsual(BuildingType type, MapPoint pos, unsigned char player, Nation
 
     ordered_wares.resize(BLD_WORK_DESC[bldType_].waresNeeded.getNum());
 
-    // Arbeiter bestellen
-    GamePlayer& owner = gwg->GetPlayer(player);
-    owner.AddJobWanted(BLD_WORK_DESC[bldType_].job, this);
-
     // Tür aufmachen,bis Gebäude besetzt ist
     OpenDoor();
 
+    GamePlayer& owner = gwg->GetPlayer(player);
     // New building gets half the average productivity from all buildings of the same type
     productivity = owner.GetBuildingRegister().CalcAverageProductivity(type) / 2u;
     // Set last productivities to current to avoid resetting it on first recalculation event
--- a/libs/s25main/GamePlayer.cpp
+++ b/libs/s25main/GamePlayer.cpp
@@ -49,6 +49,7 @@
 #include "gameTypes/JobTypes.h"
 #include "gameTypes/PactTypes.h"
 #include "gameTypes/VisualSettings.h"
+#include "gameData/BuildingConsts.h"
 #include "gameData/BuildingProperties.h"
 #include "gameData/MilitaryConsts.h"
 #include "gameData/SettingTypeConv.h"
@@ -435,6 +436,14 @@ void GamePlayer::AddBuilding(noBuilding* bld, BuildingType bldType)
     RTTR_Assert(bld->GetPlayer() == GetPlayerId());
     buildings.Add(bld, bldType);
     ChangeStatisticValue(STAT_BUILDINGS, 1);
+
+    // Order a worker if needed
+    const auto& description = BLD_WORK_DESC[bldType];
+    if(description.job != JOB_NOTHING && description.job != JOB_PRIVATE)
+    {
+        AddJobWanted(description.job, bld);
+    }
+
     if(bldType == BLD_HARBORBUILDING)
     {
         // Schiff durchgehen und denen Bescheid sagen


Flamefire:
You are correct. I'm just thinking *what*  the bug is. Technically the assertion is not wrong as it is an actual error to pass in anything but a shipyard. The rest of the class relies on that. IMO we should rather move the job ordering out of the constructor.

Idea:
- Buildings are added to the player anyway
- On adding of a nobUsual an order for that is placed

horazont:
> Technically the assertion is not wrong as it is an actual error to pass in anything but a shipyard.

Ah yes, I was thinking that. Unfortunately, I don’t have the expertise in the codebase to follow through with the other suggestion.

> IMO we should rather move the job ordering out of the constructor.

I think this sounds very sane, also from a theoretical/architectural standpoint, but I’m quite the layman in this project ;-).

Flamefire:
Fresh and sane ideas are always welcome. Often "outsiders"/laymans can provide valuable input for a project as they are not bound by "we have always done this".

Anyway thanks a lot for finding the issue. If you don't want to try taking the above, I'll take it over

horazont:
Go ahead, since I can’t make any promises :). I might come back to this issue if I have some time to spare to further develop it if there hasn’t been any progress.

horazont:
Ok, nevermind. I took another stab at the issue, it was much less complicated than I anticipated—but of course, I might’ve gone the entire wrong way to do this :).

Thanks for taking a look.

libs/s25main/factories/BuildingFactory.cpp
```diff
@@ -42,6 +43,17 @@ noBuilding* BuildingFactory::CreateBuilding(GameWorldBase& gwg, const BuildingTy
         case BLD_SHIPYARD: bld = new nobShipYard(pt, player, nation); break;
         default: bld = new nobUsual(type, pt, player, nation); break;
     }
+    GamePlayer &player_obj = gwg.GetPlayer(player);
```
Flamefire: Could you move this either below the `AddBuilding` (and reuse player_obj, btw: `lowerCamelCase` please) or into `GamePlayer::AddBuilding`? I think the latter is the better place ( I don't like doing to much in the factory, actually it should just create the building and nothing else)

libs/s25main/buildings/nobUsual.cpp
```diff
@@ -49,7 +49,6 @@ nobUsual::nobUsual(BuildingType type, MapPoint pos, unsigned char player, Nation
 
     // Arbeiter bestellen
```
Flamefire: Remove comment and move owner definition down before first use

libs/s25main/factories/BuildingFactory.cpp
```diff
@@ -42,6 +43,17 @@ noBuilding* BuildingFactory::CreateBuilding(GameWorldBase& gwg, const BuildingTy
...
+    const auto &description = BLD_WORK_DESC[type];
+    // We're filtering out JOB_PRIVATE because military buildings do their own
+    // worker stuff.
+    // We also can't do this in the nobUsual constructor, because at the point
+    // the respective nof* object is created through JobFactory, the building
+    // would not be fully initialized yet (which may invoke undefined
+    // behaviour).
+    if (description.job != JOB_NOTHING && description.job != JOB_PRIVATE) {
+        player_obj.AddJobWanted(BLD_WORK_DESC[type].job, bld);
```
Flamefire: use `description.job` please ;)

horazont:
I think I addressed all comments; I made a quick test by building up to a Shipyard and it seems good! :)

libs/s25main/factories/BuildingFactory.cpp
```diff
@@ -43,8 +44,19 @@ noBuilding* BuildingFactory::CreateBuilding(GameWorldBase& gwg, const BuildingTy
         default: bld = new nobUsual(type, pt, player, nation); break;
     }
     gwg.SetNO(pt, bld);
+    GamePlayer &playerObject = gwg.GetPlayer(player);
     // Don't do this in ctor as building might not be fully initialized yet
-    gwg.GetPlayer(player).AddBuilding(bld, type);
+    playerObject.AddBuilding(bld, type);
+    const auto &description = BLD_WORK_DESC[type];
+    // We're filtering out JOB_PRIVATE because military buildings do their own
```
Flamefire: Put on 1 line

libs/s25main/factories/BuildingFactory.cpp
```diff
@@ -43,8 +44,19 @@ noBuilding* BuildingFactory::CreateBuilding(GameWorldBase& gwg, const BuildingTy
...
+    // worker stuff.
+    // We also can't do this in the nobUsual constructor, because at the point
```
Flamefire: Remove as already covered by line 46/48

Flamefire:
Good, just a minor nitpick.

When you are done, please rebase on current master (pull master, rebase your branch, force push your branch) so github does not show the "out-of-date" message

libs/s25main/factories/BuildingFactory.cpp
horazont: I’d like to replace the comment in line 46/48 then, because I think my comment is more useful. What do you think?

libs/s25main/factories/BuildingFactory.cpp
horazont: Is there documentation on the maximum line length acceptable in comments in this project?

horazont:
There we go, hopefully!

horazont:
> Idea:
> 
>     * Buildings are added to the player anyway
> 
>     * On adding of a nobUsual an order for that is placed

Hm, do you prefer the way of handling the Job in AddBuilding? I think doing it in AddBuilding (instead of what I’m doing here now) would be cleaner. If you agree, I’ll change it to do the job ordering in AddBuilding (which already does things based on the type of building).

Flamefire:
> Hm, do you prefer the way of handling the Job in AddBuilding? I think doing it in AddBuilding (instead of what I’m doing here now) would be cleaner. If you agree, I’ll change it to do the job ordering in AddBuilding (which already does things based on the type of building).

Yes, go ahead :)

horazont:
There we go, finally, hopefully :)

libs/s25main/factories/BuildingFactory.cpp
Flamefire: There is a .clang-format file and a `make clangFormat` target so don't worry about it, just don't manually break it and run the target. Length should be 120 or so.

horazont:
Ran clangFormat, it made some whitespace changes and re-ordered the includes.

Issue #1050, opened 2019-03-09
Cannot run game on Windows (OpenGL)


FROGGS:
Hi, I only see a blank (white) window ang get the following output:
```
Return To The Roots v20190223-2f33ff0075597ff1310e0ba3954262fc1476a3d8
Compiled with Gnu GCC C/C++ (7.3.0) for Microsoft Windows 64 Bit

Starting in C:\s25rttr

WARNING: Your application version has changed - please recheck your settings!

Suche nach Treibern in C:\s25rttr\driver/video
3 video drivers found!
Loaded video driver "(SDL) OpenGL via SDL-Library"
OpenGL 2.0 supported

FATAL ERROR: OpenGL  2.1 is not supported. Try updating your GPU drivers or hardware!

FATAL ERROR: Failed to initialize the OpenGL context!

ERROR: Failed to start the game
Press ENTER to close this window . . .
```

As you can see I'm running the latest version.
This might be related to #1037, but I'm just guessing.
I cannot tell when I played the game for the last time, so I cannot guess at what time it might broke.

Thanks for any pointers.

Flow86:
Hi,

> Try updating your GPU drivers or hardware!

you will need a OpenGL 2.1 compliant video driver. Try updating your driver via the manufactures website. ( i.e www.nvidia.com, www.amd.com, www.intel.com )

The standard windows graphic driver is most of the time extremely outdated and does even not support OpenGL 2.1 (for your information, OpenGL 4.x is the current one ...)

FROGGS:
Okay, it looks like I have to replace my son's laptop then. Since it seems to only support OpenGL 1.5 (Intel 965GM Express Graphics).

This will take a while, but thanks for your help.

Flamefire:
As per the output it seems to at least support 2.0.

At the moment we don't use any OpenGL 2.1 features so we could drop back to 2.0. Will prepare a commit for that, but note that we may switch to 2.1 later on. Just watch the error output as above.

For devs:

OpenGL 2.1 introduced Pixel buffer objects and Non-square matrices in GLSL. When those are needed, update the required version. (OpenGL 2.1 is 13 years old)

Flamefire:
@FROGGS It should work on your sons laptop starting with the next nightly again. Happy playing! :)

FROGGS:
Hi, thank you for the patch!
I'll try and report back here as soon as the nightly covers this commit.

PR #1051, opened 2019-03-11
Allow OpenGL 2.0

Flamefire:
No OpenGL 2.1 features (Pixel buffer objects and Non-square matrices in
GLSL) are used ATM so reduce requirements in favor of old Intel
intgrated graphics. See #1050


Flow86:
> /source/extras/videoDrivers/SDL2/VideoSDL2.cpp:24:10: fatal error: openglCfg.hpp: No such file or directory 
> #include "openglCfg.hpp"

PR #1059, opened 2019-04-05
Load music from memory instead of from file

### LINKED ISSUE #893: Minilag when a new song is played

Leuchtefix75:
Fixes #893

Hi Alex,

Ich versuch grad mal deine Gabel zu bauen
und stoß auf was, wo ich nicht weiß mit umzugehen:

CMake Error at /usr/share/cmake-3.7/Modules/FindPackageHandleStandardArgs.cmake:138 (message):
  Could NOT find SDL (missing: SDL_LIBRARY) (found suitable version "2.0.5",
  minimum required is "1.2")
Call Stack (most recent call first):
  /usr/share/cmake-3.7/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
  /usr/share/cmake-3.7/Modules/FindSDL.cmake:188 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  src/test/mockupDrivers/CMakeLists.txt:1 (FIND_PACKAGE)


Flamefire:
Please open an issue for things like this not a pull request. And these changes are already in the upstream master, so no need to build my fork (its master is outdated anyway)

The error message states that SDL could not be found. This is SDL 1.2 in this case. Just install its development libs and reconfigure. I opened https://github.com/Return-To-The-Roots/s25client/pull/1060 to allow also SDL2 for tests which would also solve your issue. So you can try that branch too.

PR #1063, opened 2019-05-17
Fix UPnP

Flamefire:
The UPnP class closed the port on destructor which was not though of in the copy/move of `Socket` which lead to the port being closed on first copy/move (likely even on construction).

- Reworked the class to an auto-close and static functions taking a port
- Use the port only in `Socket` and close it when socket is closed (ref counted!)
- Don't use locale dependent int-to-string conversion for port (5314 could become "5.314")
- Add helper for IP-String conversion
- Add Tests
- Fix some shaky tests with bitmaps (random based tests so failures were sporadic [as they should])

### DIFF
--- a/tests/s25Main/UI/testSmartBitmap.cpp
+++ b/tests/s25Main/UI/testSmartBitmap.cpp
@@ -216,22 +216,24 @@ BOOST_AUTO_TEST_CASE(MultiPlayerBitmap)
     glSmartBitmap smartBmp;
     for(int i = 0; i < 3; i++)
     {
-        bmps.emplace_back(createRandPlayerBmp(70));
+        bmps.emplace_back(createRandPlayerBmp(50));
         smartBmp.add(bmps.back().get());
     }
+
+    Extent size = smartBmp.GetSize();
+    BOOST_TEST(smartBmp.getRequiredTexSize() == Extent(size.x * 2, size.y));
+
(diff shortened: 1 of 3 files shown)


tests/s25Main/UI/testSmartBitmap.cpp
```diff
@@ -216,22 +216,24 @@ BOOST_AUTO_TEST_CASE(MultiPlayerBitmap)
     glSmartBitmap smartBmp;
     for(int i = 0; i < 3; i++)
     {
-        bmps.emplace_back(createRandPlayerBmp(70));
+        bmps.emplace_back(createRandPlayerBmp(50));
         smartBmp.add(bmps.back().get());
     }
+
+    Extent size = smartBmp.GetSize();
```
Flow86: diese änderungen haben jetzt aber nicht wirklich was mit Upnp zu tun?

Flamefire: Hatte hauptsächlich in den Submodules (libutil) was gemacht, aber dann sind die CI tests gefailt. Da ich dort einige Random Tests (werte sind nicht fest sondern zufällig, vgl "fuzzy testing") eingebaut habe war das fast schon zu erwarten. Kann aber nen extra MR nur mit den Testfixes machen, wenn gewünscht.

PR #1067, opened 2019-06-22
Fixed crash during selection of screen resolution.

vlnovikov:
With some locales std::stringstream adds thousand delimiter to
resolution text which makes utf8 resolution string invalid.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,6 @@ addons:
       - sdl_mixer
       - gettext
       - miniupnpc
-    update: true
 
 cache:
   - ccache
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,7 @@ addons:
       - sdl_mixer
       - gettext
       - miniupnpc
+    update: true
 
 cache:
   - ccache
--- a/libs/common/src/RTTR_Assert.cpp
+++ b/libs/common/src/RTTR_Assert.cpp
@@ -19,9 +19,9 @@
 #include "RTTR_Assert.h"
 #include "RTTR_AssertError.h"
 #include "libutil/Log.h"
+#include "libutil/StringConversion.h"
 #include "libutil/System.h"
 #include <iostream>
-#include <sstream>
 #ifdef _WIN32
 #include <windows.h>
 #endif
@@ -57,7 +57,7 @@ void RTTR_AssertFailure(const char* condition, const char* file, const int line,
         if(pos != std::string::npos && filePath.size() > pos && filePath.substr(0, pos) == thisFilePath.substr(0, pos))
             filePath = filePath.substr(pos + 1);
 
-        std::stringstream sMsg;
+        s25util::ClassicImbuedStream<std::stringstream> sMsg;
         sMsg << "Assertion failure";
         if(function)
             sMsg << " in \"" << function << "\"";
--- a/libs/common/src/RTTR_Assert.cpp
+++ b/libs/common/src/RTTR_Assert.cpp
@@ -19,9 +19,9 @@
 #include "RTTR_Assert.h"
 #include "RTTR_AssertError.h"
 #include "libutil/Log.h"
-#include "libutil/StringConversion.h"
 #include "libutil/System.h"
 #include <iostream>
+#include <sstream>
 #ifdef _WIN32
 #include <windows.h>
 #endif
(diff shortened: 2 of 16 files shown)


Flamefire:
Can you rebase on current master please?

On the issue: I'm curious WHY this fails. If it fails here it can fail in other places too. Which locale are you using? The thousand separator should be a valid UTF-8 string. Maybe we need to enforce/prefer an UTF-8 locale?

vlnovikov:
> On the issue: I'm curious WHY this fails. If it fails here it can fail in other places too. Which locale are you using? The thousand separator should be a valid UTF-8 string. Maybe we need to enforce/prefer an UTF-8 locale?

Originally application stopped on assert at glArchivItem_Font.cpp:210 RTTR_Assert(isValidUTF8(text));
where text is '3�840x2�160 (16:9)' at (?) symbol is 0xE2.

locale output:
LANG=en_US.UTF-8
LANGUAGE=
LC_CTYPE="en_US.UTF-8"
LC_NUMERIC=uk_UA.UTF-8
LC_TIME=uk_UA.UTF-8
LC_COLLATE="en_US.UTF-8"
LC_MONETARY=uk_UA.UTF-8
LC_MESSAGES="en_US.UTF-8"
LC_PAPER=uk_UA.UTF-8
LC_NAME=uk_UA.UTF-8
LC_ADDRESS=uk_UA.UTF-8
LC_TELEPHONE=uk_UA.UTF-8
LC_MEASUREMENT=uk_UA.UTF-8
LC_IDENTIFICATION=uk_UA.UTF-8
LC_ALL=

I haven't observed this issue anywhere else and with different conversion approaches (e.g. std::to_string), it seems the issue is valid only for std::stringstream. Also, 0xE2 symbol is garbage in the locale I use.

vlnovikov:
> Can you rebase on current master please?

Done, but it doesn't look like I did it properly:(

Flamefire:
> Done, but it doesn't look like I did it properly:(

Just seen you did not create a new branch for the change.

Make sure you have `origin` remote set to the RttR repo. You can check with `git remote -v`. If `origin` is your fork then issue `git remote add rttr git@github.com:Return-To-The-Roots/s25client.git` and replace `origin` by `rttr` below.

Quick summary:
- Checkout your local branch: `git checkout master`
- Fetch changes: `git fetch origin` (or `git fetch rttr`)
- Rebase: `git rebase -i origin/master`
- Make sure only your commits are shown in the list. If not, delete the lines with other commits
- Save and close
- Push to your remote: `git push --force-with-lease [your-remote]` (`[your-remote]` is what is listed with your repo when calling `git remote -v`)

vlnovikov:
Unfortunately, game crashed during gameplay (attempt to open message window) with the same issue.
I've made a big commit in fork:
https://github.com/vlnovikov/s25client/commit/4d8e6ac1d792c146c25a2be928ecd3662f9edc37
Looks like more general fix is required.

vlnovikov:
https://github.com/vlnovikov/libutil/commit/3fa293cf6b8e311c52d71ae91d0bfc17886e81c5
This looks like a valid fix.

Flamefire:
I'm unsure if the good fix is to simply not print thousand separators. Being more general the digit separator could suffer from the same problem. We also run into this problem if the system locale is not UTF-8 (like `uk_UA.koi8u`), see https://vstinner.readthedocs.io/unicode.html

However for you this issue shouldn't even exist. The thousand separator for your locale should be a non-breaking space (`0x202F`): https://lh.2xlibre.net/locale/uk_UA/. Especially as the locale returns an invalid UTF-8 sequence although it claims to be UTF8 is puzzling me. Maybe something in your locales is broken.
- Why are you using mixed locales (English lang, uk format)? Is this on purpose?
- Can you check if an update solves this? (`sudo apt-get update && sudo  apt-get upgrade && sudo  locale-gen`, then restart (or relogin))
- If I try this on my Linux Mint machine I get a simple space as the separator (`0x20`), which is also strange...

Solution I'm thinking about is to ditch locale-aware formatting and use the C-default (`1000`, `3.14`, ...) see #1067

vlnovikov:
I've tried to switch country settings (or how we should call it) to UK and the issue gone. None of update/locale-gen didn't help.
Probably, something indeed wrong with the locale itself. Using numpunct from std::locale also fixed the problem: https://github.com/vlnovikov/libutil/commit/098bda68516e55763049e116e858d030028093e6
And I thought this one is just a small issue :(

Flamefire:
Can you check my branch Flamefire:locale_fixes (don't forget the updated submodules)

This should be a less invasive fix

vlnovikov:
With Flamefire:locale_fixes issue has gone. Thank you very much.

PR #1070, opened 2019-06-25
Fix crash if system locale returns non-UTF8 symbols during number format

Flamefire:
Alternative to #1067:
- Format a number (>1000, with decimals) to a string with the system locale.
- If that string is not valid UTF8, use classic locale

Additionally updates the used UTF8 library which contains simplifications, support for `std::u32string` and bugfixes (especially correct handling of empty strings)


extras/videoDrivers/SDL2/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ find_package(SDL2 2.0.4)
 
 if(SDL2_FOUND)
 	add_library(videoSDL2 SHARED ${RTTR_DRIVER_INTERFACE} VideoSDL2.cpp VideoSDL2.h icon.h icon.cpp)
-    target_link_libraries(videoSDL2 PRIVATE videodrv s25util glad utf8 SDL2::SDL2)
+    target_link_libraries(videoSDL2 PRIVATE videodrv s25util glad utf8::cpp SDL2::SDL2)
```
Flow86: looks like the indentation is wrong?

extras/videoDrivers/SDL2/VideoSDL2.cpp
```diff
@@ -23,7 +23,7 @@
 #include "helpers/containerUtils.h"
 #include "icon.h"
 #include "openglCfg.hpp"
-#include "libutf8/utf8.h"
+#include "utf8.h"
```
Flow86: shouldn't it be <utf8.h>?

libs/libGamedata/lua/LuaInterfaceBase.cpp
```diff
@@ -19,7 +19,7 @@
 #include "LuaInterfaceBase.h"
 #include "mygettext/mygettext.h"
 #include "mygettext/utils.h"
-#include "libutf8/utf8.h"
+#include "utf8.h"
```
Flow86: same here

libs/s25main/controls/ctrlEdit.cpp
```diff
@@ -23,6 +23,7 @@
 #include "ogl/FontStyle.h"
 #include "ogl/glArchivItem_Font.h"
 #include "libutil/StringConversion.h"
+#include <utf8.h>
```
Flow86: here you included it with <>

libs/s25main/ogl/glArchivItem_Font.cpp
```diff
@@ -22,11 +22,11 @@
 #include "drivers/VideoDriverWrapper.h"
 #include "glArchivItem_Bitmap.h"
 #include "helpers/containerUtils.h"
+#include "utf8.h"
```
Flow86: same here

tests/s25Main/IO/testFileIO.cpp
```diff
@@ -17,12 +17,12 @@
 
 #include "rttrDefines.h" // IWYU pragma: keep
 #include "ListDir.h"
-#include "libutil/ucString.h"
 #include <boost/filesystem.hpp>
 #include <boost/iostreams/device/mapped_file.hpp>
 #include <boost/iostreams/stream.hpp>
 #include <boost/nowide/fstream.hpp>
 #include <boost/test/unit_test.hpp>
+#include <utf8.h>
```
Flow86: and here again with <>

Flamefire:
> Perhaps we should check all includes. Also to match the common and sane order:

This is quite hard to achieve but can be done by manually specifying 3rd parties in the clang-format file

About `""` vs `<>` includes: This is a hard one. We could say all external (from a modules view) includes via `<>` (which includes utf8 and also libutil) and internal with `""`, but as you can't really enforce it... I guess we are left checking it manually.

Will fix the marked parts

Flow86:
Yea I know it has to be done manually.

> We could say all external (from a modules view) includes via `<>`

exactly

Flamefire:
Where do we put in boost?

Flow86:
thats 3rdparty, so after project but before system

Flamefire:
Ok. I was just thinking about giving boost a more special position (like "almost system") but let's not make it to complicated

Flow86:
yep in most projects we did it like that - boost was included directly before system

Flamefire:
Ok done: https://github.com/Return-To-The-Roots/s25client/pull/1070/files#diff-69a68f7ce1ea3687a38d8e4705e6fa48

BTW: I'd not require GPG signed commits. Just makes it harder for contributors and I don't see a benefit

Flow86:
thats why its not set as "required"

Flamefire:
But it shows up as a failure. So why have it at all?

Flow86:
because I want to start using it. It does have some benefits

Flamefire:
What stops you from using it while having the check disabled? Even I was confused about the red "Some checks were not successful" and thought I'd had to fix something. It even shows up in the title as "failed" which hides the other success/failure status. So I imagine for new users/contributors it will be very confusing/disturbing.

So what are the benefits of the (optional) check?   
I don't even get the real benefit of the commit signing anyway. I pushed it to my github account, so I provably have access to it. Why prove, that I have another access key (which I can change once I have access to my github account)? Both keys will be stored on the same machine so where is the additional security?

Flow86:
benefits of gpg signing:

https://mikegerwitz.com/2012/05/a-git-horror-story-repository-integrity-with-signed-commits

Issue #1073, opened 2019-07-03
Speed Up build by placing #pragma once at the correct place


Flow86:
We put our `#pragma once` in the header files at the wrong place.
We have to put it before the include guard defines, otherwise it does not have any purpose.

Flamefire:
Better: Remove it.

It is not supported by all compilers as it is non-standard and modern compilers do the same optimization for the header guards as they would for `#pragma once`. I personally do like the `#pragma once` more but I guess compatibility is more important

Flow86:
> modern compilers do the same optimization for the header guards as they would for `#pragma once`
thats not true.
at least for gcc and Visual Studio using `#pragma once` has a significant impact. The fact is, the compiler does not need to continue parsing the whole file when reaching a `#pragma once`. otherwise the compiler always to do it even if the include guard afterwards will avoid double processing of the content.

Flow86:
oh and mostly only very old compilers or very specialized compiler do not have support for `#pragma once`

Flamefire:
Sorry but that does not seem to be true

- For gcc it is measured to not matter: https://web.archive.org/web/20080930061318/http://www.gamesfromwithin.com/articles/0501/000067.html
- For VS 2003 it was around 20% faster
- "Yet, since include guards appear very often and the overhead of opening files is significant, it's common for compilers to optimize the handling of include guards, making them as fast as #pragma once" https://en.wikipedia.org/wiki/Pragma_once
- Cray compilers stopped supporting it: http://pubs.cray.com/content/S-2179/8.6/cray-c-and-c++-reference-manual-s-2179-86/miscellaneous-directives

As the cost of reintroducing header guards is exceptionally high I'd suggest not to remove them and instead rely on header guards and the compilers optimization for them

Flamefire:
Newer measurements: https://tinodidriksen.com/2011/08/cpp-include-speed/

MSVC:
- Pragma only | 3.597
- Ifndef only | 5.153

GCC:
- Pragma only | 2.005
- Ifndef only | 0.309

Clang:
- Pragma only | 0.532
- Ifndef only | 0.568

This is for ~50000 files/30000 includes. Hence speed does not matter.

Flow86:
nobody spoke about removing the include guards. I don't know what they tested there, but here we definitly have an improvement since we used pragma once (and include guards) everywhere in the correct order - and also reordered all includes (perhaps the reason it got faster is the reordering) but that way every file was structured the same.

[...]

PR #1074, opened 2019-07-03
WIP: ENH: started updating copyright headers

Flow86:
- reorder includes and update include guards + pragma once
- add SPDX-License-Identifier


extras/macosLauncher/starter.cpp
```diff
@@ -1,11 +1,33 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Return To The Roots. If not, see <http://www.gnu.org/licenses/>.
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+
 #include "RttrConfig.h"
+
```
Flamefire: Don't use newlines. I was already hunting for some of them to remove them. Use clang-format to order includes

extras/macosLauncher/starter.cpp
```diff
@@ -1,11 +1,33 @@
...
 #include <libutil/System.h>
+
 #include <boost/algorithm/string/replace.hpp>
 #include <boost/filesystem.hpp>
 #include <boost/nowide/fstream.hpp>
-#include <cstdio>
+
 #include <iostream>
 #include <stdexcept>
+#include <cstdio>
```
Flamefire: Not sorted properly

extras/videoDrivers/SDL/VideoSDL.cpp
```diff
@@ -14,21 +14,30 @@
 //
 // You should have received a copy of the GNU General Public License
 // along with Return To The Roots. If not, see <http://www.gnu.org/licenses/>.
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
```
Flamefire: Why do we require this? Makes the license header even longer :/

extras/audioDrivers/SDL/SoundSDL_Music.h
```diff
@@ -14,24 +14,40 @@
...
+        : SoundDesc(SD_MUSIC)
+        , music(nullptr)
+        , data(std::move(data))
+    {}
 
     void setInvalid() { isValid_ = false; }
-    /// Handle to the sound. Managed and freed by the driver
-    Mix_Music* music; /// Das Music-Handle.
+
+    Mix_Music* music; ///< Handle to the sound. Managed and freed by the driver
```
Flamefire: I'd rather have them above the decl. Long lines make reading this harder. the  tripple-slash already marks it as doc

libs/common/include/variant.h
```diff
@@ -14,42 +14,47 @@
...
-#define variant_h__
+#pragma once
+#ifndef libs_common_include_variant_h
+#define libs_common_include_variant_h
 
 #include <boost/variant.hpp>
+
 #include <type_traits>
 
-namespace detail {
```
Flamefire: What happened to the formatting here?

extras/audioDrivers/SDL/AudioSDL.cpp
```diff
@@ -14,16 +14,21 @@
...
 
-#include "commonDefines.h" // IWYU pragma: keep
 #include "AudioSDL.h"
 #include "SoundSDL_Effect.h"
 #include "SoundSDL_Music.h"
-#include "driver/AudioInterface.h"
-#include "driver/IAudioDriverCallback.h"
-#include "driver/Interface.h"
+
+#include <commonDefines.h> // IWYU pragma: keep
```
Flamefire: NO! This include was purposely kept at the front with the pragma. This will break things

Flow86: its still WIP - but why should it break things? the order of the includes in a header file should not be critical

extras/audioDrivers/SDL/SoundSDL_Music.h
Flow86: I could move it back above if you want to. for variables its sometimes easier if its behind the name

extras/macosLauncher/starter.cpp
Flow86: sure I will use newlines. thats what newlines are for. this way you properly see which is local/module, project and system header.

extras/macosLauncher/starter.cpp
Flow86: oops

extras/audioDrivers/SDL/AudioSDL.cpp
Flamefire: It is for the *defines file which is why the clang-format file has a regex for that and the line an IWYU pragma. It defines common aliases used throught the project (`bfs`) and includes `cstdint` to be able to use the sized integer types like built-ins.

Some time ago ALL files needed the `*defines` include, I'm reducing this step by step. The `commonDefines` is a slim version of the old include that was used there and included a lot...

extras/videoDrivers/SDL/VideoSDL.cpp
Flow86: see https://spdx.org/ids-how

extras/audioDrivers/SDL/AudioSDL.cpp
Flow86: I will resolve that locally as soon as I'm finished with the files

extras/macosLauncher/starter.cpp
Flamefire: Although I've seen this used before, but don't like it. It is not really required as namespacing and naming should make things clear (`"local.h", <module/h.h>, <boost/h.h>, <system>`). Besides that I don't understand the benefit in seeing that distinction. Why would it be important WHERE a header lives (except for ordering which is done by clang-format) On the other hand I find scrolling annoying and hence try to reduce vertical space. And again you will need to painfully manually check that newlines are consistently used.

extras/videoDrivers/SDL/VideoSDL.cpp
Flamefire: Ok that is how, but why? The license texts are already "standard"/fixed-and-known. So what benefit do we get for adding these 2 lines to 3k files?

Flamefire: Just seen that Linux 5.2 completely replace the license boilerplate with this 1 liner. See https://www.kernel.org/doc/html/latest/process/license-rules.html for rules and https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/diff/arch/arm/boot/dts/hi3519.dtsi?id=2c1212de6f9794a7becba5f219fa6ce8a8222c90 for an example. Not sure if we need the `Copyright...` line (e.g. https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/alpha/include/asm/xor.h?id=86c2f5d653058798703549e1be39a819fcac0d5d does not have it), so I'd at least remove all license text, have the SPDX as first line and potentially also get rid of the Copyright lines if they don't provide any actual benefit. If they do, keep them as a 2nd one-liner.

Flamefire:
Voting to close this:
- Heavily outdated. I guess trying to rebase and solve conflicts will be more work than doing a Search&Replace over the current tree
- Include guards OR `pragma once` not decided, see #1073 (TLDR: modern compilers optimize so both are equivalent, measurements done by others show no time difference, maintaining both is s development burden)
- copyright headers should be 3 lines only: https://github.com/Return-To-The-Roots/s25client/issues/1073#issuecomment-508078262
- lot's of formatting changes not backed by clang-format
- include reordering not backed by clang-format including erroneous moving of `#include <commonDefines.h> // IWYU pragma: keep` which explicitly is instructed to be the first include

Issue #1066, opened 2019-06-09
2560x1440 resolution does not fit the screen with the native resolution of 2560x1440


Alexander-Serov:
See attached fullscreen-mode screenshots. All other apps work fine.

Additional info: I'm using standard Windows 125% scaling, but it never affects full-screen apps. Worth checking?

Thanks for your efforts!

![bug_view](https://user-images.githubusercontent.com/22743277/59159584-452c3f80-8acc-11e9-8640-d569dc36c689.jpg)
![screen resolution](https://user-images.githubusercontent.com/22743277/59159587-4c534d80-8acc-11e9-901f-b54102c33d7e.png)
![bug_view2](https://user-images.githubusercontent.com/22743277/59159611-a9e79a00-8acc-11e9-9a5d-a5ef325925ff.jpg)

Flow86:
Its not a fullscreen app, its only borderless window. We have to check. There should be some kind of flag to automatically disable screen scaling for this app

Flamefire:
Can you try SDL2 as the graphics driver?

@Flow86 I think it is now really fullscreen after some fixes a while ago. At least what SDL considers fullscreen

Alexander-Serov:
I have checked with SDL2. The window then fits the screen, however the resolution is wrong. The options say it's still 2560x1440, but the text looks grained and when I take screenshots, they are 1920x1440. Also, other screen formatting bugs appear like this one (shifted text):
![image](https://user-images.githubusercontent.com/22743277/60394019-eaf82a80-9b1d-11e9-87cc-75babc0b238c.png)
or with black sides of the screen.

Spikeone:
I also use WQHD and I tried out the latest nightly build and can't reproduce that fault. I tried it with SDL and SDL2, no error there.

But I got another problem, when running the game with WINAPI the game doesn't even start anymore.

Starting in F:\s25rttr_20190627 
Searching for drivers in F:\s25rttr_20190627\driver/video
3 video drivers found!
Loaded video driver "(WinAPI) OpenGL via the glorious WinAPI"


FATAL ERROR: Failed to initialize the OpenGL context!



ERROR: Failed to start the game


Press ENTER to close this window . . . `

Alexander-Serov:
I confirm: 

- openGL didn't work at all,

- SDL does not fit the screen when Windows zoom is on,

- SDL2 fits the window, but the resolution is incorrect.

Flow86:
@Flamefire 
[...]

PR #1076, opened 2019-07-04
Fix videodrivers

Flamefire:
Follow-up for issues mentioned in #1066

This fixes WinAPI+OpenGL, wrongly selected resolutions and improves usability with SDL (window centred and possibly resized to fit screen)


Flow86:
what about SDL1 video driver?

Flamefire:
> what about SDL1 video driver?

What do you mean? I'd deprecate SDL1 and WinAPI as we have SDL2 working which is superior. So only bugs are fixed in SDL1/WinAPI but no new features added

Flow86:
so SDL1 does not need fixing?

Flamefire:
I'm not aware of a bug that needs fixing there

PR #1077, opened 2019-07-05
Fix crash on close

Flamefire:
When the textures stored in LOADER are destroyed the videodriver was
already shut down so the OGL library might already be unloaded. An
access to glDeleteTextures will then result in an AV.
Fix: Check if textures are already freed and don't free again.

### DIFF
--- a/libs/s25main/drivers/VideoDriverWrapper.cpp
+++ b/libs/s25main/drivers/VideoDriverWrapper.cpp
@@ -23,6 +23,7 @@
 #include "Settings.h"
 #include "WindowManager.h"
 #include "driver/VideoInterface.h"
+#include "helpers/containerUtils.h"
 #include "helpers/roundToNextPow2.h"
 #include "mygettext/mygettext.h"
 #include "ogl/DummyRenderer.h"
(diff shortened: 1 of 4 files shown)


libs/s25main/drivers/VideoDriverWrapper.cpp
```diff
@@ -185,11 +186,10 @@ bool VideoDriverWrapper::DestroyScreen()
         return false;
     }
 
-    // Texturen aufräumen
-    LOG.write("Saeubere Texturespeicher: ");
+    LOG.write("Clearing textures: ");
     unsigned ladezeit = GetTickCount();
     CleanUp();
-    LOG.write("fertig (nach %dms)\n") % (GetTickCount() - ladezeit);
+    LOG.write("finished in %dms\n") % (GetTickCount() - ladezeit);
```
Flow86: Finished

tests/s25Main/UI/testVideoDriver.cpp
```diff
@@ -39,3 +43,58 @@ BOOST_AUTO_TEST_CASE(FindClosestVideoMode)
...
+    for(unsigned i = 1u; i <= 5u; ++i)
+        BOOST_TEST(VIDEODRIVER.GenerateTexture() == i);
+    BOOST_TEST_REQUIRE(rttrOglMock2::activeTextures.size() == 5u);
+    VIDEODRIVER.DestroyScreen();
+    BOOST_TEST_REQUIRE(rttrOglMock2::activeTextures.empty());
+    // Next cleanup call is a no-op (validated inside glDeleteTextures)
+    VIDEODRIVER.CleanUp();
+
+    VIDEODRIVER.CreateScreen(VideoMode(800, 600), false);
+    glGenTextures = rttrOglMock2::glad_glGenTextures;
```
Flow86: isn't it called glGenTextures and not glad_glGenTextures ?

Flamefire: true. My IDE was smart enough to look through the macro ^^

PR #1078, opened 2019-07-15
WIP: Snap packaging (Ubuntu)

Flow86:
@lool created a snap for ubuntu.

Some parts I'd like to have discussed:

- the S2 path could/should perhaps be in the users home instead of the global path. I don't like that one needs admin privileges to run the game
- travis and http://build.snapcraft.io/ needs to be integrated

---
from original email:

Currently, the snap:
- has working audio and graphics and network server; I can load a map and start playing; MIDI playback and sound effects work
- embeds all dependencies – perhaps too much
- has been tested under a clean 18.04 desktop VM
- uses almost minimal permissions; I think desktop-legacy could probably be dropped
- editor currently segfaults on startup unless you set LC_ALL=C
- has no .desktop file – need to add one

### DIFF
--- a/snap-utils/snap-launcher
+++ b/snap-utils/snap-launcher
@@ -1,0 +1,16 @@
+#!/bin/sh
+
+pulse=`realpath $SNAP/usr/lib/*/pulseaudio`
+dri=`realpath $SNAP/usr/lib/*/dri`
+
+LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$dri:$pulse"
+export LD_LIBRARY_PATH
+LIBGL_DRIVERS_PATH="$LIBGL_DRIVERS_PATH:$dri"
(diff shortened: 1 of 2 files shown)


Flow86:
@lool Theres a desktop file at https://github.com/Return-To-The-Roots/s25client/blob/master/tools/release/debian/s25rttr.desktop

Flow86:
> Not sure about that. What about automatic updates? Can this be really done with the nightlies (and more important: Should it?`) It is additional work since we have the auto-updater

Well, thats the CI for: http://build.snapcraft.io/ which will automatically create updated snap-packages for it. but the integrated updater should be disabled (not writable)

Flamefire:
> but the integrated updater should be disabled (not writable)

I think we introduced a CMake flag for that. This should be set in the SNAP file

PR #1081, opened 2019-07-21
WIP: Preliminary Haiku support

extrowerk:
zpqE


cmake/haiku.cmake
```diff
@@ -0,0 +1,4 @@
+# set compiler flags
+set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -lnetwork -lexecinfo")
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -O2 -g -std=c11 -D_BSD_SOURCE")
```
Flamefire: `-O2 -q` should not be required, `-D_BSD_SOURCE` looks suspicious and the standard should not be set here. I'm also not aware of C11 features used.

cmake/haiku.cmake
```diff
@@ -0,0 +1,4 @@
...
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O2 -g -std=c++11 -D_BSD_SOURCE")
```
Flamefire: Similar here: Nothing should be required. Especially the `-std` is wrong

cmake/toolchains/c.haiku.common.cmake
```diff
@@ -0,0 +1,9 @@
+set(CMAKE_SYSTEM_NAME Haiku)
+
+if(NOT CMAKE_SYSTEM_VERSION)
+    message(FATAL_ERROR "Need to set CMAKE_SYSTEM_VERSION to target version (uname -r) on cmd line")
+endif()
+
+find_program(CMAKE_C_COMPILER NAMES ${COMPILER_PREFIX}-gcc)
```
Flamefire: Where is `COMPILER_PREFIX` set?

cmake/haiku.cmake
extrowerk: _BSD_SOURCE was required so far because Haiku had no features.h, but this patchset is almost 1 year old (i just had no time to upstream it), so thanks for the input, i will recheck.

extrowerk:
Here is my current recipe: https://github.com/haikuports/haikuports/pull/3999

cmake/toolchains/c.haiku.common.cmake
Flamefire: this whole file is not required and can be removed. You are not crosscompiling and hence don't need a toolchain file.

Flamefire:
Just checked this again:   
- Why did you add the c.*.cmake files? They don't seem to be used. (They are only used for cross-compiling but your recipe compiles natively) -> Remove?
- What happens when you compile without the `-D_BSD_SOURCE`?
- What happens when you compile without the `-lnetwork -lexecinfo`?

Flow86:
any progress here?

extrowerk:
New release expected at the end of may. Will check this.

extrowerk:
New Haiku release expected at the end of may with the needed support for s25. Will re-check and fix after the release.

Flamefire:
> New Haiku release expected at the end of may with the needed support for s25

I'm wondering what you mean by "needed support". What are they missing?

extrowerk:
> I'm wondering what you mean by "needed support". What are they missing?

Boost was broken, so had to add the elf headers to Haiku from FreeBSD, etc-etc-etc.
More info here: https://github.com/haikuports/haikuports/pull/3999
https://review.haiku-os.org/c/haiku/+/1635
https://review.haiku-os.org/c/haiku/+/1620

This was after Beta1, and the buildbots won't include this changes til the beta2 release.

PR #1083, opened 2019-07-28
Replace resampler by libsamplerate

### LINKED ISSUE #1082: Replace bulky sound-converter
There is a lot of old C code for the sound-converter used:
- Load all kinds of sound files via sndlib
- Convert sounds via resample 1.8.1

Especially the first is not required as we only have WAVE files of known format. A better solution than
- load from LST
- add header
- write to wav
- convert to wav
- load from wav
- write to LST

would be:
- load from LST
- convert in memory
- write to LST

For this I found a list of libraries which can do resampling: https://ccrma.stanford.edu/~jos/resample/Free_Resampling_Software.html

Especially appealing to me are:
- [smarc](http://audio-smarc.sourceforge.net/)
- [libsamplerate](https://github.com/erikd/libsamplerate) aka **S**ecret **R**abbit **C**ode

I'd try the latter which already has some CMake scripts (but needs patching though) and tests.
It should actually be added as a submodule to `s-c` but then we need recursive submodule init... Maybe add to main submodules?

Flamefire:
Uses SRC: http://www.mega-nerd.com/SRC/index.html

Included as a submodule, got a couple PRs open for upstream to improve it: https://github.com/erikd/libsamplerate/pulls/Flamefire   
Only minor things, in general it works well.

I hope to get the C++ wrapper for libsamplerate into upstream: https://github.com/erikd/libsamplerate/issues/71

Closes #1082


.gitmodules
```diff
@@ -31,3 +28,8 @@
 [submodule "contrib/turtle"]
 	path = external/turtle
 	url = https://github.com/mat007/turtle.git
+[submodule "external/libsamplerate"]
+
```
Flow86: seems like theres something wrong with line endings here?

libs/libsamplerate/include/samplerate.hpp
```diff
@@ -0,0 +1,181 @@
...
+    Result process(Data&& data) { return process(static_cast<Data&>(data)); }
+};
+
+class StateCallback : public detail::StateBase
+{
+    static SRC_STATE* createOrThrow(Converter converter, int channels, src_callback_t func, void* data)
+    {
+        int error;
+        auto* state = src_callback_new(func, static_cast<int>(converter), channels, &error, data);
+        if(!state)
```
Flow86: shouldn't you use `throwOnError` here? instead of the if

libs/libsamplerate/include/samplerate.hpp
```diff
@@ -0,0 +1,181 @@
...
+    return simple(static_cast<Data&>(data), converter, channels);
+}
+
+class State : public detail::StateBase
+{
+    static SRC_STATE* createOrThrow(Converter converter, int channels)
+    {
+        int error;
+        auto* state = src_new(static_cast<int>(converter), channels, &error);
+        if(!state)
```
Flow86: shouldn't you use `throwOnError` here? instead of the if

Flamefire: Yes and no. I want to throw an error if state is NULL, which *SHOULD* be the case if `error!=0`, but to be safe I'll also throw for `state=NULL && error=0`

Flamefire:
@Flow86 Is this good to go?

PR #1093, opened 2019-08-13
Fix windows deployment (missing DLLs)

### LINKED ISSUE #1092: DLLs missing in nightly builds
To run RttR you need the "lua52.dll" but this one is missing. 
To run the rttr.bat you need the "libgcc_s_seh-1.dll" but that one is missing too.

It's not possible to run the nightly at the moment.

Flamefire:
Fixes #1092

The main problem was a bug in the prepareRelease script which stopped execution (and copying of DLLs) after an error (`RTTR/{s25update}.exe not found`) but this did not abort the CMake installation. This would now happen after https://github.com/Return-To-The-Roots/s25client/commit/0e1a25e5f6f04d9d39099ba3b50c3ff3954c099b

To fix the bug I reworked the whole script to make it minimal. The remaining code could now probably be even ported to CMake (or use BundleUtilities)


CMakeLists.txt
```diff
@@ -380,6 +380,8 @@ set(CPACK_IGNORE_FILES "\\\\.a;/\\\\.ccache/;/\\\\.git/;${RTTR_OUTPUT_DIR};${PRO
 set(CPACK_SOURCE_IGNORE_FILES "${CPACK_IGNORE_FILES}")
 set(CPACK_BUILD_SOURCE_DIRS "${PROJECT_SOURCE_DIR}")
 
+set(CPACK_STRIP_FILES ON)
```
Flow86: shouldn't it only do that if not explicitly build as debug?

tools/release/bundleWinFiles.sh.cmake
```diff
@@ -0,0 +1,145 @@
+#!/bin/bash
+
+set -euo pipefail
```
Flow86: hmm can you fix the rename? it seems git didnt caught that correctly (`git mv a b`)

CMakeLists.txt
Flamefire: It will only strip the files when running `make install/strip` or when using cpack via `make package`. IMO this is ok.

Making it only do that in Release mode will not work with multi-generator targets.

Flow86: ah okay then its okay

tools/release/bundleWinFiles.sh.cmake
Flamefire: Maybe not in the full PR but the commit has it:
```
git show 92bfc140090764e15164a995f17ec87a4ab59aed
diff --git a/tools/release/prepareRelease.sh.cmake b/tools/release/bundleWinFiles.sh.cmake
similarity index 98%
rename from tools/release/prepareRelease.sh.cmake
rename to tools/release/bundleWinFiles.sh.cmake
```

Flow86: hmm interesting.

PR #1097, opened 2019-08-15
Don't use Win32 resource include on non-windows builds

Flamefire:
The resource folder for win32 resources was unnecessarily  included in non-windows builds of videoSDL.

This is now conditionally. Additionally the formatting was made uniform.


Flamefire:
*ping*

PR #1104, opened 2019-08-20
Fix inclusion of libvorbisfile

Flamefire:
libvorbisfile-3.dll is a dynamic dependency of SDL_mixer.dll and was not found via dependency analysis. This is now included

I also slightly reworked the logic to be closer to CMakes BundleUtilites although that can't be used as it seems to ignore executables in subfolders and pts all DLLs into the main folder -.- Didn't spend more time as debugging this is almost impossible


stefson:
Can you please edit the Readme.md and raise the needed `libbz2-dev` accordingly?

Flamefire:
> Can you please edit the Readme.md and raise the needed libbz2-dev accordingly?

No version is stated in the readme. The linux repos should all have the non-vulnerable version

Flamefire:
> I don't know why you changed .rar to .7z - but nevermind

I had to update the archive anyway and in the process I tried 7z which yielded smaller file sizes. As it is a binary file :( it doesn't matter if I update or replace it. So I went for the smaller one

PR #1110, opened 2019-09-11
Always create gui for controls in Addons window

### LINKED ISSUE #1001: Default button in Addon Settings not working
Sadly the "Default" button in the "Addon Settings" is not working as expected. The button only resets addons if they have been shown once. Means you can not reset the settings by just opening the window and hitting "Default" as this only resets the visible addons. If you scroll down and hit "Default" everything is working as expected.

The "Default" button should always work, regardless of shown addons.

options

iceslab:
Always create gui for all addon controls to allow resetting to default values. Move shifting `y` position from `createGui` to `UpdateView`. Fixes #1001.

### DIFF
--- a/libs/s25main/ingameWindows/iwAddons.cpp
+++ b/libs/s25main/ingameWindows/iwAddons.cpp
@@ -155,18 +155,22 @@ void iwAddons::UpdateView(const unsigned short selection)
             continue;
         unsigned groups = addon->getGroups();
 
+        bool isReadOnly = policy == READONLY || (policy == HOSTGAME_WHITELIST && !helpers::contains(addonIds, addon->getId()));
+        addon->createGui(this, id, y, isReadOnly, status);
+
         if((groups & selection) == selection)
             ++numAddonsInCurCategory;
+
         // hide addon's gui if addon is beyond selected group or is beyond current page scope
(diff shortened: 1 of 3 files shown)


iceslab:
My commit has failed in Jenkins, but I get 404 error when accessing details. Could you look at this or retrigger it?

libs/s25main/ingameWindows/iwAddons.cpp
```diff
@@ -155,18 +155,22 @@ void iwAddons::UpdateView(const unsigned short selection)
...
         // hide addon's gui if addon is beyond selected group or is beyond current page scope
         if(((groups & selection) != selection) || numAddonsInCurCategory < scrollbar->GetScrollPos() + 1
            || numAddonsInCurCategory > (unsigned)(scrollbar->GetScrollPos() + scrollbar->GetPageSize()) + 1)
         {
             addon->hideGui(this, id);
-            continue;
+        } else
+        {
+            // if current button is displayed make sure next one will be below
+            y += 30;
```
Flamefire: The reasoning to increase `y` in the addon is to allow the addon to choose its height/step. I wouldn't change this. Maybe make `createGui` return it's y-increase (make y a non-reference param then) and add it to y if the addon is shown?

PR #1125, opened 2019-10-24
 Make building window active if it is already open 

Flamefire:
This already uses the newly split libutil (split into common, log, network, which reduces the dependencies [e.g. the updater no longer needs libupnpc])

The fix is to expose `SetActiveWindow` and use it when an open window was found.


tests/common/testPoint.cpp
```diff
@@ -248,10 +248,10 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(unscale_point, T, SignedTypes)
     BOOST_TEST(resultUnsigned == pt2 / UnsignedPoint::all(scale), epsilon % tolerance());
 
     pt /= scale;
-    BOOST_TEST(pt == result);
+    BOOST_TEST(pt == result, epsilon % tolerance());
 
     pt2 /= scale;
-    BOOST_TEST(pt2 == resultUnsigned);
+    BOOST_TEST(pt2 == resultUnsigned, epsilon % tolerance());
```
Flow86: i thought floating point test should be done via i.e `BOOST_TEST(x == y, tolerance(0.0005));` ? or is that not available in our boost versions?

Flamefire: It seems `epsilon % tolerance()` is equivalent to `tolerance(epsilon/100)`. I think the first came up in a google search/boost docu first, so that's why I used it.

Flamefire:
@Flow86 Can this be merged? It updates submodules which can conflict with other changes.

Issue #977, opened 2018-12-27
Mapselection visible text bug


Spikeone:
I've got a strange text bug everytime I list my "other" maps folder.

![image](https://user-images.githubusercontent.com/1276342/50494174-f4df8a80-0a21-11e9-8a3c-97c492c8baa0.png)

Here my other folder is (20mb): 
[OTHER.zip](https://github.com/Return-To-The-Roots/s25client/files/2713348/OTHER.zip)

iceslab:
It appears that's not a UI bug, but malformed author name in map itself.
![malformed_author](https://user-images.githubusercontent.com/10535944/64729761-bddcd680-d4d5-11e9-8550-209c0fa07ebd.png)

Flamefire:
Ok so the problem is a `\n` (newline) in a text coming from users. The (minor) problem: Parsing of newlines is a feature. We would need some infrastructure to explicitly disallow this or check for in things like the map name, author, ... This can be done on loading the map or just on some controls.
I guess for the list control used here we can do it on the control: On *NO* list control we want newlines in the rows.

Question: What to do with them? Print as `\n` literal or remove?

Spikeone:
In my opinion I'd simply (if thats simply) go for literal as for example if you have to authors who call themselves horst and ninja they might label the autor as horst\ninja - well thats not very probable but - well better see something than just removing it in my opinion.

Flamefire:
Kinda simple. I'll take a look at it once @Flow86 is back to merge PRs or allows to merge without his explicit review.

Flamefire:
I'm currently about to remove newline handling from the font rendering completely. We have ctrlMultiline and text splitting for creating multiple strings out of a string containing possible newlines.

The side effect is, that because `\n` does not exist as a glyph it will be rendered as `?` instead. I don't really want to implement replacement of such rare occurrences and rather use the "non-printable-character" replacement (in our case `?`) for that. As it is an error to include a newline in the map title I think this is appropriate.

What do you think?

Spikeone:
For me this sounds good enough. Althoug one could think about flipping it like ¿ just to make sure everybody sees this is an errorneous character and not a missplaced question mark.

Flamefire:
Yes. But maybe we use the proper one: https://www.fileformat.info/info/unicode/char/fffd/index.htm
Could you make them?

Spikeone:
[...]

PR #1137, opened 2019-11-17
Rework line break handling

Flamefire:
To fix https://github.com/Return-To-The-Roots/s25client/issues/977 this adds the U+FFFD glyph to show unprintable chars and treats newlines in fonts as such. So all line breaks must be handled outside of `font.Draw()` e.g. via `ctrlMultiline`

Also the font class has been refactored: Only UTF8 text is handled now (other text (mostly) wasn't used), which needed changes in ctrlEdit. I added tests for that too. Also the Font class now is constructed from a Font-ArchivItem instead of being one. This initializes the class eagerly (on ctor) instead of lazily allowing use as const (why would a `Draw` be non-const?)

Especially the line-break-handling removal brought a ~5% performance gain when drawing text. I expect the new class to perform generally better due to smaller memory usage and less checks to be done as well as being non-virtual avoiding a `dynamic_cast` in the loader

For the future I'd model all classes like that: Separate ArchivItem and actual usage. E.g. we don't need 6 different `gl*Bitmap` classes, but only 1 (or 2 for player) sprite or texture class. So during loading the classes are "transformed"/loaded into GPU or whatever and then used like that. Cut down on `dynamic_cast`s at runtime.


libs/s25main/Loader.cpp
```diff
@@ -264,6 +269,25 @@ bool Loader::LoadSounds()
     return true;
 }
 
+bool Loader::LoadFonts()
+{
+    if(!LoadFile(RTTRCONFIG.ExpandPath("<RTTR_RTTR>/LSTS/OUTLINE_FONTS.LST"), GetPaletteN("pal5")))
+        return false;
+    fonts.clear();
+    const auto& loadedFonts = GetArchive("outline_fonts");
+    for(int i = 0; i < 3; i++)
```
Flow86: we should make sure that the archive contains 3 fonts here

Flamefire: It does: See 3 lines below

Flow86: ah okay. I thought about a good way to not hardcode the "3" here - perhaps at least add some kind of constant `FONT_COUNT` or similar

libs/s25main/Loader.cpp
```diff
@@ -303,13 +327,11 @@ void Loader::LoadDummyGUIFiles()
         io.push(std::move(bmp));
     }
     // Fonts
-    libsiedler2::Archiv& fonts = files_["outline_fonts"].archiv;
     auto* palette = GetPaletteN("colors");
-    fonts.alloc(3);
     libsiedler2::PixelBufferBGRA buffer(15, 16);
     for(unsigned i = 0; i < 3; i++)
```
Flow86: same point is here with the "3"

libs/s25main/Loader.cpp
Flamefire: I added an enum `FontSize` and changed the signature of `GetFont` accordingly. This also makes sure you can only get valid fonts.

PR #1144, opened 2019-12-11
Fix crash on chat input of format strings

Flamefire:
The LOG class expects a format string, so do not pass a user-provided string to it which may or may not contain a format specifier/special char.

Also fix misbehaving edit on clearing where the internal string was not kept in sync with the actual text so clearing it still showed the old text.


Flow86:
so you can't log empty lines anymore? ;-)

Flamefire:
No, you can't log nothing anymore. An empty line is "\n" ;)

PR #1155, opened 2020-01-02
Fix font issues

Flamefire:
- Show log message if lua translation is unavailable
- Always create newline when wrapping font and encountering `\n`
- Fix bug in tooltip that uses min instead of max width


Flamefire:
@Flow86 Ready for merge. Travis failure is (IMO) a wrong configuration on their side -.-

Got a few semi-related commits in this though...

PR #1184, opened 2020-01-19
Lua fixes

Flamefire:
Some lua fixes like the victory dialog and UTF-8 BOM in lua scripts and some related work as UB due to specializing  `std` traits.

Also fixes the version in the ingame settings

### DIFF
--- a/libs/libGamedata/lua/LuaInterfaceBase.cpp
+++ b/libs/libGamedata/lua/LuaInterfaceBase.cpp
@@ -53,37 +53,32 @@ void LuaInterfaceBase::Register(kaguya::State& state)
 
 void LuaInterfaceBase::ErrorHandlerNoThrow(int /*status*/, const char* message)
 {
-    LOG.write(_("Lua error: %s\n")) % (message ? message : "Unknown");
+    LOG.write(_("Lua error: %s\n")) % (message ? message : _("Unknown"));
     errorOccured_ = true;
 }
 
 void LuaInterfaceBase::ErrorHandler(int status, const char* message)
 {
     ErrorHandlerNoThrow(status, message);
-    throw LuaExecutionError(message ? message : "Unknown error");
+    throw LuaExecutionError(message ? message : _("Unknown error"));
 }
 
 bool LuaInterfaceBase::LoadScript(const std::string& scriptPath)
 {
-    script_.clear();
     bnw::ifstream scriptFile(scriptPath.c_str());
     std::string tmpScript;
     tmpScript.assign(std::istreambuf_iterator<char>(scriptFile), std::istreambuf_iterator<char>());
     if(!scriptFile)
     {
-        LOG.write("Failed to read lua file '%1%'\n") % scriptPath;
+        LOG.write(_("Failed to read lua file '%1%'\n")) % scriptPath;
         return false;
     }
-    if(!ValidateUTF8(tmpScript))
-        return false;
-    try
+    // Remove UTF-8 BOM if present
+    if(tmpScript.substr(0, 3) == "\xef\xbb\xbf")
+        tmpScript.erase(0, 3);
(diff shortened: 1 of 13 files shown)


Flamefire:
@Flow86 Done. There still is this error on travis which I don't understand. I openend a thread in their support forum, so let's see what happens.

> dyld: Library not loaded: /System/Library/Frameworks/ImageIO.framework/Versions/A/Resources/libGIF.dylib
> 
>   Referenced from: /System/Library/Frameworks/ImageIO.framework/Versions/A/ImageIO
> 
>   Reason: Incompatible library version: ImageIO requires version 1.0.0 or later, but libGIF.dylib provides version 0.0.0

Flamefire:
@Flow86 Can this be merged?

libs/libGamedata/lua/LuaInterfaceBase.cpp
```diff
@@ -53,37 +53,32 @@ void LuaInterfaceBase::Register(kaguya::State& state)
 
 void LuaInterfaceBase::ErrorHandlerNoThrow(int /*status*/, const char* message)
 {
-    LOG.write(_("Lua error: %s\n")) % (message ? message : "Unknown");
+    LOG.write(_("Lua error: %s\n")) % (message ? message : _("Unknown"));
```
Flow86: should'nt be this also "Unknown error"? to simplify translation?

libs/s25main/ingameWindows/iwOptionsWindow.cpp
```diff
@@ -66,114 +85,83 @@ iwOptionsWindow::iwOptionsWindow()
...
-    AddTextButton(17, DrawPoint(67, 443), Extent(168, 24), TC_RED1, _("Surrender"), NormalFont);
-    // "Spiel beenden"
-    AddTextButton(0, DrawPoint(67, 474), Extent(168, 24), TC_RED1, _("End game"), NormalFont);
+    AddTextButton(ID_btMusicPlayer, DrawPoint(100, 380), Extent(160, 22), TC_GREEN2, _("Music player"), NormalFont);
+    AddTextButton(ID_btAdvanced, DrawPoint(67, 412), Extent(168, 24), TC_GREEN2, _("Advanced"), NormalFont);
+    AddTextButton(ID_btSurrender, DrawPoint(67, 443), Extent(168, 24), TC_RED1, _("Surrender"), NormalFont);
+    AddTextButton(ID_btEndGame, DrawPoint(67, 474), Extent(168, 24), TC_RED1, _("End game"), NormalFont);
 }
 
 void iwOptionsWindow::Msg_ButtonClick(const unsigned ctrl_id)
```
Flow86: we should perhaps some time migrate these multiplexing stuff to std::function/bind?

Flamefire: I was thinking about the same so you can do `AddTextButton(...).OnClick([this](){ this->doSomething(); }` and get rid of the IDs altogether. But this has downsides too and is a lot of work for not much gain

libs/libGamedata/lua/LuaInterfaceBase.cpp
Flamefire: Yes, done

libs/s25main/ingameWindows/iwOptionsWindow.cpp
Flow86: exactly. The only downside (besides changing a lot of code) which comes in my mind is that you "pollute" your constructor instead? The whole (de)multiplexing would get removed by such an approach?

Flamefire: You'll also need complete types (But then we could get rid of the `AddFoo` functions) and you might want more than 1 line of code which doesn't look good in lambdas, hence you'll get many OnFooClicked functions or so.

PR #1186, opened 2020-01-28
Rework of random map generator

vader1986:
Hi guys, hope you're all well. 

I know, this looks like a massive PR, but I only rewrote code of the random map generator I originally added. Therefore, I deleted the "mapGenerator" folders and added a folder named "randomMaps" instead. If you search for "readme" you'll find an explanation of the folder structure and how the new map generator works. 

I worked on this for quite a while, but unfortunately with a lot of breaks in between. Probably best, if you check out the branch and give it a try, you'll see the generated maps are a lot better than before. Also I tried to split up the code to make it more testable, so every single class has unit tests as well. 

The things I had issues with:
* keeping it in sync with master. Initially I tried to rebase occasionally, but there were some large structural changes, so this became really tricky - therefore, you see a single commit with all the changes here. It's not really touching any in-game related code, so I don't see a massive risk with this rework.
* using existing RTTR data structures which were introduced to the original map generator. That's what I mean: https://github.com/Return-To-The-Roots/s25client/blob/master/libs/s25main/mapGenerator/RandomConfig.cpp#L82

There might be a few more change's I'll do, because I haven't checked if the CI passes. 

image
image
image
image
image
image


Flamefire:
First you'll need to to a `git rebase -i master` and remove all the unrelated commits which are probably from some old changes. Make sure your local master is up to date with origin master before doing so.

Just some cursory look for now:
- Why the rename to the rather non-descriptive "randomMaps"? I find "mapGenerator" better suited.
- See if GridUtility can use/be replaced by the MapBase class which handles distance in nodes, neighbours, ..., See also MapGeometry
- IMO the utilities are not required. We have a bitmap class in libsiedler2 already and the visualizer might not be required due to the map preview classes
- NextPowerOfTwo is already somewhere (I guess in common)
- Use std algorithms instead of raw loops, e.g.https://github.com/vader1986/s25client/blob/254a5cf05c9f63e56e897e702399141b4e4b3d00/libs/s25main/randomMaps/elevation/Level.cpp#L24-L31 and https://github.com/vader1986/s25client/blob/254a5cf05c9f63e56e897e702399141b4e4b3d00/libs/s25main/randomMaps/elevation/Level.cpp#L60-L65 can use the constructor
- Use namespaces instead of classes full of static members
- in fact: namespace everything (I know we have been sloppy with that but it avoids name clashes) Basically: root namespace is rttr and every folder is a namespace
- See if you can map your own texture types (if you really need them) to the rttr ones, see https://github.com/Return-To-The-Roots/s25client/blob/master/libs/s25main/mapGenerator/RandomConfig.cpp#L82 vs own traits https://github.com/vader1986/s25client/blob/feature/rewrite-random-map-generation/libs/s25main/randomMaps/terrain/TextureType.h
- use `using foo = bar` instead of typedef: https://github.com/vader1986/s25client/blob/feature/rewrite-random-map-generation/libs/s25main/randomMaps/terrain/CoastlineCalculator.h#L24
- Thinks like this class: https://github.com/vader1986/s25client/blob/feature/rewrite-random-map-generation/libs/s25main/randomMaps/terrain/CoastlineCalculator.h can be simple functions
-  remove the "rttrDefines.h" includes from new code wherever possible. It used to be a "include everything" type include. Just include what you really need.
- Instead of making folders of related stuff, check what can be combined into files. Compilation time is already very high and each file creates some overhead, so the less files, the better. Using free functions instead of classes with a few static members might help (e.g. the above linked Coastline class can be `calculateCoastline(...)`)
- Check your constructors. I've seen one (but forgotten) which copies vectors needlessly. Also Ctors like https://github.com/vader1986/s25client/blob/1c29de3a9673e1a054c26f42d2b53567efea2094/libs/s25main/randomMaps/MapSettings.h#L30-L40 can use in-class initializers (remove the constructor and assign the values to the class members directly in the class definition)
- Things like https://github.com/vader1986/s25client/blob/1c29de3a9673e1a054c26f42d2b53567efea2094/libs/s25main/randomMaps/MapSettings.cpp#L36 can use `helpers::clamp`
- Code like https://github.com/vader1986/s25client/blob/1c29de3a9673e1a054c26f42d2b53567efea2094/libs/s25main/randomMaps/Migration.cpp#L80-L84 looks questionable. Sizes can be mostly arbitrary. Why specialize for 2 values? Shouldn't those be ranges instead?
- Can the Creator classes be combined into 1 that has parameters only? The general flow is the same, maybe the specific code can be abstracted to depend only on parameters like lake sizes etc? Or put the common parts (smooth, set HQs, add rivers, place resources) in 1 class and use plugin classes to do the specific stuff
- Don't use `new` or `delete` but `make_unique`

I'll stop for now. Don't be discouraged by this, it sounds really good what you do. I only try to hold to high implementation standards so the code stays maintainable which isn't easy with the existing code but certainly possible for new code.

vader1986:
> First you'll need to to a `git rebase -i master` and remove all the unrelated commits which are probably from some old changes. Make sure your local master is up to date with origin master before doing so.

Looks like there's something messed up with my fork. 

> 
> Just some cursory look for now:
> 
> * Why the rename to the rather non-descriptive "randomMaps"? I find "mapGenerator" better suited.

That was only to avoid overlapping changes, while re-writing the random map generator. Just forgot to rename it back I guess. 

> * See if GridUtility can use/be replaced by the MapBase class which handles distance in nodes, neighbours, ..., See also MapGeometry
> * IMO the utilities are not required. We have a bitmap class in libsiedler2 already and the visualizer might not be required due to the map preview classes
> * NextPowerOfTwo is already somewhere (I guess in common)

Without any help I find it quite difficult to find the myself through the code base (I tried to fix some errors in the past while playing in debug mode). Thanks for the hints, I'll take a took a look at those classes you mentioned in the upcoming weeks. 

> * Use std algorithms instead of raw loops, e.g.https://github.com/vader1986/s25client/blob/254a5cf05c9f63e56e897e702399141b4e4b3d00/libs/s25main/randomMaps/elevation/Level.cpp#L24-L31 and https://github.com/vader1986/s25client/blob/254a5cf05c9f63e56e897e702399141b4e4b3d00/libs/s25main/randomMaps/elevation/Level.cpp#L60-L65 can use the constructor

Sorry, which constructor? 

> * Use namespaces instead of classes full of static members
> * in fact: namespace everything (I know we have been sloppy with that but it avoids name clashes) Basically: root namespace is rttr and every folder is a namespace

I definitely prefer using namespaces, I wasn't aware that you were using those at all (apart from the libsiedler namespace). I'll change that as well.

> * See if you can map your own texture types (if you really need them) to the rttr ones, see https://github.com/Return-To-The-Roots/s25client/blob/master/libs/s25main/mapGenerator/RandomConfig.cpp#L82 vs own traits https://github.com/vader1986/s25client/blob/feature/rewrite-random-map-generation/libs/s25main/randomMaps/terrain/TextureType.h

I had a look at those before, but don't really get how that works. I'll take another look at that, for now I'll add it to my todo list as well, but I might come back with a few questions. 

> * use `using foo = bar` instead of typedef: https://github.com/vader1986/s25client/blob/feature/rewrite-random-map-generation/libs/s25main/randomMaps/terrain/CoastlineCalculator.h#L24
> * Thinks like this class: https://github.com/vader1986/s25client/blob/feature/rewrite-random-map-generation/libs/s25main/randomMaps/terrain/CoastlineCalculator.h can be simple functions

Sounds good, will also add it to my todos. 

> * remove the "rttrDefines.h" includes from new code wherever possible. It used to be a "include everything" type include. Just include what you really need.

What's this one for anyway? I don't remember why I added it in the first place, but probably to avoid some linker or compile error. 

> * Instead of making folders of related stuff, check what can be combined into files. Compilation time is already very high and each file creates some overhead, so the less files, the better. Using free functions instead of classes with a few static members might help (e.g. the above linked Coastline class can be `calculateCoastline(...)`)

I see, I didn't no that compile time is an issue. On my local machine, it doesn't take long at all. But happy to merge things together where it makes sense ... 

> * Check your constructors. I've seen one (but forgotten) which copies vectors needlessly. Also Ctors like https://github.com/vader1986/s25client/blob/1c29de3a9673e1a054c26f42d2b53567efea2094/libs/s25main/randomMaps/MapSettings.h#L30-L40 can use in-class initializers (remove the constructor and assign the values to the class members directly in the class definition)
> * Things like https://github.com/vader1986/s25client/blob/1c29de3a9673e1a054c26f42d2b53567efea2094/libs/s25main/randomMaps/MapSettings.cpp#L36 can use `helpers::clamp`
> * Code like https://github.com/vader1986/s25client/blob/1c29de3a9673e1a054c26f42d2b53567efea2094/libs/s25main/randomMaps/Migration.cpp#L80-L84 looks questionable. Sizes can be mostly arbitrary. Why specialize for 2 values? Shouldn't those be ranges instead?

That looks questionable indeed, probably some left-over from when I tried to fine tune the map type. 

> * Can the Creator classes be combined into 1 that has parameters only? The general flow is the same, maybe the specific code can be abstracted to depend only on parameters like lake sizes etc? Or put the common parts (smooth, set HQs, add rivers, place resources) in 1 class and use plugin classes to do the specific stuff

Sure, I can do that. 

> * Don't use `new` or `delete` but `make_unique`
> 
> I'll stop for now. Don't be discouraged by this, it sounds really good what you do. I only try to hold to high implementation standards so the code stays maintainable which isn't easy with the existing code but certainly possible for new code.

Feedback is appreciated, I'm not working with C++ at all apart from this project. Most of the "best practice" related comments, I can probably go through the code again and fix them in all places, that shouldn't take long. Some are probably more effort (e.g. re-using rttr code), but I'm happy to give it a try eventually. I'll just keep the PR open for the time being, unless that's causing too much work for the CI server.

Spikeone:
@vader1986 Does this also fix the "palm trees on winter maps" issue?

Flamefire:
The mess can come from a rebase to the wrong branch or a WIP. Nothing a rebase -i can't fix.

The constructor I meant was for the set, but you are adding indices, not the values, my bad. 

Namespaces were not originally used. I added them over time. Basically: Boy scout approach: Leave the place cleaner than you found it ;)

The idea with the DescIdx stuff is: You don't have an enum for terrain types and some hardcoded switches for properties but you have a "description" for each terrain type which has those properties. This description is loaded from a lua file which allows easier adjusting and extension later (e.g. adding new terrains or mixing them in 1 map). Then you have a DescIdx which is the replacement for the enum: An index which is just an integer to get to the description. So no hardcoded values. Downside is that you can't assume anything about a terrain index anymore but need to query the description, see https://github.com/Return-To-The-Roots/s25client/blob/f5e720bc1b985d3beae3f3cdea4cd3aa4ef1713e/libs/s25main/mapGenerator/RandomConfig.cpp#L82-L83

Don't worry about rttrDefines, just remove it. If anything fails to compile include the correct header. If anything not written by you fails, tell me.

> Feedback is appreciated, I'm not working with C++ at all apart from this project.

Great! It is actually similar for me: I don't do much C++ anymore besides this and use this project to improve my skills.

Another (optional) suggestion: Const everything. All local variables that don't change can be made const. This helps when reading the code later and avoids accidentally assigning to the wrong variable

If anything is unclear, feel free to ask. As for rttr code: Check libutil, libsiedler2 (mostly the map class) and the common folder in the main repo. Just scan over it to know what is there. Some things come up at different places so its good to have single, well tested implementations of them.

As for combining sources: The general rule of 1 class per file is fine and combining multiple related things into 1 file is too. There is a tradeoff there where files get to big to be easily read and to small to easily navigate and also compile time overhead for many small files. So combine stuff that is always used together in 1 file and for the rest decide on case by case.

vader1986:
> @vader1986 Does this also fix the "palm trees on winter maps" issue?

It's not fixed yet, but I'll definitely do that within this PR. I goal is to close this PR until end of Q2 (since I expected quite a few change requests before creating it), so it will take a while. But I can commit some changes to fix it and if you like you can try out this branch. Then you can just copy over generated maps to the master version. I'll just update my ToDo list with Flamefire's comments and add this one as well, so I'll continuously work on it.

vader1986:
> The mess can come from a rebase to the wrong branch or a WIP. Nothing a rebase -i can't fix.

I'll try to update my master again, but I think I did merge the RTTR master into my origin once which messed it up. I might re-fork the repository, but will try one more thing before, because that will require deleting my current fork (and along with it the PR I believe). 

Thanks for the explanations, very helpful indeed.

Flamefire:
>  I might re-fork the repository, but will try one more thing before, because that will require deleting my current fork (and along with it the PR I believe).

Don't!   
Just to make things clear: Your setup should be that you have locally 2 remotes: "origin" pointing to rttr and "<name>" pointing your own. You can check that with `git remote -v`. Your local "master" branch should track the "origin" master. So that when you checkout your master and do `git pull` it pulls in the origin master and updates your local one so they are equal. There should not be any changes at that point (`git status`)

From this setup you can work. It is advised (by some) to do the following before opening a PR: checkout master, pull, checkout your branch, `git rebase master`. Then all your changes are on top of the current master in a nice linear history. Note that you need to force-push after a rebase: `git push --force-with-lease`. Be extra careful when doing so as it overwrites your remote branch, so check `git log` first to see if this is what you want.

Feel free to ask!

vader1986:
> Don't!
> Just to make things clear: Your setup should be that you have locally 2 remotes: "origin" pointing to rttr and "" pointing your own. You can check that with `git remote -v`. Your local "master" branch should track the "origin" master. So that when you checkout your master and do `git pull` it pulls in the origin master and updates your local one so they are equal. There should not be any changes at that point (`git status`)

`s25client timm$ git remote -v
origin	git@github.com:vader1986/s25client.git (fetch)
origin	git@github.com:vader1986/s25client.git (push)
rttr	ssh://git@github.com/Return-To-The-Roots/s25client.git (fetch)
rttr	ssh://git@github.com/Return-To-The-Roots/s25client.git (push)
s25client timm$ git status
On branch master
Your branch is up to date with 'origin/master'.

nothing to commit, working tree clean`

I don't think the setup is the problem. The actual problem is that I messed up my remote master branch long time ago and I was branching from this messed up master. 

> From this setup you can work. It is advised (by some) to do the following before opening a PR: checkout master, pull, checkout your branch, `git rebase master`. Then all your changes are on top of the current master in a nice linear history. Note that you need to force-push after a rebase: `git push --force-with-lease`. Be extra careful when doing so as it overwrites your remote branch, so check `git log` first to see if this is what you want.

Yeah, I see that how this would help if my remote master was fine.

I'm using sourcetree (usually just for cloning / visualizing history, other than that command line). So that's what it looks like on my local machine:
image

It looks wrong to me, that there's a merge from rttr/master into origin/master. The last commit ("Handle TAB do switch between ...") before the merge already appeared in the origin/master when I scroll down:

image

Will have more time tomorrow to hopefully figure out this issue.

Flamefire:
> The actual problem is that I messed up my remote master branch long time ago and I was branching from this messed up master.

That was the mistake: don't use your forks master. Just pretend it doesn't exist. Make your lokal master track rttr master and use that  (happens automatically if you clone rttr first which registers it as origin)

vader1986:
> > The actual problem is that I messed up my remote master branch long time ago and I was branching from this messed up master.
> 
> That was the mistake: don't use your forks master. Just pretend it doesn't exist. Make your lokal master track rttr master and use that (happens automatically if you clone rttr first which registers it as origin)

Thanks, I just created a new branch based on the original RTTR master and cherry-picked my commits. Therefore, I had to create a new PR, so I closed this one. Sorry for the inconvenience.
https://github.com/Return-To-The-Roots/s25client/pull/1187

Flamefire:
For the future: Instead of creating a new branch, rebase your branch with `git rebase -i`. You can select the commits you want to keep and then force push

vader1986:
> For the future: Instead of creating a new branch, rebase your branch with `git rebase -i`. You can select the commits you want to keep and then force push

Yeah sorry, makes sense now after reading about the "-i" argument.

Flamefire:
No problem. I'll probaly add a git primer to rttr to help newcomers :)

PR #1187, opened 2020-02-02
WIP: Improve random map generation

vader1986:
Same as previous PR (https://github.com/Return-To-The-Roots/s25client/pull/1186) only based on RTTR master branch (instead of messed up local master). Going to work on the review comments, feel free to add more as well (this gonna take a while anyway.


vader1986:
> but this was the correct to do it? otherwise it must be copied twice?

I tried to fix the "warning: pass by value and use std::move [modernize-pass-by-value]" but didn't really know what I was doing. Reverting it for now and reading up on it.

Flamefire:
> I tried to fix the "warning: pass by value and use std::move [modernize-pass-by-value]" but didn't really know what I was doing. Reverting it for now and reading up on it.

You only applied halve of the suggestion. Note that you can search for the warning id and find https://clang.llvm.org/extra/clang-tidy/checks/modernize-pass-by-value.html which explains it pretty well.

It boils down to: If you use `a.foo(std::vector<int>{1,2,3})` and `foo` simply copies the vector into an internal member then you have a copy. If `foo` takes the vector by instances and then moves it, then you have only 1 move which is way cheaper. Hence: `ItemFilter(std::vector<Position> remove) : remove_(std::move(remove)) {}`

Your loop can also be simplified: 

        std::vector<Position> result;
        std::copy_if(input.begin(), input.end(), std::back_inserter(result), [](const Position& pt){
            return std::find(remove_.begin(), remove_.end(), point) == remove_.end();
        });

Or if you use a `set` for the `remove_`: `return remove_.find(point) == remove_.end();`

And there is even a method in containerHelpers.h: `return !helpers::contains(remove_, point)` which works for either (but with set it can be faster if there are many elements)

stefson:
I just pulled your branch and went for a test, the default values do work, but if you tinker a bit with them the s25client binary segfaults: 

```
Thread 1 "s25client" received signal SIGABRT, Aborted.
0x00007ffff77740f1 in raise () from /lib64/libc.so.6
(gdb) bt
#0  0x00007ffff77740f1 in raise () from /lib64/libc.so.6
#1  0x00007ffff775d535 in abort () from /lib64/libc.so.6
#2  0x00007ffff7aef703 in ?? () from /usr/lib/gcc/x86_64-pc-linux-gnu/9.2.0/libstdc++.so.6
#3  0x00007ffff7b1f9c6 in ?? () from /usr/lib/gcc/x86_64-pc-linux-gnu/9.2.0/libstdc++.so.6
#4  0x00007ffff7b1fa11 in std::terminate() () from /usr/lib/gcc/x86_64-pc-linux-gnu/9.2.0/libstdc++.so.6
#5  0x00007ffff7b1fc63 in __cxa_throw () from /usr/lib/gcc/x86_64-pc-linux-gnu/9.2.0/libstdc++.so.6
#6  0x00007ffff7aef35c in ?? () from /usr/lib/gcc/x86_64-pc-linux-gnu/9.2.0/libstdc++.so.6
#7  0x00007ffff7bc460f in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_assign(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) () from /usr/lib/gcc/x86_64-pc-linux-gnu/9.2.0/libstdc++.so.6
#8  0x0000555555890f29 in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::assign (__str=..., this=0x555558396ec0)
    at /usr/lib/gcc/x86_64-pc-linux-gnu/9.2.0/include/g++-v9/bits/basic_string.h:1364
#9  std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::operator= (__str=..., this=0x555558396ec0)
    at /usr/lib/gcc/x86_64-pc-linux-gnu/9.2.0/include/g++-v9/bits/basic_string.h:696
#10 libsiedler2::ArchivItem::setName (name=..., this=0x555558396eb0)
    at /var/tmp/portage/games-strategy/s25rttr-9999-r4/work/s25rttr-9999/external/libsiedler2/include/libsiedler2/ArchivItem.h:40
#11 Map::CreateArchiv (this=0x555555d23a20) at /var/tmp/portage/games-strategy/s25rttr-9999-r4/work/s25rttr-9999/libs/s25main/randomMaps/Map.cpp:51
#12 0x000055555568b0e2 in MapGenerator::Create (filePath=..., settings=...)
    at /var/tmp/portage/games-strategy/s25rttr-9999-r4/work/s25rttr-9999/libs/s25main/randomMaps/MapGenerator.cpp:33
#13 0x000055555560cdcf in dskSelectMap::CreateRandomMap (this=this@entry=0x55555625e970)
    at /var/tmp/portage/games-strategy/s25rttr-9999-r4/work/s25rttr-9999/libs/s25main/desktops/dskSelectMap.cpp:295
#14 0x0000555555611bdf in dskSelectMap::Msg_ButtonClick (this=0x55555625e970, ctrl_id=<optimized out>)
    at /var/tmp/portage/games-strategy/s25rttr-9999-r4/work/s25rttr-9999/libs/s25main/desktops/dskSelectMap.cpp:269
#15 0x0000555555743943 in ctrlButton::Msg_LeftUp (this=0x555558356db0, mc=...) at /var/tmp/portage/games-strategy/s25rttr-9999-r4/work/s25rttr-9999/libs/s25main/Window.h:123
#16 0x0000555555715058 in Window::RelayMouseMessage (this=0x55555625e970, msg=<optimized out>, mc=...)
    at /var/tmp/portage/games-strategy/s25rttr-9999-r4/work/s25rttr-9999/libs/s25main/Window.cpp:133
#17 0x00005555555ee5f7 in WindowManager::Msg_LeftUp (this=0x555555adf5f0, mc=...) at /usr/lib/gcc/x86_64-pc-linux-gnu/9.2.0/include/g++-v9/bits/unique_ptr.h:352
#18 0x00007ffff7fc0b35 in VideoSDL2::MessageLoop (this=0x555555adf090) at /var/tmp/portage/games-strategy/s25rttr-9999-r4/work/s25rttr-9999/extras/videoDrivers/SDL2/VideoSDL2.cpp:383
#19 0x0000555555623710 in VideoDriverWrapper::Run (this=<optimized out>)
    at /var/tmp/portage/games-strategy/s25rttr-9999-r4/work/s25rttr-9999/libs/s25main/drivers/VideoDriverWrapper.cpp:295
#20 0x00005555555cf1f3 in GameManager::Run (this=0x555555acc130)
    at /var/tmp/portage/games-strategy/s25rttr-9999-r4/work/s25rttr-9999/external/libutil/libs/common/include/s25util/SingletonImp.hpp:72
#21 0x00005555555bea51 in (anonymous namespace)::RunProgram (options=...)
    at /var/tmp/portage/games-strategy/s25rttr-9999-r4/work/s25rttr-9999/external/libutil/libs/common/include/s25util/SingletonImp.hpp:72
#22 0x00005555555bed64 in main (argc=<optimized out>, argv=0x7fffffffdcb8) at /var/tmp/portage/games-strategy/s25rttr-9999-r4/work/s25rttr-9999/libs/s25client/s25client.cpp:440
(gdb) 
```

triggered by clicking on the apply button with these settings: 

2020-02-04-141540_1680x1050_scrot

vader1986:
> I just pulled your branch and went for a test, the default values do work, but if you tinker a bit with them the s25client binary segfaults:

Interesting, I don't think I changed much in the UI for the "apply" button. I'll take a look at this, really shouldn't happen. Do you remember which settings you selected? I tried quite a lot but can't reproduce it on my machine.

stefson:
The settings are those in the screenshot. It segfaults when I hit apply, and with latest push from you it segfaults with the same backtrace when clicking on `Zufällige Karte`, I mean the button which generates the map.

vader1986:
> The settings are those in the screenshot. It segfaults when I hit apply, and with latest push from you it segfaults with the same backtrace when clicking on `Zufällige Karte`, I mean the button which generates the map.

The stack trace looks like something goes wrong copying the const string reference. Also looks a bit odd having a const member in the Map class, so I replaced it by using std::string and copy it via std::move in the constructor.

stefson:
that solved the segfault, thank you! 

do you think you're able to implement something as: if there's at least one big island, or the world consists of two islands, to then allow the player to build a harbour?

vader1986:
> that solved the segfault, thank you!
> 
> do you think you're able to implement something as: if there's at least one big island, or the world consists of two islands, to then allow the player to build a harbour?

Oh okay, I thought that's already in there. Could you let me know the map type which generated a map with islands but no harbor positions?

stefson:
well, I searched a spot during game play for two hours on a 256x256 map, can this be done from within the editor?

vader1986:
> well, I searched a spot during game play for two hours on a 256x256 map, can this be done from within the editor?

Actually, I found the problem. I set the maximum number of harbors per island to the number of players. So for a 256x256 with 2 players, that is not very much I think. I'll change this!

Flamefire:
> Also looks a bit odd having a const member in the Map class, so I replaced it by using std::string and copy it via std::move in the constructor.

Nothing wrong with having a const member. The problem was having a **reference** member. What happened was: Create temporary string and pass its reference into the ctor. After the ctor ran, the temporary gets destroyed because nobody owns the instance (hence "temporary") and the map class contained a dangling reference pointing to garbage. Feel free to have a `const std::string foo` member in your class to avoid accidental changes but be also aware that such a class cannot be copy-assigned (obviously, because the member is const)

vader1986:
> well, I searched a spot during game play for two hours on a 256x256 map, can this be done from within the editor?

The harbor positioning should be fixed now.

vader1986:
@Flamefire I would like to add namespace everywhere (and move some static functions out of classes). Are there any conventions where to put the namespaces? 

For example like this?
`
...
#ifndef Continent_h__
#define Continent_h__

#include "mapGenerator/algorithm/RandomUtility.h"
...

namespace rttr {
namespace mapGenerator {

class Continent
{
...
}

}
}
#endif // Continent_h__
`

vader1986:
Also quick question (hopefully). Do you know why job "#697.7" keeps failing?
https://travis-ci.com/Return-To-The-Roots/s25client/jobs/284447525

The warning in the end looks like this - couldn't really find anything on google:
> LD:/usr/local/lib:
2696DYLD:/usr/local/lib:
2697dyld: Library not loaded: /System/Library/Frameworks/ImageIO.framework/Versions/A/Resources/libGIF.dylib
2698  Referenced from: /System/Library/Frameworks/ImageIO.framework/Versions/A/ImageIO
2699  Reason: Incompatible library version: ImageIO requires version 1.0.0 or later, but libGIF.dylib provides version 0.0.0

vader1986:
Also (sorry, last comment for today!), I looked into the existing classes. It looks to me like I could create a `World` object instead of using the custom `Map` class. Then I could probably use the existing serializer to save the map? Does that make sense? 

I think then it would be easier to replace classes like the `GridUtility` and use existing ones. Currently all those custom classes are fitted exactly my own data structures, so it will be difficult to replace only a little part.

libs/s25main/mapGenerator/objects/Harbors.cpp
```diff
@@ -36,8 +36,9 @@ std::vector<double> Harbors::NextHarborDistance(const Coastline& coastline,
                                                 const std::vector<Position>& harbors,
                                                 const MapExtent& size)
 {
-    double currentDistance = double(size.x + size.y);
-    int coastNodes = coastline.size();
+    auto currentDistance = double(size.x + size.y);
```
Flamefire: Can you avoid C-Style casts (`(type)(expr)` or `type(expr)`) please? Use `static_cast<double>(...)` etc. The current code might do more than one wants (don't ask...) and I'd like to completely disallow C-Style casts in the code base sometime in the future.

Flamefire:
> Are there any conventions where to put the namespaces?

I don't think there are any options. There is exactly 1 place to put them that is syntactically right, your example code looks like that, so yes.

> Do you know why job "#697.7" keeps failing?

Yes: Travis is broken: https://travis-ci.community/t/imageio-framwork-unusable/6938/
For now I'll just allow this test to fail, see https://github.com/Return-To-The-Roots/s25client/pull/1184/commits/7d4c8a540a852b008ba0d3463cba84b5aa2f92db

> It looks to me like I could create a World object instead of using the custom Map class. Then I could probably use the existing serializer to save the map? Does that make sense?

This is kind of correct. But what you really want is `glArchivItemMap`. `World` is the Rttr map and can be saved/loaded to the savegame format while the ArchivItemMap is more or less a direct representation of the settlers 2 file format and can be saved to/loaded from it. You could even create a convenience class for you which inherits from `glArchivItemMap` (to have the data) and `MapBase` to have the functions like getting neighbours etc. You'll need to overwrite the load and resize function: On load you'd want to tell `MapBase` the new size and on `resize` (if you allow this at all in the public part) you might want to adjust the layers. But maybe simply disallow it (put into the `private:` part a `using MapBase::Resize`)

libs/s25main/mapGenerator/Random.cpp
```diff
@@ -36,6 +36,13 @@
 #include "mapGenerator/elevation/HeightMap.h"
 #include "mapGenerator/elevation/Level.h"
 
+namespace rttr {
+namespace mapGenerator {
+
+using namespace objects;
```
Flamefire: Looks good so far. 👍 
It seems you are doing that already but just to make it explicit: `using namespace` is disallowed in headers (as it affects more than it should) and should be avoided generally. Sometimes it avoids repetitive mentioning of namespaces but often it helps making it clear what you are doing. E.g. `resources::check_placement(...)` is better than a `using namespace resources; check_placement(...)`.

libs/s25main/mapGenerator/Random.h
```diff
@@ -22,14 +22,19 @@
...
+
 class Random
 {
 private:
-    RandomUtility rnd_;
+    algorithm::RandomUtility rnd_;
 
 public:
-    Random(RandomUtility& rnd) : rnd_(rnd) {}
+    Random(algorithm::RandomUtility& rnd) : rnd_(rnd) {}
```
Flamefire: You are copying the instance. Is that on purpose? It looks wrong because if you are using the original somewhere else you'll get the same random numbers (guessing what the class does here)

libs/s25main/mapGenerator/Random.h
```diff
@@ -22,14 +22,19 @@
...
     Map* Create(const MapSettings& settings);
```
Flamefire: This is a single method of a class with a generic name `Random`. I know we programmers are bad with names but of a class `Random` I'd expect something to create random numbers with...
In this case: Get rid of the class and make it a single method `std::unique_ptr<Map> createRandomMap(RandomUtility rnd, const MapSettings& settings);` I recommend watching https://www.youtube.com/watch?v=WLDT1lDOsb4. The title says it all: "Free your functions!" ;)
This is a good example for a free function, because how is it used? You create 1 instance, then call Create once and then destroy the class, right? Why the need to store state then? If you expect the state to be kept for some reason, then go for a class. But maybe name it `RandomMapCreator` or so.

libs/s25main/mapGenerator/Random.h
```diff
@@ -22,14 +22,19 @@
 #include "mapGenerator/MapSettings.h"
```
Flamefire: Another hint: Check what includes you really need. Here you need none. Everything can be forward declared and included in the cpp file. Well if you really meant to keep a RandomUtility  instance then you need 1 include.

In general: Double check all headers for includes that you may not need and replace them by forward declarations wherever possible. Keeps compiletimes and recompilations down.

Flamefire:
Found the issue for the CI failures on OSX. Fix in https://github.com/Return-To-The-Roots/s25client/pull/1188

libs/s25main/mapGenerator/README.md
```diff
@@ -1,41 +0,0 @@
-# Random Map Generation
```
Flamefire: Feel free to keep that. If it is up to date t might help future developers :)

vader1986: Fair enough. For now I removed it because it doesn't seem to contain information which are worth the additional effort for maintaining it. I might add it later after refactoring the classes (and moving some stuff to free functions as suggested) to explain how to add new "random map types" or tweak the generation process. 

Let me know if you find any commits unreasonable, your comments in between helped quite a bit (if you have time to check ofc). The video on free functions was quite fun to watch.

libs/s25main/mapGenerator/DistanceByProperty.h
```diff
@@ -170,7 +170,7 @@ std::vector<int> DistanceByProperty(const T& args,
         
         queue.pop();
         
-        const auto& neighbors = GridNeighbors(subset[subsetIndex], size);
+        const auto neighbors = GridNeighbors(subset[subsetIndex], size);
         
         for (auto neighbor: neighbors)
```
Flamefire: Sometimes it might be worth having the type and not `auto`, e.g. here or for `neighbors`, I guess they are Position or MapPoint? Without an idea its hard to tell, so don't overuse `auto` :)

Flamefire: Similar for below `auto subsetIndex = subsetIndexMap.find(neighborIndex);`: The naming isn't great, actually this isn't the index, it is an iterator, so using `itSubsetIndex` as the name would help when reading the code. Of course one could write the iterator type but meh...

libs/s25main/mapGenerator/DistanceByProperty.h
```diff
@@ -0,0 +1,201 @@
...
+/**
+ * Computes a vector field describing the distance of each grid position (by index) to the closest position
+ * for which the evaluator returned true.
+ * @param args generic parameter which is passed to the evaluator along with the index of the grid index
+ * @param size absolute size of the current map in x- and y-direction
+ * @param evaluator function used evaluate whether or not the property is given
+ * @return distance of each grid position to closest point with the evaluated property being true.
+ */
+template<typename T>
+std::vector<int> DistanceByProperty(const T& args,
```
Flamefire: Suggestion for better design (IMO): Use `template<typename T> std::vector<int> DistanceByProperty(const MapExtent& size, T&& evaluator)`. This avoids the "old-shool" callback-with-param based approach and you can use lambdas instead as well as raw function pointers. Downside is that you need to document the call syntax. Or you can use `std::function<bool(int)>` and get rid of the template but pay for a bit of runtime cost due to a virtual call, but branch prediction should rescue some.

vader1986:
I'll be on vacation for 3 weeks so don't be surprised that there'll be no progress for a while.

libs/s25main/mapGenerator/Random.h
vader1986: Absolutely no need to keep the state. Since I'm just in the process of removing some (pretty useless) classes, I created different functions instead. 

Also the name Random was stupid indeed. Originally the name came from the fact that I named all the classes for the different random map styles according to their enum value. 

I remember you had another note on the random map styles to merge them into a single function which just takes parameters (like lake size, ...). I leans against the random map generation in Age of Empires where the user can just select different map styles rather than having to deal with lots of parameters, I quite like the idea. However, it might be nice to have the parameters available, I see the point as well, but I would rather prefer to leave it like this for now (as it's already usable and produces good results) and focus on the other review changes.

vader1986:
@Flamefire I'm just trying to figure out how to replace my custom texture type by existing code. Maybe you can help me out understanding the current approach. I'm looking into the existing `CreateDefaultTextures` and a question.

https://github.com/Return-To-The-Roots/s25client/blob/master/libs/s25main/mapGenerator/RandomConfig.cpp#L79

The following part is looking for all water textures I guess:
`
std::vector<DescIdx<TerrainDesc>> terrains =
      FindAllTerrains([](const auto& desc) { return desc.kind == TerrainKind::WATER && desc.Is(ETerrain::Shippable); });
if(!terrains.empty())
{
    for(int i = 0; i < 4; i++)
        textures.push_back(terrains[Rand(terrains.size())]);
}`
The search is more or less clear, but why are we choose exactly 4 water textures? And why random ones? 

The other code is mostly clear, I like the ordering by humidity!

vader1986:
Sorry for the formatting.

Flamefire:
This is the code for the `textures` definition in RandomConfig.h:

```
    /**
     * Mapping of elevation (= height values) to terrain. The number of elements defines the
     * maximum height value used to generate the random map.
     * For example, following statement will assign water textures to landscape with the
     * height of "0": textures[0] = TT_WATER;
     */
    std::vector<DescIdx<TerrainDesc>> textures;
```

So my guess: Height 0-3 should be (shippable)  water but it doesn't matter which. So to create more random looking maps, randomize the water textures used.

I only translated the original code to use the `DescIdx` without thinking much about it, so AFAIK this is basically what it was before but without the hardcoded indices.

vader1986:
@Flamefire 
Just working on my PR again. I noticed I can't pull latest changes (3 commits) from master due to following error message:
```
From ssh://github.com/Return-To-The-Roots/s25client
 * branch                master     -> FETCH_HEAD
Updating ebde28025..2c089af3d
Downloading data/RTTR/LSTS/GAME/AFR_ICON.LST (549 B)
Error downloading object: data/RTTR/LSTS/GAME/AFR_ICON.LST (9a854dc): Smudge error: Error downloading data/RTTR/LSTS/GAME/AFR_ICON.LST (9a854dccda5fb14df435b8185bc01a47efe3dcf080a9b8fae45f225b0b809c6e): batch response: This repository is over its data quota. Account responsible for LFS bandwidth should purchase more data packs to restore access.

Errors logged to /Users/timmhoffmeister/Repositories/s25client/.git/lfs/logs/20200412T134547.371204.log
Use `git lfs logs last` to view the log.
error: external filter 'git-lfs filter-process' failed
fatal: data/RTTR/LSTS/GAME/AFR_ICON.LST: smudge filter lfs failed
Completed with errors, see above
```

Google gave me following explanation:
https://github.com/nabla-c0d3/nassl/issues/17

Happy to pay for the 10$ for the next two month if that resolves the issue for now via gift card:
https://github.myshopify.com/products/gift-card

Just let me know the email address.

Flamefire:
Ouch -.- Thanks for the notice. Opened https://github.com/Return-To-The-Roots/s25client/issues/1204

libs/s25main/mapGenerator/objects/Harbors.cpp
vader1986: **Fixed**
I replaced all C-Style casts by using `static_cast` instead. Hopefully I didn't miss any.

libs/s25main/mapGenerator/Random.cpp
vader1986: ** Fixed**
There was no need for all the subfolders and therefore nested namespaces. I moved functions which fit together into a single file in mapGenerator folder and `rttr::mapGenerator` namespace.

libs/s25main/mapGenerator/Random.h
vader1986: **Fixed**
There was no need copying the `RandomUtility` instance - also it lead to weird behavior as you said.

libs/s25main/mapGenerator/Random.h
vader1986: I tried to improve this a bit by removing duplicate code.

libs/s25main/mapGenerator/Random.h
vader1986: Removed lots of includes which are not required.

libs/s25main/mapGenerator/DistanceByProperty.h
vader1986: **Fixed**
Good point, I updated the `DistanceByProperty ` function.

vader1986:
I removed the `TextureType` enum entirely and replaced it by using rttr type `DescIdx<TerrainDesc>`. 

There're still some ToDos I've got on my list:
- try to replace GridUtility by using MapBase class if possible
- check if all functions are covered by unit tests (pretty sure I missed some)

Anything else I missed @Flamefire ?

libs/s25main/mapGenerator/Reshaper.cpp
```diff
@@ -0,0 +1,215 @@
...
+        case Random:
+            ElevateRandom(z, size, scale);
+            break;
+            
+        default:
+            break;
+    }
+}
+
+double Reshaper::ScaleCorner(int x, int y, ElevationCorner corner, const MapExtent& size) const
```
Flamefire, line 151: Please avoid taking a `x,y` pair and take a `Position` instead. Same performance, better semantics and readability.

libs/s25main/mapGenerator/Reshaper.h
```diff
@@ -0,0 +1,139 @@
...
+#define Reshaper_h__
+
+#include "mapGenerator/HeightMap.h"
+#include "mapGenerator/RandomUtility.h"
+#include "gameTypes/MapCoordinates.h"
+
+namespace rttr {
+namespace mapGenerator {
+
+enum ElevationMode
```
Flamefire, line 28: Please use `enum class` for all enums and no prefix for the values. Avoids name clashes

libs/s25main/mapGenerator/Resources.cpp
```diff
@@ -0,0 +1,342 @@
...
+#include "mapGenerator/Resources.h"
+#include "mapGenerator/Textures.h"
+#include "mapGenerator/DistanceByProperty.h"
+
+namespace rttr {
+namespace mapGenerator {
+
+using namespace libsiedler2;
+
+bool IsHeadQuarterOrHarborPosition(const Map& map, int index)
```
Flamefire, line 27: What is this doing in Resources? There are already functions like `IsHeadQuarter` somewhere else.

Also use `contains` :)

libs/s25main/mapGenerator/TextureMapping.cpp
```diff
@@ -0,0 +1,272 @@
...
+    auto buildableTerrains = FindTerrains([](const auto& desc) {
+        return desc.kind == TerrainKind::LAND && desc.Is(ETerrain::Buildable);
+    });
+    
+    if (buildableTerrains.empty())
+    {
+        throw std::runtime_error("no buildable terrain found");
+    }
+    
+    bool swapped;
```
Flamefire, line 85: This is `std::sort`

libs/s25main/mapGenerator/Rivers.cpp
```diff
@@ -0,0 +1,279 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with Return To The Roots. If not, see <http://www.gnu.org/licenses/>.
+
+#include "mapGenerator/Rivers.h"
+
+namespace rttr {
+namespace mapGenerator {
+
+bool RiverBrush::IsAnyOf(const Textures& textures, const Texture& texture)
```
Flamefire, line 23: aka `contains`, use it :)

vader1986:
After rebasing on current master, I got following linker error on Mac OS:
```
Undefined symbols for architecture x86_64:
  "_lua_remove", referenced from:
      kaguya::class_userdata::get_metatable(lua_State*, std::type_info const&) in libgamedata.a(GameDataLoader.o)
      kaguya::class_userdata::newmetatable(lua_State*, std::type_info const&, char const*) in libgamedata.a(GameDataLoader.o)
      kaguya::class_userdata::newmetatable(lua_State*, std::type_info const&, char const*) in libgamedata.a(LuaInterfaceBase.o)
      kaguya::class_userdata::get_metatable(lua_State*, std::type_info const&) in libgamedata.a(LuaInterfaceBase.o)
      kaguya::LuaStackRef kaguya::detail::LuaTableOrUserDataImpl<kaguya::LuaRef>::getField<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > >(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) const in libgamedata.a(LuaInterfaceBase.o)
      kaguya::class_userdata::get_metatable(lua_State*, std::type_info const&) in libs25Main.a(LuaInterfaceGame.o)
      kaguya::class_userdata::newmetatable(lua_State*, std::type_info const&, char const*) in libs25Main.a(LuaInterfaceGame.o)
      ...
  "_lua_replace", referenced from:
      kaguya::Metatable::call_constructor_function(lua_State*) in libgamedata.a(GameDataLoader.o)
      kaguya::Metatable::call_constructor_function(lua_State*) in libgamedata.a(LuaInterfaceBase.o)
      kaguya::Metatable::call_constructor_function(lua_State*) in libs25Main.a(LuaInterfaceGame.o)
      kaguya::Metatable::call_constructor_function(lua_State*) in libs25Main.a(LuaInterfaceGameBase.o)
      kaguya::Metatable::call_constructor_function(lua_State*) in libs25Main.a(LuaInterfaceSettings.o)
      kaguya::Metatable::call_constructor_function(lua_State*) in libs25Main.a(LuaServerPlayer.o)
      kaguya::Metatable::call_constructor_function(lua_State*) in libs25Main.a(LuaWorld.o)
      ...
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

Any idea if this could be related?

vader1986:
Thanks for removing lfs by the way

Spikeone:
I'd really love to see this new version in 0.9 - I played some games using the old generator but the mountains only had few resources. Screenshots I saw using your changed version looked better :)

libs/s25main/mapGenerator/Map.h
```diff
@@ -18,133 +18,164 @@
...
+    
+    Features();
+    Features(Texture defaultTexture);
+};
 
-/**
- * Data type for reading, writing and generating maps.
- */
-struct Map
+class Map : public FeatureMap<Features>
```
vader1986, line 49: Hey @Flamefire if you have some time, could you have a quick look at this comment, please? I just want to avoid going the wrong direction. So currently, I try to move the index computation into a single place (by overriding MapBase class). Therefore I added a template type `FeatureMap<TFeature>` which is basically a `MapBase` which stores a single value per MapPoint (maybe ValueMap is a better name?). It will (eventually) lead to remove many classes/functions, like Brush/BrushSettings, Tile & GridUtilities - basically all which are doing weird calculation based on a wrong understanding of the geometry (https://settlers2.net was giving some useful insight what the grid is actually looking like). Does this look okay so far?

Anyway, now I had two possible approaches - either I keep references to multiple FeatureMap types (option 1) within the Map class or the Map class itself inherits from FeatureMap (as currently implemented, option 2). I personally prefer option 1) - even though 2 looked a bit nicer initially. What do you think?

vader1986:
> I'd really love to see this new version in 0.9 - I played some games using the old generator but the mountains only had few resources. Screenshots I saw using your changed version looked better :)

Thanks, always motivating to have people looking for this! It might take a while though (my personal plan is to get the PR finished until end of Q3). Not sure what's the plan when to release 0.9.

libs/s25main/mapGenerator/Map.h
Flamefire: This sounds a lot like `NodeMapBase`, have you seen that? Name might not be optimal either but it does what you want and should be equal to your current `FeatureMap`. You can still have structs as your "nodes" or single values.

Your question is 2 fold:
- Store 1 value/node and have N maps for the N features or Store N values/node in 1 map
- Inheritance or composition

The first is mainly an optimization but also allows to limit the influence of code parts. If you need to work with 1 feature in isolation (e.g. only the animals, then only the resources) then it makes sense to have N maps. It will also be faster due to caching (check "AoS vs SoA"). (S2 does this: the "layers" in the map). Just check your interfaces what is best, e.g. `adjustHeights(heightMap)` or `adjustHeights(heightAnimalsResourcesAndEverythingMap)` is easy to decide what you need, but I guess there is also `setResources(terrain, height, resources)` where individual maps are a bit bulky. But having those individual maps you can set some of them `const` and others not when passing to functions, which might be easier to read and reason about. But the `MapBase` contains also `GetPointsInRadius` and `CheckPointsInRadius` which won't really work when you have multiple maps and need to check conditions on multiple ones (e.g "texture is mountain and height is bigger than 5"), but that could be improved

I wouldn't use multiple references. Use multiple instances.

About the second part: Inheriting from multiple feature maps is not possible (well... not useful), so if you go for that, the question is answered. If you go for 1 featuremap: Decide if the map "has a" featuremap or "is-a" featuremap from a logicial PoV. It really depends how you are working with it.

Side note: The `Map_` class looks pretty much like the `libsiedler2::ArchivItemMap`, maybe you can use or wrap that or the `glArchivItemMap` (which is badly named and should be merged/moved to `libsiedler2::ArchivItemMap`)

Ok, as you see I can't give a clear recommendation. Just check the options you have, the (intended) usage and then decide. You'll learn something either way and in the end both ways work.

vader1986: Haven't seen the `NodeMapBase ` class - yeah it is exactly the same. 

I probably wasn't clear about my point, but my question was just about 1 value / node or N values / node. I'll go for 1 value / node:
* seems to me like the better choice since you can function inputs can be specific (const)
* and does not always require the entire `Map` instance
* CheckPointsInRadius` is a valid point, but I'll see how I can work around it - `GetPointsInRadius` returns MapPoints anyway (assuming all maps have the same size it would still work). 

For the `Map_` class it's going to be deleted. This was just an intermediate commit and I didn't want to break everything, so I just renamed it for now to keep things compiling.

Flamefire: About `CheckPointsInRadius`: My idea is to make that a free function which takes a variadic argument of `NodeMapBase`s and will call your functor with 1 value (ref) of each map. Only problem there is the default values, so they don't get mixed up with the variadic stuff

stefson:
Do you follow a branch which includes the upgrade to boost-1.73.0? Then I might test it and give feedback, now that I've made the upgrade.

vader1986:
> Do you follow a branch which includes the upgrade to boost-1.73.0? Then I might test it and give feedback, now that I've made the upgrade.

Sorry, just saw your comment. I'm not sure, how recent are those changes? I haven't done a rebase for a while (mostly because it's a huge pain). I should be 29 commits behind master. However, I'm rewriting some parts at the moment, so things are not really testable anymore, but I can let you know when it's back in a good state.

stefson:
yes, please ping me if you have rebased on current master.

.clang-tidy
```diff
@@ -19,7 +19,9 @@ Checks: >-
   -modernize-use-default-member-init,
   -modernize-use-emplace,
   -modernize-use-override,
+  -modernize-use-trailing-return-type,
```
vader1986, line 22: seems like something went wrong with my rebase :(

vader1986: I wonder if there's a way around this. Yesterday, when I did the interactive rebase, I had to go through > 200 commits (1-by-1) - and there were many merge conflict within every single one. This took be about 6 hours. I've got no clue what caused those conflicts, because there were no commits on master really affecting this area of the code, unless I missed something (which is very possible). 

I'll do some investigation next weekend, but right now, I don't know what to do with this. I'm using SourceTree (git client with UI) rather than command line git - mostly because when I tried to do it via command line it would take even longer to go through it, so I stopped at some point. Either way, hopefully I'll find out what went wrong then ...

vader1986: Found out quite a few issues, I definitely shouldn't use SourceTree the way I did. Might be actually easier to get it done properly via command line after all.

stefson:
Since you rebased to boost-1.73, I gave it a shot. 

The branch doesn't compile with gcc-9 on linux-x64, here is the log: [gcc-9-build.log.zip](https://github.com/Return-To-The-Roots/s25client/files/4788825/gcc-9-build.log.zip) 

clang-9 works

vader1986:
@Flamefire the current master branch crashes for me at startup:

```
Loading ""/.../Repositories/s25client/build/Debug/Contents/MacOS/share/s25rttr/S2/DATA/SOUNDDAT/SOUND.LST"": done in 2ms
Starting sound conversion: failed: Error at line 10: Could not convert copy
ERROR: Some files failed to load.
Please ensure that the Settlers 2 Gold-Edition is installed 
in the same directory as Return to the Roots.
```

Weirdly, the nightly build seems to work. Any idea what I've done wrong? I did reset all changes and double-checked that the S2 files are in the expected folder (especially the failing one).

Flamefire:
The `sound.scs` is still the old one. Make sure to either rebuild which will copy that file to the build folder or copy that manually (data folder)

vader1986:
> The `sound.scs` is still the old one. Make sure to either rebuild which will copy that file to the build folder or copy that manually (data folder)

Thanks for the quick response!

Flamefire:
A little warning: We'll be upgrading to use clang-format-9 "soon": #1265. This may cause conflicts with your changes if clang-format does format things differently that you touched here too. Rebasing on that can be tedious, but e.g. `git mergetool` with `meld` does a pretty good job resolving the conflicts automatically. I'd suggest to do a `git rebase -i master` before that is done (actually do a `git rebase master` first) and combine (`fixup/f`) commits containing fixes for previous commits with them so you have fewer commits.

Do you have an estimate when you'll be ready with this PR? If it isn't to far we can hold back the clang-format changes. Besides that it would be great to have this is the upcoming release

vader1986:
> A little warning: We'll be upgrading to use clang-format-9 "soon": #1265. This may cause conflicts with your changes if clang-format does format things differently that you touched here too. Rebasing on that can be tedious, but e.g. `git mergetool` with `meld` does a pretty good job resolving the conflicts automatically. I'd suggest to do a `git rebase -i master` before that is done (actually do a `git rebase master` first) and combine (`fixup/f`) commits containing fixes for previous commits with them so you have fewer commits.
> 
> Do you have an estimate when you'll be ready with this PR? If it isn't to far we can hold back the clang-format changes. Besides that it would be great to have this is the upcoming release

Thanks for the heads up. I don't think this will be ready for release within the next weeks, I'm very busy at work and just relocating. I'd say go ahead, the last couple of rebases were a lot easier just by not using Source Tree. I'm sure I'll find my way through ...

vader1986:
The rebase was fine, only I can't run `clangFormat` anymore from Xcode ("Unknown key 'AlignConsecutiveMacros' YAML" and "Error reading .clang-format: Invalid argument"). But that's probably a local issue (also upgraded Xcode very recently).

Flamefire:
That is because your CMake cache still has a clang-format 5. I think we need to handle this in CMake. For now: Just remove the clang-format cache entry and reconfigure or set it directly to clang-format 9

tests/s25Main/mapGenerator/testTextures.cpp
```diff
@@ -0,0 +1,238 @@
...
+template<class T_Test>
+void RunTest(T_Test test);
+
+template<class T_Test>
+void RunTest(T_Test test)
+{
+    MapExtent size(8, 8);
+    DescIdx<LandscapeDesc> landscape(1);
+    WorldDescription worldDesc;
+    loadGameData(worldDesc);
```
vader1986, line 36: I guess this does not work if S2 data is not available. I will try to populate `WorldDescription` with test data instead manually, unless you guys have a better idea?

Flamefire: It does. The game data is is in `RTTR/gamedata` in the data folder of the repo root. It doesn't load the actual graphics but only the description of them

vader1986: That's great. I guess I asked because my build fails for clang-10, so I thought this would be related:
https://travis-ci.com/github/Return-To-The-Roots/s25client/jobs/365891251

It looks a bit like this issue:
https://stackoverflow.com/questions/58867953/address-sanitizer-sometimes-misses-heap-use-after-free

I wouldn't be entirely surprised if there's a memory issue in the unit test code, but the error is a bit cryptic.

vader1986:
Thanks @stefson I removed trees from lava (and snow). I'm trying to work out the issues with harbor placements next weekend. The issues I noted are:
1) don't place harbor positions at (tiny) inland lakes
2) don't place invalid harbor positions (insufficient water nearby)

stefson:
On greenland, there's also deer spawned onto impassable ice barries (snow caps of mountains) when starting a new map, but it dies quickly within 10seconds or so and vanishes quickly. It also spawns on passable mountain terroir as I just noticed, but also vanishes within the first 300 game frames or so: 
[Random.swd.zip](https://github.com/Return-To-The-Roots/s25client/files/5114506/Random.swd.zip)

I tried to generate a 1024 x 1024 map, but I cancelled when it took over a minute. Do you have an idea of how long this is supposed to take? For me, anything below 256 x 256 takes what feels like a second, and 512 x 512 takes maybe 10 seconds. 1024 x 1024 is four time that size, but takes a really long time, doesn't it?

hoffmeistert:
I've been looking through the code how harbor positions are loaded in rttr, because I couldn't figure out why some of the random map's harbor positions are removed while loading. I found this piece of code:
```
    std::vector<bool> hasCoastAtSea(world.seas.size() + 1, false);
    bool foundCoast = false;
    for(const auto dir : helpers::EnumRange<Direction>{})
    {
        // Skip point at NW as often there is no path from it if the harbor is north of an island
        unsigned short seaId = (dir == Direction::NORTHWEST) ? 0 : world.GetSeaFromCoastalPoint(world.GetNeighbour(it->pos, dir));
        // Only 1 coastal point per sea
        if(hasCoastAtSea[seaId])
            seaId = 0;
        else
            hasCoastAtSea[seaId] = true;

        it->cps[dir].seaId = seaId;
        if(seaId)
            foundCoast = true;
    }
```
https://github.com/Return-To-The-Roots/s25client/blob/master/libs/s25main/world/MapLoader.cpp#L447

I might be wrong, but this is the way I understand it:
image

vader1986:
Sorry, logged in with the wrong account!

vader1986:
I pushed an update to remove harbor positions from (small) inner lakes @stefson .

About performance, I impressed 512 x 512 in 10 seconds? Well, guess I have no gaming PC, for me it takes about a minute or more. 1024 x 1024 takes about 15 minutes for me. There's definitely room for improvement considering performance, but that will take a major effort. I could focus on that on a separate PR maybe, unless 1024 x 1024 is really essential (it does work, but it only takes ages)? I could do some profiling when I get some more time to see what's causing the bottleneck.

stefson:
I'm sorry, but your solution for the harbour problem didn't work out: there are at least three of these inner lakes where it didn't work. One is at `275:143` , the others nearby. Saveame: 
[Random2.swd.zip](https://github.com/Return-To-The-Roots/s25client/files/5151819/Random2.swd.zip)

I'm at 7e5dc3bf7ae47b92a0a5ee4c609e5ceab3596719

From my layman terms, it seems as if you're wasting a lot of cpu cycles on bigger maps with emerging a super fragmented coastline, with hundreds of fjords per island. It's not a top priority though, I believe those 1024x1024 maps are not officially supported, but if the situation doesn't improve much near release date I'd suggest to disable 1024x1024 in the gui and thus disable it, leaving the code open for improvements later or by other users.

vader1986:
> I'm sorry, but your solution for the harbour problem didn't work out: there are at least three of these inner lakes where it didn't work. 

Forgot to apply it properly for water maps, only did that for mixed. Should be fixed in a commit after. I'll take a look at the performance as soon as I get the chance. Might have a free weekend if I'm lucky. Once I know the bottleneck it should be straight forward to improve bits & pieces.

vader1986:
Performance should be a bit better now for 1024x1024 maps (and also other map sizes). Also I removed the huge number of fjords for larger maps sizes.

vader1986:
@stefson I did some profiling and fixed some performance related issues. Still takes me about 50 seconds to build a 1024 x 1024 map (before it was > 2 minutes for me).

vader1986:
@Flamefire Finished refactoring for now I think. Hope you find your suggestions applied in the code changes.

vader1986:
@Flamefire Just rebased my branch after the pragma once changes etc.

libs/s25main/mapGenerator/RandomMap.cpp
```diff
@@ -0,0 +1,292 @@
...
+        }
+    }
+
+    void SmoothHeightMap(ValueMap<uint8_t>& z, const ValueRange<uint8_t>& range)
+    {
+        int radius = GetSmoothRadius(z.GetSize());
+        int iterations = GetSmoothIterations(z.GetSize());
+
+        Smooth(iterations, radius, z);
+        Scale(z, range.minimum, range.maximum);
```
vader1986, line 111: This doesn't really work 100%. Sometimes this leads to a map with not very well distributed z-values which finally results in a map without mountains. I'll have a think about a smarter approach here this weekend.

vader1986: After a bit of debugging I resolved this problem. Wasn't really related to smoothing/scaling at all.

vader1986:
Hey guys, any review update on this PR? Any update would be appreciated (even if it's just a statement that there's no time at the moment). Just want to make sure the changes are not entirely in vain. 

Just a note: I did try using a struct which contains the information for one node (textures, z-value, resource, ...) because it would fit better design-wise. However, performance was considerably worse. That's why I kept separate vectors for separate types - it looked like accessing neighbors was much faster this way (probably more efficient caching).

libs/s25main/mapGenerator/Algorithms.h
```diff
@@ -0,0 +1,298 @@
...
+     * @param minimum minimum value to consider
+     *
+     * @returns a value between the specified minimum and the maximum value of the map.
+     */
+    template<typename T, class T_Area>
+    unsigned Count(const ValueMap<T>& values, const T_Area& area, T minimum, T maximum)
+    {
+        unsigned valuesInRange = 0;
+
+        if(area.empty())
```
Flamefire, line 237: I'd not use that. Use 2 functions because the count of an empty are being equal to the count of the whole map feels wrong. And as you essentially switch algorithms, making it 2 functions (one not taking an area) makes more sense. This tickles further so the `WholeMap` function which returns an empty vector can go too

libs/s25main/mapGenerator/RandomUtility.h
```diff
@@ -0,0 +1,87 @@
...
+
+#pragma once
+
+#include "random/XorShift.h"
+#include "gameTypes/MapCoordinates.h"
+#include <vector>
+
+namespace rttr { namespace mapGenerator {
+
+    using UsedRNG = XorShift;
```
Flamefire, line 26: I think you could just use one of the std:: PRNGs here, or what was the reason to using that here?

libs/s25main/mapGenerator/Resources.cpp
```diff
@@ -0,0 +1,235 @@
...
+        Tree palm1(196, 0xF0);
+        Tree palm2(197, 0x30);
+        Tree fir(198, 0x30);
+        Tree oak(196, 0xB0);
+        Tree birch(196, 0x70);
+        Tree cherry(197, 0xF0);
+        Tree pine(196, 0x30);
+        Tree cypress(197, 0xB0);
+
+        if(landscapeId == 0x0) // greenland
```
Flamefire, line 41: Why is that not using the enum?

libs/s25main/mapGenerator/Resources.cpp
```diff
@@ -0,0 +1,235 @@
...
+    {
+        const auto& textures = map.textures;
+        auto& resources = map.resources;
+
+        RTTR_FOREACH_PT(MapPoint, map.size)
+        {
+            if(textures.All(pt, IsMinableMountain))
+            {
+                int randomNumber = rnd.RandomInt(1, 100);
+                int ratio = settings.ratioGold;
```
Flamefire, line 177: Wouldn't this lead to clusters of gold in 1 corner and stone in the other?

libs/s25main/mapGenerator/Textures.cpp
```diff
@@ -0,0 +1,266 @@
...
+#include "mapGenerator/Algorithms.h"
+#include "mapGenerator/TextureHelper.h"
+
+#include <algorithm>
+#include <set>
+#include <stdexcept>
+
+namespace rttr { namespace mapGenerator {
+
+    void TextureMap::Set(const MapPoint& pt, const DescIdx<TerrainDesc>& texture)
```
Flamefire, line 28: You can take DescIdx by instance/copy. It is a simple number on purpose :)

libs/s25main/mapGenerator/Triangles.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+    Triangle::Triangle(bool rsu, const Position& position, const MapExtent& size)
+        : rsu(rsu), position(MakeMapPoint(position, size))
+    {}
+
+    Triangle Triangle::Inverse() const { return Triangle(!rsu, position); }
+
+    Triangle Triangle::Inverse(const MapPoint& position) const { return Triangle(!rsu, position); }
+
+    std::vector<Triangle> GetTriangles(const MapPoint& p, const MapExtent& size, Direction direction)
```
Flamefire, line 33: Maybe make this a struct `TrianglePair` with named members? Or would that make the using code more complicated? Then maybe a std::array?

libs/s25main/mapGenerator/Triangles.cpp
```diff
@@ -0,0 +1,199 @@
...
+            //   U    1/-1 (NE)
+            //
+            //   L    0/ 0 (SE)
+            //   S   -1/ 0 (SW)
+            //   D    0/-1 (N)
+
+            switch(direction.native_value())
+            {
+                case Direction::WEST:
+                    return {Triangle(true, Position(p.x, p.y - 1), size),
```
Flamefire, line 50: Can't that whole logic use the GetNeightbour stuff? This MapGeometry is complicated enough so I'd avoid scattering it around

libs/s25main/mapGenerator/Triangles.cpp
```diff
@@ -0,0 +1,199 @@
...
+                case Direction::NORTHWEST:
+                    return {Triangle(true, Position(p.x - 1, p.y - 1), size),
+                            Triangle(false, Position(p.x - 1, p.y - 1), size)};
+            }
+        }
+
+        throw std::invalid_argument("direction not supported");
+    }
+
+    std::vector<Triangle> GetTriangles(const MapPoint& p, const MapExtent& size)
```
Flamefire, line 111: Same here: array of 6?

libs/s25main/mapGenerator/Triangles.cpp
```diff
@@ -0,0 +1,199 @@
...
+        const auto& rsu1 = MakeMapPoint(Position(p.x - !(p.y & 1), p.y - 1), size);
+        const auto& rsu2 = MakeMapPoint(Position(p.x + (p.y & 1), p.y - 1), size);
+        const auto& lsd1 = MakeMapPoint(Position(p.x - 1, p.y), size);
+        const auto& lsd2 = MakeMapPoint(Position(p.x - !(p.y & 1), p.y - 1), size);
+
+        return {Triangle(true, p),  Triangle(true, rsu1),  Triangle(true, rsu2),
+                Triangle(false, p), Triangle(false, lsd1), Triangle(false, lsd2)};
+    }
+
+    std::vector<Triangle> GetTriangleNeighbors(const Triangle& t, const MapExtent& size)
```
Flamefire, line 137: Same

libs/s25main/mapGenerator/ValueMap.h
```diff
@@ -0,0 +1,174 @@
...
+
+#include "world/NodeMapBase.h"
+#include <cmath>
+
+namespace rttr { namespace mapGenerator {
+
+    template<typename T_Value>
+    struct ValueRange
+    {
+        ValueRange(T_Value minimum, T_Value maximum) : minimum(minimum), maximum(maximum) {}
```
Flamefire, line 28: Maybe assert the min<=max?

libs/s25main/world/MapBase.cpp
```diff
@@ -100,6 +100,21 @@ MapPoint MapBase::GetNeighbour2(const MapPoint pt, unsigned dir) const
     return MakeMapPoint(::GetNeighbour2(Position(pt), dir));
 }
 
+std::vector<MapPoint> MapBase::GetNeighbours(const MapPoint& pt) const
+{
+    unsigned yplus1 = pt.y == size_.y - 1 ? 0 : pt.y + 1;
```
Flamefire, line 105: This looks wrong. What if pt.y+1==MapSize? Why not use the GetNeighbour?

libs/s25main/world/MapBase.h
```diff
@@ -52,6 +52,8 @@ class MapBase
     MapPoint GetNeighbour2(MapPoint, unsigned dir) const;
     // Convenience functions for the above function
     MapCoord GetXA(MapPoint pt, Direction dir) const;
+    // Gets all neighbors (in all directions) for given position
+    std::vector<MapPoint> GetNeighbours(const MapPoint& pt) const;
```
Flamefire, line 56: So this is GetPointsInRadius(pt, 1)?

libs/s25main/mapGenerator/ValueMap.h
```diff
@@ -0,0 +1,174 @@
...
+        {
+            double slope = 1. * (size - 1) / difference;
+            return static_cast<unsigned>(round(slope * (value - range.minimum)));
+        }
+
+        return 0u;
+    }
+
+    template<typename T_Value>
+    class ValueMap : public NodeMapBase<T_Value>
```
Flamefire, line 67: This looks like it could be free functions and NodeMapBase be final again, or what was the reason?

libs/s25main/world/NodeMapBase.h
```diff
@@ -24,14 +24,17 @@
 /// Base class for a map with the geometry of the world (MapBase)
 /// Provides nodes of a given type with accessors to the node accepting flat indices or points
 template<typename T_Node>
-class NodeMapBase final : public MapBase
+class NodeMapBase : public MapBase
```
Flamefire, line 27: See above. The final makes also the resize final and devirtualize it

libs/s25main/world/NodeMapBase.h
```diff
@@ -61,3 +64,11 @@ void NodeMapBase<T_Node>::Resize(const MapExtent& newSize)
     nodes.clear();
     nodes.resize(prodOfComponents(newSize));
 }
+
+template<typename T_Node>
+void NodeMapBase<T_Node>::Resize(const MapExtent& newSize, const T_Node& defaultValue)
```
Flamefire, line 69: Hm... This looks like it might invalidate assumptions and can only be valid if this class is final as otherwise inheritants could overwrite Resize but don't get called for this

Flamefire (changes requested):
Some comments. This PR is to huge to really grasp it. Just see if there are ways to simplify and maybe just merge and improve over time

libs/s25main/mapGenerator/RandomUtility.h
```diff
@@ -64,14 +64,19 @@ namespace rttr { namespace mapGenerator {
...
          * @param min minimum value for the random number
          * @param max maximum value for the random number
          *
          * @return a random number between min and max (inclusive).
          */
-        int RandomInt(int min, int max);
+        template<typename T>
+        T RandomValue(T min, T max)
+        {
+            std::uniform_int_distribution<T> distr(min, max);
```
Flamefire, line 77: Please add a static_assert that T is an integer/not a floating point type

libs/s25main/mapGenerator/Algorithms.h
vader1986: Fixed (removed "WholeMap" and split up into separate functions)

libs/s25main/mapGenerator/Algorithms.h
```diff
@@ -0,0 +1,298 @@
...
+            RTTR_FOREACH_PT(MapPoint, values.GetSize())
+            {
+                if(values[pt] >= minimum && values[pt] <= maximum)
+                {
+                    valuesInRange++;
+                }
+            }
+        } else
+        {
+            for(const MapPoint& pt : area)
```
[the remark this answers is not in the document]
vader1986: Fixed

libs/s25main/mapGenerator/Algorithms.h
```diff
@@ -0,0 +1,298 @@
...
+    T LimitFor(const ValueMap<T>& values, const T_Area& area, double coverage, T minimum)
+    {
+        if(coverage < 0 || coverage > 1)
+        {
+            throw std::invalid_argument("coverage must be between 0 and 1");
+        }
+
+        const T maximum = area.empty() ? values.GetMaximum() : values.GetMaximum(area);
+
+        if(minimum == maximum)
```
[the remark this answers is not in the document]
vader1986: Fixed, it is definitely possible (shouldn't occur in the current usage though).

libs/s25main/mapGenerator/HeadQuarters.h
```diff
@@ -0,0 +1,116 @@
...
+    {
+        auto isHeadQuarter = [&map](const MapPoint& pt) {
+            return map.objectInfos[pt] == libsiedler2::OI_HeadquarterMask;
+        };
+
+        auto isObstacle = [&map](MapPoint point) {
+            return map.textures.Any(point, [](auto t) { return !t.Is(ETerrain::Buildable); });
+        };
+
+        auto quality = Distances(map.size, isHeadQuarter);
```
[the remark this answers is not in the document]
vader1986: Yeah, I see your point. Renamed the variable & added the type and a comment.

libs/s25main/mapGenerator/HeadQuarters.h
```diff
@@ -0,0 +1,116 @@
...
+        const auto& obstacleDistance = Distances(map.size, isObstacle);
+
+        auto minDistance = std::max(std::min(obstacleDistance.GetMaximum(area), 4u), 2u);
```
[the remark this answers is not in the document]
vader1986: Fixed

libs/s25main/mapGenerator/Islands.cpp
```diff
@@ -0,0 +1,115 @@
...
+        const MapPoint mapOrigin(0, 0);
+        const MapPoint mapCenter(map.size.x / 2, map.size.y / 2);
+        const unsigned mapMaxDist = distances.CalcDistance(mapCenter, mapOrigin);
+
+        auto compare = [&distances, mapMaxDist, center](const MapPoint& rhs, const MapPoint& lhs) {
+            // computes prefered extension points for the island by considering distance to
+            // center of the island and maximizing distance to other land
+
+            const auto ratioRhs = static_cast<double>(distances[rhs]) / mapMaxDist
+                                  + (1. - static_cast<double>(distances.CalcDistance(center, rhs)) / mapMaxDist);
```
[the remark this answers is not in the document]
vader1986: Absolutely right. I initially came up with this approach but didn't really simplify it. Fixed now.

libs/s25main/mapGenerator/Islands.cpp
```diff
@@ -0,0 +1,115 @@
...
+                                        [&island](const MapPoint& pt) { return !helpers::contains(island, pt); });
+        };
+
+        const auto& coastDistance = Distances(map.size, island, 0, isCoast);
+
+        for(const MapPoint& pt : island)
+        {
+            const int base = map.height.minimum + coastDistance[pt];
+            map.z[pt] =
+              std::min(static_cast<unsigned>(rnd.RandomInt(base, base + 1)), static_cast<unsigned>(map.height.maximum));
```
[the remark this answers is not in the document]
vader1986: Fixed

libs/s25main/mapGenerator/Map.cpp
```diff
@@ -14,75 +14,121 @@
...
+        // First 7 players go into the header
+        for(unsigned i = 0; i < 7; i++)
+        {
+            header->setPlayerHQ(i, hqPositions_[i].x, hqPositions_[i].y);
+        }
+
+        std::vector<uint8_t> z(numNodes);
+        std::vector<uint8_t> rsu(numNodes);
+        std::vector<uint8_t> lsd(numNodes);
+        std::vector<uint8_t> road(numNodes, 0x0);
```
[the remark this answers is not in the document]
vader1986: Fixed all

libs/s25main/mapGenerator/Map.cpp
```diff
@@ -14,75 +14,121 @@
...
+            {
+                rsu[t.position.x + t.position.y * size.y] |= libsiedler2::HARBOR_MASK;
+            } else
+            {
+                lsd[t.position.x + t.position.y * size.y] |= libsiedler2::HARBOR_MASK;
+            }
+        }
+
+        map->push(std::move(header));
+        map->push(std::make_unique<libsiedler2::ArchivItem_Raw>(z));
```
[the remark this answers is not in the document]
vader1986: Fixed, does work.

libs/s25main/mapGenerator/MapSettings.cpp
```diff
@@ -15,23 +15,25 @@
...
-    // Even map size
-    size.x &= ~1;
-    size.y &= ~1;
-    size = elMax(size, MapExtent::all(16));
-    if(ratioGold + ratioIron + ratioCoal + ratioGranite == 0)
-        ratioCoal = 1;
-    minPlayerRadius = std::min(1. - 1e-5, std::max(minPlayerRadius, 0.01));
-    maxPlayerRadius = std::min(minPlayerRadius + 1e-5, 1.);
-}
+    void MapSettings::Validate()
```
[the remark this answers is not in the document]
vader1986: Fixed

libs/s25main/mapGenerator/MapSettings.cpp
```diff
@@ -15,23 +15,25 @@
...
+    {
+        if(name.empty())
+            name = "Random";
+        if(author.empty())
+            author = "AutoGenerated";
+
+        numPlayers = std::min(MAX_PLAYERS, std::max(numPlayers, 1u));
```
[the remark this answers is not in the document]
vader1986: Fixed

libs/s25main/mapGenerator/RandomMap.cpp
```diff
@@ -0,0 +1,313 @@
...
+#include "libsiedler2/libsiedler2.h"
+
+#include <iostream>
+#include <stdexcept>
+
+namespace rttr { namespace mapGenerator {
+
+    unsigned GetMaximumHeight(const MapExtent& size)
+    {
+        switch(size.x + size.y)
```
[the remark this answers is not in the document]
vader1986: Agreed, fixed.

libs/s25main/mapGenerator/RandomMap.cpp
```diff
@@ -0,0 +1,313 @@
...
+
+    RandomMap::RandomMap(RandomUtility& rnd, Map& map) : rnd_(rnd), map_(map), texturizer_(map.z, map.textures) {}
+
+    void RandomMap::Create(const MapSettings& settings)
+    {
+        auto defaultHeight = map_.height.minimum + map_.height.GetDifference() / 2;
+
+        map_.z.Resize(settings.size, defaultHeight);
+
+        if(settings.style == MapStyle::Water)
```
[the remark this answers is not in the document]
vader1986: Fixed

libs/s25main/mapGenerator/RandomUtility.h
vader1986: I just used what it was before. The original reason not to use std random number generators was that they could produce different numbers depending on operating system. So I used the existing class instead, as far as I remember.

libs/s25main/mapGenerator/Resources.cpp
vader1986: done

libs/s25main/mapGenerator/Resources.cpp
vader1986: fixed

libs/s25main/mapGenerator/Textures.cpp
vader1986: Ah okay, that makes sense. Just replaced all `const DescIdx<TerrainDesc>&` by `DescIdx<TerrainDesc>`.

libs/s25main/world/MapBase.cpp
vader1986: I used `GetNeighbour` before but the performance is much worse for larger map sizes.
This specific case should work, for example:
Assuming `size_.y = 16` and `pt.y = 15`. Then `pt.y + 1 = 16 (= size_.y)`. Therefore `(pt.y + 1) % size.y = 0`. `pt.y` cannot be `16` assuming `MapPoint` is always within the range `[0, size.y - 1]`. Or maybe I didn't get the point?

libs/s25main/world/MapBase.h
vader1986: Yes, I tried that as well. But again, performance was much worse and I used it quite a few times looping over the whole map. Also tried to pre-calculate neighbors once for the whole map. But then the memory access seems to slow things down.

libs/s25main/mapGenerator/ValueMap.h
vader1986: Yes agreed & fixed!

libs/s25main/world/NodeMapBase.h
vader1986: Right, I changed it back to `final`. Just added a `Map` member function for computing max/min values using standard functions (looping over the vector instead the map).

libs/s25main/mapGenerator/RandomUtility.h
vader1986: done

vader1986:
> Some comments. This PR is to huge to really grasp it. Just see if there are ways to simplify and maybe just merge and improve over time

I thought it was still an improvement to the code before. But I'll take another look through the code changes. Maybe I can simplify some things, especially around `Textures`, I think the code is not straight forward. Also, I'll look through the `Triangle` stuff, as I left some comment "unfixed" so far.

libs/s25main/mapGenerator/ValueMap.h
vader1986: done

libs/s25main/world/MapBase.h
Flamefire: Then maybe we should use your method in other places as well. Just checked it, looks indeed good and performant

Flamefire:
Just seen your harbor question from https://github.com/Return-To-The-Roots/s25client/pull/1187#issuecomment-683323981

- lsd textures are ignored, because a mark on the rsu already marks that point, so no need to check lsd
- and no the code does something different. In short: It checks each potential harbor points and assigns seas to coastal points. Every neighbor (all 6) is a potentially coastal point (i.e. where ships can park), the code allows exactly 1 coastal point per sea (a harbor could be at 2 seas) and assigns the sea id to the coastal point. If any coastal point was found, the harbor is valid (not shown but thats `foundCoast`) The NW invalid thing is: The point NW of the harbor (not of a sea point as in the drawing) should be ignored, this is because the harbor is right next to it and may block a path. So basically a workaround for when ships with soldiers land (spawn at the coastal point) but then they can't move because the path is blocked.

I think all issues I raised are resolved? Or anything still open?

vader1986:
> Just seen your harbor question from [#1187 (comment)](https://github.com/Return-To-The-Roots/s25client/pull/1187#issuecomment-683323981)
> 
> * lsd textures are ignored, because a mark on the rsu already marks that point, so no need to check lsd
> * and no the code does something different. In short: It checks each potential harbor points and assigns seas to coastal points. Every neighbor (all 6) is a potentially coastal point (i.e. where ships can park), the code allows exactly 1 coastal point per sea (a harbor could be at 2 seas) and assigns the sea id to the coastal point. If any coastal point was found, the harbor is valid (not shown but thats `foundCoast`) The NW invalid thing is: The point NW of the harbor (not of a sea point as in the drawing) should be ignored, this is because the harbor is right next to it and may block a path. So basically a workaround for when ships with soldiers land (spawn at the coastal point) but then they can't move because the path is blocked.

Thanks for the explanation. Makes more sense now. 

> I think all issues I raised are resolved? Or anything still open?

I think so. I'm still looking for some improvements here and there. But it's probably better to do them in separate PRs, because it's easier to review smaller updates (unless you go through the single commits).

Flamefire:
> I'm still looking for some improvements here and there. But it's probably better to do them in separate PRs, because it's easier to review smaller updates (unless you go through the single commits).

Exactly. And I have a PR almost ready which changes a bit which could be touched and hence broken by this: #1295

So any chance you can make the PR ready for merging (passes CI) soon? Appveyor shows a missing `<random>` include and the static-analysis thing is something I broke, but is only a naming of the parameter (mismatch between header and cpp file) which you can simply fix here

vader1986:
The CI failures are fixed. Also I did a rebase.

stefson:
Do you need some runtime testing?

vader1986:
> Do you need some runtime testing?

Definitely wouldn't hurt. I went through the different map sizes, map types & landscape types a couple of days ago and it looked fine. But it's easy to miss something, especially since I didn't play all the maps (just opened and checked for obvious problems).

Flamefire (approved):
Thanks!
The remaining issues can be fixed by small follow-ups

JonathanSteinbuch:
Hi, @vader1986,

I just tried this one out. Some observations (Since I noticed them, I wanted to write them down, but feel free to ignore this)

- The coastlines look very beautiful.
- The rivers look wonderful.
- I really like the design of the different height zones, especially regarding different tree types growing at different heights.
- Regarding performance for large sizes: It seems to me that using "helpers::contains" or "helpers::contains_if" inside lambda functions that are passed to the Distances function might be the main culprit. I think that it would be much less expensive computationally to just pass a container of points that would evaluate to true to the Distances function. Especially since in the cases where helpers::contains is used the data is stored in a container before already.
- I think it would be more interesting if not everything (landscape, trees) only depended on distance to the coastline. Maybe have another round of focuse points (maybe even of different types or weights) to adjust the more fine grained aspects of the landscape
- Especially I would like the density of trees and/or rocks to vary so that there are relatively clear areas and more forested areas.
- I would prefer for there to be more even ground where you can build large buildings (maybe the slope function being less steep at some grass heights)
- With the large mountain ranges that are often generated it would be useful for there to be some way to get to their interior areas.
- What would be awesome would be more options in the map generator window. For example sliders determining the density and (average) weights of focus points for example, but also for some of the other constants that are used.

Overall good work, nice project!

vader1986:
> * Regarding performance for large sizes: It seems to me that using "helpers::contains" or "helpers::contains_if" inside lambda functions that are passed to the Distances function might be the main culprit. I think that it would be much less expensive computationally to just pass a container of points that would evaluate to true to the Distances function. Especially since in the cases where helpers::contains is used the data is stored in a container before already.

I tried quite a bit, like pre-calcuating neighbors etc to improve performance. In regards to your suggestion, to create a container with points (instead of the lambda function which returns true/false for a point), it would still require to iterate over the entire map to compute this container. Unless I missed some cases where this contains was already available, I'll definitely add it to my ToDo list to check that!

> * I think it would be more interesting if not everything (landscape, trees) only depended on distance to the coastline. Maybe have another round of focuse points (maybe even of different types or weights) to adjust the more fine grained aspects of the landscape

Yes I noticed that things are looking very boring on larger maps. What I had in mind, was a recursive approach for adding more details (hills, mountains, maybe swamps) within the non-water areas of the map. But I'll probably raise an issue in GitHub first to discuss this more in detail before starting to implement.

> * Especially I would like the density of trees and/or rocks to vary so that there are relatively clear areas and more forested areas.

That's a great idea and probably not too difficult to implement!

> * I would prefer for there to be more even ground where you can build large buildings (maybe the slope function being less steep at some grass heights)

Noted. 

> * With the large mountain ranges that are often generated it would be useful for there to be some way to get to their interior areas.

Yeah, that's a bit related to the other point you mentioned (at least the solution I've got in mind). 

> * What would be awesome would be more options in the map generator window. For example sliders determining the density and (average) weights of focus points for example, but also for some of the other constants that are used.

Noted as well. 

> Overall good work, nice project!

Thanks a lot for the feedback! :-)

JonathanSteinbuch:
> I tried quite a bit, like pre-calcuating neighbors etc to improve performance. In regards to your suggestion, to create a container with points (instead of the lambda function which returns true/false for a point), it would still require to iterate over the entire map to compute this container. Unless I missed some cases where this contains was already available, I'll definitely add it to my ToDo list to check that!

To expand and clarify here. I don't think the problem lies with lambda functions per se or iterating over the whole map, but with the unnecessary uses of helpers::contains and contains_if for every map point.

For example in the Restructure function in Terrain.cpp you have
```
        auto inFocusArea = [&focusArea](const MapPoint& pt) { return helpers::contains(focusArea, pt); };

        auto& z = map.z;
        auto distances = Distances(size, inFocusArea);
```
where focusArea is a vector of MapPoints while in the Distances algorithm you have 
```
        RTTR_FOREACH_PT(MapPoint, size)
        {
            if(evaluator(pt))
            {
                distances[pt] = 0;
                queue.push(pt);
            }
        }
```
This at least could be simplified by just using a for each loop over the focusArea (as for example in the other Distances algorithm).

For another example I also noticed it a bit less straight forward in AddObjects of Resources.cpp. I would try to remove the contains_if by iterating over the forbidden points and adding all elements in a radius to a forbidden area container. Then iterating over the whole map to add snow-or-lava points should be fine, I feel.

PR #1189, opened 2020-02-12
Add support for SDL2_mixer using the same driver code as SDL_mixer

chewi:
Only one line actually required an #if macro to support both versions.

Supporting both versions in AudioSDL.cpp was easy. Supporting both versions in CMake was much trickier. This isn't because the same source files are being used but because of how CMake stores the variables resulting from `find_package(SDL_mixer)`. Perhaps you can think of a better approach.

### DIFF
--- a/cmake/Modules/FindSDL_mixer.cmake
+++ b/cmake/Modules/FindSDL_mixer.cmake
@@ -11,7 +11,7 @@ This module defines:
 
 ::
 
-  SDL_Mixer::SDL_Mixer target to link to
+  SDL{1,2}_Mixer::SDL_Mixer target to link to
   SDL_MIXER_FOUND, if false, do not try to link against
   SDL_MIXER_VERSION - human-readable string containing the
                              version of SDL_mixer
@@ -26,13 +26,16 @@ module, but with modifications to recognize OS X frameworks and
 additional Unix paths (FreeBSD, etc).
 #]=======================================================================]
 
+unset(SDL_MIXER_LIBRARY CACHE)
+unset(SDL_MIXER_INCLUDE_DIR CACHE)
+
 if(CMAKE_SIZEOF_VOID_P EQUAL 8)
   set(VC_LIB_PATH_SUFFIX lib/x64)
 else()
   set(VC_LIB_PATH_SUFFIX lib/x86)
 endif()
 
-if(PACKAGE_FIND_VERSION_MAJOR EQUAL "2")
+if(SDL_mixer_FIND_VERSION VERSION_GREATER_EQUAL 2)
   set(libName SDL2_mixer)
   set(pathSuffixes SDL2 include/SDL2 include)
 else()
@@ -80,7 +83,6 @@ if(SDL_MIXER_INCLUDE_DIR AND EXISTS "${SDL_MIXER_INCLUDE_DIR}/SDL_mixer.h")
   unset(SDL_MIXER_VERSION_MAJOR_LINE)
   unset(SDL_MIXER_VERSION_MINOR_LINE)
   unset(SDL_MIXER_VERSION_PATCH_LINE)
-  unset(SDL_MIXER_VERSION_MAJOR)
   unset(SDL_MIXER_VERSION_MINOR)
   unset(SDL_MIXER_VERSION_PATCH)
 else()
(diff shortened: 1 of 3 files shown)


cmake/Modules/FindSDL_mixer.cmake
```diff
@@ -26,13 +26,16 @@ module, but with modifications to recognize OS X frameworks and
 additional Unix paths (FreeBSD, etc).
 #]=======================================================================]
 
+unset(SDL_MIXER_LIBRARY CACHE)
```
Flamefire: Don't do that. This will search again on every configure
I suggest using a variable for holding the prefix or whole varname (SDL_MIXER_LIBRARY vs SDL2_MIXER_LIBRARY)

cmake/Modules/FindSDL_mixer.cmake
```diff
@@ -104,21 +106,22 @@ find_package_handle_standard_args(SDL_mixer
                                   REQUIRED_VARS SDL_MIXER_LIBRARY SDL_MIXER_INCLUDE_DIR ${additionalVars}
                                   VERSION_VAR SDL_MIXER_VERSION)
 
-if(NOT TARGET SDL_mixer::SDL_mixer)
-  add_library(SDL_mixer-SDL_mixer INTERFACE)
-  add_library(SDL_mixer::SDL_mixer ALIAS SDL_mixer-SDL_mixer)
-  target_link_libraries(SDL_mixer-SDL_mixer INTERFACE ${SDL_MIXER_LIBRARY})
-  target_include_directories(SDL_mixer-SDL_mixer SYSTEM INTERFACE ${SDL_MIXER_INCLUDE_DIR})
+if(NOT TARGET SDL${SDL_MIXER_VERSION_MAJOR}_mixer::SDL_mixer)
```
Flamefire: needs to be wrapped inside an `if(SDL_mixer_FOUND)`

cmake/Modules/FindSDL_mixer.cmake
```diff
@@ -104,21 +106,22 @@ find_package_handle_standard_args(SDL_mixer
...
+  add_library(SDL${SDL_MIXER_VERSION_MAJOR}_mixer-SDL_mixer INTERFACE)
```
Flamefire: Use a variable for `SDL${SDL_MIXER_VERSION_MAJOR}_mixer-SDL_mixer` to avoid the repetition. It hard to read...

chewi:
> Is it really possible to use SDL1 with SDL_mixer2 or SDL2 with SDL_mixer?

Well it seems to load both SDL1 and SDL2 at the same time, one for the video and one for the audio. I wouldn't have expected it to work either but also figured this is what you have been doing. I did end up checking just to make sure and indeed it does.

> SDL2 seems to work stable. Maybe we drop SDL1 and SDL_mixer1 completely and use the 2 for everything?

I almost suggested this myself. While there a lot of old commercial games and dormant open source games still tied to SDL1, the actively maintained games switched to using SDL2 exclusively years ago. I'll hold off fixing up CMake until you've made a decision on this.

chewi:
I forgot that s25edit still uses SDL1 so perhaps that's an issue for you. I suspect it wouldn't be hard to migrate it though.

Flamefire:
@Flow86 agreed to drop sdl1? I suggested to drop the driver dll approach completely some time ago which would make it also easier (ubsan issues...) and dropping the abstraction over sdl/winapi could give a bit more performance, but having the driver in the exe should already improve it a bit due to less duplicated machine code

Flow86:
we could drop the sdl1, but we should replace it everywhere (i.e also on the editor)

chewi:
Well the good news is [SGE has been ported to SDL2](https://github.com/flibitijibibo/libSGE). I had a quick stab at making this work but I'm not quite there yet. You could probably do it quicker.

Flamefire:
Well I changed the SGE code quite a lot to squeeze out more performance, also the version used was modified already, so we can't just update from upstream :(

But I'll take a look

chewi:
On February 18, 2020 2:03:00 PM GMT, Alexander Grund <notifications@github.com> wrote:
>Well I changed the SGE code quite a lot to squeeze out more
>performance, also the version used was modified already, so we can't
>just update from upstream :(
>
>But I'll take a look

I had noticed that you'd made some performance improvements but I was rather optimistically hoping it would just be better under SDL2 anyway. 😁


-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Flamefire:
I have ported the editor to SDL2 and testing now, wasn't exactly easy as SGE has parts which simply won't work with SDL2. I deleted them similar to what the mentioned "port" did: They commented the code and added an assert -.-

PR will follow, closing  this. But thanks for the fixes to FindSDL_mixer :+1:

PR #1191, opened 2020-02-17
Addon changes

### LINKED ISSUE #1001: Default button in Addon Settings not working

Flamefire:
Refactor the addons to fix #1001 in a more complete way than #1110 although it uses the same approach: Create all GUI controls on start of the iwAddons window. Also (same as there) the shifting of the addons is done in the window, not the addon classes). Thanks @iceslab !

Reasoning: The pagesize of the scrollbar assumes a height of 30. Factored that out into a constant.

@Flow86: After the (minor) refactoring of the Addon classes themselves I don't see why we really need to have different classes. My suggestion:
- Only have the `Addon` class and instances of it
- (Opt) Have Addon store its current value (to avoid the stuff with AddonWithStatus) or make a struct `AddonData` (containing the description) and `Addon` containing the AddonData, status and accessors with current interface.
- Store an enum whether this is a bool or list addon (others aren't really possible)
- Create the GUI in iwAddons based on that enum. we have implicit restrictions on height and width anyway so Addon::createGui was never really flexible.
- Create a factory for creating a vector of all addons (currently in GlobalGameSettings) instead of the addons.h to have all addon stuff inside the addon folder. No one needs to know the types that exist, only the AddonId and the status
- Also cut down on the number of required steps to add an addon, currently:
  - Add new (really small file) with just some data
  - Add new AddonId with implicit author
  - Add to addons.h
  - Add to GlobalGameSettings
- Better:
  - Add all "Addons" (basically their data) into a single file which serves as the factory (probably a class)
  - So now you need to add a new AddonId and add a call to `registerAddon(AddonId, data)`, done
  - I'd also add the authors name to the data (although it wouldn't be currently used)
  - Downside: File becomes quite big, but that should be faster to compile than 40 small files. 
- Finally: Add an assert/exception when `isEnabled` is called on a non-bool addon and rework the handling of missing addons. IMO this shouldn't ever happen but currently some fallback returning a default or arbitrary value is implemented.

### DIFF
--- a/libs/s25main/Window.h
+++ b/libs/s25main/Window.h
@@ -233,7 +233,7 @@ public:
     virtual void Msg_ProgressChange(unsigned /*ctrl_id*/, unsigned short /*position*/) {}
     virtual void Msg_ScrollChange(unsigned /*ctrl_id*/, unsigned short /*position*/) {}
     virtual void Msg_ScrollShow(unsigned /*ctrl_id*/, bool /*visible*/) {}
-    virtual void Msg_OptionGroupChange(unsigned /*ctrl_id*/, int /*selection*/) {}
+    virtual void Msg_OptionGroupChange(unsigned /*ctrl_id*/, unsigned /*selection*/) {}
     virtual void Msg_Timer(unsigned /*ctrl_id*/) {}
     virtual void Msg_TableSelectItem(unsigned /*ctrl_id*/, int /*selection*/) {}
     virtual void Msg_TableChooseItem(unsigned /*ctrl_id*/, unsigned /*selection*/) {}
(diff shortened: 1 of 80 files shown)


libs/s25main/Window.h
```diff
@@ -253,7 +253,7 @@ class Window
     virtual void Msg_Group_CheckboxChange(unsigned /*group_id*/, unsigned /*ctrl_id*/, bool /*checked*/) {}
     virtual void Msg_Group_ProgressChange(unsigned /*group_id*/, unsigned /*ctrl_id*/, unsigned short /*position*/) {}
     virtual void Msg_Group_ScrollShow(unsigned /*group_id*/, unsigned /*ctrl_id*/, bool /*visible*/) {}
-    virtual void Msg_Group_OptionGroupChange(unsigned /*group_id*/, unsigned /*ctrl_id*/, int /*selection*/) {}
+    virtual void Msg_Group_OptionGroupChange(unsigned /*group_id*/, unsigned /*ctrl_id*/, unsigned /*selection*/) {}
```
Flow86: wasn't it signed because of "-1" as "no selection"?

Flamefire: The docu once said 0xFFFF was "no selection". As you see it was really confusing. There now is no "no selection" as this was a strange corner case anyway... And as "selection" is any ID there is no "invalid marker" left. I started using optional<unsigned> but that got quite messy either for no current use case. Hence the only way to get "no selection" is if you create an option group with no buttons (or don't initially call SetSelection) and then call GetSelection. But the callback can never return "no selection" as it is a callback called when the selection changes which only happens if something else gets selected

Flow86: oh okay. shouldnt the rest be changed too then?

Flamefire: What rest? I changed all `OptionGroupChange` messages

Flamefire:
@Flow86 anything holding this up?

PR #1210, opened 2020-04-18
Use Boost.Nowide

Flamefire:
I took over maintenance of Boost.Nowide to fix outstanding issues and get it finally included into Boost 1.73. This will help fix possibly undetected issues.

For starters:
- Faster startup times due to fixed buffering in the Boost.Nowide filestreams (it used to use a 1Byte buffer)
- Some minor quirks fixed
- UTF converter included. It is basically the one from Boost.Locale but now included in Boost.Nowide and outstanding issues (mostly annoying warnings) fixed.

This also allowed me to remove utf8cpp which is nice (iterator interface) but turns out the basic conversions in Boost.Nowide are enough for 95% of the cases. Hence another dependency less.

There are 2 way to get Boost.Nowide: Either use Boost 1.73 which has it included or by installing it using any other Boost version. The current search finds both with a fallback via FetchContent (very convenient, IMO better than recursive submodules) and there is an option to require the system version. Hope that works? @stefson 

This also includes the fix for the issue reported by @Spikeone :
>C:\Program Files (x86)\Windows Kits\8.1\Include\um\shlobj.h(1054): warning C4091: "typedef ": Ignoriert auf der linken Seite von "tagGPFIDL_FLAGS", wenn keine Variable deklariert wurde

This is a bug in the MSVC headers... Not sure why it generates a warning though, as these are system headers, but well, now the warning is disabled for this file.


stefson:
are we talking about these headers and libs? 

```
/usr/include/boost/nowide
/usr/include/boost/nowide/args.hpp
/usr/include/boost/nowide/config.hpp
/usr/include/boost/nowide/convert.hpp
/usr/include/boost/nowide/cstdio.hpp
/usr/include/boost/nowide/cstdlib.hpp
/usr/include/boost/nowide/detail
/usr/include/boost/nowide/detail/convert.hpp
/usr/include/boost/nowide/detail/utf.hpp
/usr/include/boost/nowide/filebuf.hpp
/usr/include/boost/nowide/filesystem.hpp
/usr/include/boost/nowide/fstream.hpp
/usr/include/boost/nowide/iostream.hpp
/usr/include/boost/nowide/replacement.hpp
/usr/include/boost/nowide/stackstring.hpp
/usr/include/boost/nowide/utf8_codecvt.hpp
/usr/include/boost/nowide/windows.hpp
/usr/lib64/libboost_nowide-mt.so
/usr/lib64/libboost_nowide.so
/usr/lib64/libboost_nowide.so.1
/usr/lib64/libboost_nowide.so.1.73
/usr/lib64/libboost_nowide.so.1.73.0
```

I just build them from 1.73.0_beta1

stefson:
regarding git clone, something is still off: 

```
git clone --recursive https://github.com/Flamefire/s25client.git
Klone nach 's25client' ...
remote: Enumerating objects: 107, done.
remote: Counting objects: 100% (107/107), done.
remote: Compressing objects: 100% (83/83), done.
remote: Total 66244 (delta 29), reused 48 (delta 24), pack-reused 66137
Empfange Objekte: 100% (66244/66244), 188.90 MiB | 10.78 MiB/s, Fertig.
Löse Unterschiede auf: 100% (49548/49548), Fertig.
git-lfs filter-process: git-lfs: Kommando nicht gefunden.
fatal: Die Gegenseite hat unerwartet abgebrochen.
warning: Klonen erfolgreich, Auschecken ist aber fehlgeschlagen.
Sie können mit 'git status' prüfen, was ausgecheckt worden ist
und das Auschecken mit 'git restore --source=HEAD :/' erneut versuchen.
```
you might have to rebase?

Flamefire:
> are we talking about these headers and libs?

As you build them from Boost 1.73, yes

> regarding git clone, something is still off:

"git-lfs: Kommando nicht gefunden." --> Seems you have uninstalled the git-lfs binary but not the hooks. Try `git lfs uninstall` in the repo

stefson:
well, it's correct that I have uninstalled git-lfs, but that error is from a fresh clone of your fork

anyway, checking out with the package manager was less of a problem, but it runs into various errors with lib detection. I'll add the compressed log for you: [build.log.zip](https://github.com/Return-To-The-Roots/s25client/files/4497312/build.log.zip)

if you can't find the error straight away, it might be that boost-1.73 needs to be ported in some way, or that it's still buggy with only beta1

Flamefire:
`git lfs install` installs the hooks globally, so you have to globally uninstall it.

I can reproduce the issue. Seems to be with rttr. Will check tomorrow.

Flamefire:
Ok found: The found library is only available in the current dir and subdirs. -.- So searching for it in the top-level file now.

I also added RTTR_USE_SYSTEM_LIBS which sets the default for RTTR_USE_SYSTEM_*

.travis.yml
```diff
@@ -34,10 +34,10 @@ before_install:
 
 install:
   - if [ "${TRAVIS_OS_NAME}" = "osx" ]; then export PATH="/usr/local/opt/gettext/bin:$PATH"; fi
-  # Minimum requirements: CMake 3.8.2, Boost 1.64
+  # Minimum requirements: CMake 3.9, Boost 1.64
```
Flow86, line 37: we should write that into the README.md too

Flamefire: We don't have any versions in the readme, so I would not add it there. As usual: "Documentation is always out of date", so I prefer code be the docu. It is right at the top of the CML and running CMake will give a very clear error if it is to low. It is only a comment here to explain the below line(s) (why that versions were chosen and not any other)

Flamefire:
@Flow86 Ready to merge?

stefson:
it doesn't work for me yet:  

```
-- Configuring done
CMake Error at libs/s25client/CMakeLists.txt:19 (add_executable):
  Target "s25client" links to target "Samplerate::Samplerate" but the target
  was not found.  Perhaps a find_package() call is missing for an IMPORTED
  target, or an ALIAS target is missing?


CMake Error at libs/s25main/CMakeLists.txt:44 (ADD_LIBRARY):
  Target "s25Main" links to target "Samplerate::Samplerate" but the target
  was not found.  Perhaps a find_package() call is missing for an IMPORTED
  target, or an ALIAS target is missing?


-- Generating done
CMake Generate step failed.  Build files cannot be regenerated correctly.
```

there are also many warnings about this and that, during configure. please have a look at the full log: [build.log.zip](https://github.com/Return-To-The-Roots/s25client/files/4514200/build.log.zip)

I'm using cmake-3.17.1

Flamefire:
That is strange but not introduced here. Will ask the CMake guys as this is supposed to work.

Also the warnings are solely about the new Boost. Make sure you install the boost cmake configs in a path CMake can find. Or use cmake 3.17.2+ once available.

Hence nothing to hold this PR back :)

stefson:
ok, got it compiled and runtime tested with gcc-9 and clang-9 on linux-x86_64

your idea of using cmake fetch is not really acceptable distrowise, however since this is only a temporarily fix needed when building against boost <=1.72.0 , it should be okay, given that you don't release the stable-0.9.0 before there's a final of boost-1.73.0 that is tested against. 

boost has become such a highly complex framework, that it often silently breaks things at runtime, this is not a hypothetical case for this project - there have been cases of segfaults and other things in the past with new boost versions.

Flamefire:
> your idea of using cmake fetch is not really acceptable distrowise

Distros can install Boost.Nowide outside of Boost as some currently do already using the old source code and use that system version with rttr. Once they Bundle 1.73 they can use that.

The FetchContent is a convenience package-manager replacement for developers at least.

PR #1212, opened 2020-04-19
Add --version command

### LINKED ISSUE #1146: please add --version switch to s25client binary
It would be great to have this feature, so that a simple

`/usr/bin/s25client --version`

would throw out the version, and given that it's a snapshot, the git commit id 

at the moment it's not implemented: 
```
/usr/bin/s25client --version
terminate called after throwing an instance of 'boost::wrapexcept<boost::program_options::unknown_option>'
  what():  unrecognised option '--version'
```

Flamefire:
Also catches invalid options passed and displays the help
Closes #1146

Example output:
>$ Debug/s25client.exe --version
Return To The Roots v20200419-784eb58fc1eb42751042d82f16cc92617fc6c2ae
Compiled with Microsoft Visual C++ version 1925 for Win32 64 Bit

@stefson Does this fit your needs? I'm reusing the exact same output that is also written to log/console on start


libs/s25client/s25client.cpp
```diff
@@ -56,6 +55,8 @@
 #include <csignal>
 #endif
 
+namespace bfs = boost::filesystem;
```
Flow86, line 58: is bfs used?

Flamefire: Yes. I only removed the include that used to define this (want to get rid of the "keep" headers)

stefson:
A simple `0.9.0-git` would have been enough, so that's even better :-)

PR #1213, opened 2020-04-20
Fix text not being shown in tool-window and main menu

### LINKED ISSUE #1207: Tool ordering invisible numbers
When using the tool ordering addon, the number 10 is not visible, 11 is visible and any higher number is invisible in the ordering screen

Flamefire:
The text/button were to small, adjusted sizes.
Also limit number of tools to order to 99 not 100 (causing this again)
Fixes #1207

### DIFF
--- a/libs/s25main/ingameWindows/iwMainMenu.cpp
+++ b/libs/s25main/ingameWindows/iwMainMenu.cpp
@@ -21,6 +21,7 @@
 #include "Loader.h"
 #include "WindowManager.h"
 #include "addons/const_addons.h"
+#include "controls/ctrlTextButton.h"
 #include "iwAIDebug.h"
 #include "iwBuildOrder.h"
 #include "iwBuildingProductivities.h"
@@ -81,7 +82,11 @@ iwMainMenu::iwMainMenu(GameWorldView& gwv, GameCommandFactory& gcFactory)
     bool enableAIDebug = true;
 #endif
     if(gwv.GetViewer().GetPlayer().isHost && enableAIDebug)
-        AddTextButton(13, DrawPoint(80, 210), Extent(22, 22), TC_GREY, _("AI"), NormalFont, _("AI Debug Window"));
+    {
+        auto* bt = static_cast<ctrlTextButton*>(
+          AddTextButton(13, DrawPoint(80, 210), Extent(0, 22), TC_GREY, _("AI"), NormalFont, _("AI Debug Window")));
+        bt->ResizeForMaxChars(bt->GetText().size());
+    }
 
     // Optionen
     AddImageButton(30, DrawPoint(12, 231), Extent(165, 32), TC_GREY, LOADER.GetImageN("io", 37), _("Options"));
--- a/libs/s25main/ingameWindows/iwMainMenu.cpp
+++ b/libs/s25main/ingameWindows/iwMainMenu.cpp
@@ -15,7 +15,6 @@
 // You should have received a copy of the GNU General Public License
 // along with Return To The Roots. If not, see <http://www.gnu.org/licenses/>.
 
-#include "rttrDefines.h" // IWYU pragma: keep
 #include "iwMainMenu.h"
 #include "GamePlayer.h"
 #include "GlobalGameSettings.h"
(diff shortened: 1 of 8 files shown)


libs/s25main/ingameWindows/iwTools.cpp
```diff
@@ -56,9 +54,9 @@ iwTools::iwTools(const GameWorldViewer& gwv, GameCommandFactory& gcFactory)
         for(unsigned i = 0; i < NUM_TOOLS; ++i)
         {
             Extent btSize = Extent(20, 13);
-            ctrlButton* bt = AddImageButton(100 + i * 2, DrawPoint(174, 25 + i * 28), btSize, TC_GREY, LOADER.GetImageN("io", 33), "+1");
+            ctrlButton* bt = AddImageButton(100 + i * 2, DrawPoint(174, 27 + i * 28), btSize, TC_GREY, LOADER.GetImageN("io", 33), "+1");
             AddImageButton(101 + i * 2, bt->GetPos() + DrawPoint(0, btSize.y), btSize, TC_GREY, LOADER.GetImageN("io", 34), "-1");
-            AddTextDeepening(200 + i, DrawPoint(151, 4 + bt->GetPos().y), Extent(20, 18), TC_GREY, "", NormalFont, COLOR_YELLOW);
```
Flow86, line 61: is it possible to get the size from the font itself?

libs/s25main/ingameWindows/iwMainMenu.cpp
```diff
@@ -82,7 +81,7 @@ iwMainMenu::iwMainMenu(GameWorldView& gwv, GameCommandFactory& gcFactory)
     bool enableAIDebug = true;
 #endif
     if(gwv.GetViewer().GetPlayer().isHost && enableAIDebug)
-        AddTextButton(13, DrawPoint(80, 210), Extent(20, 20), TC_GREY, _("AI"), NormalFont, _("AI Debug Window"));
+        AddTextButton(13, DrawPoint(80, 210), Extent(22, 22), TC_GREY, _("AI"), NormalFont, _("AI Debug Window"));
```
Flow86, line 84: same here

libs/s25main/ingameWindows/iwTools.cpp
Flamefire: Yes and no. There is `font.getDx()` which gets the largest width. However we'd have to add the border the element adds, so we'd need to query the element. I guess the only viable solution would be to add a function `extendWidthForNChars(numberOfChars)` to those classes to make it handle it itself.

Flamefire: @Flow86 There is a problem: The value used is to big. The max width is 12 for all chars but the max width for the numbers is 7. So the field gets too big. How'd you handle this? 

My idea: A a function to query the min size of the control and a free function to get the max size for a given char array. Then one can do: `txt->Resize(text->getbordersize() + Extent(2 * getMaxSize(font, {'0', '1', ... '9'}), ...))`

That might look strange but works. Also for the AI button it would be `bt->Resize(bt->getbordersize() + font->getextend(bt->gettext()))`

Flow86: yes I thought we have some "how big is it in real"-method in the font-draw stuff

Flamefire: Yes we do. But a) currently missing is a "how much space is there for the text" part and b) something to calculate the max size for a given set of possible texts.

Asking the font "get me the size for '0' to '99'" isn't feasible, but getting the max char width of '0'-'9' and multiplying that by 2 is ok.

There are also 2 design issues:
- Currently the text is part of the textdeepening or textbutton. But IMO it should rather be a child of that. Then it would just be a ctor overload.
- This means that a ctrlText should be sized too. This would also make more sense for calling the Font.Draw method with the alignment: It should pass a rect instead of a point. Avoids stuff like this: https://github.com/Return-To-The-Roots/s25client/blob/d122e7e3f19b86196d02407ebedd485016ec3acf/libs/s25main/controls/ctrlTextDeepening.cpp#L40-L56

If we make ctrlText a (sized) child, then we can generically get the border size by calling `bt->getSize() - bt->getCtrl<ctrlText>(0)->getSize()`

Flamefire: Or even better: Make the Font.Draw way more low-level: Let it simply draw the given text at a given position. The alignment and trimming to a size can be done by ctrlText. So everywhere you want that behavior you got to use a ctrlText.

Flamefire:
@Flow86 As mentioned in Slack I'd rather merge this now and do the proper Font handling later as this does fix the bug

Flamefire:
Rebased.

PR #1216, opened 2020-04-29
Update submodules, include anonymization of EMail address

### LINKED ISSUE #1214: Remove emailaddress from logfile
Inside the logfile the user's emailaddress gets logged. We should anonymize it.
Printing the first + last letter of alias and domain should be visible so we can verify it, but like this less accidents can happen, that the user posts or uploads its emailadress to a post or issue

Flamefire:
Fixes #1214

@Flow86 Check the tests if that is reasonable.


Flow86:
looks good to me, 
perhaps one very picky point: the "at_pos" can/should be const I think

Flamefire:
Good point. I also un-pythonized the name

PR #1219, opened 2020-05-04
Fix crash on map param

### LINKED ISSUE #1218: crash when s25client help
So I know the command should be `s25client --help`, but if you write `s25client help` without the **--** you get a crash.

Version info:
`>s25client.exe --version
Return To The Roots v20200430-1960d86e3e4c4f9fb4c8c3976d9eab4bc8e5b837
Compiled with GNU C++ version 7.3-win32 20180506 for Win32 64 Bit`

Flamefire:
Fixes #1218 

Larger refactoring to get loading of data into a reusable class. Also avoid the use of the Singletons (want to get rid of them eventually)

Not fully satisfied with the solution as there is still duplicated code. More maybe in future PRs

### DIFF
--- a/libs/s25main/MusicPlayer.cpp
+++ b/libs/s25main/MusicPlayer.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2005 - 2017 Settlers Freaks (sf-team at siedler25.org)
+// Copyright (c) 2005 - 2020 Settlers Freaks (sf-team at siedler25.org)
 //
 // This file is part of Return To The Roots.
 //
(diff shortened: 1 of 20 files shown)


libs/s25main/MusicPlayer.cpp
```diff
@@ -15,184 +15,31 @@
...
-#include "ingameWindows/iwMusicPlayer.h"
 #include "ogl/MusicItem.h"
 #include "libsiedler2/Archiv.h"
 #include "libsiedler2/ArchivItem.h"
 #include "libsiedler2/ErrorCodes.h"
 #include "libsiedler2/prototypen.h"
 #include "s25util/Log.h"
 #include "s25util/StringConversion.h"
+
+/*
```
Flow86, line 29: dead code

Flamefire: Thanks, fixed

PR #1224, opened 2020-05-14
Renamed the library paths on Linux

Mailaender:
While `/usr/lib64/driver/audio/libaudioSDL.so` probably still adheres to FHS it is confusing. To me it sounds like a system wide hardware driver. I prefixed it with the project name like the data directory does. You can do whatever you want inside that.


CMakeLists.txt
```diff
@@ -156,9 +156,9 @@ else()
         set(RTTR_DOCDIR "${DEFAULT_DOCDIR}" CACHE STRING "Directory for documentation files")
         if("${RTTR_LIBDIR}" STREQUAL "")
             # Avoid using absolute path accidentally
-            set(DEFAULT_DRIVERDIR "driver")
+            set(DEFAULT_DRIVERDIR "s25rttr/driver")
```
Flow86, line 159: this will create weird directories on windows.

Flamefire: No it actually makes perfect sense. Windows is handled above. So this is linux only.

Flow86: ah you're right. then its good

Issue #96, opened 2015-07-31
Source tarball


nE0sIghT:
As requested [here](https://bugs.launchpad.net/s25rttr/+bug/1069546) please create a source tarball for current (stable) s25rttr version and any feature version as well.

Flow86:
the current stable source tarball can be downloaded here: https://github.com/Return-To-The-Roots/s25client/archive/0.8.zip

hasufell:
These don't contain the submodules, so for packaging these seem broken.

nE0sIghT:
Also they are outdated

Flamefire:
The missing submodules are a very valid point.
Regarding outdated: The latest stable release is 0.8.1, so it is not outdated as nightlys are no releases. However I think we need more releases.
For example the current version is pretty stable.  We could tag that as 0.8.2 and move all remaining issues up to 0.8.3 or higher. There are a few remaining kinda high priority bugs which should be fixed, but in general the nightly is much better than 0.8.1.
I'd even name this 0.9.0 following the major=Really big change, medium(?)=feature change, minor=patches strategy and allow also 0.10.0 ;-)

hasufell:
> The missing submodules are a very valid point.

When you do releases (not just tags) you can also upload source tarballs that have everything necessary. E.g. [flex](https://github.com/westes/flex/releases/tag/v2.6.1) does that where the non-generated tarballs have a bootstrapped autotools.

Mailaender:
This is also a major inconvenience for @openSUSE so I would appreciate a 0.8.2 tag. For a network game I think it is essential that everyone packages the same version and not random git commits, because stable is so broken and outdated.

Flow86:
Hi,
unfortunately github does not support submodules inside the (automatic) tarballs "releases" page.

to get the full source, you have to git clone it. 
A release in github is also a git tag, so the following works to the the "full" source:

```
git clone ...
git checkout 0.8.1  # < or use any other "release"-tag here
git submodule update --init
```

Mailaender:
This isn't a huge problem for @openSUSE as http://openbuildservice.org/2016/04/08/new_git_in_27/ exists, but probably other distributors. Uploading a source tarball would be nice. Tagging a new release after 2 years is a must. For now I packaged the current master with some patches as the sub-modules  seem to lack behind. https://build.opensuse.org/package/show/games/s25rttr

Flamefire:
[...]

PR #1225, opened 2020-05-15
Run s25client instead of a wrapper script on Linux desktops

Mailaender:
In https://github.com/Return-To-The-Roots/s25client/issues/96 you hinted that this patch may be wanted for inclusion in general. I don't know enough about your custom integrated Debian package to decide, but here is my patch in a convenient form.


Flamefire:
@Flow86 As mentioned in #96 I'm in favor of inclusion. I'll leave it to you

PR #1227, opened 2020-05-17
Do static analysis on GHA

Flamefire:
This adds the static analysis (currently only formatting and encoding) to Github Actions. This is faster than doing this on travis.

I'll also move the clang-tidy checks to GHA to remove 2 travis jobs, but that needs a whole build setup on GHA


libs/s25main/Loader.cpp
```diff
@@ -569,7 +573,7 @@ void Loader::fillCaches()
                                 id += NATION_RTTR_TO_S2[nation] * 6;
                             } else if(nation == NAT_BABYLONIANS) //-V547
                             {
-                                id += NATION_RTTR_TO_S2[nation] * 6;
+                                id += NATION_RTTR_TO_S2[NAT_ROMANS] * 6;
```
Flow86, line 576: this fixes a different bug?

libs/s25main/Loader.h
```diff
@@ -188,6 +187,8 @@ class Loader : public Singleton<Loader, SingletonPolicies::WithLongevity>
         RTTR_Assert(!item || res);
         return res;
     }
+    Log& logger_;
```
Flow86, line 190: I can't recommend storing a reference here, same problem, you don't know when it will be destroyed.
Use a shared_ptr on `ILog`

libs/s25main/Loader.h
```diff
@@ -188,6 +187,8 @@ class Loader : public Singleton<Loader, SingletonPolicies::WithLongevity>
...
+    const RttrConfig& config_;
```
Flow86, line 191: same here, "you shall not store references"

libs/s25main/Timer.cpp
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2005 - 2018 Settlers Freaks (sf-team at siedler25.org)
+// Copyright (c) 2005 - 22020018 Settlers Freaks (sf-team at siedler25.org)
```
Flow86, line 1: wtf?

libs/s25main/ai/aijh/AIConstruction.cpp
```diff
@@ -550,9 +550,7 @@ BuildingType AIConstruction::ChooseMilitaryBuilding(const MapPoint pt)
                 bld = BLD_CATAPULT;
             else
             {
-                if(randmil % 2 == 0)
-                    bld = biggestBld;
-                else if(aii.CanBuildBuildingtype(BLD_WATCHTOWER))
+                if(randmil % 2 == 0 && aii.CanBuildBuildingtype(BLD_WATCHTOWER))
```
Flow86, line 553: thats wrong?

libs/s25main/Loader.cpp
Flamefire: Not really. clang-tidy has a new warning "duplicated branches". I just made the "difference" here explicit to silence it.

libs/s25main/Loader.h
Flamefire: References are non-owning references to existing objects. shared_ptr implies ownership and even worse: multiple ownerships. So references are appropriate here (besides being the least amount of change)

libs/s25main/Timer.cpp
Flamefire: C&P -.-

libs/s25main/ai/aijh/AIConstruction.cpp
Flamefire: Yes and no. `randmil % 2 == 0` is semantically: with a 50% chance. I just changed the "sides of the coin"
Check the whole if-elseif-else chain to verify that it is correct

libs/s25main/notifications/NotificationManager.h
```diff
@@ -35,22 +35,22 @@ class NotificationManager
...
-    inline NotificationManager();
-    inline ~NotificationManager();
+    NotificationManager();
+    ~NotificationManager();
 
     /// Subscribe to a specific notification.
     /// Unsubscribes when the subscription has no references left
     template<class T_Note>
-    inline Subscribtion subscribe(std::function<void(T_Note)> callback);
+    Subscribtion subscribe(std::function<void(T_Note)> callback);
```
Flow86, line 44: can we fix that typo too? `Subscri**p**tion`

libs/s25main/ai/aijh/AIConstruction.cpp
Flow86: hmm okay yes you're right. but changing that will break existing replays

libs/s25main/Loader.h
Flow86: but we have an implicit ownership here now. and you can't mock it.

of course its the least amount of change, but we should reconsider that on the next step then

libs/s25main/Loader.cpp
Flow86: but  `NAT_BABYLONIANS` isnt `NAT_ROMANS`, is it?

libs/s25main/notifications/NotificationManager.h
Flamefire: Good point: https://github.com/Return-To-The-Roots/s25client/pull/1227/commits/5ba8395eed50885e5c2ef22f3d2d1a0c107eacc8

libs/s25main/ai/aijh/AIConstruction.cpp
Flamefire: This is AI. AI decisions are not part of the replay, only the GCs

libs/s25main/Loader.h
Flamefire: No ownership is here. That is the whole model of references: You don't own it, you reference it
And you can mock with references just as well.

libs/s25main/Loader.cpp
Flamefire: There is no babylonians in S2, this makes it clear we are reusing the roman figures until actual new ones exist. See the comment below.

Flow86: ok

PR #1231, opened 2020-05-18
Update wasteland.lua

### LINKED ISSUE #1182: Most edges wrong on wasteland maps
image

image

image

Spikeone:
Still not perfect, but it's not possible to fully fix the issue with the current system.

fixes #1182


Flow86:
```
s25client/tests/libGameData/testGameData.cpp(69): fatal error: in "GameData/LoadGameData": critical check desc.s2Id == TerrainData::GetTextureIdentifier(t) has failed [0x11 != 0x13]
```

Flamefire:
@Spikeone This is an imperfection of our test and/or a bug in your file:
https://github.com/Return-To-The-Roots/s25client/pull/1231/files#diff-5b98a88249d31fa5837abd8cf2f35708R274

We haven't used 0x11 as the s2 ID before. This is what the test uses: https://github.com/Return-To-The-Roots/s25client/blob/master/tests/legacyFiles/src/legacy/TerrainData.cpp

Are those values wrong?

Spikeone:
Well, looks like I can't simply upload the file without breaking a test :x
0x11 is a valid ID but shouldn't be used (normally) see https://settlers2.net/documentation/world-map-file-format-wldswd/

Flamefire:
> Well, looks like I can't simply upload the file without breaking a test :x

Just if your uploaded file breaks a test. ;)

Well we never used that texture did we? It isn't even textured and likely a bug in the original game anyway. Why did you want to include it? IIRC for this test the order of the textures is important (which is not really the case for the data but only to compare it to the manually created data which we had before), so I guess if you really want this texture you could add it to the end after all existing ones

Spikeone:
Hope I didn't break anything by "@Spikeone Merge branch 'master' into master" but moving the 0x11 texture hopefully helps :x

Flow86:
you should use "rebase" onto master, and then force push your feature branch, then your feature commits are added on top of the master, and no merge commit is necessary.

PR #1237, opened 2020-05-24
update credit skip to work again, update credits

### LINKED ISSUE #1236: Credit skip doesn't work
When you click, you don't skip a name but see the same again.

Flow86:
fixes #1236

@Flamefire  perhaps you can change it to use the animation system? I'm honest I don't have the time to look into it to find out how that works

### DIFF
--- a/libs/s25main/desktops/dskCredits.cpp
+++ b/libs/s25main/desktops/dskCredits.cpp
@@ -54,7 +54,219 @@ dskCredits::dskCredits() : Desktop(LOADER.GetImageN("setup013", 0)), itCurEntry(
     // "Credits"
     AddText(2, DrawPoint(400, 33), _("Credits"), COLOR_YELLOW, FontStyle::CENTER, LargeFont);
 
-    itCurEntry = entries.end();
+    // order by last name (alphabetical)
+    CreditsEntry entry = CreditsEntry("Alexander Grund (Flamefire):", GetCreditsImgOrDefault("flamefire"));
+    entry.lines.push_back(_("Programming"));
+    entry.lines.push_back(_("Quality Assurance"));
+    entries.push_back(entry);
+
+    entry = CreditsEntry("Patrick Haak (Demophobie):", GetCreditsImgOrDefault("demophobie"));
+    entry.lines.push_back(_("Website Administration"));
+    entry.lines.push_back(_("Quality Assurance"));
+    entries.push_back(entry);
+
+    entry = CreditsEntry("Jan-Henrik Kluth (jh):", GetCreditsImgOrDefault("jh"));
+    entry.lines.push_back(_("Programming"));
+    entry.lines.push_back(_("Artificial Intelligence (AI)"));
+    entries.push_back(entry);
+
+    entry = CreditsEntry("Christopher Kuehnel (Spikeone):", GetCreditsImgOrDefault("spikeone"));
+    entry.lines.push_back(_("Additional graphics"));
+    entry.lines.push_back(_("Quality Assurance"));
+    entry.lines.push_back(_("Mapping"));
+    entries.push_back(entry);
+
+    entry = CreditsEntry("Stefan Schüchl (Z-Stef):", GetCreditsImgOrDefault("z-stef"));
+    entry.lines.push_back(_("Website Administration"));
+    entry.lines.push_back(_("Website Programming"));
+    entries.push_back(entry);
+
+    entry = CreditsEntry("Marcus Ströbel (Maqs):", GetCreditsImgOrDefault("maqs"));
+    entry.lines.push_back(_("Programming"));
+    entry.lines.push_back(_("Quality Assurance"));
+    entries.push_back(entry);
+
+    entry = CreditsEntry("Jonas Trampe (NastX):", GetCreditsImgOrDefault("nastx"));
+    entry.lines.push_back(_("Quality Assurance"));
+    entry.lines.push_back(_("Mapping"));
+    entries.push_back(entry);
+
+    // founder members
+    entry = CreditsEntry("Oliver Siebert (Oliverr):", GetCreditsImgOrDefault("oliverr"));
+    entry.lines.push_back(_("Project founder"));
+    entry.lines.push_back(_("Project management"));
+    entry.lines.push_back(_("Programming"));
+    entry.lines.push_back(_("Quality Assurance"));
+    entries.push_back(entry);
+
+    entry = CreditsEntry("Florian Doersch (FloSoft):", GetCreditsImgOrDefault("flosoft"));
+    entry.lines.push_back(_("Project founder"));
+    entry.lines.push_back(_("Project management"));
+    entry.lines.push_back(_("Server management"));
+    entry.lines.push_back(_("Programming"));
+    entry.lines.push_back(_("Website Administration"));
+    entry.lines.push_back(_("Website Programming"));
+    entry.lines.push_back(_("Quality Assurance"));
+    entry.lines.push_back(_("Finances"));
+    entries.push_back(entry);
+
+    // alphabetical again
+    entry = CreditsEntry(_("Additional Programming:"));
+    entry.lines.push_back("Ikhar Beq (PoC)");
+    entry.lines.push_back("Cat666");
+    entry.lines.push_back("Devil");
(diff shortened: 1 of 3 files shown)


libs/s25main/desktops/dskCredits.cpp
```diff
@@ -174,16 +290,31 @@ void dskCredits::Draw_()
...
+    timer.restart();
+}
+
+void dskCredits::GotoPrevPage()
+{
+    // Prev page
+    if(this->itCurEntry == entries.begin())
+        this->itCurEntry = std::prev(entries.end());
+    else
+        --this->itCurEntry;
```
Flamefire, line 308: To mirror the other function:

```suggestion
    if(this->itCurEntry == entries.begin())
        this->itCurEntry = entries.end();
    --this->itCurEntry;
```

BTW: Can you remove the trivial comments? `GotoPrevPage()` --> `// Prev page` is just noise

libs/s25main/desktops/dskCredits.cpp
```diff
@@ -174,16 +290,31 @@ void dskCredits::Draw_()
...
+    timer.restart();
+}
+
 void dskCredits::DrawCredit()
 {
-    unsigned time = VIDEODRIVER.GetTickCount() - startTime;
+    const unsigned time = static_cast<unsigned>(std::chrono::duration_cast<std::chrono::milliseconds>(timer.getElapsed()).count());
```
Flamefire, line 314: Just define `PAGE_TIME` as  a chrono type and don't convert at all

libs/s25main/desktops/dskCredits.cpp
```diff
@@ -322,7 +453,20 @@ glArchivItem_Bitmap* dskCredits::GetCreditsImgOrDefault(const std::string& name)
...
+    return true;
+}
+
+bool dskCredits::Msg_LeftUp(const MouseCoords& /*mc*/)
+{
+    GotoNextPage();
+    return true;
+}
+
+bool dskCredits::Msg_RightUp(const MouseCoords& mc)
```
Flamefire, line 466:
```suggestion
bool dskCredits::Msg_RightUp(const MouseCoords&)
```
or the warning will trigger

libs/s25main/desktops/dskCredits.h
```diff
@@ -57,10 +64,10 @@ class dskCredits : public Desktop
         std::string lastLine;
         glArchivItem_Bitmap* pic;
         std::vector<Line> lines;
-        explicit CreditsEntry(std::string title, std::string lastLine = "")
+        explicit CreditsEntry(std::string title, std::string lastLine = std::string())
```
Flamefire, line 67: Why this change? This is just more noise. Use `{}` if you don't like an empty string

libs/s25main/desktops/dskCredits.cpp
```diff
@@ -82,69 +89,178 @@ dskCredits::dskCredits() : Desktop(LOADER.GetImageN("setup013", 0))
...
-    entry.lines.push_back(CreditsEntry::Line("Marius Loewe", 1));
-    entry.lines.push_back(CreditsEntry::Line("Eric Lutter", 1));
-    entry.lines.push_back(CreditsEntry::Line("Bob Kromonos Achten", 1));
-    entry.lines.push_back(CreditsEntry::Line("morlock", 1));
-    entry.lines.push_back(CreditsEntry::Line("Hans Gabathuler", 1));
-    entry.lines.push_back(CreditsEntry::Line("Jan Montag", 1));
-    entry.lines.push_back(CreditsEntry::Line("Patrick Schefczyk", 1));
-    entries.push_back(entry);
+
+    const size_t page_size = 15;
```
Flamefire, line 243: You use 2 columns but 15 names? Why not 16 so it is 8 per column?

Also: Maybe alternate the names (1 left, 1 right, 1 left, ...) so the last page is filled top to bottom. But doesn't really matter

libs/s25main/desktops/dskCredits.cpp
```diff
@@ -82,69 +89,178 @@ dskCredits::dskCredits() : Desktop(LOADER.GetImageN("setup013", 0))
...
+      // V                              // 
+      // W                              // 
+      "Niels Wiederanders",             // 
+      "Philipp Wohlleben"               // 
+      // X                              // 
+      // Y                              // 
+      // Z                              // 
+    };
+
+    // alphabetical again, two rows
```
Flamefire, line 240:
```suggestion
    // alphabetical again, two columns
```

libs/s25main/desktops/dskCredits.h
Flow86: you shall not use `""` - it will/can create a copy from the constant `""`. but `{}` is propably the  best idea. will change that

libs/s25main/desktops/dskCredits.cpp
Flow86: because its not filled that way.
we have on the left 2 additional lines with the red "bottom" text - and below that the bobs will start walking around.
this way we have a quite good layout with the names and the red text.

libs/s25main/desktops/dskCredits.h
Flamefire: So what is the difference? And don't see any problem.

Flamefire: At any optimization level (hence above `-O0`) the code is literally the same: https://godbolt.org/z/7dGAza

But readability is different:
- `void foo(std::string s = "")` - "Default value is an empty string"
- `void foo(std::string s = std::string())` - "Default value is string...with no value"
- `void foo(std::string s = {})` - "Default value is an default constructed...string"

So the first is easiest to parse visually. the second needs you to go through a lot of chars to see what the actual value is and the extra parens don't help and the last needs you to go back to see what thing you are creating and then see it's an empty string.

libs/s25main/desktops/dskCredits.cpp
```diff
@@ -335,6 +479,7 @@ void dskCredits::SetActive(bool active)
     Desktop::SetActive(active);
     if(active)
     {
-        startTime = bobTime = bobSpawnTime = VIDEODRIVER.GetTickCount();
+        bobTime = bobSpawnTime = VIDEODRIVER.GetTickCount();
```
Flamefire, line 482: This needs to change too as it is the current source of the bug which @Spikeone mistakenly thought of a "broken skip": There was no skipping it (re)activated the desktop on clicking it which restarts the timers.

It should only happen on the first call, or (now easier): When the timer isn't already running (as we never stop it it works)

Flamefire:
@Flow86 Shall I finish this PR so we can get a release out or will you have time to do that soonish?

Flow86:
I can do that soonish. Give me some time ;-)

libs/s25main/desktops/dskCredits.cpp
Flow86: doesnt work, you cant do an `a>b` comparison or similar with `std::chrono::milliseconds`?

libs/s25main/desktops/dskCredits.cpp
Flow86: doesnt happen for me now anymore?

Flamefire: It still should as it is still wrong: The SetActive may happen multiple times caused by clicking. The `timer.restart` below will cause the `time` in `DrawCredits` to reset and the bobTime/bobSpawnTime reset above will cause an additional delay in spawning bobs and a lack in their animation (which likely won't be noticed though)
I'd suggest to use a timer for all 3, and here check if the timer has been started and only if not call start

Flow86: the only thing which will happen is that you stop spawning new bobs if you click. Thats even kind of an easter egg ;-)
Credits

Flamefire: It also has this skip-like behavior @Spikeone mentioned which is rather odd. However it looks like it is actually skipping now, is it? I'm surprised by that as from the code I think it just restarts the current credit. If it does skip, it is fine

Flow86: it steps to the next credit on click now (and resets the time a new bob will be spawned, so its working

Issue #1234, opened 2020-05-22
Intro button not working


Spikeone:
The intro button is not working

Flow86:
did we have an intro? perhaps we should remove the button

Spikeone:
@Flow86 right, it sadly never did - are you sure that we shouldn't add the intro?

Flow86:
we never implemented playing a video file. but perhaps we could do that somehow @Flamefire

Flamefire:
Is it worth the work? We'd just use it for the intro...

Spikeone:
Well, never watched the intro (is it even included or is it only if you have the cd?) - but having it would be quite nice for the feeling (e.g. you watch it on first start and then it autoskips) as it brings back some memories.
But I don't know how much work it really is.

Flow86:
perhaps we should disable the button for 0.9, so its clear that its not working (yet) . then we can think about it. I also would like to have it if I'm honest

Flamefire:
> perhaps we should disable the button for 0.9, so its clear that its not working (yet)

Yep just remove it.

>  I also would like to have it if I'm honest

Implementing a video player is non-trivial. I don't even know where that intro is located at and haven't found any hint. So I'd say no, but feel free ;)

Spikeone:
[VIDEO.zip](https://github.com/Return-To-The-Roots/s25client/files/4673360/VIDEO.zip)

It's actually located in the VIDEO folder :D which is often not included in any online version as it's nearly 15Mb. It's a *.SMK File.

Flamefire:
Good luck: https://en.wikipedia.org/wiki/Smacker_video

> In Smacker video, a frame is split into 4×4 blocks in raster-scan order.[2] Each block can be coded in one of six coding modes: skip, fill, mono, and three full modes. Each mode can be signaled for multiple blocks in a run-length encoding scheme. 

It goes on with more cases. https://wiki.multimedia.cx/index.php/Smacker has more info with huffman trees for video and audio.

As mentioned: non-trivial

Spikeone:
Well, there is a lib for smacker video: https://git.ffmpeg.org/gitweb/ffmpeg.git/tree/HEAD:/libavcodec

Flamefire:
@Flow86 Didn't even want to include a zip library as a dependency to allow for compressed resources. And just the library isn't enough either...

Flow86:
[...]

PR #1238, opened 2020-05-24
disable intro button

Flow86:
I do not remove it, since we want the intro perhaps for 1.0

see #1234


Flamefire:
As mentioned writing the decoder for that format and adding a video playback function is very much work, but if you want to do that....

Merging

PR #1239, opened 2020-05-24
 Fix credit image query and transparency 

### LINKED ISSUE #1235: Credit images not working (again)
The credits aren't working - thats maybe the same reason why overriding graphics doesn't work anymore.

Flamefire:
Fixes #1235


Flamefire:
Fixed and tested on Clang and GCC 8, let's see what CI says. Also please merge #1227 which updates both submodules already halfway to the point this PR updates them

PR #1240, opened 2020-05-30
set locale before parsing playlist

jgilje:
i've got a loading error in Playlist.cpp on my machine. I traced the fault down to Playlist loading with my environment's locale. My suggested fix is to apply "C" locale before starting to parse.

Before applying this patch, it should refuse loading by running
`LC_ALL=nb_NO.UTF-8 bin/s25client`
while running
`LC_ALL=C bin/s25client`
successfully loads

### DIFF
--- a/libs/s25main/Playlist.cpp
+++ b/libs/s25main/Playlist.cpp
@@ -25,6 +25,7 @@
 #include <boost/nowide/fstream.hpp>
 #include <algorithm>
 #include <random>
+#include <sstream>
 
 namespace bnw = boost::nowide;
 
--- a/libs/s25main/Playlist.cpp
+++ b/libs/s25main/Playlist.cpp
@@ -79,7 +79,7 @@ bool Playlist::SaveAs(const std::string& filename, const bool overwrite)
         }
     }
 
-    bnw::ofstream out(filename.c_str());
+    s25util::ClassicImbuedStream<bnw::ofstream> out(filename.c_str());
     if(!out.good())
         return false;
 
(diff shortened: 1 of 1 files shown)


Flamefire:
Thanks for the PR! Can you tell what exactly went wrong? I guess it happens in the line https://github.com/Return-To-The-Roots/s25client/pull/1240/files#diff-23013cba4d8d3e5cb745e59a339619b5R125 but fail to see why it is locale dependent there.

Anyway I'd suggest the following improvement:
- add `#include "s25util/StringConversion.h"`
- Use `s25util::ClassicImbuedStream<std::stringstream>` instead of the `std::stringstream` which is a wrapper doing what you do manually. Yes it is slightly verbose, but more explicit and allows searching for such instances. Additionally it does not require the large `<locale>` header

Could you also do the same for the saving? I'll likely happen there too.

jgilje:
Hi. The actual problem is in `if(!(sline >> repeats >> random_str))`, I further narrowed it down to the `sline >> repeats` part. _WHY_ this would ever fail depending on the locale is beyond me :)

I updated the PR to use `s25util::ClassicImbuedStream<std::stringstream>` as suggested. Just realized I forgot to switch saving part. I'll add that soon.

libs/s25main/Playlist.cpp
```diff
@@ -19,12 +19,12 @@
 #include "ingameWindows/iwMusicPlayer.h"
 #include "mygettext/mygettext.h"
 #include "s25util/Log.h"
+#include "s25util/StringConversion.h"
 #include <boost/algorithm/string/trim.hpp>
 #include <boost/filesystem/path.hpp>
 #include <boost/nowide/fstream.hpp>
 #include <algorithm>
 #include <random>
-#include <sstream>
```
Flamefire, line 27: This include is still required as `std::stringstream` is used. It might work without as another header could include this, but it is better to include everything you use. So please re-add this :)

jgilje: done :)

PR #1242, opened 2020-05-31
Fix loading of overrides

### LINKED ISSUE #1112: Overriding graphics seems no longer to function
After years I continued working on the babyolian scout. So I created a 8 graphic animation for left-top and wanted to test them. So I tried to replace the original roman graphics by creating a JOBS.BOB folder here `s25rttr_20190820\RTTR\LSTS\GAME\JOBS.BOB` Sadly those graphics seem not to replace the original graphics at all. So tried to unpack the following file: `s25rttr_20190820\RTTR\LSTS\GAME\Babylonier\JOBS.BOB.LST` and repalced the graphics in that folder.

There still seems to be no effect at all which is strange in two ways as I expected the second file to be in use.
The game says it loads the graphics but seems to not overwrite existing files anymore (or loads the jobs file after).

Has there been a change I missed on how to replace graphics or am I doing something wrong?

Flamefire:
This fixes #1112

The main points are:
- resources (images, sounds, ...) are identified by a `ResourceId`
- That is a simple name, all lowercase so extensions and casing do not matter
- Folders should still have an extension (usually either .lst or .bob) to be considered for loading when scanning the folder, but not duplicate extensions
- files may have duplicate extensions. This is only useful for bob files, like foo.bob.lst where the .lst means to load it as lst format, then handle it as a bob resource to add to resource foo
- bob resources have a single archive as their first entry containing the graphics. For folders or files not named `*.bob` this is done by the loader to be compatible with `*.bob` files

Some related changes are included:
- Better naming for Loader::Load methods
- Usage of boost::filesystem::path where a real path is wanted
- introduce toLower function as `std::tolower` is potentially dangerous to use directly (see cppreference)
- fix a long standing TODO related to `nation_gfx` which should only contain nations actually loaded. This improves loading times probably a lot as the loop at https://github.com/Return-To-The-Roots/s25client/compare/master...Flamefire:bob_loading?expand=1#diff-f26a4af6ca06758a8f91142fea989e52R495 can be fully skipped for those not trying to load hundreds of non-existing images

All of those came up in the refactoring to understand the code and implement the feature. IMO it is much cleaner now.


libs/s25main/Loader.cpp
```diff
@@ -280,7 +288,7 @@ bool Loader::LoadSounds()
 
 bool Loader::LoadFonts()
 {
-    if(!LoadFile(config_.ExpandPath("<RTTR_RTTR>/LSTS/fonts.LST"), GetPaletteN("pal5")))
+    if(!Load(config_.ExpandPath("<RTTR_RTTR>/LSTS/fonts.LST"), GetPaletteN("pal5")))
```
Flow86, line 291: should the path here possibly also added to the FILE_PATHS or at least as a global constant in the same file as the rest of the paths, that way we have them in one place.

libs/s25main/Loader.cpp
```diff
@@ -357,47 +365,46 @@ void Loader::LoadDummyGUIFiles()
...
  *
- *  @return @p true bei Erfolg, @p false bei Fehler.
+ *  @return @p true on success
  */
-bool Loader::LoadFilesAtGame(const std::string& mapGfxPath, bool isWinterGFX, const std::vector<bool>& nations)
+bool Loader::LoadFilesAtGame(const std::string& mapGfxPath, bool isWinterGFX, const std::vector<Nation>& nations)
 {
-    if(NAT_BABYLONIANS < nations.size() && nations[NAT_BABYLONIANS])
+    if(helpers::contains(nations, NAT_BABYLONIANS))
         AddOverrideFolder("<RTTR_RTTR>/LSTS/GAME/Babylonier", false);
```
Flow86, line 379: same for here

libs/s25main/Loader.cpp
```diff
@@ -930,140 +949,156 @@ bool Loader::MergeArchives(libsiedler2::Archiv& targetArchiv, libsiedler2::Archi
...
-    filesAndFolders = ListDir(path, "idx", true, &filesAndFolders);
-    filesAndFolders = ListDir(path, "bmp", true, &filesAndFolders);
-    filesAndFolders = ListDir(path, "txt", true, &filesAndFolders);
-    filesAndFolders = ListDir(path, "ger", true, &filesAndFolders);
-    filesAndFolders = ListDir(path, "eng", true, &filesAndFolders);
-    filesAndFolders = ListDir(path, "ini", true, &filesAndFolders);
+    std::vector<bfs::path> filesAndFolders;
+    for(const auto ext : {"lst", "lbm", "bob", "idx", "bmp", "txt", "ger", "eng", "ini"})
+    {
+        std::vector<bfs::path> curFiles = ListDir(path, ext, true);
```
Flow86, line 1094: result can be const (or simply const auto?)

libs/s25main/ogl/glTexturePacker.cpp
```diff
@@ -97,7 +97,7 @@ bool glTexturePacker::packHelper(std::vector<glSmartBitmap*>& list)
             }
             // free texture packer, as it is not needed any more
             root->destroy(list.size());
```
Flow86, line 99: is it possible to wrap that call into the node destructor itself?

tests/s25Main/locale/testLocalization.cpp
```diff
@@ -34,7 +34,7 @@ struct LocaleFixture
     {
         rttr::test::LogAccessor logAcc;
         LANGUAGES.setLanguage("en");
-        LOADER.LoadFile(RTTRCONFIG.ExpandPath(FILE_PATHS[95]) + "/languages.ini", nullptr, true);
+        LOADER.Load(RTTRCONFIG.ExpandPath(FILE_PATHS[95]) + "/languages.ini", nullptr, true);
```
Flow86, line 37: probably same here? to add the full path to the files-constants

libs/s25main/Loader.cpp
Flamefire: I am very much against that list of files because it's incredibly hard to work with especially as the indices into some array are used. This leads to things like `Load(FILEPATHS[2]); // foo.lst` which is bad for 2 reasons: Duplication of the name (in array and comment) and it can possibly become out of sync.

In the resourceLoading branch I have I started by adding the folders with resources to a registry and then load resources by plain resourceIDs (filename w/o any extension) which avoids the issues.

So I'd wait till we introduce something like that and not move even more into that array than there already is.

libs/s25main/ogl/glTexturePacker.cpp
Flamefire: No. The destroy function does a depth-first loop to destruct the tree. The dtor would do a breath-first recursive call likely killing the stack memory.

I was thinking about improving this but due to the tree-structure this packer is very slow and unusable in debug builds of MSVC (takes minutes to build the texture). Hence I'm planning to rewrite it using a flat allocation and less smart distribution strategy but which seems to be still very good

libs/s25main/Loader.cpp
Flamefire: Good catch!

I'd only use auto where the type is mostly clear from the RHS or complicated (iterators) as IMO the type is a vital part for knowing the code.

libs/s25main/ogl/glTexturePacker.cpp
Flow86: ok :-)

libs/s25main/Loader.cpp
Flow86: yes but perhaps we can add some kind of FILE_PATH_FONTS_LST as a constant

libs/s25main/Loader.cpp
Flow86: yes I often do that too, but at least the const ;-)

Flamefire: Done: https://github.com/Return-To-The-Roots/s25client/pull/1242/commits/c7e4b87cf09eb1d50f0fe7a2e50f518cae0690eb

libs/s25main/Loader.cpp
Flamefire: I don't see the advantage. IIRC this is only used once and the name of the constant is almost as long as the path. So I'd keep the changeset low and fix all those paths later.

Flow86: the advantage is that we have all paths together in one file - so if we need to change them, we have them at one place

Flamefire: Question then: Shouldn't that be a separate PR? This PR doesn't introduce new path constants so IMO this whole discussion is valuable but out of scope of this PR.

Flow86: I like that commit. its okay if its a separate PR though

Flamefire: Ok. Removed it from this PR so this can be merged now

Issue #935, opened 2018-11-12
Game does not handle invalid player position


Spikeone:
If player 1 has an invalid start position, the problem is shown in the console but not handled properly. The game keeps showing the loading screen.

![image](https://user-images.githubusercontent.com/1276342/48371278-c13be180-e6bb-11e8-8b08-b577782eb467.png)

Map: Campaign Miss200.wld

Flamefire:
Ok I found the issue. What happens:
- Game start
- switch to loading screen desktop (enqueued for next draw call)
- game loading starts
- failure -> loading screen gets notified -> shows msgbox with the failure (closing this would go back to menu)
- desktop switch executed, all open windows get closed
- Result: Loading desktop shown expecting the msgbox to return but it doesn't

So the problem is a more general one: Opening a new window while a desktop switch is underway does not work unless the function (showAfterSwitch) is used which was introduced as a workaround for issues like this. However it requires the desktop to know, that it is switched (from/to).

Ideas:
1. If a desktop switch is under way when calling `Show(window)` the window is shown after the switch
  - Possible problem: This window might be intended for the current desktop
2. `Show` takes the desktop on which the window is shown as a param (or the window constructor requires the parent to be set to the parent desktop) and the window will get shown and closed with its desktop
  - Will be quite some work 
  - Maybe problem for windows opening windows (e.g. help windows from buildings) but they can open it for the same desktop (they know their parent)
  - Lua system would also need to know the current desktop. Not sure if this coupling is great...

2. seems to be the best soultion though. @Flow86 What do you think?

Flamefire:
@Spikeone How does S2 treat the case where there are N players but not all have start positions?

Spikeone:
@Flamefire The game trusts the editor - the editor sets a flag either 1 = valid, or 0 = invalid. https://settlers2.net/documentation/world-map-file-format-wldswd/

The editor always tested start positions when saving (BQ I guess). If anything failed, you got a window showing:
![image](https://user-images.githubusercontent.com/1276342/83942281-79773080-a7f2-11ea-9325-d7f88baec0d5.png)

A world was valid when you got "World is O.K."
![image](https://user-images.githubusercontent.com/1276342/83942295-8bf16a00-a7f2-11ea-8dab-51d5fcd4a661.png)

I'd not test what happens if you use an invalid world in Settlers 2 as they surely thought this was impossible with their editor (not regarding altered maps).

Flamefire:
[...]

PR #1248, opened 2020-06-05
Map load error handling

### LINKED ISSUE #1243: Maps silently not shown/loaded
Reported by Aden in Discord.

> My son just made a map using the RTTR editor.  It's saved in the 'worlds' folder (in .SWD format), but when we go in-game, we can't find it.  It otherwise opens fine in the editor.  Two questions:
> 
> 1. What do we need to do to get the game to recognize the map?
> 2. A number of years ago there was a separate .exe program I used to 'fix' the coastal harbor places.  Is there something similar we need to do to the map files to get it to accept harbor locations, or is that not an issue in rttr?

Map is https://cdn.discordapp.com/attachments/524755374331658240/716579549118332928/Lugees_Map.swd

@Spikeone mentioned:

> looks like we silently don't load maps with players > 8

Hence the editor shouldn't create 16 player maps and RTTR should show a message/warning at least

@Spikeone How verbose should such a warning be? When showing the map selection a message box for each faulty map? It usually shouldn't happen at all and the user might then want to fix or remove the map. Or just a log entry? Any other ideas?

Flamefire:
As explained in https://github.com/Return-To-The-Roots/s25client/issues/1243#issuecomment-636804549 this fixes #1243 by showing an error in the log for completely broken maps (e.g. wrong file format) and a message box when selecting an unsupported map.

To avoid this triggering without the user selecting anything (e.g. when there is only 1 map in a folder) I changed it so no map is selected when changing "folders"

Also some related fixes and refactoring:
- When no map was selected it was still drawn with size zero
- ctrlTable: Sorting behavior has now a better interface instead of some strange bool pointer

I have no idea how to unittest the dskSelectMap but did add some tests for table sorting

### DIFF
--- a/libs/s25main/controls/ctrlPreviewMinimap.cpp
+++ b/libs/s25main/controls/ctrlPreviewMinimap.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2005 - 2017 Settlers Freaks (sf-team at siedler25.org)
+// Copyright (c) 2005 - 2020 Settlers Freaks (sf-team at siedler25.org)
 //
 // This file is part of Return To The Roots.
 //
@@ -15,7 +15,6 @@
 // You should have received a copy of the GNU General Public License
 // along with Return To The Roots. If not, see <http://www.gnu.org/licenses/>.
 
-#include "rttrDefines.h" // IWYU pragma: keep
 #include "ctrlPreviewMinimap.h"
 #include "ogl/glArchivItem_Map.h"
 #include "libsiedler2/ArchivItem_Map_Header.h"
@@ -33,13 +32,16 @@ ctrlPreviewMinimap::ctrlPreviewMinimap(Window* parent, const unsigned id, const 
  */
 void ctrlPreviewMinimap::Draw_()
 {
+    if(mapSize == Extent::all(0))
+        return;
+
     // Button drumrum zeichnen
     Draw3DBorder(GetBoundaryRect(), TC_GREY, true);
 
     // Map ansich zeichnen
     DrawMap(minimap);
 
-    Extent playerPxlSize(4, 4);
+    const Extent playerPxlSize(4, 4);
     const DrawPoint basePos = GetDrawPos();
     // Startpositionen zeichnen
     for(const Player& player : players)
@@ -67,8 +69,8 @@ void ctrlPreviewMinimap::SetMap(const glArchivItem_Map* const s2map)
         return;
     }
 
-    unsigned short map_width = s2map->getHeader().getWidth();
-    unsigned short map_height = s2map->getHeader().getHeight();
+    const unsigned short map_width = s2map->getHeader().getWidth();
+    const unsigned short map_height = s2map->getHeader().getHeight();
     SetMapSize(Extent(map_width, map_height));
     minimap.SetMap(*s2map);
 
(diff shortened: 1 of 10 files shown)


libs/s25main/controls/ctrlPreviewMinimap.cpp
```diff
@@ -80,8 +82,8 @@ void ctrlPreviewMinimap::SetMap(const glArchivItem_Map* const s2map)
             // Startposition eines Spielers an dieser Stelle?
             if(s2map->GetMapDataAt(MAP_TYPE, x, y) != 0x80)
                 continue;
-            unsigned player = s2map->GetMapDataAt(MAP_LANDSCAPE, x, y);
-            if(player < MAX_PLAYERS)
+            const unsigned player = s2map->GetMapDataAt(MAP_LANDSCAPE, x, y);
+            if(player < players.size())
```
Flow86, line 86: I thought the numbers must not be continuous on some maps?

libs/s25main/desktops/dskSelectMap.cpp
```diff
@@ -172,12 +176,34 @@ void dskSelectMap::Msg_OptionGroupChange(const unsigned /*ctrl_id*/, unsigned se
         FillTable(ListDir(worldsPath, "wld"));
     }
 
+    if(brokenMapPaths.size() > numFaultyMapsPrior)
+    {
+        std::string errorTxt =
+          helpers::format(_("%1% map(s) could not be loaded. Check the log for details"), brokenMapPaths.size() - numFaultyMapsPrior);
```
Flow86, line 182: should be logged to the log too

libs/s25main/desktops/dskSelectMap.cpp
```diff
@@ -200,33 +224,32 @@ void dskSelectMap::Msg_TableSelectItem(const unsigned ctrl_id, const int selecti
...
-                preview.SetMap(map);
+                std::unique_ptr<glArchivItem_Map> map = loadAndVerifyMap(path);
+                RTTR_Assert(map);
+                preview.SetMap(map.get());
                 txtMapName.SetText(s25util::ansiToUTF8(map->getHeader().getName()));
                 txtMapPath.SetText(path);
                 btContinue.SetEnabled(true);
+            } catch(const std::runtime_error& e)
+            {
+                std::string errorTxt = helpers::format(_("Could not load map:\n%1%\n%2%"), path, e.what());
```
Flow86, line 243: same here

libs/s25main/desktops/dskSelectMap.cpp
```diff
@@ -425,14 +448,15 @@ void dskSelectMap::FillTable(const std::vector<bfs::path>& files)
             continue;
         // Karteninformationen laden
         libsiedler2::Archiv map;
-        if(libsiedler2::loader::LoadMAP(filePath.string(), map, true) != 0)
+        if(int ec = libsiedler2::loader::LoadMAP(filePath.string(), map, true))
+        {
+            LOG.write(_("Failed to load map %1%: %2%")) % filePath % libsiedler2::getErrorString(ec);
```
Flow86, line 453: newline missing on error message

libs/s25main/controls/ctrlPreviewMinimap.cpp
Flamefire: I'd be surprised by that. The editor only allows you to set them contiguous. But it doesn't matter for this code. This simply places a colored square at each position where a player HQ is set. I only changed the `MAX_PLAYERS` to `players.size()` where `players` is a `MAX_PLAYERS` sized array, so no change here. I only wanted to avoid the duplication and make it clear no OOB access is happening

BUT: Maybe we should validate that all players are set. For that I'd move the `Point` class to libutil and let an s2map in libsiedler2 return an array with player positions for each player in the map. This allows to validate in libsiedler2 that each player has a unique start position and in s25client that there are as many start positions as in the header and each is set. I'd do the latter in s25client because it might be that we want players without a start position. See related issue #935 where a mission seems to have that case

libs/s25main/desktops/dskSelectMap.cpp
Flamefire: The log already contains a message for each map. IMO logging this would be redundant

PR #1254, opened 2020-06-07
Use named constants instead of FILE_PATHS

Flamefire:
As started earlier this now uses organized, named constants instead of opaque indices.

The Babylonians overwrite folder had to be renamed to match the name used. IMO this is correct anyway as using the German name for the folder might be confusing.


libs/rttrConfig/src/files.h
```diff
@@ -24,132 +24,89 @@
...
-  /*101 */ "",                                  // unused
-  /*102 */ "<RTTR_GAME>/GFX/PICS/SETUP013.LBM", // Optionen
-  /*103 */ "<RTTR_GAME>/GFX/PICS/SETUP015.LBM"  // Freies Spiel
-}};
+RTTR_IGNORE_DIAGNOSTIC("-Wunused-variable")
 
-constexpr unsigned FILE_SPLASH_ID = 104;
+namespace s25 {
+namespace folders {
+    constexpr auto config = "<RTTR_CONFIG>";                   // Einstellungsordner
```
Flow86, line 31: the comments should be translated

libs/s25client/s25client.cpp
```diff
@@ -296,7 +298,7 @@ bool InitDirectories()
 #endif
     if(!oldSettingsDir.empty() && bfs::is_directory(oldSettingsDir))
     {
-        const std::string newSettingsDir = RTTRCONFIG.ExpandPath(FILE_PATHS[94]);
+        const std::string newSettingsDir = RTTRCONFIG.ExpandPath(s25::folders::config);
         if(bfs::exists(newSettingsDir))
         {
             s25util::error(std::string("Old and new settings directory found. Please delete the one you don't want to keep!\nOld: ")
```
Flow86, line 304: This text should be translateable (but separate issue)

libs/s25client/s25client.cpp
```diff
@@ -313,10 +315,13 @@ bool InitDirectories()
             return false;
         }
     }
+    const std::string soundLSTPath = RTTRCONFIG.ExpandPath(s25::folders::lstsUser) + "/SOUND.LST";
```
Flow86, line 318: this full file path should be added to the files-section too

libs/s25main/desktops/dskSelectMap.cpp
```diff
@@ -313,7 +315,7 @@ void dskSelectMap::Msg_TableChooseItem(const unsigned /*ctrl_id*/, const unsigne
 void dskSelectMap::CreateRandomMap()
 {
     // setup filepath for the random map
-    std::string mapPath = (bfs::path(RTTRCONFIG.ExpandPath(FILE_PATHS[48])) / "Random.swd").string();
+    std::string mapPath = (bfs::path(RTTRCONFIG.ExpandPath(s25::folders::mapsPlayed)) / "Random.swd").string();
```
Flow86, line 318: should perhaps also be a separate constant

libs/s25main/ingameWindows/iwAIDebug.cpp
```diff
@@ -119,10 +118,11 @@ iwAIDebug::iwAIDebug(GameWorldView& gwv, const std::vector<const AIPlayer*>& ais
     overlays->AddString("Borderland");
     overlays->AddString("Fish");
 
-    text = AddText(ID_Text, DrawPoint(15, 120), "", COLOR_YELLOW, FontStyle::LEFT | FontStyle::TOP | FontStyle::NO_OUTLINE, NormalFont);
-
     // Show 7 lines of text and 1 empty line
-    SetIwSize(Extent(GetIwSize().x, text->GetPos().y + 8 * text->GetFont()->getHeight()));
+    text =
```
Flow86, line 122: this should be a separate issue

libs/s25client/s25client.cpp
Flamefire: I'm planning on putting this into a function so it is clear what it does: migration from old folders to new (changed names, no longer used...)
I wouldn't put those into a global file as those are not meant to be used anywhere but here

libs/s25main/ingameWindows/iwAIDebug.cpp
Flamefire: Woops, yes. Slipped in as I worked on that before

libs/s25main/desktops/dskSelectMap.cpp
Flamefire: I'd say this is a path that is local to this code only. The folder is a reusable constant but the filename is pretty much arbitrary so I'd say it can stay here.

libs/s25main/ingameWindows/iwAIDebug.cpp
Flamefire: --> #1255

libs/s25client/s25client.cpp
Flamefire: Done in: https://github.com/Return-To-The-Roots/s25client/pull/1254/commits/45104bf8cf7b3f32c506a33387fe4b23aa3c2299

libs/s25client/s25client.cpp
Flamefire: I Implemented a change to this in https://github.com/Return-To-The-Roots/s25client/pull/1254/commits/45104bf8cf7b3f32c506a33387fe4b23aa3c2299

As the code executes before anything is loaded the string must not be translatable and LOG must not be used yet.

The new code does this and is nicely extendable if we chose to do the renaming proposed in Discord

libs/s25client/s25client.cpp
```diff
@@ -278,45 +280,85 @@ void SetAppSymbol()
...
+    const std::vector<MigrationEntry> migrations = {
 #ifdef _WIN32
-    oldSettingsDir = RTTRCONFIG.ExpandPath("~/Siedler II.5 RttR");
+      {"~/Siedler II.5 RttR", s25::folders::config, true},
 #elif defined(__APPLE__)
-    oldSettingsDir = RTTRCONFIG.ExpandPath("~/.s25rttr");
+      {"~/.s25rttr", s25::folders::config, true},
 #endif
-    if(!oldSettingsDir.empty() && bfs::is_directory(oldSettingsDir))
+      {std::string(s25::folders::lstsUser).append("/SOUND.LST"), "", false},
```
Flow86, line 299: we should add the old SDL drivers here too

Flamefire: Done although it is a bit dirty due to os-specific extensions

libs/s25client/s25client.cpp
```diff
@@ -278,45 +280,95 @@ void SetAppSymbol()
...
+            {
+                const bfs::path newName = RTTRCONFIG.ExpandPath(entry.newName);
+                if(exists(newName))
+                {
+                    throw std::runtime_error(
+                      helpers::format("Old and new %1% found. Please delete the one you don't want to keep!\nOld: %2%\nNew: %3%",
+                                      is_directory(oldName) ? "directory" : "file", oldName, newName));
+                }
+                boost::system::error_code ec;
+                rename(oldName, newName, ec);
```
Flow86, line 338: perhaps we should be verbose here, that the user nows whats going on.

libs/s25client/s25client.cpp
```diff
@@ -278,45 +280,95 @@ void SetAppSymbol()
...
+            if(!exists(oldName))
+                continue;
+            if(entry.isFolder && !is_directory(oldName))
+                throw std::runtime_error(helpers::format("Expected folder but %1% is not a folder", oldName));
+            else if(!entry.isFolder && !is_regular_file(oldName))
+                throw std::runtime_error(helpers::format("Expected file but %1% is not a file", oldName));
+            if(entry.newName.empty())
+            {
+                boost::system::error_code ec;
+                if(!remove(oldName, ec))
```
Flow86, line 326: perhaps we should be verbose here, that the user nows whats going on.

libs/s25client/s25client.cpp
Flamefire: done: https://github.com/Return-To-The-Roots/s25client/pull/1254/commits/ddc68f779ab096f1b7aff5b7501c28ac8d83fd09

libs/s25client/s25client.cpp
Flamefire: https://github.com/Return-To-The-Roots/s25client/pull/1254/commits/ddc68f779ab096f1b7aff5b7501c28ac8d83fd09

PR #1258, opened 2020-06-13
Improve sound conversion

Flamefire:
Use the new loadMapping function from libsiedler2 and adapt the conversion script

Found also a bug: It used the wave header instead of sound.scs for getting the source samplerate.

To help with improving the sounds I added a command line switch `--convert-sounds` which converts the sounds into wave files into `convertedSoundEffects` inside the user settings folder

### DIFF
--- a/libs/s25main/convertSounds.cpp
+++ b/libs/s25main/convertSounds.cpp
@@ -17,42 +17,32 @@
 #include "convertSounds.h"
 #include <libsiedler2/Archiv.h>
 #include <libsiedler2/ArchivItem_Sound_Wave.h>
-#include <boost/nowide/fstream.hpp>
+#include <libsiedler2/loadMapping.h>
+#include <s25util/StringConversion.h>
 #include <cmath>
 #include <samplerate.hpp>
 #include <sstream>
 #include <stdexcept>
 #include <vector>
 
-namespace bnw = boost::nowide;
-
-bool convertSounds(libsiedler2::Archiv& sounds, const bfs::path& scriptPath)
+void convertSounds(libsiedler2::Archiv& sounds, const boost::filesystem::path& scriptPath)
 {
-    bnw::ifstream file(scriptPath); // script
     samplerate::State converter(samplerate::Converter::SincFastest, 1);
     std::vector<float> input, output;
-    std::string line;
-    const unsigned targetFrequency = 44100;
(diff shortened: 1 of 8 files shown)


libs/s25main/convertSounds.cpp
```diff
@@ -17,42 +17,32 @@
...
-            continue;
         if(header.numChannels != 1)
-            throw std::runtime_error("Unexpected number of channels for item " + std::to_string(item));
+            throw std::runtime_error("Unexpected number of channels for item " + std::to_string(idx));
         if(header.frameSize != 1 || header.bitsPerSample != 8)
-            throw std::runtime_error("Unsupported format for item " + std::to_string(item));
+            throw std::runtime_error("Unsupported format for item " + std::to_string(idx));
         converter.reset();
-        double rate = static_cast<double>(targetFrequency) / sound->getHeader().samplesPerSec;
+        const double rate = static_cast<double>(targetFrequency) / frequency;
```
Flow86, line 45: this was wrong all the way through already on the old code?

Flamefire: No, in https://github.com/Return-To-The-Roots/s-c/blob/c605631fc69a51a3a0ec5634e9a33b9fcdb8cc3c/src/sound-convert.cpp it is correct. Seems to be a mistake on translating it to libsamplerate

Flow86: so the header has still the wrong samplerate then?

Flow86:
```
header.samplesPerSec = frequency;
header.bytesPerSec = frequency * blockAlign;
```

Flamefire: The header is the one that libsiedler2 returns which is a hardcoded guess. So yes potentially wrong samplerate.

After conversion it is (now) correct by using the sound.scs and setting the header below

Flow86: is it reset below?

Flow86: but `header.bytesPerSec` i set wrongly

Flamefire: `blockAlign` (as it was called, now `frameSize`) is `1`, so the values are correct. Added #1260 to make that clearer

PR #1259, opened 2020-06-14
Test language files and fix russian language

Flamefire:
Fixes #1257 and avoids it in the future

The test is not very solid as the determination what a format string is is a bit heuristic. But for now it works well and it found all issues in the russian language file

PR #1261, opened 2020-06-27
Allow overriding link mapping for bob files

Flamefire:
This allows to create `*.links` (or txt) files to append or overwrite mappings for bobs (e.g. soldiers and scouts)

For this the job-consts had to be changed so the job-bob-id is nation specific at this level and all code consistently uses either this or the cache.

It removes the proprietary files (which were unused) and adds a mapping.links file which uses Roman graphics for the Babylonian soldiers/scout. So @Spikeone can continue here ;)

What is still missing: Way to override the fight animations. Those are in ROM_BOBS.LST and constants in FIGHT_ANIMATIONS (MilitaryConsts.h).   
Idea: To avoid including placeholder graphics we could change the loaded to check those for every nation and use the Roman graphics (copy the elements to the expected positions) when the required ones are missing.

- [x] Conflicts with #1259 due to mygettext submodule changes. Merge the other one first, then rebase this!


data/RTTR/LSTS/GAME/Babylonians/jobs.bob/mapping.links
```diff
@@ -0,0 +1,585 @@
+# Soldier and scout replacements for Babylonians
```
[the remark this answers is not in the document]
Flamefire: Done: https://github.com/Return-To-The-Roots/s25client/pull/1261/commits/cb8a7902c86d3991d1f0ad500bdb2af9a5ca3023#diff-5ce90d43b502645f7415ead79310a5f8R2-R5

Is that enough?

Flamefire:
I did a good refactoring of the BOB file class. It was really annoying that the "links" had to have `96` added because we placed the bodies at those indices. This would have been hard to explain in a short file header comment. Hence I modified libsiedler2 so this is no longer required and the links get the 96 added directly on loading.

Another good thing: As the soldiers and scouts have no fat version (Why did they use the bob format for the jobs.bob anyway? The links array is half wasted) I removed those entries from the links file which makes it smaller and more readable.

The remaining refactoring was basically cleaning up of things that I found hard to understand/potentially confusing. It is now much easier to use and read the BOB file (class)

Flamefire:
@Flow86 ready

libs/s25main/Loader.cpp
```diff
@@ -604,15 +604,8 @@ void Loader::fillCaches()
...
                     const libsiedler2::ImgDir imgDir = toImgDir(dir);
 
-                    unsigned overlayIdx = bob_jobs->getOverlayIdx(id, fat, imgDir, ani_step);
-                    if(overlayIdx == 188 && fat)
-                    {
-                        // No fat version(?)
-                        overlayIdx = bob_jobs->getOverlayIdx(id, false, imgDir, ani_step);
-                        fat = false;
-                    }
                     bmp.add(dynamic_cast<glArchivItem_Bitmap_Player*>(bob_jobs->getBody(fat, imgDir, ani_step)));
```
Flow86, line 607: was this a bug with non-gold-edition files? (i.e 1.55+mission-cd) - anyhow we could fix it with an override file then :-D

Flamefire: No idea that's why I asked. With gold edition it works fine so no problem here ;)

Flow86: perhaps it was some different bug with animation system or so, which one worked around a while ago.

PR #1262, opened 2020-07-03
Handle error on loading game resources

### LINKED ISSUE #1244: Failing to load resources does not abort map loading
The loader returns a bool for LoadFilesAtGame but it is ignored by the game loader. This should probably be handled although it is unlikely to occur.

Flamefire:
Fixes #1244

This can occur when the mapping.links override has the wrong format


Flamefire:
@Flow86 Can you merge this please. Travis ran through but is hanging with notifications

PR #1263, opened 2020-07-03
 Use new IDs for the fighting soldiers of babylonians

Flamefire:
Fallback to romans if they don't exist
Also refactoring to use FigAnimationSprites which is common enough to justify a new type

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -42,19 +42,19 @@ endif()
 
 option(RTTR_USE_SYSTEM_LIBSAMPLERATE "Use system installed libsamplerate. Fails if not found!" "${RTTR_USE_SYSTEM_LIBS}")
 if(RTTR_USE_SYSTEM_LIBSAMPLERATE)
-	find_package(Samplerate 0.1.9 REQUIRED)
-	unset(samplerate_LIB_DEPENDS CACHE)
-	add_library(samplerate INTERFACE)
-	target_link_libraries(samplerate INTERFACE Samplerate::Samplerate)
+    find_package(Samplerate 0.1.9 REQUIRED)
+    unset(samplerate_LIB_DEPENDS CACHE)
+    add_library(samplerate INTERFACE)
+    target_link_libraries(samplerate INTERFACE Samplerate::Samplerate)
 else()
     include(FetchContent)
     FetchContent_Declare(
         LibSamplerate
-        GIT_REPOSITORY https://github.com/erikd/libsamplerate
-        GIT_TAG        727504026ee51c6fe29afe6255b70513504a6265
+        GIT_REPOSITORY https://github.com/Flamefire/libsamplerate
+        GIT_TAG        b5d39fbfc86b656fed68c935ac7b0a06349e4aa1
     )
-	set(LIBSAMPLERATE_TESTS OFF CACHE INTERNAL "")
-	set(LIBSAMPLERATE_INSTALL OFF CACHE INTERNAL "")
+    set(LIBSAMPLERATE_TESTS OFF CACHE INTERNAL "")
+    set(LIBSAMPLERATE_INSTALL OFF CACHE INTERNAL "")
     FetchContent_MakeAvailable(LibSamplerate)
 endif()
 
--- a/libs/s25main/Loader.cpp
+++ b/libs/s25main/Loader.cpp
@@ -40,6 +40,7 @@
 #include "gameTypes/Direction.h"
 #include "gameTypes/DirectionToImgDir.h"
 #include "gameData/JobConsts.h"
+#include "gameData/MilitaryConsts.h"
 #include "gameData/NationConsts.h"
 #include "libsiedler2/ArchivItem_Font.h"
 #include "libsiedler2/ArchivItem_Palette.h"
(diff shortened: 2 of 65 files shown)


external/CMakeLists.txt
```diff
@@ -42,19 +42,19 @@ endif()
...
+    unset(samplerate_LIB_DEPENDS CACHE)
+    add_library(samplerate INTERFACE)
+    target_link_libraries(samplerate INTERFACE Samplerate::Samplerate)
 else()
     include(FetchContent)
     FetchContent_Declare(
         LibSamplerate
-        GIT_REPOSITORY https://github.com/erikd/libsamplerate
-        GIT_TAG        727504026ee51c6fe29afe6255b70513504a6265
+        GIT_REPOSITORY https://github.com/Flamefire/libsamplerate
```
Flow86, line 53: we should put it to the common RttR location

libs/s25main/Loader.cpp
```diff
@@ -613,14 +614,59 @@ void Loader::fillCaches()
...
 
-        bmp.add(GetNationPlayerImage(nation, 0));
-        bmp.addShadow(GetNationImage(nation, 1));
+        libsiedler2::Archiv& romBobs = GetArchive("rom_bobs");
+        const auto getAlternative = [&romBobs](unsigned id, unsigned altId) {
+            auto* bmp = convertChecked<glArchivItem_Bitmap_Player*>(romBobs[id]);
+            return bmp ? bmp : convertChecked<glArchivItem_Bitmap_Player*>(romBobs[altId]);
+        };
+        // Special handling for non-native nats: Use roman animations if own are missing
+        const unsigned fallbackNation = nation < NUM_NATIVE_NATS ? nation : NAT_ROMANS;
```
Flow86, line 633: can we perhaps rename NATS to NATIONS? `NUM_NATIVE_NATIONS` would make it more readable

external/CMakeLists.txt
Flamefire: Why? This is a fork I maintain until my PRs are accepted into upstream. It is unrelated to Rttr and I would rather not spend even more work on that than I already have

libs/s25main/Loader.cpp
Flamefire: Good point. Done for this and `NUM_NATS`

external/CMakeLists.txt
Flow86: ah okay. It looked like we have a special adapted libsamplerate in there.

Flamefire: No I've just been waiting for a year to get those PRs merged and decided to make my patches available for others to use

Flamefire:
Done for now. Remaining TODO: Remove GD_INVALID and GD_NOTHING, see 92f1f4a .

Note that I'm a fan of optional and variant but the current solution is not perfect. IIRC those are based on the pre-C++11 stuff and hence might pull in more dependencies than required (e.g. boost.type_traits), but well... For variant there is Boost.Variant2 in Boost 1.71 which is superior but that's quite a big leap

stefson:
With these changes merged, I can't load savegames from current master anymore: 

```
SERVER: Using gameframe length of 50ms
SERVER: Using networkframe length of 1 GFs (50ms)
SerializedGameData::PopObject_: ERROR: After loading Object(obj_id = 5.788, got = 2); Code is wrong!
Error when loading game: Invalid safety code after PopObject
server state changed to stop
```

Flamefire:
@stefson great find! This is either a bug introduced here or surfacing here. I didn't change any serialization of saved games knowingly, only replays. Can you attach the savegame for me to have a look?

stefson:
sure, here is one example: 
[Afrika.zip](https://github.com/Return-To-The-Roots/s25client/files/4879380/Afrika.zip)

the error isn't very verbose though :/

Flamefire:
Ok found. I accidentally changed the number of ware types, fixed that. We should probably also address the GD_NOTHING issue: We have a "good" which is not a good in the enum so at usage site it should be an optional if it can be nothing. But this likely changes the serialization of saved games. I could workaround this be defining `MaxEnumValue_v<Good>` in the saved game as `None` (for the optional) which incidentally works because that's the current value of GD_NOTHING. Might be even a good idea in general to save the bool but fails if we add more wares... But I guess that is almost impossible anyway

stefson:
I just wanted to note you my old savegame is now working again with your latest changes to the branch.

tests/s25Main/integration/testTrading.cpp
```diff
@@ -26,6 +24,7 @@
 #include "gameData/JobConsts.h"
 #include <rttr/test/LogAccessor.hpp>
 #include <boost/test/unit_test.hpp>
+#include <boost/variant/variant.hpp>
```
Flow86, line 27: where is this variant used?

libs/s25main/nodeObjs/noFighting.cpp
```diff
@@ -313,19 +296,13 @@ void noFighting::StartAttack()
         switch(gwg->GetGGS().getSelection(AddonId::ADJUST_MILITARY_STRENGTH))
         {
             case 0: // Maximale Stärke
-            {
                 results[i] = RANDOM.Rand(__FILE__, __LINE__, GetObjId(), soldiers[i]->GetRank() + 6);
-            }
-            break;
+                break;
```
Flow86, line 300: I don't like the different style here?
shouldn't be the `break` consistent - either on one line or on multiple lines?

tests/s25Main/integration/testTrading.cpp
Flamefire: in the call to `this->TradeOverLand(players[0]->GetHQPos(), GD_BOARDS, 2);`
The `GD_BOARDS` implicitely constructs a variant, but it is only forward declared in the header, so users need to include it. This is done because the variant header is huge.

libs/s25main/nodeObjs/noFighting.cpp
Flamefire: Only the default case is different. Clang-Format detects that it is a short line and puts it all in one line.

I guess we could move the comment to after the default: so clang-format has to use multiple lines. But well, not really keen on doing that now as it will retrigger a whole CI cycle and I don't think it's worth it. Next time we touch this...

PR #1264, opened 2020-07-13
Remove rttrDefines and similar "forced" includes

Flamefire:
Avoids overhead where not required
Requires to declare what is needed

Also it is an unnecessary step required for new files and might be confusing to newcomers.

Also some more unique_ptrs to avoid the deletePtr usages


Flow86:
probably you forgot to include `#include <memory>` ? for those unique_ptr stuff?

Flamefire:
Correct. Tested locally and found a new issue with clang-9  -.- Should work now

PR #1267, opened 2020-07-17
Use Clang format 9

### LINKED ISSUE #1265: Update clang-format
We are currently using clang-format 5.0 which is pretty dated now. It also isn't possible for developers to use a newer one as different versions format differently and they may not even understand the clang-format style file due to changes.

Furthermore Ubuntu 20 / Linux Mint 20 doesn't even have clang-format 5 in its sources anymore. There it starts at 6.

I hence propose that we switch to a newer version. 10 is the current one, if we want to be conservative we could use 9, which should be available in most repos.

The update is almost trivial (changing versions at 2-3 places and rerun the format target) but may lead to conflicts for open PRs and ongoing work. To mitigate for changes where a rebase is not feasible the author can reformat the code with the new clang-format then merge the master to his branch so conflicts will be minimal if at all.

I suggest to do it before the release so it is a clean point

Flamefire:
Closes #1265


.clang-format
```diff
@@ -1,26 +1,30 @@
...
 Language:        Cpp
-BasedOnStyle:  LLVM
 AccessModifierOffset: -4
 AlignAfterOpenBracket: Align
+AlignConsecutiveMacros: false
 AlignConsecutiveAssignments: false
 AlignConsecutiveDeclarations: false
 AlignEscapedNewlines: Left
-AlignOperands: true
+AlignOperands:   true
```
Flow86, line 9: easter egg adding a wrongly formatted formatter config file? :-D

.clang-format
```diff
@@ -38,12 +43,13 @@ BraceWrapping:
...
 BreakBeforeBraces: Custom
 BreakBeforeInheritanceComma: false
+BreakInheritanceList: AfterColon
 BreakBeforeTernaryOperators: false
 BreakConstructorInitializersBeforeComma: false
 BreakConstructorInitializers: BeforeColon
 BreakAfterJavaFieldAnnotations: false
 BreakStringLiterals: true
-ColumnLimit: 140
+ColumnLimit:     140
```
Flow86, line 52: same here

libs/s25main/desktops/dskGameInterface.cpp
```diff
@@ -689,7 +689,8 @@ bool dskGameInterface::Msg_KeyDown(const KeyEvent& ke)
         case KT_F9: // Readme
             WINDOWMANAGER.ToggleWindow(std::make_unique<iwTextfile>("readme.txt", _("Readme!")));
             return true;
-        case KT_F10: {
+        case KT_F10:
+        {
```
Flow86, line 693: I think for this there is an option to indent the case labels correctly

libs/s25main/figures/noFigure.cpp
```diff
@@ -397,7 +393,9 @@ void noFigure::HandleEvent(const unsigned id)
         switch(fs)
         {
             case FS_GOHOME:
-            case FS_GOTOGOAL: { WalkToGoal();
+            case FS_GOTOGOAL:
+            {
```
Flow86, line 397: IndentCaseLabels=false and probably  IndentCaseLabels=false ?

.clang-format
Flamefire: Nope, that is what `clang-format-9 --dump-config` outputs. Don't want to touch that for easier updates

Flow86: strange enough

libs/s25main/figures/noFigure.cpp
Flamefire: C&P bug in your message? ;)

libs/s25main/desktops/dskGameInterface.cpp
Flamefire: What do you mean by "correctly"? They are correct aren't they? An opening brace is on a new line on the same level as the outer code and code inside braces is indented by 1 level. Now all is perfectly fine

libs/s25main/figures/noFigure.cpp
Flow86: second one should be `IndentCaseBlocks=false` ;-)

Flamefire: That doesn't exist in 9: https://releases.llvm.org/9.0.0/tools/clang/docs/ClangFormatStyleOptions.html

Anyway I'd keep it like it is as it's the most consistent and IMO most readable.

Flamefire:
Can this be merged ASAP please? It will conflict with other PRs which I'd like to avoid :)

Issue #1185, opened 2020-01-20
s25client fails to compile with clang-10 in libsiedler2/src/FileEntry.cpp:44:21: error: loop variable 'it' is always a copy because the range of type


stefson:
first of all: this version of clang is a snapshot from current git, I bootstrapped it out of curiosity. So yeah, you can discard this if you want to and claim it's a bug in clang, but you still might want to check the error: 

```
cd /var/tmp/portage/games-strategy/s25rttr-9999-r4/work/s25rttr-9999_build/external/libsiedler2 && /usr/lib/llvm/10/bin/clang++-10 -DBOOST_NOWIDE_USE_FILESYSTEM -I/var/tmp/portage/games-strategy/s25rttr-9999-r4/work/s25rttr-9999/external/libsiedler2/include/libsiedler2 -I/var/tmp/portage/games-strategy/s25rttr-9999-r4/work/s25rttr-9999/external/libutil/libs/common/include -I/var/tmp/portage/games-strategy/s25rttr-9999-r4/work/s25rttr-9999/external/libendian/include -I/var/tmp/portage/games-strategy/s25rttr-9999-r4/work/s25rttr-9999/external/libutil/nowide/include   -Og -ggdb   -mfpmath=sse -msse2 -ffast-math -fomit-frame-pointer -Wall -pedantic -Wextra -Werror -Wno-unused-command-line-argument -fno-strict-aliasing -pedantic-errors -Qunused-arguments -Wcast-qual -Wdisabled-optimization -Wfloat-conversion -Wformat-nonliteral -Wformat-security -Wformat=2 -Wimport -Winit-self -Wmissing-declarations -Wmissing-field-initializers -Wmissing-format-attribute -Wmissing-include-dirs -Wmissing-noreturn -Wpacked -Wparentheses -Wpedantic -Wpointer-arith -Wstrict-aliasing=2 -Wundef -Wunused -Wunused-parameter -Wwrite-strings -Wno-unknown-pragmas -Wctor-dtor-privacy -Woverloaded-virtual -Wno-error=inconsistent-missing-override -Wredundant-decls -o CMakeFiles/siedler2.dir/src/StandardAllocator.cpp.o -c /var/tmp/portage/games-strategy/s25rttr-9999-r4/work/s25rttr-9999/external/libsiedler2/src/StandardAllocator.cpp
[...]

PR #1274, opened 2020-07-25
Improve docu and tests, add rttr:FormatNumGFs

Flamefire:
This bundles some improvements to the documentation, tests and adds the requested rttr:FormatNumGFs from #1185 

- Add a `doc` folder for documentation
- Convert the Lua doc from the wiki to files and fix references
- Ensure consistent Markdown via linter
- Fix a minor issue where lua code is ran twice for the equality check which can make mocks fail
- Refactor Lua interfaces to work with a new ILocalGameState which avoids some hacks required for testing and is the base for the  new feature which would otherwise be untestable
- Replace usage of Boost features with std features and Rttr features with Boost features where they exist

### DIFF
--- a/libs/s25main/network/GameClient.cpp
+++ b/libs/s25main/network/GameClient.cpp
@@ -296,7 +296,7 @@ void GameClient::StartGame(const unsigned random_init)
 
     GameWorld& gameWorld = game->world_;
     if(mapinfo.savegame)
-        mapinfo.savegame->sgd.ReadSnapshot(game);
+        mapinfo.savegame->sgd.ReadSnapshot(game, *this);
     else
     {
         RTTR_Assert(mapinfo.type != MAPTYPE_SAVEGAME);
@@ -304,7 +304,7 @@ void GameClient::StartGame(const unsigned random_init)
         for(unsigned i = 0; i < gameWorld.GetNumPlayers(); ++i)
             gameWorld.GetPlayer(i).MakeStartPacts();
 
-        if(!gameWorld.LoadMap(game, mapinfo.filepath, mapinfo.luaFilepath))
+        if(!gameWorld.LoadMap(game, *this, mapinfo.filepath, mapinfo.luaFilepath))
         {
             OnError(CE_INVALID_MAP);
             return;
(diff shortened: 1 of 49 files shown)


libs/s25main/network/GameClient.cpp
```diff
@@ -758,7 +758,7 @@ bool GameClient::OnGameMessage(const GameMessage_Chat& msg)
 {
     if(msg.destination == CD_SYSTEM)
     {
-        SystemChat(msg.text, msg.player);
+        SystemChat(msg.text, (msg.player < game->world_.GetNumPlayers()) ? msg.player : GetPlayerId());
```
Flow86, line 761: shouldnt this the "system-player"? or do I understand correctly that this is then then the local player?

cant we simply write `SystemChat(msg.text)` here?

libs/s25main/network/GameClient.cpp
```diff
@@ -1577,18 +1572,20 @@ void GameClient::SkipGF(unsigned gf, GameWorldView& gwv)
...
 
-void GameClient::SystemChat(const std::string& text, unsigned char player)
+void GameClient::SystemChat(const std::string& text)
 {
-    if(!ci)
-        return;
-    if(player == 0xFF)
-        player = mainPlayer.playerId;
-    ci->CI_Chat(player, CD_SYSTEM, text);
+    SystemChat(text, mainPlayer.playerId);
```
Flow86, line 1577: whats the difference between `mainPlayer.playerId` and `GetPlayerId()`?

libs/s25main/network/GameClient.cpp
Flamefire: We have 1 example were this is used: When a player saves the game (which will cause a lag) he sends a message to be displayed to all players. In that case `msg.player` is set to that player and must be used. However we can also just send a message which should be displayed as `SystemChat(msg.text)` which is equivalent to `SystemChat(msg.text, GetPlayerId());` and done by setting `msg.player=0xFF`.

Kinda unusual but that's what it is currently.

libs/s25main/network/GameClient.cpp
Flamefire: Nothing: `unsigned GetPlayerId() const override { return mainPlayer.playerId; }`

Flow86: then we should probably use `GetPlayerId()` there too, so its clear which one is used

Flamefire: Done: 7d27ec83c

PR #1275, opened 2020-07-26
Allow dot in showVersion test

chewi:
The existing regular expression seems to expect that the test will
only ever be run from a git checkout. Gentoo would like to eventually
run it from a release source tarball.

### DIFF
--- a/tests/s25client/CMakeLists.txt
+++ b/tests/s25client/CMakeLists.txt
@@ -7,7 +7,7 @@ set_tests_properties(s25client_showHelp PROPERTIES
 add_test(NAME s25client_showVersion COMMAND s25client --version
          WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
 set_tests_properties(s25client_showVersion PROPERTIES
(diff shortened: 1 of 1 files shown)


Flow86:
I thought the release tar ball also contains a revision file?

Flamefire:
Ok, there are 2 changes here: First one (correctly) allows versions like `0.8.1` instead of only the nightly format of `20200715` and the second one removes the need to have a revision.

@chewi Why would you want to remove the revision? And how would you even do that? Do you have custom patches? I'm asking because if there is a real need for this we can have that upstream to not require any patches. In general: If you patch anything let us know and we'll find a way to avoid that.   
Note that the revision can be set via `-DRTTR_REVISION` or by having a `revision.txt` file which the official release source archives will have.

chewi:
Ah, sorry for being dumb here. Initially I did just add the dot and submitted that here, then quickly assumed there would be no revision in the release tarball and panicked.

DON'T PANIC

:sweat_smile:

Flow86:
ah yes. the "." should be allowed in there! :+1:

PR #1276, opened 2020-07-28
Fix wrong image used for workers walking with ware

Flamefire:
Regression from f3257cfca78d8c9ce589726c7f87ade6a3592276


libs/s25main/figures/nofWellguy.h
```diff
@@ -32,7 +32,7 @@ class nofWellguy : public nofWorkman
     void DrawWorking(DrawPoint drawPt) override;
     /// Gibt die ID in JOBS.BOB zurück, wenn der Beruf Waren rausträgt (bzw rein)
     // TODO:der Brunnentyphat keine ID in JOBS.BOB
```
Flow86, line 34: do you want to translate the comment to english? (and perhaps add a name for the `100` (`JOBS_TO_CARRIER_ID_OFFSET` ?)

Flamefire: Done

Flamefire:
@Flow86 Fixed a warning on MSVC. Good to go now

PR #1277, opened 2020-08-01
Enum refactoring

Flamefire:
The aim is to increase the (type) safety of RttR by using enum classes. In the current codebase there are lot's of `unsigned char`s and similar floating around with magic numbers for "invalid" or special purpose. At some places some directions are valid, but others are not and conversions are done by adding/subtracting numbers.   
This is hard to understand and possible to get wrong.

Solution:

- Use OptionalEnum and dedicated (enum) types to express correct semantics and represent valid values uniquely
- Use EnumArray, (Enum)Range for typesafe range-based for loops
- Convert between direction types with enforced correctness

TODO: 

- [ ] Make Direction a regular strong enum to remove the `is_enum` workarounds
- [ ] Convert ShipDirection
- [ ] Convert Species
- [ ] Convert Visibility
- [ ] Make sure enums are contiguous (->mainly the MaxEnumValue enabled ones)

Note: This can be merged but I can also finish my work on this. However due to the size of this PR subsequent PRs could be better suited and this should be merged soon to avoid conflicts with others.

This PR got so big due to the introduction of the new enums and removal of `Direction.toUInt` and `fromInt` which led to most places where "magic" was done. Using the `EnumRange` (Search&Replace over the codebase) removed many of them, as did `EnumArray`.

Most other replacements are magic numbers which no longer work e.g. `ware->GetNextDir() == 1` --> `ware->GetNextDir() == RoadPathDirection::NorthWest`

I did pay attention to maintain savegame compatibility so existing savegames and replays will continue to work. In fact no functional changes should have happened and performance should not be impacted (and potentially even improve due to guaranteed conversions instead of range checks). Some work is required to polish the types which requires changing the savegames etc. E.g. Making the enums contiguous (easy: Mapping old savegame values to new ones) and we could consider matching the `Direction` enum with S2 (we have an offset of 3 which leads to required conversions).   
I'd also like to move the MaxEnumValue to libutil and check the range in the deserialization of enums (See SerializedGameData.PopEnum)


Flamefire:
@Flow86 Can this be merged? Would rather sooner than later (see above)

Also from Jenkins: 
> CMake Error: The source directory "/source" does not appear to contain CMakeLists.txt

Flow86:
I had to reinstall my server, jenkins builds are a bit broken yet. I need to check that

PR #1283, opened 2020-08-10
Require MSVC 2017+

### LINKED ISSUE #1282: Drop support for MSVC 2015
Continuation of #897 and https://github.com/Return-To-The-Roots/s25client/issues/897#issuecomment-480494023

C&P:

```
Coming across this again. The "constexpr does not imply const in C++14" comes up again. The C4814 is not even listed on the MS pages -.-

Although it is not really required playing with constexpr would be great. This needs MSVC 2017 which also has the /std:c++14 flag with improved compatibility. [@Flow86 ] Do you have any references on the "most used VS"? Is it really required with MSVC 2019 just released and 2017 community edition available?
```

Others found the same issue: https://gitmemory.com/issue/Return-To-The-Roots/s25client/897/480494023

Being able to "just use C++14" would be nice and will avoid having to fix valid C++14 code to be compatible with the C++11/14 mix of MSVC 2015

Is MSVC 2015 still relevant in 2020?

Flamefire:
Closes #1282

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ Then just run (tests or application) as usual.
 
 - cmake (i.e from <http://www.cmake.org/download/>)
 - boost (i.e from <http://www.boost.org/>)
-- Visual Studio (at least 2015, you can get the community edition for free)
+- Visual Studio (at least 2017, you can get the community edition for free)
 - Git Client (i.e TortoiseGit)
 
 #### Steps
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,11 +1,12 @@
 version: 0.8.3.{build}
+
 branches:
   only:
   - master
 
 image:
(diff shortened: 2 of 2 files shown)


Flow86:
but this currently only changes the ci pipelines, no cmake c++ version changes?

Flamefire:
It also updates the requirements in the readme. So after this we can make full use of C++14.

In CMake we already set C++14 as the requirement. I searched for "(MSVC )2015" and found no instances, if you see any code that has workarounds or so for MSVC 2015 which can be removed, tell me.

I didn't go and started updating code to use C++14 features where we couldn't before because how can I search for that? That's got to be incremental over time.

Flamefire:
@Flow86 Agreed or is anything missing?

PR #1286, opened 2020-08-12
Fix musicplayer

### LINKED ISSUE #1278: Removing custom tracks results in strange glyphs
Simply add a custom track and remove it using the interface:
grafik

### LINKED ISSUE #1281: Custom tracks not saved
Possibly related to #1280 but if you add custom music tracks to the music player, they wont be saved and you'll have to readd them.

Flamefire:
This combines some refactoring in order to fix some issues:

- Use optional to be able to distinguish between selected item and "nothing selected". This forces to check for that and makes the semantic clear. This was likely the cause for and fixes #1278 : Integer promotion led to not detecting the invalid value
- Fix #1281 : The playlist was saved in the current folder, not where it should be
- Add a button to manually save the playlist as all progress will be lost when switching between playlists. It is still saved when closing the window
- As the default playlist cannot be saved disable the buttons to do so
- We were saving the playlists into the program folder which is wrong and potentially read-only. Create new folder in user data for that.
- Refactor the playlist class to remove the circular reference to the iwMusicPlayer. This allows to add tests for it

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -29,6 +29,7 @@ Checks: >-
   -readability-named-parameter,
   -readability-static-accessed-through-instance,
   -readability-uppercase-literal-suffix,
+  -readability-identifier-naming,
   -clang-diagnostic-inconsistent-missing-override
 CheckOptions:
   - key: performance-for-range-copy.AllowedTypes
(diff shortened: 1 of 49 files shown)


Flow86:
```
/source/libs/s25main/controls/ctrlTable.cpp:301:28: error: comparison of unsigned expression >= 0 is always true
```

Flamefire:
Fixed

PR #1289, opened 2020-08-15
Make SoundHandle a RAII wrapper around a RawSoundHandle

### LINKED ISSUE #1280: Game crashes on closing with custom tracks
When you load a custom track in the music player and close the game, the game will crash uppon closing

Flamefire:
Previously we returned an allocated struct from the driver which the driver can change (to mark an unloaded sound) but which is deallocated from the exe
This is a problem with e.g. MSVC which may have separate heaps so all instances allocated from a DLL are deleted when the DLL is unloaded. This results in a dangling pointer.

Solution: Return an opaque DriverHandle (void*) from the DLL which is allocated and freed by the driver only. Then wrap this ASAP into a SoundHandle which calls the driver to free it on destruction of the last SoundHandle reference and also calls a register function for the contained pointer to have the driver reset the driverData pointer once the sound is freed to avoid calling into the driver when it is unloaded already.

Fixes #1280


extras/audioDrivers/SDL/AudioSDL.cpp
```diff
@@ -101,15 +101,21 @@ bool AudioSDL::Initialize()
         std::cerr << Mix_GetError() << std::endl;
         return false;
     }
-    SetNumChannels(Mix_AllocateChannels(MAX_NUM_CHANNELS));
+    SetNumChannels(Mix_AllocateChannels(DEFAULT_NUM_CHANNELS));
     Mix_SetMusicCMD(nullptr);
+    currentInstance = this;
```
Flow86, line 106: shouldn't the instance deinitialized on unload?

extras/audioDrivers/SDL/AudioSDL.cpp
```diff
@@ -101,15 +101,21 @@ bool AudioSDL::Initialize()
...
     Mix_HookMusicFinished(AudioSDL::MusicFinished);
 
     initialized = true;
 
     return true;
 }
 
+void AudioSDL::MusicFinished()
+{
+    currentInstance->driverCallback->Msg_MusicFinished();
```
Flow86, line 116: shouldn't we check on nullptr?

extras/audioDrivers/SDL/AudioSDL.cpp
```diff
@@ -119,6 +127,7 @@ void AudioSDL::CleanUp()
     AudioDriver::CleanUp();
 
     Mix_CloseAudio();
+    currentInstance = nullptr;
```
Flow86, line 130: should we check if its the same instance, not that we clean up the wrong one?

Flamefire: If you have 2 instances active and running here you'll have bigger problems I guess. Mix_CloseAudio() likely cancels everything already so there won't be any calls to the callback and hence this won't be used again

PR #1295, opened 2020-08-21
More enum refactoring

Flamefire:
- Convert (almost) all enums to scoped enums 
- Use a constexpr function for maxEnumValue to avoid includes and allow nested enums
- remove the is_enum workarounds
- Make TradeDirection contiguous


libs/common/include/helpers/EnumArray.h
```diff
@@ -44,6 +45,32 @@ struct EnumArray
     iterator end() noexcept { return elems + size(); }
     constexpr const_iterator end() const noexcept { return elems + size(); }
 
+    constexpr bool operator==(const EnumArray& rhs)
```
Flow86, line 48: before implementing everything by ourselfs, shouldnt we probably use something like https://github.com/Neargye/magic_enum instead? (header only, MIT license?)

doc/lua/methods.md
```diff
@@ -122,7 +122,7 @@ Set the settings with a table of keys:
 - speed: GS_VERYSLOW, GS_SLOW, GS_NORMAL, GS_FAST, GS_VERYFAST
 - objective: GO_NONE, GO_CONQUER3_4, GO_TOTALDOMINATION
 - startWares: SWR_VLOW, SWR_LOW, SWR_NORMAL, SWR_ALOT
-- fow: EXP_DISABLED, EXP_CLASSIC, EXP_FOGOFWAR, EXP_FOGOFWARE_EXPLORED
+- fow: EXP_DISABLED, EXP_CLASSIC, EXP_FOGOFWAR, EXP_FOGOFWAREXPLORED
```
Flow86, line 125: shouldnt this not be called `EXP_FOGOFWAR_EXPLORED` to make it readable? (so "only" fix the typo)

libs/s25main/world/MapLoader.cpp
```diff
@@ -98,7 +98,7 @@ void MapLoader::SetMapExplored(World& world)
         for(unsigned i = 0; i < MAX_PLAYERS; ++i)
         {
             // If we have FoW here, save it
-            if(world.GetNode(pt).fow[i].visibility == VIS_FOW)
+            if(world.GetNode(pt).fow[i].visibility == Visibility::FoW)
```
Flow86, line 101: we probalby should name `FoW` `FogOfWar` too

libs/s25main/world/MapGeometry.cpp
```diff
@@ -32,15 +34,17 @@ Position GetNeighbour(const Position& p, const Direction dir)
...
     {
         case Direction::WEST: return Position(p.x - 1, p.y);
         case Direction::NORTHWEST: return Position(p.x - !(p.y & 1), p.y - 1);
         case Direction::NORTHEAST: return Position(p.x + (p.y & 1), p.y - 1);
         case Direction::EAST: return Position(p.x + 1, p.y);
         case Direction::SOUTHEAST: return Position(p.x + (p.y & 1), p.y + 1);
-        default: RTTR_Assert(dir == Direction::SOUTHWEST); return Position(p.x - !(p.y & 1), p.y + 1);
+        case Direction::SOUTHWEST: return Position(p.x - !(p.y & 1), p.y + 1);
     }
+    RTTR_Assert(false);
```
Flow86, line 46: we should probably add the possibility to add a text to the asserts, so we can make it more readable whats happening?
like
```
RTTR_Assert(false) << "Unknown direction occured";
```
or something like that? (like the GTEST assertions?)

libs/common/include/helpers/EnumArray.h
Flamefire: AFAIK the `EnumArray` is new and simple enough, we could replace the enum string output operators and such though and our own string-enum (`ENUM_WITH_STRING`) 
Yeah the magic_enum and the like is nice but they all use some brute-force approach to guess the enum names (it is mostly for enum<->string conversion) so it is likely heavy on compile-time where used and might break (it even says it uses "a compiler specific hack").
And this one is C++17, which is to recent for us(?)

Oh and we can't even use it in the place where we use `ENUM_WITH_STRING` (for `AddonId`) because for some reason we encode the author in the enumerator value and hence get values up to `0x00E0002` which is impossible for the magic_enum approach. So we'd need to convert that to regular enums without explicit values (which I wouldn't mind)

doc/lua/methods.md
Flamefire: I wanted to a) keep it consistent with everything else (`VERYFAST` not `VERY_FAST`, ...) and b) allow auto-generation of the names by simply upper-casing the names which saves quite some work (and can even be more automated than it is now by using the output operators)

libs/s25main/world/MapLoader.cpp
Flamefire: I see no need to make it longer. Even the comment only uses `FoW` which in combination with `Visibility` is expressive enough, isn't it?

libs/s25main/world/MapGeometry.cpp
Flamefire: That would make it much more heavy-weight. You can already do `RTTR_Assert(!"Unknown direction occured");` similar to `assert` on which this is based on.
And I think the usages are very rare. Usually it is a simple precondition that is checked which is expressive enough, if not a comment helps. And here it is above a `BOOST_UNREACHABLE_RETURN`, so that should be clear too.

Flow86: ok thats true in that case

libs/s25main/world/MapLoader.cpp
Flow86: but it would be more consistent, longer is not bad

libs/common/include/helpers/EnumArray.h
Flow86: ah okay havent thought about the c++17 part

libs/s25main/world/MapLoader.cpp
Flamefire: Good point. So I checked what we use so far: SaveFOWNode, fowVisibility, GetNode(pt).fow, GetYoungestFOWNode, isFoW, FoWNode, ...
So only the exploration enum uses the full name. Hence if we renamed this we'd need to rename all IMO, but I would rather not do that: GetYoungestFogOfWarNode looks overly verbose.
So I'd keep it as-is with the argument that `Exploration` is rather user facing so should not be abbreviated but everything else is internal and has enough context to make it clear what FoW means.

Or should we just change the enum so (only) the enums are consistent? Personally I like the shorter version better but no strong opinion for the enum

Flow86: I would make the enums consistent, the rest are bad, but okayish (one can still understand it) even if some of the other names are also a bit questionable (i.e isFow - shouldnt it be hasFow?) - the enums are more important I think (and one step by another)

libs/s25main/world/MapGeometry.cpp
Flamefire: Just seen I already did that: RTTR_Assert_Msg which is a readable shortcut for the above workaround

Flamefire:
Pending CI this is now ready and strong enums are used pretty much everywhere. The remaining ones are not worth replacing.

Only one that needs to be done is the Team enum but that will be in a follow up and is blocked by #1310

Flow86 (approved):

PR #1301, opened 2020-09-08
Fix crash in transport window when nothing is selected

### LINKED ISSUE #1300: S2 crashes in transport order settings
When pressing the up/down-Buttons in the transport order settings while no goods are chosen, the client crashes:

`terminate called after throwing an instance of 'std::logic_error'
  what():  No button selected in option group`

-> should be catched

Used version: v20200905 24b1e7

Flamefire:
Select the first button on creation, fixes #1300
Also some refactoring to unify the arrays

### DIFF
--- a/libs/s25main/ingameWindows/iwTransport.cpp
+++ b/libs/s25main/ingameWindows/iwTransport.cpp
@@ -63,7 +63,6 @@ iwTransport::iwTransport(const GameWorldViewer& gwv, GameCommandFactory& gcFacto
                    {LOADER.GetMapTexN(2250 + GD_HAMMER), gettext_noop("Tools")},
                    {LOADER.GetMapTexN(2250 + GD_BOAT), WARE_NAMES[GD_BOAT]}}};
     // Positionen der einzelnen Buttons
-    constexpr auto numButtons = std::tuple_size<decltype(buttonData)>::value;
     const std::array<DrawPoint, numButtons> BUTTON_POS = {{{20, 25},
                                                            {52, 42},
                                                            {84, 59},
--- a/libs/s25main/ingameWindows/iwTransport.cpp
+++ b/libs/s25main/ingameWindows/iwTransport.cpp
@@ -48,20 +48,20 @@ iwTransport::iwTransport(const GameWorldViewer& gwv, GameCommandFactory& gcFacto
     // Buttons der einzelnen Waren anlegen
     ctrlOptionGroup* group = AddOptionGroup(6, ctrlOptionGroup::ILLUMINATE);
 
-    buttonData = {{{LOADER.GetMapTexN(2250 + GD_COINS), WARE_NAMES[GD_COINS]},
+    buttonData = {{{LOADER.GetMapTexN(WARES_TEX_MAP_OFFSET + GD_COINS), WARE_NAMES[GD_COINS]},
                    {LOADER.GetTextureN("io", 111), gettext_noop("Weapons")},
-                   {LOADER.GetMapTexN(2250 + GD_BEER), WARE_NAMES[GD_BEER]},
-                   {LOADER.GetMapTexN(2250 + GD_IRON), WARE_NAMES[GD_IRON]},
-                   {LOADER.GetMapTexN(2250 + GD_GOLD), WARE_NAMES[GD_GOLD]},
-                   {LOADER.GetMapTexN(2250 + GD_IRONORE), WARE_NAMES[GD_IRONORE]},
-                   {LOADER.GetMapTexN(2250 + GD_COAL), WARE_NAMES[GD_COAL]},
-                   {LOADER.GetMapTexN(2250 + GD_BOARDS), WARE_NAMES[GD_BOARDS]},
-                   {LOADER.GetMapTexN(2250 + GD_STONES), WARE_NAMES[GD_STONES]},
-                   {LOADER.GetMapTexN(2250 + GD_WOOD), WARE_NAMES[GD_WOOD]},
-                   {LOADER.GetMapTexN(2250 + GD_WATER), WARE_NAMES[GD_WATER]},
+                   {LOADER.GetMapTexN(WARES_TEX_MAP_OFFSET + GD_BEER), WARE_NAMES[GD_BEER]},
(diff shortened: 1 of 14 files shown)


libs/s25main/ingameWindows/iwTransport.cpp
```diff
@@ -53,42 +48,44 @@ iwTransport::iwTransport(const GameWorldViewer& gwv, GameCommandFactory& gcFacto
...
-    TRANSPORT_SPRITES[6] = LOADER.GetMapTexN(2250 + GD_COAL);
-    TRANSPORT_SPRITES[7] = LOADER.GetMapTexN(2250 + GD_BOARDS);
-    TRANSPORT_SPRITES[8] = LOADER.GetMapTexN(2250 + GD_STONES);
-    TRANSPORT_SPRITES[9] = LOADER.GetMapTexN(2250 + GD_WOOD);
-    TRANSPORT_SPRITES[10] = LOADER.GetMapTexN(2250 + GD_WATER);
-    TRANSPORT_SPRITES[11] = LOADER.GetTextureN("io", 80);
-    TRANSPORT_SPRITES[12] = LOADER.GetMapTexN(2250 + GD_HAMMER);
-    TRANSPORT_SPRITES[13] = LOADER.GetMapTexN(2250 + GD_BOAT);
-
+    buttonData = {{{LOADER.GetMapTexN(2250 + GD_COINS), WARE_NAMES[GD_COINS]},
```
Flow86, line 51: WARE_NAMES is still gettext_noop?

libs/s25main/ingameWindows/iwTransport.h
```diff
@@ -28,9 +29,15 @@ class GameCommandFactory;
...
+    struct ButtonData
+    {
+        ITexture* sprite;
+        std::string tooltip;
+    };
     const GameWorldViewer& gwv;
     GameCommandFactory& gcFactory;
-    ITexture* TRANSPORT_SPRITES[14]; /// Einzelne Bilder für die Waren.
+
+    std::array<ButtonData, 14> buttonData;
```
Flow86, line 40: can we have a constant for the 14?

Flow86 (changes requested):

libs/s25main/ingameWindows/iwTransport.h
Flow86: (numButtons?)

libs/s25main/ingameWindows/iwTransport.cpp
Flamefire: Yes, see libs/s25main/gameTypes/GoodTypes.h

I just wanted to avoid copying all the texts here

libs/s25main/ingameWindows/iwTransport.h
Flamefire: I did it the other way round: Using `buttonData.size` as `numButtons`, which IMO makes more sense: https://github.com/Return-To-The-Roots/s25client/pull/1301/files#diff-b140bedff094aa8fd2769a0841d341a4R66

But actually we should use `TransportOrders.size`... Adding a commit

libs/s25main/ingameWindows/iwTransport.cpp
Flow86: okay, only wanted to make sure that this is not the only place where these are translated :-D

libs/s25main/ingameWindows/iwTransport.cpp
```diff
@@ -53,42 +48,43 @@ iwTransport::iwTransport(const GameWorldViewer& gwv, GameCommandFactory& gcFacto
...
+                   {LOADER.GetTextureN("io", 111), gettext_noop("Weapons")},
+                   {LOADER.GetMapTexN(2250 + GD_BEER), WARE_NAMES[GD_BEER]},
```
Flow86, line 53: do we get an name for the 2250 here?

libs/s25main/ingameWindows/iwTransport.cpp
```diff
@@ -131,13 +127,11 @@ void iwTransport::Msg_ButtonClick(const unsigned ctrl_id)
...
 
-            //// Tooltips in der der Standardbelegung
-            // memcpy(tooltip_indices,STD_TOOLTIP_INDICES,14*sizeof(unsigned short));
-
-            for(unsigned char i = 0; i < 14; ++i)
+            for(unsigned char i = 0; i < buttonData.size(); ++i)
             {
-                group->GetCtrl<ctrlImageButton>(i)->SetImage(TRANSPORT_SPRITES[i]);
-                group->GetCtrl<ctrlImageButton>(i)->SetTooltip(_(TOOLTIPS[i]));
+                const auto& data = buttonData[i];
```
Flow86, line 132: its a pity that you can't get the auto-loop-index somehow :-D (on `for(x : list)`)

Flamefire: Hm, you can: boost::adapters::indexed IIRC :)

Flow86: so we could do this instead there? :-D

Flamefire: I was thinking about it (used it before), but what I really dislike about it (or Boost.Range) is that it is still using C++98. Checked that header and it has includes to Boost.Tuple and Boost.MPL, both are heavy IIRC. And then there is this:

```
// Why yet another "pair" class:
// - std::pair can't store references
```

Which isn't correct. So well, for the sake of compiletimes I'd rather not use it

libs/s25main/ingameWindows/iwTransport.cpp
Flamefire: > WARES_TEX_MAP_OFFSET

libs/s25main/ingameWindows/iwTransport.cpp
Flow86: yes that makes sense

libs/s25main/ingameWindows/iwTransport.cpp
Flow86: perfect :-)

Flow86 (approved):

Issue #1247, opened 2020-06-04
Override Bug: Skipping duplicate


Flow86:
With the latest nightly I'll get this:

```
Skipping duplicate override file "D:\SIEDLER2\s25rttr-nightly\RTTR\LSTS\fonts.LST" for fontsLade ""D:\SIEDLER2\s25rttr-nightly\RTTR\LSTS\fonts.LST"": fertig in 14ms
...
Loading LST,LBM,BOB,IDX,BMP,TXT,GER,ENG,INI files from ""D:\SIEDLER2\s25rttr-nightly\RTTR\LSTS""
Skipping duplicate override file "D:\SIEDLER2\s25rttr-nightly\RTTR\LSTS\CREDITS.LST" for creditsLade Ordner "D:\SIEDLER2\s25rttr-nightly\RTTR\LSTS\CREDITS.LST"
  Lade "6" Einträge: fertig in 5ms
Skipping duplicate override file "D:\SIEDLER2\s25rttr-nightly\RTTR\LSTS\OUTLINE_FONTS.LST" for outline_fontsLade ""D:\SIEDLER2\s25rttr-nightly\RTTR\LSTS\OUTLINE_FONTS.LST"": fertig in 4ms
Skipping duplicate override file "D:\SIEDLER2\s25rttr-nightly\RTTR\LSTS\RTTR.LST" for rttrLade ""D:\SIEDLER2\s25rttr-nightly\RTTR\LSTS\RTTR.LST"": fertig in 1ms
Skipping duplicate override file "D:\SIEDLER2\s25rttr-nightly\RTTR\LSTS\firework.LST" for fireworkLade Ordner "D:\SIEDLER2\s25rttr-nightly\RTTR\LSTS\firework.LST"
  Lade "32" Einträge: fertig in 19ms
Skipping duplicate override file "D:\SIEDLER2\s25rttr-nightly\RTTR\LSTS\fonts.LST" for fontsSkipping duplicate override file "D:\SIEDLER2\s25rttr-nightly\RTTR\LSTS\io_new.LST" for io_newLade Ordner "D:\SIEDLER2\s25rttr-nightly\RTTR\LSTS\io_new.LST"
  Lade "14" Einträge: fertig in 7ms
[...]

PR #1303, opened 2020-09-11
Resource handling rework

Flamefire:
This is a complete overhaul of the resource system. The major change is that no files are loaded implicitly but every files has to be loaded explicitly at some point. This solves the main problem in #1247 where override files could be loaded twice potentially breaking an already loaded file by omitting the base file and also avoids loading random files that are in those folders.

Full description: https://github.com/Flamefire/s25client/blob/overide_handling_rework/doc/ResourceLoading.md

Additionally I introduced a distinction between loading from a path (e.g. for S2 files and almost always used) and from a resource id. The idea is from my other branch: https://github.com/Flamefire/s25client/tree/ressourceHandling/libs/s25main/resources The final goal is that we only ever use the resource IDs and add all folders from S2 etc as resource folders, see https://github.com/Flamefire/s25client/blob/8ce5fd4a7b0b5a0634becf3704e0adeb144e856f/libs/s25main/resources/ArchiveLocator.cpp#L30-L41

But that needs some more work and thinking I guess. So for now only the rttr base resource folder is added and one can load a resource from there with e.g. `"map_new"_res`.

I also made the `ResourceId` fully `constexpr` instead of using a `std::string`. The size of a `std::string` is 32 bytes. Even our longest name ("charburner_bobs" which I want gone anyway) is only 15 chars. Hence we could use an array of 15 chars + an uint8_t for the size instead which cuts the length in half to 16 bytes. Then most conversions can be made at compiletime and the sanity checks go away or error at compiletime (note: only for GCC/Clang though, MSVC lacks here and there is no consteval yet...)


Flow86 (changes requested):
needs rebase now

Flamefire:
Rebased and conflicts resolved. @Flow86 can this be merged?

PR #1304, opened 2020-09-14
Use #pragma once and update to Clang-format 10

### LINKED ISSUE #1302: Update to clang-format-10 and #pragma once issues
I noticed that clang-format-10 is available in all ubuntu repos where clang-format-9 is so to be more future proof I'd like to update that. I just checked that no changes from the update alone happen.

However I'd like to change 2 options:

**IndentPPDirectives:**

```
#ifdef WIN32
#    include <windows.h>
#    define IS_OK 1
#endif
```

And also change the line length to 120 (from 140) as I noticed that the few lines above 120 we have cause readability issues in conflict resolving tools and some IDEs with sidebars. With 120 it seems to be much better and there won't be many lines affected, so that one is easy.

However the first cause problems when we have:

```
#pragma once
#ifndef foo_h
#define foo_h
```

I already suspected that using both is a pessimization because compilers detect the include-guard pattern and handle that automatically making at as fast as `#pragma once` (see #1073) and having include-guards and the pragma is more work for maintainers for no (noticeable) gain. This is confirmed by clang-format not detecting the above as include guards and trying to format it as 
```
#pragma once
#ifndef foo_h
#    define foo_h
```
while it does not do so (detecting include guard pattern) without the pragma once.

@Flow86 I'd hence like to press for a decision to choose either include guards or the pragma, not both. I'd go for the pragma as it is shorter and C&P. No need to come up with a name or keep it updated when renaming/extracting files. We even have some faulty include guards which I only noticed due to the formatter. The pragma once is easier to check for. And I don't think we support any compiler not understanding it.

CCing @vader1986 @stefson if there is any strong argument for either

Flamefire:
Closes #1302


Flow86:
```
fatal error: in "SystemTestSuite/GetUsername": critical check !System::getUserName().empty() has failed
```

.github/workflows/static-analysis.yml
```diff
@@ -7,39 +7,39 @@ on:
...
-    - name: Clang-Tidy Main
-      run: tools/ci/runClangTidy.sh main
-    - name: Clang-Tidy External
-      run: tools/ci/runClangTidy.sh external
+      - uses: actions/checkout@v2
+      - run: git submodule update --init
+      - name: Install dependencies
+        run: |
+          sudo apt-get update && sudo apt-get install -y \
+            clang-9 clang-tidy-9 \
```
Flow86, line 45: shouldn't it install clang 10?

Flamefire: Not yet. clang and clang-tidy must match and clang-tidy-10 contains checks that I want to enable in the near future but which would fail now. Note that clang-format-10 is not dependent on clang-10

Flamefire:
@Flow86 Why would `$USER` be empty on jenkins? Even `getpwuid` doesn't seem to work.

Flow86:
Hmm. docker environment? I don't know. I will check

Flamefire:
Yep, looks very much like it. I think we should even throw an exception if the user name is empty, similar to Windows. When we use $USER then for docker you can just pass something to the docker env as $USER. Is likely the easiest. Then I'd revert my commit  using getpwuid

Flamefire:
Added a workaround in #1303: Set $USER manually if not set. Should be ok, will remove the commit to libutil using the syscall

Flamefire:
Extracted the jenkins fix to https://github.com/Return-To-The-Roots/s25client/pull/1305, I can rebase the other 2 PRs onto that.

For this one I updated clang-tidy to version 10 too to resolve the version disparity.

@Flow86 Please review and merge as appropriate

PR #1310, opened 2020-11-10
Introduce new "smart" range-based team building options

jdoerfert:
The old team building options where either fixed (1,2,3,4) or purely
random (?). The new options (1-2,1-3,1-4) allow players to be placed
in ranges of teams while they also ensure team sizes to be as even as
possible. That is, a player with the team `1-3` is placed in team 1,
2, or 3 in a way that minimizes the absolute difference in team size.

### DIFF
--- a/libs/s25main/network/GameServer.cpp
+++ b/libs/s25main/network/GameServer.cpp
@@ -21,6 +21,7 @@
 #include "GameMessage_GameCommand.h"
 #include "GameServerPlayer.h"
 #include "GlobalGameSettings.h"
+#include "JoinPlayerInfo.h"
 #include "RTTR_Version.h"
 #include "RttrConfig.h"
 #include "Savegame.h"
@@ -31,7 +32,9 @@
 #include "network/CreateServerInfo.h"
 #include "network/GameMessages.h"
 #include "ogl/glArchivItem_Map.h"
+#include "random/Random.h"
 #include "gameTypes/LanGameInfo.h"
+#include "gameTypes/TeamTypes.h"
 #include "gameData/GameConsts.h"
 #include "gameData/LanDiscoveryCfg.h"
 #include "liblobby/LobbyClient.h"
@@ -46,6 +49,7 @@
 #include <cmath>
 #include <helpers/chronoIO.h>
 #include <iomanip>
+#include <iterator>
 #include <mygettext/mygettext.h>
 
 inline std::ostream& operator<<(std::ostream& os, const AsyncChecksum& checksum)
@@ -430,12 +434,109 @@ void GameServer::Stop()
     LOG.write("server state changed to stop\n");
 }
 
+// Check if there are players that have not been assigned a team but only a
+// range. Those players are assigned a team now while we try to balanace the
+// number of players per team. Returns true iff players have been assigned.
+bool GameServer::assignPlayersOfRandomTeams(std::vector<JoinPlayerInfo>& playerInfos)
+{
+    static_assert(NUM_TEAMS == 4, "Expected exactly 4 playable teams!");
+
+    std::set<unsigned> potentialPlayers[NUM_TEAMS];
+    unsigned nPlayers[NUM_TEAMS] = {0};
+    unsigned unassignedPlayers = 0;
+
+    // First collect fixed players and potential ones.
+    for(unsigned player = 0; player < playerInfos.size(); ++player)
+    {
+        JoinPlayerInfo& playerInfo = playerInfos[player];
+        switch(playerInfo.team)
+        {
+            case TM_RANDOMTEAM:
+            case TM_TEAM1: ++nPlayers[0]; break;
+            case TM_RANDOMTEAM2:
+            case TM_TEAM2: ++nPlayers[1]; break;
+            case TM_RANDOMTEAM3:
+            case TM_TEAM3: ++nPlayers[2]; break;
+            case TM_RANDOMTEAM4:
+            case TM_TEAM4: ++nPlayers[3]; break;
+            case TM_TEAM_1_TO_2:
+                ++unassignedPlayers;
+                potentialPlayers[0].insert(player);
+                potentialPlayers[1].insert(player);
+                break;
+            case TM_TEAM_1_TO_3:
+                ++unassignedPlayers;
+                potentialPlayers[0].insert(player);
+                potentialPlayers[1].insert(player);
+                potentialPlayers[2].insert(player);
+                break;
+            case TM_TEAM_1_TO_4:
+                ++unassignedPlayers;
+                potentialPlayers[0].insert(player);
+                potentialPlayers[1].insert(player);
+                potentialPlayers[2].insert(player);
+                potentialPlayers[3].insert(player);
+                break;
+            case TM_NOTEAM: break;
+        }
+    }
+    // Check for unassigned players first.
+    if(unassignedPlayers == 0)
+        return false;
+
+    while(unassignedPlayers)
+    {
+        // Set of potential teams for the next player.
+        std::vector<int> teamsForNextPlayer;
+
+        // Determine the minimal team size for teams that can take an unassigned player.
+        unsigned minNextTeamSize = std::numeric_limits<unsigned>::max();
+        for(unsigned team = 0; team < NUM_TEAMS; ++team)
+        {
+            // Check if we can add a player to this team at all.
+            if(!potentialPlayers[team].empty())
+                minNextTeamSize = std::min(minNextTeamSize, nPlayers[team]);
+        }
+
+        for(unsigned team = 0; team < NUM_TEAMS; ++team)
+        {
+            // Check if we can add a player to this team at all.
+            if(potentialPlayers[team].empty())
+                continue;
+            // Check if this is a team with the minimal number of players amongst teams that can have unassigned ones.
+            if(nPlayers[team] <= minNextTeamSize)
+                teamsForNextPlayer.push_back(team);
+        }
+
+        RTTR_Assert_Msg(!teamsForNextPlayer.empty(), "Expected to have teams with potential players!");
+        int nextTeam = teamsForNextPlayer[rand() % teamsForNextPlayer.size()];
+        RTTR_Assert_Msg(!potentialPlayers[nextTeam].empty(), "Expected next team to have potential players!");
+
+        // Pick a random player that can go into this team.
+        int nextPlayer = *getRandomElement(potentialPlayers[nextTeam]);
+
+        // The player is now assigned and the team size increased.
+        for(auto& PotentialPlayer : potentialPlayers)
+            PotentialPlayer.erase(nextPlayer);
+        ++nPlayers[nextTeam];
+
+        playerInfos[nextPlayer].team = Team(TM_TEAM1 + nextTeam);
+        --unassignedPlayers;
+    }
+
+    return true;
+}
+
 /**
  *  startet das Spiel.
  */
 bool GameServer::StartGame()
 {
     lanAnnouncer.Stop();
+
+    // Finalize the team selection for unassigned players.
+    if(assignPlayersOfRandomTeams(playerInfos))
+        SendToAll(GameMessage_Player_List(playerInfos));
 
     // Bei Savegames wird der Startwert von den Clients aus der Datei gelesen!
     unsigned random_init;
--- a/libs/s25main/random/Random.h
+++ b/libs/s25main/random/Random.h
@@ -25,6 +25,7 @@
 #include <cstddef>
 #include <iosfwd>
 #include <limits>
+#include <random>
 #include <string>
 #include <utility>
 #include <vector>
@@ -139,3 +140,45 @@ public:
 
 /// Shortcut for creating an instance of RandomFunctor
 #define RANDOM_SHUFFLE(container) RandomFunctor::shuffleContainer(container, __FILE__, __LINE__)
+
+/// Return a random number in the interval [lowerBound, upperBound].
+template<typename RandomT>
+static unsigned getRandomNumber(const RandomT& rng, unsigned lowerBound, unsigned upperBound)
+{
+    std::uniform_int_distribution<typename RandomT::result_type> dist(lowerBound, upperBound);
+    return dist(rng);
+}
+
+/// Return a random number in the interval [0, upperBound).
+template<typename RandomT>
+static unsigned getRandomNumber(const RandomT& rng, unsigned upperBound)
+{
+    RTTR_Assert(upperBound > 0);
+    return getRandomNumber(rng, 0, upperBound - 1);
+}
+
+/// Return a random number in the interval [0, upperBound).
+static unsigned getRandomNumber(unsigned upperBound)
+{
+    std::mt19937 rng;
+    std::random_device rnd_dev;
+    rng.seed(rnd_dev());
+    RTTR_Assert(upperBound > 0);
+    return getRandomNumber(rng, upperBound);
+}
+
+/// Return an iterator to a random element of \p container. If \p isEnd is nonnull, set it to true if the returned
+/// iterator is the "end" iterator and false otherwise.
+template<typename ContainerT>
+static typename ContainerT::iterator getRandomElement(ContainerT& container, bool* isEnd = nullptr)
+{
+    unsigned size = container.size();
+    bool empty = size == 0;
+    if(isEnd)
+        *isEnd = empty;
+    if(empty)
+        return container.end();
+    auto it = container.begin();
+    std::advance(it, getRandomNumber(size));
+    return it;
+}
--- a/tests/s25Main/integration/testGameServerJointPlayerDistribution.cpp
+++ b/tests/s25Main/integration/testGameServerJointPlayerDistribution.cpp
@@ -1,0 +1,129 @@
+// Copyright (c) 2016 - 2018 Settlers Freaks (sf-team at siedler25.org)
+//
+// This file is part of Return To The Roots.
+//
+// Return To The Roots is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 2 of the License, or
+// (at your option) any later version.
+//
+// Return To The Roots is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Return To The Roots. If not, see <http://www.gnu.org/licenses/>.
+
+#include "BasePlayerInfo.h"
+#include "JoinPlayerInfo.h"
+#include "network/GameServer.h"
+#include "gameTypes/TeamTypes.h"
+#include <boost/test/unit_test.hpp>
+
+static unsigned getMaxTeamSizeDifference(std::vector<JoinPlayerInfo>& playerInfos, unsigned nTeams)
+{
+    unsigned nPlayers[4] = {0, 0, 0, 0};
+    for(auto& playerInfo : playerInfos)
+    {
+        switch(playerInfo.team)
+        {
+            case TM_RANDOMTEAM:
+            case TM_TEAM1: ++nPlayers[0]; break;
+            case TM_RANDOMTEAM2:
+            case TM_TEAM2: ++nPlayers[1]; break;
+            case TM_RANDOMTEAM3:
+            case TM_TEAM3: ++nPlayers[2]; break;
+            case TM_RANDOMTEAM4:
+            case TM_TEAM4: ++nPlayers[3]; break;
+            default: return -1;
+        };
+    }
+
+    std::sort(&nPlayers[0], &nPlayers[nTeams]);
+    BOOST_TEST_REQUIRE(nPlayers[0] <= nPlayers[nTeams - 1]);
+    return nPlayers[nTeams - 1] - nPlayers[0];
+}
+
+BOOST_AUTO_TEST_CASE(JoinPlayerAssignment)
+{
+    BasePlayerInfo BPI1, BPI2, BPI3, BPI4, BPI1_2, BPI1_3, BPI1_4, BPIRT1, BPIRT2, BPIRT3, BPIRT4;
+    BPI1.team = TM_TEAM1;
+    BPI2.team = TM_TEAM2;
+    BPI3.team = TM_TEAM3;
+    BPI4.team = TM_TEAM4;
+    BPIRT1.team = TM_RANDOMTEAM;
+    BPIRT2.team = TM_RANDOMTEAM2;
+    BPIRT3.team = TM_RANDOMTEAM3;
+    BPIRT4.team = TM_RANDOMTEAM4;
+    BPI1_2.team = TM_TEAM_1_TO_2;
+    BPI1_3.team = TM_TEAM_1_TO_3;
+    BPI1_4.team = TM_TEAM_1_TO_4;
+
+    std::vector<JoinPlayerInfo> playerInfos;
+
+    playerInfos.push_back(JoinPlayerInfo(BPI1_4));
+    playerInfos.push_back(JoinPlayerInfo(BPI1));
+    playerInfos.push_back(JoinPlayerInfo(BPIRT2));
+    playerInfos.push_back(JoinPlayerInfo(BPI1_3));
+    BOOST_REQUIRE(GameServer::assignPlayersOfRandomTeams(playerInfos));
+    BOOST_TEST(2u >= getMaxTeamSizeDifference(playerInfos, 4));
+
+    playerInfos.clear();
+    playerInfos.push_back(JoinPlayerInfo(BPI1_4));
+    playerInfos.push_back(JoinPlayerInfo(BPI1));
+    playerInfos.push_back(JoinPlayerInfo(BPIRT2));
+    playerInfos.push_back(JoinPlayerInfo(BPI1_4));
+    BOOST_REQUIRE(GameServer::assignPlayersOfRandomTeams(playerInfos));
+    BOOST_TEST(0u == getMaxTeamSizeDifference(playerInfos, 4));
+
+    playerInfos.clear();
+    playerInfos.push_back(JoinPlayerInfo(BPI1_3));
+    playerInfos.push_back(JoinPlayerInfo(BPI1));
+    playerInfos.push_back(JoinPlayerInfo(BPIRT2));
+    playerInfos.push_back(JoinPlayerInfo(BPI1_3));
+    BOOST_REQUIRE(GameServer::assignPlayersOfRandomTeams(playerInfos));
+    BOOST_TEST(1u == getMaxTeamSizeDifference(playerInfos, 3));
+
+    playerInfos.clear();
+    playerInfos.push_back(JoinPlayerInfo(BPI1_4));
+    playerInfos.push_back(JoinPlayerInfo(BPI1_4));
+    playerInfos.push_back(JoinPlayerInfo(BPI1_4));
+    playerInfos.push_back(JoinPlayerInfo(BPI1_4));
+    playerInfos.push_back(JoinPlayerInfo(BPI1_4));
+    playerInfos.push_back(JoinPlayerInfo(BPI1_4));
+    playerInfos.push_back(JoinPlayerInfo(BPI1_4));
+    playerInfos.push_back(JoinPlayerInfo(BPI1_4));
+    BOOST_REQUIRE(GameServer::assignPlayersOfRandomTeams(playerInfos));
+    BOOST_TEST(0u == getMaxTeamSizeDifference(playerInfos, 4));
+
+    playerInfos.clear();
+    playerInfos.push_back(JoinPlayerInfo(BPI1_4));
+    playerInfos.push_back(JoinPlayerInfo(BPI1_4));
+    playerInfos.push_back(JoinPlayerInfo(BPI1_4));
+    playerInfos.push_back(JoinPlayerInfo(BPI1_3));
+    playerInfos.push_back(JoinPlayerInfo(BPI1_3));
+    playerInfos.push_back(JoinPlayerInfo(BPI1_3));
+    playerInfos.push_back(JoinPlayerInfo(BPI1_3));
+    playerInfos.push_back(JoinPlayerInfo(BPI1_3));
+    BOOST_REQUIRE(GameServer::assignPlayersOfRandomTeams(playerInfos));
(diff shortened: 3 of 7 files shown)


libs/s25main/network/GameServer.cpp
```diff
@@ -21,17 +21,20 @@
...
 #include "GameServerPlayer.h"
 #include "GlobalGameSettings.h"
+#include "JoinPlayerInfo.h"
 #include "RTTR_Version.h"
 #include "RttrConfig.h"
 #include "Savegame.h"
 #include "Settings.h"
 #include "commonDefines.h"
 #include "files.h"
+#include "helpers/MultiArray.h"
```
Flamefire, line 31:
```suggestion
```
Unused?

libs/s25main/network/GameServer.cpp
```diff
@@ -430,13 +434,112 @@ void GameServer::Stop()
...
+                break;
+            case TM_TEAM_1_TO_4:
+                ++unassignedPlayers;
+                nPotentialPlayers[0].insert(player);
+                nPotentialPlayers[1].insert(player);
+                nPotentialPlayers[2].insert(player);
+                nPotentialPlayers[3].insert(player);
+                break;
+            case TM_NOTEAM: break;
+            default: RTTR_Assert_Msg(false, "Unexpected team!");
```
Flamefire, line 475: What about the random teams?

libs/s25main/network/GameServer.cpp
```diff
@@ -430,13 +434,112 @@ void GameServer::Stop()
...
+        }
+    }
+    // Check for unassigned players first.
+    if(unassignedPlayers == 0)
+        return false;
+
+    RTTR_Assert_Msg(!nPotentialPlayers[0].empty() || !nPotentialPlayers[1].empty() || !nPotentialPlayers[2].empty()
```
Flamefire, line 482: I'd remove that due to size of the condition. The same issue will be caught a bit below by the next assertion(s)

libs/s25main/network/GameServer.cpp
```diff
@@ -430,13 +434,112 @@ void GameServer::Stop()
...
+                      || !nPotentialPlayers[3].empty(),
+                    "Expected at least one team!");
+
+    while(unassignedPlayers)
+    {
+        // Set of potential teams for the next player.
+        std::vector<int> teamsForNextPlayer;
+
+        // Determine the minimal team size for teams that can take an unassigned player.
+        unsigned minNextTeamSize = -1;
```
Flamefire, line 492: This is a bug, I'd suggest to write tests to catch such issues.

libs/s25main/network/GameServer.cpp
```diff
@@ -430,13 +434,112 @@ void GameServer::Stop()
...
+        for(int team = 0; team < 4; ++team)
+        {
+            // Check if we can add a player to this team at all.
+            if(!nPotentialPlayers[team].empty())
+              minNextTeamSize = std::min(minNextTeamSize, nPlayers[team]);
+        }
+        RTTR_Assert_Msg(minNextTeamSize != -1, "Expected to find team for an unassigned player!");
```
Flamefire, line 499: This can be removed if you initialize `minNextTeamSize=0`

libs/s25main/network/GameServer.cpp
```diff
@@ -430,13 +434,112 @@ void GameServer::Stop()
...
+            // Check if we can add a player to this team at all.
+            if(nPotentialPlayers[team].empty())
+                continue;
+            // Check if this is a team with the minimal number of players amongst teams that can have unassigned ones.
+            if(nPlayers[team] > minNextTeamSize)
+                continue;
+            teamsForNextPlayer.push_back(team);
+        }
+
+        int nextTeam = teamsForNextPlayer[rand() % teamsForNextPlayer.size()];
```
Flamefire, line 512: An assert here makes sense: teamsForNextPlayer must not be empty. In meaning that assert is the same as the one above (line 499)

libs/s25main/network/GameServer.cpp
```diff
@@ -430,13 +434,112 @@ void GameServer::Stop()
...
+        RTTR_Assert_Msg(!nPotentialPlayers[nextTeam].empty(), "Expected next team to have potential players!");
+
+        // Pick a random player that can go into this team.
+        auto it = nPotentialPlayers[nextTeam].begin();
+        std::advance(it, rand() % nPotentialPlayers[nextTeam].size());
+        int nextPlayer = *it;
+
+        // The player is now assigned and the team size increased.
+        for(int i = 0; i < 4; ++i)
+            nPotentialPlayers[i].erase(nextPlayer);
```
Flamefire, line 522: Maybe:
```suggestion
        for(auto& playerSet: nPotentialPlayers)
            playerSet.erase(nextPlayer);
```

libs/s25main/network/GameServer.cpp
```diff
@@ -430,13 +434,112 @@ void GameServer::Stop()
     LOG.write("server state changed to stop\n");
 }
 
+// Check if there are players that have not been assigned a team but only a
+// range. Those players are assigned a team now while we try to balanace the
+// number of players per team. Returns true iff players have been assigned.
+static bool adjustSmartTeams(std::vector<JoinPlayerInfo>& playerInfos)
+{
+    std::set<unsigned> nPotentialPlayers[4];
```
Flamefire, line 442: Could this get a better name? It isn't the number of players but player indices so remove the "n" prefix?

Flamefire (changes requested):
I do like the idea. While you are at it: I think it makes sense to overhaul the whole random team thing so we have either
- Team 1, 2, 3 or 4
- Random
- Fair Random with restriction (yours)

We currently "fake" the Random thing by using different enum values. Having your code already let's us do that properly in a similar way: Just iterate over the "Random" values and assign a random team, then broadcast.
This could be done here or in a follow-up PR.

For this PR I'd like to see some tests. If you make that function non-static it should be pretty easy to test: Throw in some players with teams set and check that the distribution of teams matches (i.e. no overfull team or a team with no players) and no player is in a team where he shouldn't be (when using the restriction to team 1-3 etc). For this I suggest a helper function which creates a vector of JoinPlayerInfo out of a vector of Teams or even fully wrap it in a (test) function which takes a vector of teams, calls your function and returns a vector of teams. That is then easily testable :)

libs/s25main/network/GameServer.cpp
jdoerfert: But that would simply hide a broken invariant, right? The point was to ensure we do not end up here without having potential players to distribute, though, I can obviously remove the asserts if that is preferred.

libs/s25main/network/GameServer.cpp
jdoerfert: Sure. I'll add one here, remove the others.

libs/s25main/network/GameServer.cpp
jdoerfert: sure.

libs/s25main/network/GameServer.cpp
jdoerfert: I don't get it. What is a bug? The assertions should never trigger, as the code is written. They are safeguards for future changes that break invariants. Unsure what you want to test and catch here.

libs/s25main/network/GameServer.cpp
jdoerfert: They don't "exist" this late. Whenever the host clicks "random" they are assigned, for now, to a team (IIRC). A follow up could change that and distribute random teams here as well, maybe make ? behave like 1-4 does and remove the latter. That would be a breaking change and I wanted to separate things so people can determine if they want to keep the old behavior at all or simply go with "smart" teams.

Flamefire: IIRC when "random" is selected a TM_RANDOMTEAM1-TM_RANDOMTEAM4 is assigned which shows up as random but is a separate enum. Should be easy enough to test by trying it.

libs/s25main/network/GameServer.cpp
Flamefire: Converting `-1` to unsigned makes it a large value. I got confused by the `std::min`, so that is not a bug, but confusing to read. You can use `std::numeric_limits<unsigned>::max()` instead or use a boost::optional

libs/s25main/network/GameServer.cpp
jdoerfert: My bad, they are assigned early but keep the TM_RANDOMTEAMX value still. I'll add them here and keep the old behavior for now.

jdoerfert:
> For this PR I'd like to see some tests. If you make that function non-static it should be pretty easy to test: Throw in some players with teams set and check that the distribution of teams matches (i.e. no overfull team or a team with no players) and no player is in a team where he shouldn't be (when using the restriction to team 1-3 etc). For this I suggest a helper function which creates a vector of JoinPlayerInfo out of a vector of Teams or even fully wrap it in a (test) function which takes a vector of teams, calls your function and returns a vector of teams. That is then easily testable :)

I could not find any tests referencing joint players or RANDOMTEAM, where should they be placed?

Spikeone:
Dropdowns for team (and color) selection would be great as well then :) less clicking for everyone

Flamefire:
This folder is probably the right place: https://github.com/Return-To-The-Roots/s25client/tree/master/tests/s25Main/integration
You can use https://github.com/Return-To-The-Roots/s25client/blob/master/tests/s25Main/integration/testGameServerPlayer.cpp as a starting point and ping me for questions

jdoerfert:
Looking at the CI error, I'm unsure I caused this.

tests/s25Main/integration/testGameServerJointPlayerDistribution.cpp
```diff
@@ -0,0 +1,136 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Return To The Roots. If not, see <http://www.gnu.org/licenses/>.
+
+#include "BasePlayerInfo.h"
+#include "JoinPlayerInfo.h"
+#include "network/GameServer.h"
+#include "gameTypes/TeamTypes.h"
+#include <boost/test/tools/old/interface.hpp>
```
Flamefire, line 22: Not required
```suggestion
```

tests/s25Main/integration/testGameServerJointPlayerDistribution.cpp
```diff
@@ -0,0 +1,136 @@
...
+#include <boost/test/unit_test.hpp>
+#include <cstdio>
```
Flamefire, line 24: Not required(?)
```suggestion
```

tests/s25Main/integration/testGameServerJointPlayerDistribution.cpp
```diff
@@ -0,0 +1,136 @@
...
+            case TM_RANDOMTEAM3:
+            case TM_TEAM3: ++nPlayers[2]; break;
+            case TM_RANDOMTEAM4:
+            case TM_TEAM4: ++nPlayers[3]; break;
+            default: return -1;
+        };
+    }
+
+    std::sort(&nPlayers[0], &nPlayers[nTeams]);
+    BOOST_REQUIRE_LE(nPlayers[0], nPlayers[nTeams - 1]);
```
Flamefire, line 48: I found it easier to switch to the new Boost test macros:
```suggestion
    BOOST_TEST_REQUIRE(nPlayers[0] < nPlayers[nTeams - 1]);
```

tests/s25Main/integration/testGameServerJointPlayerDistribution.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+    auto& playerInfos = getPlayerInfos();
+
+    playerInfos.clear();
+    playerInfos.emplace_back(BPI1_4);
+    playerInfos.emplace_back(BPI1);
+    playerInfos.emplace_back(BPIRT2);
+    playerInfos.emplace_back(BPI1_3);
+    BOOST_REQUIRE(assignPlayersOfRandomTeams());
+    BOOST_REQUIRE_GE(2, getMaxTeamSizeDifference(playerInfos, 4));
```
Flamefire, line 75: This is the source of the CI error (unsigned 2):

```suggestion
    BOOST_TEST(2u >= getMaxTeamSizeDifference(playerInfos, 4));
```

tests/s25Main/integration/testGameServerJointPlayerDistribution.cpp
```diff
@@ -0,0 +1,136 @@
...
+    BPIRT3.team = TM_RANDOMTEAM3;
+    BPIRT4.team = TM_RANDOMTEAM4;
+    BPI1_2.team = TM_TEAM_1_TO_2;
+    BPI1_3.team = TM_TEAM_1_TO_3;
+    BPI1_4.team = TM_TEAM_1_TO_4;
+
+    auto& playerInfos = getPlayerInfos();
+
+    playerInfos.clear();
+    playerInfos.emplace_back(BPI1_4);
```
Flamefire, line 70: No need for emplace, push_back is enough for all

Flamefire (changes requested):
Hm, in the previous iteration you had a static method which received a vector of players. That would have been easier to test as you wouldn't even need the fixture to create a GameServer instance. What was the reason for switching that?
In general such pure functions are preferable as they don't require any setup for tests and are easier to reason about as they don't have state

tests/s25Main/integration/testGameServerJointPlayerDistribution.cpp
jdoerfert: Any particular reason? I mean, I don't care much but it seems rather counter intuitive.

jdoerfert:
> Hm, in the previous iteration you had a static method which received a vector of players. That would have been easier to test as you wouldn't even need the fixture to create a GameServer instance. What was the reason for switching that?
> In general such pure functions are preferable as they don't require any setup for tests and are easier to reason about as they don't have state

As long as it is static (=internal linkage), that all makes sense. Adding global externally visible functions is usually not what people advocate for, but, y'know, I can go with a global one and pass a vector instead.

jdoerfert:
> Looking at the CI error, I'm unsure I caused this.

^ still the only failure

jdoerfert:
ping

libs/s25main/network/GameServer.cpp
```diff
@@ -430,13 +433,111 @@ void GameServer::Stop()
...
+        }
+
+        RTTR_Assert_Msg(!teamsForNextPlayer.empty(), "Expected to have teams with potential players!");
+        int nextTeam = teamsForNextPlayer[rand() % teamsForNextPlayer.size()];
+        RTTR_Assert_Msg(!PotentialPlayers[nextTeam].empty(), "Expected next team to have potential players!");
+
+        // Pick a random player that can go into this team.
+        auto it = PotentialPlayers[nextTeam].begin();
+        std::advance(it, rand() % PotentialPlayers[nextTeam].size());
+        int nextPlayer = *it;
```
Flamefire, line 516: I feel this should go into a new function. This is rather verbose and hence hard to see what it does, and used a couple of times. So my suggestion: Create a function `element getRandomElement(collection, prng)` with another overload without the prng (which creates one internally and calls the other function)
Or better: Instead of a prng it could take a functor `getRandIndex(max)` which returns a value in `[0,max]`. We have several other uses in the code where we need a random element from a container, so that would certainly help.

Can be done in a follow-up though.

libs/s25main/network/GameServer.cpp
```diff
@@ -430,13 +433,111 @@ void GameServer::Stop()
     LOG.write("server state changed to stop\n");
 }
 
+// Check if there are players that have not been assigned a team but only a
+// range. Those players are assigned a team now while we try to balanace the
+// number of players per team. Returns true iff players have been assigned.
+bool GameServer::assignPlayersOfRandomTeams(std::vector<JoinPlayerInfo>& playerInfos)
+{
+    std::set<unsigned> PotentialPlayers[4];
```
Flamefire, line 441: I know this isn't introduced by you but could you use a named constant instead of the 4? You use it quite a lot so a constant makes sense. Maybe rename the NUM_TEAMS to NUM_TEAM_OPTIONS (careful to do it in all code) and introduce a new NUM_TEAMS=4

tests/s25Main/integration/testGameServerJointPlayerDistribution.cpp
```diff
@@ -0,0 +1,129 @@
...
+    BOOST_REQUIRE(GameServer::assignPlayersOfRandomTeams(playerInfos));
+    BOOST_TEST(2u >= getMaxTeamSizeDifference(playerInfos, 4));
+
+    playerInfos.clear();
+    playerInfos.push_back(JoinPlayerInfo(BPI1_4));
+    playerInfos.push_back(JoinPlayerInfo(BPI1));
+    playerInfos.push_back(JoinPlayerInfo(BPIRT2));
+    playerInfos.push_back(JoinPlayerInfo(BPI1_4));
+    BOOST_REQUIRE(GameServer::assignPlayersOfRandomTeams(playerInfos));
+    BOOST_REQUIRE_EQUAL(0, getMaxTeamSizeDifference(playerInfos, 4));
```
Flamefire, line 78:
```suggestion
    BOOST_TEST(0u == getMaxTeamSizeDifference(playerInfos, 4));
```
Similar below to fix the CI issue. Notice the use of the unsigned constant (which is what MSVC complains about: currently there is "signed == unsigned" used)

Flamefire (changes requested):
Ok, next round. Getting close :+1: 

Optional: I suggest you pull from RttR master and rebase your branch on top of it (that is the note from Github):
```
git fetch origin
git rebase origin/master
```

libs/s25main/network/GameServer.cpp
```diff
@@ -430,13 +433,111 @@ void GameServer::Stop()
...
+    unsigned nPlayers[4] = {0, 0, 0, 0};
+    unsigned unassignedPlayers = 0;
+
+    // First collect fixed players and potential ones.
+    for(unsigned player = 0; player < playerInfos.size(); ++player)
+    {
+        JoinPlayerInfo& playerInfo = playerInfos[player];
+        switch(playerInfo.team)
+        {
+            case TM_RANDOMTEAM:
```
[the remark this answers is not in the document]
jdoerfert: RANDOMTEAM is not random, it is `?` that was resolved to `1` (= team 1),
RANDOMTEAM2 is not random, it is `?` that was resolved to `2` (= team 2),
...

jdoerfert:
> Optional: I suggest you pull from RttR master and rebase your branch on top of it (that is the note from Github):
> 
> ```
> git fetch origin
> git rebase origin/master
> ```

I did when I created the pull request. I thought it is still mergeable, isn't it?

Flamefire:
> I did when I created the pull request. I thought it is still mergeable, isn't it?

Yes, hence why I marked it as optional. Github shows the "This branch is out-of-date with the base branch" notice, so there have been commits to master since then. Not sure what is inbetween now but it doesn't hurt and avoids conflicts if you make more changes. But as mentioned: Optional, feel free to ignore

libs/s25main/network/GameServer.cpp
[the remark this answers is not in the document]
jdoerfert: I assume this is resolved w/o change.

libs/s25main/random/Random.h
```diff
@@ -139,3 +140,45 @@ class RandomFunctor
 
 /// Shortcut for creating an instance of RandomFunctor
 #define RANDOM_SHUFFLE(container) RandomFunctor::shuffleContainer(container, __FILE__, __LINE__)
+
+/// Return a random number in the interval [lowerBound, upperBound].
+template<typename RandomT>
+static unsigned getRandomNumber(const RandomT& rng, unsigned lowerBound, unsigned upperBound)
```
Flamefire, line 146:
```suggestion
inline unsigned getRandomNumber(RandomT& rng, unsigned lowerBound, unsigned upperBound)
```

libs/s25main/random/Random.h
```diff
@@ -139,3 +140,45 @@ class RandomFunctor
...
+{
+    std::uniform_int_distribution<typename RandomT::result_type> dist(lowerBound, upperBound);
```
Flamefire, line 148: The result of the PRNG doesn't matter for the result if the distribution:
```suggestion
    std::uniform_int_distribution<unsigned> dist(lowerBound, upperBound);
```

libs/s25main/random/Random.h
```diff
@@ -139,3 +140,45 @@ class RandomFunctor
...
+    return dist(rng);
+}
+
+/// Return a random number in the interval [0, upperBound).
+template<typename RandomT>
+static unsigned getRandomNumber(const RandomT& rng, unsigned upperBound)
+{
+    RTTR_Assert(upperBound > 0);
+    return getRandomNumber(rng, 0, upperBound - 1);
```
Flamefire, line 157: Better: 
```suggestion
inline unsigned getRandomIndex(const RandomT& rng, unsigned size)
{
    RTTR_Assert(size > 0);
    return getRandomNumber(rng, 0, size - 1);
```

libs/s25main/random/Random.h
```diff
@@ -139,3 +140,45 @@ class RandomFunctor
...
+}
+
+/// Return a random number in the interval [0, upperBound).
+static unsigned getRandomNumber(unsigned upperBound)
+{
+    std::mt19937 rng;
+    std::random_device rnd_dev;
+    rng.seed(rnd_dev());
+    RTTR_Assert(upperBound > 0);
+    return getRandomNumber(rng, upperBound);
```
Flamefire, line 167:
```suggestion
inline unsigned getRandomIndex(unsigned size)
{
    std::random_device rnd_dev;
    std::mt19937 rng(rnd_dev());
    return getRandomNumber(rng, size);
```

libs/s25main/random/Random.h
```diff
@@ -139,3 +140,45 @@ class RandomFunctor
...
+}
+
+/// Return an iterator to a random element of \p container. If \p isEnd is nonnull, set it to true if the returned
+/// iterator is the "end" iterator and false otherwise.
+template<typename ContainerT>
+static typename ContainerT::iterator getRandomElement(ContainerT& container, bool* isEnd = nullptr)
```
Flamefire, line 173:
```suggestion
inline typename ContainerT::iterator getRandomElement(ContainerT& container, bool* isEnd = nullptr)
```

libs/s25main/random/Random.h
```diff
@@ -139,3 +140,45 @@ class RandomFunctor
...
+{
+    unsigned size = container.size();
+    bool empty = size == 0;
+    if(isEnd)
+        *isEnd = empty;
+    if(empty)
+        return container.end();
+    auto it = container.begin();
+    std::advance(it, getRandomNumber(size));
```
Flamefire, line 182:
```suggestion
    std::advance(it, getRandomIndex(size));
```

libs/s25main/random/Random.h
```diff
@@ -139,3 +140,45 @@ class RandomFunctor
...
+template<typename RandomT>
+static unsigned getRandomNumber(const RandomT& rng, unsigned lowerBound, unsigned upperBound)
+{
+    std::uniform_int_distribution<typename RandomT::result_type> dist(lowerBound, upperBound);
+    return dist(rng);
+}
+
+/// Return a random number in the interval [0, upperBound).
+template<typename RandomT>
+static unsigned getRandomNumber(const RandomT& rng, unsigned upperBound)
```
Flamefire, line 154:
```suggestion
inline unsigned getRandomNumber(RandomT& rng, unsigned upperBound)
```

libs/s25main/random/Random.h
jdoerfert: Why?

Flamefire (changes requested):
See comments. Mostly about inline and non-const rng.
Also a suggestion about not naming a non-inclusive range "upperBound". I think this makes it clearer

libs/s25main/random/Random.h
Flamefire: You want an unsigned number that is uniformly distributed, don't you? The change makes it say that. Internally it could call the rng multiple times to get enough entropy to fill an unsigned. E.g. the rng could return only bytes.

libs/s25main/random/Random.h
jdoerfert: > See comments. Mostly about inline and non-const rng.
> Also a suggestion about not naming a non-inclusive range "upperBound". I think this makes it clearer

- const rng was an oversight.
- inline seems to be counter productive. What is the reason you want it to have inline linkage over internal linkage?

Flamefire: Inline mainly because it is in a header. If it was a cpp file, then yes, internal linkage but this way it would be internal to any cpp file which includes this header which isn't what one usually wants. Besides that the compiler warns about unused static functions but not for inline functions, so that's what I'd use

Flamefire:
Hey @jdoerfert any progress here?

jdoerfert:
> Hey @jdoerfert any progress here?

No. I might revisit this at some point but have no intention of doing so in the near future. Feel free to use the code and adapt it yourself.

FWIW, the review process, especially given the state of the code base, should be reconsidered. I get it that you want to make things perfect, for some definition of perfect, but that is not how code reviews will encourage participation and long term engagement. Just my 2c.

Flamefire:
> FWIW, the review process, especially given the state of the code base, should be reconsidered. I get it that you want to make things perfect, for some definition of perfect, but that is not how code reviews will encourage participation and long term engagement. Just my 2c.

Point well taken. I'm working hard (every once in a while at least) to get the code base to a better state so it is also easier to maintain.
Any specific points of the review you'd have suggestions on how to do better?
I tried to provide suggestions for the change so it would be low effort from your side to use them (click 1 button) and about 80% of the comments were suggestions on how to make the code compile/work/pass CI. A few nits were with naming, e.g. variables as lowercase (we should write that down somewhere I guess) and not introducing new magic constants (one of the things I'm removing over time where they still exist)

Anyway, thanks for this very useful feature! I'll take care of the remaining few issues and get this in.

jdoerfert:
> > FWIW, the review process, especially given the state of the code base, should be reconsidered. I get it that you want to make things perfect, for some definition of perfect, but that is not how code reviews will encourage participation and long term engagement. Just my 2c.
> 
> Point well taken. I'm working hard (every once in a while at least) to get the code base to a better state so it is also easier to maintain.

Don't get me wrong, I did not want to argue "against" the code base. I wanted to argue for "imperfect" additions that can be refined in-tree.

> Any specific points of the review you'd have suggestions on how to do better?

From what I've read on this topic, I decided to accept "correct" and "reasonable" code rather earlier than later. More patches and quicker turn around time is more encouraging than bugged down on a feature till it's "perfect", it never is anyway. There can be rules, e.g.,requiring tests is fine, but non-rule based change request should be optional and limited to a minimum. For example, if something like assertions or comments are supposed to be "too much", don't require a new round of review just to remove them.  Making suggestions is generally OK, but you should aim at the least amount of review rounds. One more option is a conditional accept. If changes are necessary before it can be merged, provide details so the next version is OK. Code can be improved in-tree as well, possibly faster. Maybe even ask the author to review those follow up patches so you even get the benefit of showing how you would the code to look like.

> I tried to provide suggestions for the change so it would be low effort from your side to use them (click 1 button) and about 80% of the comments were suggestions on how to make the code compile/work/pass CI. A few nits were with naming, e.g. variables as lowercase (we should write that down somewhere I guess) and not introducing new magic constants (one of the things I'm removing over time where they still exist)

I'm not very familiar with the github API, maybe that is on me. The problem with nits is that they should not require more review rounds. People need to make time to look at comments, read them, go to the repo, modify the files, upload, reply, go back to "wait mode", which is not perfect, especially in the beginning. 

> Anyway, thanks for this very useful feature! I'll take care of the remaining few issues and get this in.

Thanks, that would be great.

---

I'll check at some point if I have more features that might be upstreamable. For now, I'll concentrate on my bot which needs a redesign to keep timing issues in check.

Flamefire:
>  I wanted to argue for "imperfect" additions that can be refined in-tree.

Experience has shown that once code is accepted it often won't be "polished"/refined. It is *much* easier to do that in the review. So some level of quality must be achieved prior to merging. Most of that is guarded by CI already, i.e. warnings and unit tests.
Especially the latter are crucial for any new feature and useful for bugfixes.

> For example, if something like assertions or comments are supposed to be "too much", don't require a new round of review just to remove them. Making suggestions is generally OK, but you should aim at the least amount of review rounds

I think I mentioned those together with actual issues, i.e. compile failures due to (at least) warnings. But alright, I will be more careful here, thanks!

> Code can be improved in-tree as well, possibly faster.

I'd disagree here. The work is the same: You commit, push, done. The benefit doing that in one PR is that you don't need to open a new PR (which is some work) and that the initial PR shows the whole changes in one diff. So you can easily see what belongs together and why things are introduced/changed. In follow-ups this is harder as you likely need to open the whole code base to see where the changes comes into play.

> I'm not very familiar with the github API, maybe that is on me. 

In case that helps: The things mentioned show up as "Suggested change". You can (I think at least when in the "files" tab) click the "commit" button below it to accept that change and commit it and even add multiple such to a single commit (see that button). You might then want to locally pull and see if any surrounding code needs updating too but often that button is enough. It is a kinda new feature and I found that useful when having my code reviewed.

Flow86:
> Code can be improved in-tree as well, possibly faster.
I also have to disagree here. Changes which belong to each other should belong to each other - specially if one already saw them in a review.

jdoerfert:
> > Code can be improved in-tree as well, possibly faster.
> 
> I also have to disagree here. Changes which belong to each other should belong to each other - specially if one already saw them in a review.

I have no intention to fight you on this or start a discussion, I tried to provide you with some "best practices" I apply in other projects. You run this one as you wish.


> > I wanted to argue for "imperfect" additions that can be refined in-tree.
> 
> Experience has shown that once code is accepted it often won't be "polished"/refined. It is _much_ easier to do that in the review. So some level of quality must be achieved prior to merging. Most of that is guarded by CI already, i.e. warnings and unit 
tests.

I did not argue to include broken code, as defined by the CI for example. Imperfect could just mean that a new helper function is static and hidden instead of exposed in a new public header and then generalized, renamed, moved to a new place, and finally used in other places as well. All of that should be a follow up.

> Especially the latter are crucial for any new feature and useful for bugfixes.

Never objected to tests or working CI.

> > Code can be improved in-tree as well, possibly faster.
> 
> I'd disagree here. The work is the same: You commit, push, done. The benefit doing that in one PR is that you don't need to open a new PR (which is some work) and that the initial PR shows the whole changes in one diff. So you can easily see what belongs together and why things are introduced/changed. In follow-ups this is harder as you likely need to open the whole code base to see where the changes comes into play.

As mentioned above, your project, you do you. I guess you have more experience with the way people contribute here anyway, my feedback described what we do elsewhere but it might not be applicable.

Flamefire:
Follow-up in #1331, so closing this. Thanks a lot!

PR #1312, opened 2020-11-25
Economy mode for Return To The Roots

JonathanSteinbuch:
Dear repository maintainers,

I kindly ask you to review the changes I propose to the project.

I added an economy mode (a la settlers 3 economy mode, i.e. collect (in teams) as much as possible of 7 good types). For use with that I also added addons for a peaceful mode and to keep geologist signs visible for longer.

Please let me know if I should make adjustments.

Best Regards
Jonathan


Flow86:
Hi,

thank you for your effort here.

I have some remarks:

You've put a lot of new features into one thing (like the additional addons) - What about making a separeate PR for these, so these can be integrated much faster?

I like the "durable geologist signs" and "peaceful" mode addons.
- for the durable geologist signs: why not have a selection how long they are visible? (like Default, 2x, 4x, 10x)

JonathanSteinbuch:
Hi Flow, 

thanks for the fast reply and let me first of all thank you for the tremendous work you guys put into this project. I'm a huge fan.

Your remarks are reasonable. I will do as you suggest. To explain why I put it all into one thing: The main part is the economy mode and the rest resulted from playtesting with my family and what came up in that.

I'm closing this pull request in order to open separate ones.

Best Regards
Jonathan

PR #1313, opened 2020-11-26
Peaceful mode and geologist sign durability addons

JonathanSteinbuch:
These addons resulted out of playtesting (with the economy mode).
As suggested by @Flow86 in https://github.com/Return-To-The-Roots/s25client/issues/1312#issuecomment-734143961_ I present them here as a separate pull request.

Please let me know of any issues

Best Regards
Jonathan


libs/s25main/addons/AddonDurableGeologistSigns.h
```diff
@@ -0,0 +1,41 @@
...
+ *  Addon makes resource signs stay visible longer than normal
+ */
+class AddonDurableGeologistSigns : public AddonList
+{
+public:
+    AddonDurableGeologistSigns()
+        : AddonList(AddonId::DURABLE_GEOLOGIST_SIGNS, AddonGroup::Economy, _("Geologist sign durability"),
+                    _("Increase the durability of geologist signs by a factor."),
+                    {
+                      std::to_string(signDurabilityFactor[0]) + "x",
```
Flow86, line 34: Please name the first one "Default", also make the rest of it translateable (use boost::format + gettext)

libs/s25main/addons/AddonPeaceful.h
```diff
@@ -0,0 +1,32 @@
...
+#include "AddonBool.h"
+
+/**
+ *  Addon disables all attacks
+ */
+class AddonPeaceful : public AddonBool
+{
+public:
+    AddonPeaceful()
+        : AddonBool(AddonId::PEACEFUL, AddonGroup::GamePlay | AddonGroup::Military, _("Peaceful"),
```
Flow86, line 29: perhaps better name it "PeacefulMode" (also the AddonId PEACEFULMODE)

libs/s25main/ingameWindows/iwAction.cpp
```diff
@@ -503,15 +504,15 @@ void iwAction::Msg_TabChange(const unsigned ctrl_id, const unsigned short tab_id
                 break;
                 case TAB_ATTACK:
                 {
-                    if(available_soldiers_count > 0)
+                    if(available_soldiers_count > 0 && !gwv.GetWorld().GetGGS().getSelection(AddonId::PEACEFUL))
```
Flow86, line 507: you should probably better "zero" the available_soldiers_count  instead

Flow86: or even better, disable `tabs.attack` and `tabs.sea_attack`

libs/s25main/world/GameWorldGame.cpp
```diff
@@ -722,7 +722,8 @@ void GameWorldGame::Attack(const unsigned char player_attacker, const MapPoint p
                            const bool strong_soldiers)
 {
     auto* attacked_building = GetSpecObj<nobBaseMilitary>(pt);
-    if(!attacked_building || !attacked_building->IsAttackable(player_attacker))
+    if(!attacked_building || !attacked_building->IsAttackable(player_attacker)
+       || (GetGGS().getSelection(AddonId::PEACEFUL)))
```
Flow86, line 726: this should not be necessary if you simply disable the attack / sea attack tabs

JonathanSteinbuch: Thank you for all your suggestions! I think it is important that in the game logic itself attacks are prohibited (and not only in the UI), because otherwise the AI will happily attack me while I can't attack it. But these reviews made me realize that IsAttackable is directly or indirectly called in all the places I checked for the peaceful mode, so I replaced all these checks with a single one in nobBaseMilitary::IsAttackable, which is more elegant, I hope you agree.

Flow86:
Okay perfect. One thing you could check: Perhaps its possible to write a small (unit) test for the peaceful mode? like "set peaceful mode, try to attack -> should not work" - but that would be the 150% solution ;-)

libs/s25main/nodeObjs/noSign.cpp
```diff
@@ -16,8 +16,12 @@
 // along with Return To The Roots. If not, see <http://www.gnu.org/licenses/>.
 
 #include "noSign.h"
+#include "../GlobalGameSettings.h"
+#include "../world/GameWorldGame.h"
```
Flamefire, line 20: Such "go up" includes should never be required. Can you try this (needs resorting the include):

```suggestion
#include "GlobalGameSettings.h"
#include "world/GameWorldGame.h"
```

Flamefire (approved):

PR #1314, opened 2020-11-26
Economy mode

JonathanSteinbuch:
I added an economy mode (a la settlers 3 economy mode, i.e. collect (in teams) as much as possible of 7 good types).

This is a new version of my previous pull request https://github.com/Return-To-The-Roots/s25client/pull/1312. As suggested by @Flow86 in https://github.com/Return-To-The-Roots/s25client/issues/1312#issuecomment-734143961_ I moved the addons changes out to a separate branch (with pull request https://github.com/Return-To-The-Roots/s25client/pull/1313), upon which this branch builds. For discussion of my implementation of the economy mode I opened this pull request while the other is still running. 

To activate the economy mode in game, choos the game objective "Economy Mode". This adds an immaterial game object EconomyModeHandler which keeps tabs on the progress the players and teams make towards collecting the 7 good types. Also the EconomymodeHandler has an event which is triggered after a set time which can be changed in the addon menu. The EconomyModeHandler is added to the GameWorldBase and as such is serialized and deserialized when saving and loading (this is necessary to remember which good types to collect).
Also I added an ingame window (accessible from the ingame main menu) that shows the economic progress the player and his team make. At the end of the game also the amounts of the opposing teams are made visible and the map is revealed.

For playing the economy mode I recommend using a large map, not too short a time frame (I mostly play at 2 hours game length) and playing in teams with human opponents. (The AI can also be played against, but since this is meant as a multiplayer mode I didn't change the AI behaviour to collect specific goods)

Please let me know about any things that you would like to see improved.

Best Regards
Jonathan


libs/s25main/Game.cpp
```diff
@@ -41,7 +43,15 @@ void Game::Start(bool startFromSave)
     if(startFromSave)
         CheckObjective();
     else
+    {
+        if(ggs_.objective == GO_ECONOMYMODE)
+        {
+            unsigned int selection = ggs_.getSelection(AddonId::GAME_LENGTH);
+            world_.econHandler =
+              new EconomyModeHandler(AddonGameLengthList[selection] / 50); // 50 is the assumed game frame length
```
Flamefire, line 51: Why assume? You could use the actual selected speed here

libs/s25main/ingameWindows/iwEconomicProgress.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+iwEconomicProgress::~iwEconomicProgress() = default;
+
+void iwEconomicProgress::Draw_()
+{
+    IngameWindow::Draw_();
+
+    // Teamfarben zeichnen
+    const std::vector<EconomyModeHandler::econTeam>& teams = gwv.GetWorld().econHandler->GetTeams();
+    for(unsigned int t = 0; t < teams.size(); t++)
```
Flamefire, line 117: Maybe: `for(const unsigned teamIdx: teamOrder)`?

libs/s25main/ingameWindows/iwEconomicProgress.cpp
```diff
@@ -0,0 +1,200 @@
...
+            } else
+            {
+                text->SetText("???");
+            }
+        }
+    }
+
+    // Refresh game progress tracker
+    unsigned elapsed = 100;
+    if(!eH->isOver() && eH->GetEndFrame() > 0)
```
Flamefire, line 195: Why is the 2nd condition required? Could there be a better named function? E.g. `isInfiniteDuration` or so?

libs/s25main/ingameWindows/iwEconomicProgress.h
```diff
@@ -0,0 +1,45 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Return To The Roots. If not, see <http://www.gnu.org/licenses/>.
+
+#pragma once
+
+#include "../EconomyModeHandler.h"
```
Flamefire, line 20: No relative includes (`../`)

libs/s25main/ingameWindows/iwMainMenu.cpp
```diff
@@ -164,6 +172,11 @@ void iwMainMenu::Msg_ButtonClick(const unsigned ctrl_id)
             WINDOWMANAGER.ToggleWindow(std::make_unique<iwDiplomacy>(gwv.GetViewer(), gcFactory));
         }
         break;
+        case 12: // Wirtschaftsmodusfortschritt
```
Flamefire, line 175: No german comments please.

libs/s25main/world/GameWorldBase.h
```diff
@@ -67,6 +68,11 @@ class GameWorldBase : public World
     std::vector<GamePlayer> players;
     const GlobalGameSettings& gameSettings;
     EventManager& em;
+
+public:
+    EconomyModeHandler* econHandler = nullptr;
```
Flamefire, line 73: Please use a `std::unique_ptr`

libs/s25main/world/GameWorldBase.cpp
```diff
@@ -42,7 +42,10 @@ GameWorldBase::GameWorldBase(std::vector<GamePlayer> players, const GlobalGameSe
       gameSettings(gameSettings), em(em), gi(nullptr)
 {}
 
-GameWorldBase::~GameWorldBase() = default;
+GameWorldBase::~GameWorldBase()
+{
+    delete econHandler;
```
Flamefire, line 47: When using `std::unique_ptr` this can be reverted. Generally: No new/delete wherever possible

libs/s25main/world/GameWorldViewer.cpp
```diff
@@ -111,6 +111,12 @@ Visibility GameWorldViewer::GetVisibility(const MapPoint pt) const
     if(GAMECLIENT.IsReplayModeOn() && GAMECLIENT.IsReplayFOWDisabled())
         return VIS_VISIBLE;
 
+    // Im Wirtschaftsmodus und Spiel vorbei? Dann auch alles sichtbar
+    if(GetWorld().econHandler && GetWorld().econHandler->globalVisibility())
```
Flamefire, line 115: Hm, maybe a better way would be to make the world update all nodes fow members and set them to visible? Because this is a "set once" at the end of the econ mode, isn't it?

libs/s25main/ingameWindows/iwEconomicProgress.h
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) 2005 - 2017 Settlers Freaks (sf-team at siedler25.org)
```
Flamefire, line 1: It is 2020 ;)

Flamefire (changes requested):
Great work, looks pretty solid overall. Added some remarks to improve readability/maintainability.
Mostly about consistency, better naming and use of types/algorithms so it is easier to see what is what.

libs/s25main/world/GameWorldViewer.cpp
JonathanSteinbuch: I had tried this because I agree with your conceptual preference. But setting every node visible individually can take multiple seconds on very large maps, while the way I decided to use here takes no noticeable time. Also this same way is used for defeated players, which is also set at most once for each player. 

However, if you can give me a hint how to solve the performance issues with updating all nodes I am happy to change this.

libs/s25main/EconomyModeHandler.cpp
```diff
@@ -0,0 +1,344 @@
...
+    else
+        for(unsigned i = 0; i < gwg->GetNumPlayers(); ++i)
+        {
+            if(bestMask & getPlayerMask(i))
+            {
+                gwg->GetGameInterface()->GI_Winner(i);
+            }
+        }
+
+    // Call function to recalculcate visibilities
```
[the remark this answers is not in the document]
JonathanSteinbuch: Because by calling the alliance-change-function visibilities are updated in an efficient way

libs/s25main/world/GameWorldViewer.cpp
Flamefire: Multiple seconds? And you simply loop over the nodes and fow entries?
Hm... that's hard then. Sure that this is optimised (release) mode already? Maybe we need to restructure the array from AoS to SoA

libs/s25main/EconomyModeHandler.h
```diff
@@ -0,0 +1,93 @@
...
+    const GameEvent* event;
+    /// Good types to collect
+    GoodType types[numGoodTypesToCollect];
+
+    // Data for economy mode progress tracking
+    std::vector<econTeam> teams;
+    unsigned maxTeamAmounts[numGoodTypesToCollect] = {0};
+    unsigned int mostWins = 0;
+    unsigned amounts[numGoodTypesToCollect][MAX_PLAYERS];
+    unsigned last_updated;
```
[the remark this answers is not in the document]
JonathanSteinbuch: The amounts are only updated when either iwEconomicProgress or the end game event request it. I save the gameframe at which we updated last, because iwEconomicProgress asks for an update every video frame and otherwise we would update more than once in a gameframe.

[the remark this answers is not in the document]
Flamefire: Makes sense, alright :)

libs/s25main/ingameWindows/iwEconomicProgress.cpp
JonathanSteinbuch: Here I also use the counter t to do the positioning of the rectangles so changing this would need to store the positioning delta together with the team order.

libs/s25main/world/GameWorldViewer.cpp
JonathanSteinbuch: I just looked at it more thoroughly. Previously I had called `GameWorldGame::MakeVisible(const MapPoint pt, const unsigned char player)` which seemed the correct way. But I now think this is slow because it updates the minimap and potentially lua scripts etc. after every single point. 

So I added a function to world directly working on the nodes and then a function to the GameInterface to explicitly update the minimap and colors afterwards. This seems to be without noticeable performance issues. (Note that I only actually implemented this interface method in dskGameInterface, because that's the only point where it's necessary for now)

libs/s25main/EconomyModeHandler.cpp
```diff
@@ -22,181 +22,178 @@
...
-  /*  5 */ GD_PICKAXE,    // Spitzhacke
-  /*  6 */ GD_SHOVEL,     // Schaufel
-  /*  7 */ GD_CRUCIBLE,   // Schmelztiegel
-  /*  8 */ GD_RODANDLINE, // Angel
-  /*  9 */ GD_SCYTHE,     // Sense
-  /* 12 */ GD_CLEAVER,    // Beil
-  /* 13 */ GD_ROLLINGPIN, // Nudelholz
-  /* 14 */ GD_BOW,        // Bogen
+std::array<GoodType, 12> specialGoodPool = {
+  /*  1 */ GD_TONGS,
```
Flamefire, line 34: Could you either remove or update the index comments too? 1-9, then 12-14 doesn't make sense to me. I guess removing them is just fine.
PS: I'd really liked if C++ had a way to auto-deduce the size of a std::array but I guess that is C++17 with deduction guides only. Maybe a `makeArray` constexpr function would be good? It could be templated to take an array and deduce type and size and return a std::array. This way mistakes like having a size of 12 but 11 entries can be avoided.

libs/s25main/EconomyModeHandler.cpp
```diff
@@ -22,181 +22,178 @@
...
-  /* 24 */ GD_STONES,  // Steine
-  /* 27 */ GD_GRAIN,   // Getreide
-  /* 28 */ GD_COINS,   // Mnzen
-  /* 29 */ GD_GOLD,    // Gold
-  /* 30 */ GD_IRONORE, // Eisenerz
-  /* 31 */ GD_COAL,    // Kohle
-  /* 32 */ GD_MEAT,    // Fleisch
-  /* 33 */ GD_HAM,     // Schinken ( Schwein )
+std::array<GoodType, 18> commonGoodPool = {
+  /*  0 */ GD_BEER,
```
Flamefire, line 49: Same here, comments are not needed I guess

libs/s25main/EconomyModeHandler.cpp
```diff
@@ -22,181 +22,178 @@
...
+  /* 28 */ GD_COINS,
+  /* 29 */ GD_GOLD,
+  /* 30 */ GD_IRONORE,
+  /* 31 */ GD_COAL,
+  /* 32 */ GD_MEAT,
+  /* 33 */ GD_HAM,
 };
 
-EconomyModeHandler::EconomyModeHandler(unsigned end_frame) : end_frame(end_frame), last_updated(0)
+const unsigned int numGoodTypesToCollect = 7;
```
Flamefire, line 69: `constexpr unsigned numGoodTypesToCollect = 7;` (remove the int for consistency and use constexpr where possible

libs/s25main/EconomyModeHandler.h
```diff
@@ -14,49 +15,58 @@ class GameEvent;
...
+        EconTeam(std::bitset<MAX_PLAYERS> playersInTeam, unsigned int numGoodTypesToCollect) noexcept
+            : playersInTeam(playersInTeam), amountsTheTeamCollected(numGoodTypesToCollect, 0), goodTypeWins(0)
+        {}
+
+        EconTeam(SerializedGameData& sgd, unsigned int numGoodTypesToCollect);
+
+        void Serialize(SerializedGameData& sgd) const;
 
+        // Returns true if the player is in the team
         bool inTeam(unsigned int playerId) const;
```
Flamefire, line 39: Nit: For such query like function adding verbs is a good idea. Here: `isInTeam` or better:  `containsPlayer`, then it really reads like english on the call site: `team.containsPlayer(playerId)`
You don't need to change it, just a note

libs/s25main/EconomyModeHandler.cpp
JonathanSteinbuch: sure.

To the PS: I read that C++20 can auto-deduce the size from an initializer list (with std::to_array). For now I changed it to a const vector, because vector can auto-deduce the size. I know it isn't necessarily done at compile time then, but here I think it doesn't really matter either way.

libs/s25main/EconomyModeHandler.cpp
```diff
@@ -0,0 +1,344 @@
...
+#include "GlobalGameSettings.h"
+#include "SerializedGameData.h"
+#include "random/Random.h"
+#include "world/GameWorld.h"
+#include "world/GameWorldGame.h"
+#include "gameTypes/JobTypes.h"
+#include "gameData/JobConsts.h"
+#include <boost/optional.hpp>
+
+auto getPlayerMask = [](unsigned playerId) { return 1u << playerId; };
```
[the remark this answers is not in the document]
JonathanSteinbuch: with bitset I didn't need this anymore anyway

libs/s25main/EconomyModeHandler.cpp
```diff
@@ -0,0 +1,344 @@
...
+        if(i == types_found)
+        {
+            types[types_found] = next_type;
+            types_found++;
+        }
+    }
+    types[numGoodTypesToCollect - 1] = types_less[RANDOM.Rand(__FILE__, __LINE__, GetObjId(), num_types_less)];
+    types_found++;
+
+    if(end_frame > 0)
```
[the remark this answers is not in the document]
JonathanSteinbuch: I realized, that EconomyModeHandler doesn't even need to keep track of the event, because the EventManager will do that for me

libs/s25main/EconomyModeHandler.cpp
```diff
@@ -0,0 +1,344 @@
...
+    {
+        sgd.PushUnsignedChar(type);
+    }
+}
+
+void EconomyModeHandler::FindTeams()
+{
+    // If we already determined who is in a team with whom skip this. For the economy mode we only count teams at game
+    // start
+    if(!teams.empty())
```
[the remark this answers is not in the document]
JonathanSteinbuch: makes it somewhat cleaner, yes

libs/s25main/desktops/dskHostGame.cpp
```diff
@@ -759,6 +764,36 @@ void dskHostGame::Msg_MsgBoxResult(const unsigned msgbox_id, const MsgboxResult
                 UpdateGGS();
         }
         break;
+        case 10: // Economy Mode - change Addon Setttings
+        {
+            if(mbr == MSR_YES)
+            {
+                gameLobby->getSettings().setSelection(AddonId::PEACEFULMODE, true);
```
Flow86, line 771: probably we should not imply peaceful mode here - still could be interesting to fight while trying to achieve the economy goal?

JonathanSteinbuch: Upon further consideration, I agree. I had originally added this, because at first I didn't see how you could keep addon settings between games. I will revert that. 

I will for now keep the other check that you don't play with peaceful mode in attack based games, unless you think that's superfluous.

Maybe in the future it would be nice to be able to save multiple addon profiles. Would be convenient if there are different game modes you like to play with

libs/s25main/EconomyModeHandler.cpp
Flamefire: Actually `to_array` does more than that, the deduction guide rather acts like the now-removed `make_array`. I actually like that one, it seems to be pretty well and usable: https://godbolt.org/z/5x5xzE

libs/s25main/desktops/dskHostGame.cpp
Flamefire: @Flow86 It isn't implied. There is a message box asking you whether to enable peaceful mode. I think that is a good idea as that might be what is often wanted. If the message box says exactly what it changes I'd keep it.

libs/s25main/ingameWindows/iwEconomicProgress.cpp
Flamefire: You mean in `DrawPoint drawPt = GetDrawPos() + DrawPoint(37 + (t + 1) * 63, 22);`?
Then you could do `DrawPoint curTeamRectPos= GetDrawPos() + DrawPoint(37 + 63, 22);` before the loop and `curTeamRectPos.x += 63;` inside the loop, couldn't you?

libs/s25main/desktops/dskHostGame.cpp
```diff
@@ -1019,3 +1036,22 @@ void dskHostGame::LC_Chat(const std::string& player, const std::string& text)
...
+bool dskHostGame::checkOptions()
+{
+    if(forceOptions)
+        return true;
+    const GlobalGameSettings& ggs = gameLobby->getSettings();
+    if(ggs.getSelection(AddonId::PEACEFULMODE)
+       && (ggs.objective == GO_CONQUER3_4 || ggs.objective == GO_TOTALDOMINATION))
+    {
+        WINDOWMANAGER.Show(std::make_unique<iwMsgbox>(
+          _("peaceful Mode"),
```
Flamefire, line 1049:
```suggestion
          _("Peaceful mode"),
```
It is a title, so capitalize the first word

libs/s25main/desktops/dskHostGame.cpp
```diff
@@ -1019,3 +1036,22 @@ void dskHostGame::LC_Chat(const std::string& player, const std::string& text)
...
+          _("You chose a war based victory condition but peaceful  mode is still active. "
+            "Would you like to disactivate peaceful  mode before you start? Choosing no will "
+            "start the game, yes will let you review the changes."),
```
Flamefire, line 1052: Minor typo fix:
```suggestion
          _("You chose a war based victory condition but peaceful mode is still active. "
            "Would you like to deactivate peaceful mode before you start? Choosing No will "
            "start the game, Yes will let you review the changes."),
```

libs/s25main/ingameWindows/iwEconomicProgress.cpp
JonathanSteinbuch: I didn't think of this. For a moment I thought this was some insanely clever idea I missed until now. Then I realized that curTeamRectPos.x basically becomes a counter variable. Still looks a bit cleaner, I agree. Thanks.

Also this made me notice that I use way too many explicit numbers in the construction of this window, so I cleaned that up.

Also made me realize that I can simply store pointers in teamOrder which makes it even simpler

libs/s25main/EconomyModeHandler.cpp
JonathanSteinbuch: It's the least I can do. After all, I should have read the coding conventions for this project before making a pull request (I just found them). :D

libs/common/include/helpers/make_array.h
```diff
@@ -0,0 +1,27 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with Return To The Roots. If not, see <http://www.gnu.org/licenses/>.
+
+#pragma once
+
+#include <array>
+
+template<class D = void, class... Types>
+constexpr auto make_array(Types&&... t)
```
Flamefire, line 23: Could you put this into the namespace (see the other files in this folder)
I started putting stuff into namespaces, hence the mix of files in the include folder and this subfolder

libs/s25main/EconomyModeHandler.cpp
```diff
@@ -33,56 +34,29 @@
...
-      GD_STONES,
-      GD_GRAIN,
-      GD_COINS,
-      GD_GOLD,
-      GD_IRONORE,
-      GD_COAL,
-      GD_MEAT,
-      GD_HAM,
-    });
+    const auto specialGoodPool =
```
Flamefire, line 37: You can use constexpr here so the array is created at compiletime. Also the template argument to make_array shouldn't be required. It is optional in case mixed values are used

JonathanSteinbuch: I would argue that while the template argument is not necessary it's actually useful here so we get a type check for the elements put inside. But you seem to care much more about these details than me, so I removed the template argument.

tests/s25Main/integration/testEconomyMode.cpp
```diff
@@ -0,0 +1,144 @@
...
+        hqPos[playerIdx] = world.GetPlayer(playerIdx).GetHQPos();
+    }
+    this->ggs.objective = GO_ECONOMYMODE;
+
+    world.econHandler = std::make_unique<EconomyModeHandler>(4);
+    BOOST_REQUIRE(world.econHandler);
+    const auto goodsToCollect = world.econHandler->GetGoodTypesToCollect();
+
+    BOOST_REQUIRE(!goodsToCollect.empty());
+    BOOST_REQUIRE_EQUAL(world.econHandler->GetEndFrame(), (unsigned)4);
```
Flamefire, line 58:
```suggestion
    BOOST_REQUIRE_EQUAL(world.econHandler->GetEndFrame(), 4u);
```

tests/s25Main/integration/testEconomyMode.cpp
```diff
@@ -0,0 +1,144 @@
...
+    for(unsigned i = 0; i < world.GetNumPlayers(); i++)
+        save.AddPlayer(world.GetPlayer(i));
+    save.ggs = ggs;
+    save.start_gf = em.GetCurrentGF();
+    save.sgd.MakeSnapshot(game);
+    TmpFile tmpFile;
+    BOOST_TEST_REQUIRE(save.Save(tmpFile.filePath, "MapTitle"));
+    Savegame loadSave;
+    BOOST_TEST_REQUIRE(loadSave.Load(tmpFile.filePath, SaveGameDataToLoad::All));
+    BOOST_TEST_REQUIRE(loadSave.GetNumPlayers() == (unsigned)3);
```
Flamefire, line 115:
```suggestion
    BOOST_TEST_REQUIRE(loadSave.GetNumPlayers() == 3u);
```

tests/s25Main/integration/testEconomyMode.cpp
```diff
@@ -0,0 +1,144 @@
...
+    BOOST_TEST_REQUIRE(newWorld.econHandler);
+    auto& goodsToCollectAfter = newWorld.econHandler->GetGoodTypesToCollect();
+    auto& econTeamsAfter = newWorld.econHandler->GetTeams();
+    BOOST_REQUIRE(goodsToCollect == goodsToCollectAfter);
+    BOOST_REQUIRE_EQUAL(econTeams.size(), econTeamsAfter.size());
+    newWorld.GetEvMgr().ExecuteNextGF();
+    newWorld.econHandler->UpdateAmounts();
+    for(auto& teamPair : boost::combine(econTeams, econTeamsAfter))
+    {
+        auto& before = teamPair.get_head();
```
Flamefire: Using `boost::combine` is pretty neat! Maybe use `boost::tie(f, s) = var;` instead of the `get_head` stuff? Especially the `teamPair.get_tail().get_head();` disturbs me, I couldn't even tell if that is correct. Or maybe `get<0>(teamPair)` already works (`tie` won't work with references)?

Also please use `const auto&` (const wherever possible to avoid accidentally changing something) and be careful with `auto`: It makes code harder to read when the type is not clear, so I tend to use it only for hard-to-type types (e.g. iterators) or when the type appears on the RHS (`auto x = static_cast<Foo>(bar)`)

tests/s25Main/integration/testEconomyMode.cpp
```diff
@@ -0,0 +1,144 @@
...
+    save.sgd.MakeSnapshot(game);
+    TmpFile tmpFile;
+    BOOST_TEST_REQUIRE(save.Save(tmpFile.filePath, "MapTitle"));
+    Savegame loadSave;
+    BOOST_TEST_REQUIRE(loadSave.Load(tmpFile.filePath, SaveGameDataToLoad::All));
+    BOOST_TEST_REQUIRE(loadSave.GetNumPlayers() == (unsigned)3);
+    std::vector<PlayerInfo> players;
+    for(unsigned j = 0; j < 3; j++)
+        players.push_back(PlayerInfo(loadSave.GetPlayer(j)));
+    std::shared_ptr<Game> sharedGame(new Game(save.ggs, loadSave.start_gf, players));
```
Flamefire, line 119: Maybe:
```suggestion
   auto sharedGame = std::make_shared<Game>(save.ggs, loadSave.start_gf, players);
```

tests/s25Main/integration/testEconomyMode.cpp
```diff
@@ -129,8 +129,8 @@ BOOST_FIXTURE_TEST_CASE(EconomyMode3Players, WorldWithGCExecution3P)
     newWorld.econHandler->UpdateAmounts();
     for(const auto& teamPair : boost::combine(econTeams, econTeamsAfter))
     {
-        const auto& before = teamPair.get_head();
-        const auto& after = teamPair.get_tail().get_head();
+        const EconomyModeHandler::EconTeam& before = boost::get<0>(teamPair);
+        const EconomyModeHandler::EconTeam& after = boost::get<1>(teamPair);
```
Flamefire, line 133: Just a hint (no need to change): You can rely on ADL here and use `get<0>(teamPair)`. This is less explicit and has the benefit that one could switch to a combine/zip iterator that returns std:: types

libs/s25main/desktops/dskHostGame.cpp
JonathanSteinbuch: I don't know who has seniority here or something. I put the message box back in and made the text much clearer. Just decide amongst yourselves and possibly revert the commit if you decide against it.

tests/s25Main/integration/testEconomyMode.cpp
JonathanSteinbuch: Ok, thanks

libs/s25main/desktops/dskHostGame.cpp
```diff
@@ -1042,8 +1057,23 @@ bool dskHostGame::checkOptions()
...
+            "peaceful mode. Would you like to adjust settings for a peaceful "
+            "game?\n"
+            "Choosing Yes will activate peaceful mode, ban catapults and disable buildings receiving coins by default. "
+            "After clicking Yes you will be able to review the changes and then start the game by clicking the Start "
+            "game button again. If you like to play economy mode with attacks, "
+            "choosing No will start the game without any "
+            "changes."),
+          this, MSB_YESNOCANCEL, MSB_QUESTIONGREEN, 10));
+        return false;
+    } else if(ggs.getSelection(AddonId::PEACEFULMODE)
```
Flamefire, line 1075: Maybe better. `ggs.isEnabled(AddonId::PEACEFULMODE)`?
Also can you reformat the message? Line breaks look arbitrary, maybe add one after `\n` and then check that the lines in code are ok? (i.e. the `"game?\n"` is odd)
Besides that the message is pretty verbose but I like it. Maybe I'd omit the  `(check the economic progress window in game)` but I guess that is good for newcomers
I'd also add a newline before `If you like to play economy mode with attacks,` and maybe make it similar to the Yes part: "Choosing No will start the game without any changes." should be enough

JonathanSteinbuch: I like the `(check the economic progress window in game)` so I kept it for now. The rest I changed

JonathanSteinbuch:
I'm not sure if there are any change requests I missed but to me this looks good now.

PS: I noticed one issue: The games take longer than calculated by the game frame length (~25% on fast speed). I also found out the reason: Video frame limiting. `ExecuteGameFrame` only gets called every 1000/60 = ~16.7 ms at 60fps. In effect every game frame is on average 8.3 ms longer than it should be, which at a game frame length of 40ms is about 25%. Of course this effect can be reduced by turning of frame limiting but in my opinion the internal game speed should ideally not depend on the video settings.

I also created a solution https://github.com/JonathanSteinbuch/s25client/tree/timing, by keeping track of these little deltas and calling subsequent frames a little earlier, which works great when the game frame length is greater than the video frame length. When the two frame lengths get close, movements can be a little less smooth than normally, but if you want smooth you can increase the frame limit, that makes sense. Of course for a game frame length lower than the video frame length this method can't keep the time accurate (because we would have to do micro skips basically) but it still is a little closer than before.

Because this affects the whole game and not just economy mode I should open a new pull request for the timing issue, right?

Flamefire:
I'll do a final review once I got some more time. Or @Flow86 if he is faster.

I'm aware of the problem and actually we have to handle it and ideally in either way. So making multiple game steps per video frame should be possible too. For a similar purpose (video frame rate, aka software VSync) I created FrameTimer and FrameLimiter classes. Those should be used for that too, but needs checking if they are generic enough.
And yes that is an extra PR. In general: The smaller a change, the better. So it can be reviewed and merged fast

libs/s25main/desktops/dskHostGame.cpp
Flow86: ah ok then it should be fine if you are asked for it.

libs/s25main/EconomyModeHandler.cpp
```diff
@@ -0,0 +1,302 @@
...
+        }
+    }
+
+    return retVal;
+}
+
+void EconomyModeHandler::UpdateAmounts()
+{
+    // Return if the game is over or we already updated the amounts this game frame
+    if(isOver() || gfLastUpdated == GetEvMgr().GetCurrentGF())
```
Flow86, line 187: do we want to update each GF? or should we also say its calculated every "now and then"?

Flow86: besides this question everything looks good!

Flamefire: If you go for the "now and then" approach: I remember that there is a "statistics step". I guess that would be a good occasion.

@Flow86 Is the "heavy load" measured? If so, maybe we can improve that?
BTW: Do our toolchains support OpenMP? That would be a simple way to get some more speed here.

JonathanSteinbuch: I looked at this. Indeed, if I open the Economic progress window the framerate (without frame limiting) drops significantly. So I analysed and optimized a bit and even changed it to only update the iwEconomicProgress content at most once every second, because that's certainly enough.

But it doesn't make a significant difference in the frame rate. Indeed, when you do some perfomance analysis, you will see that the only thing for the ingame windows that takes any significant amount of time at all is drawing. And this seems to depend on the number and type of control elements displayed in the window, for example the tool window with tool ordering enabled is even much worse than the economic progress window. So I recommend that if the goal is to reduce the load while displaying ingame windows the Draw routine should be optimized, for example by prerendering/baking the whole windows except elements that change into textures or if possible by using vector buffers instead of directly drawing quads. But this is vastly beyond the scope of this pull request.

So tldr: I checked, and no, in comparison with the whole situation of drawing an ingame window these calculations are not a heavy load at all.

JonathanSteinbuch: Should be all good now again

JonathanSteinbuch: I thought about this again, and to be honest, I don't like these last three "performance optimization" commits I made very much, because of stylistic and clarity reasons, and as I wrote it's probably unnecessary to have them. I regret them and think I will revert them tomorrow. Should also make it easier to review as @Flow86 already said before that it basically looks good.

JonathanSteinbuch:
I apologize for my imperfect handling of @Flow86's question.

Just for clarification since I didn't hear from any of you in a while here: Are you waiting for any additional input from me on this point or are you just not finding the time to answer? (which is fine by me, no worry, it's a hobby project after all)

Best Regards
Jonathan

Flamefire (dismissed):

Flamefire:
@Flow86 Any idea why jenkins didn't run? Travis is trash now, so we have to migrate that to GHA but I'd like to have at least a green from jenkins before merging

Flow86:
@Flamefire Jenkins is currently broken because of this: https://github.com/jenkinsci/docker-plugin/issues/821

Flamefire:
@Flow86 Any idea what that is? It looks new to me:

> ld: archive has no table of contents file '../s25main/libs25Main.a' for architecture x86_64

Flow86:
probably a cached broken file. I cleared the cache and rerun the build now

Flamefire:
Can you rebase please so the new CI runs?

JonathanSteinbuch:
@Flamefire, I did a merge, which seemd fine to me. If you insist on a rebase let me know.

Flow86:
> @Flamefire, I did a merge, which seemd fine to me. If you insist on a rebase let me know.

please do a rebase, feature branches should be merge-commit free

JonathanSteinbuch:
@Flow86, as requested, I pushed a rebase. 
Since I saw you working on the credits I took the liberty to add a commit putting my name under Additional Programming

libs/s25main/EconomyModeHandler.cpp
```diff
@@ -0,0 +1,300 @@
...
+
+    constexpr unsigned numGoodTypesToCollect = 7;
+
+    // Randomly determine *numGoodTypesToCollect* many good types, one of which is a special good (=tool)
+
+    static_assert(numGoodTypesToCollect > 0, "There have to be goods to be collected");
+    static_assert(commonGoodPool.size() >= numGoodTypesToCollect - 1, "There have to be enough commond goods");
+    static_assert(!specialGoodPool.empty(), "There have to be enough special goods");
+    goodsToCollect.clear();
+    goodsToCollect.resize(numGoodTypesToCollect);
```
Flamefire, line 53: Just stumbled upon this and took a little bit to long to understand it although it should be very easy. May I suggest a minor change: Don't use `resize` but `reserve` (nit: The `clear` is superflous and can be removed), use a regular `for` loop with `push_back` below and instead of `std::find` you can use `helpers::contains`

That is if you want to keep using a vector here, the code is fully reasonable for an array. Was there a reason you switched this to vector although the `numGoodTypesToCollect` is a compile time constant?

JonathanSteinbuch: Using reserve and contains is a good change, thanks for the suggestion, did it. 

Going to a regular for loop would require changing the goods pools to sets, because I don't want duplicates in goodsToCollect. With the current implementation if the random generator gives out a good we already have in goodsToCollect I just go for another round in the while loop until I have found enough unique goods. If I changed the goods pools to sets I could just remove any found goods from the set so I wouldn't have to do extra loop iterations. The overhead didn't seem worth it to me though, especially since I asserted above that there are enough goods in the pool, so I won't run into an infinite loop. If you prefer the set approach for clarity, though, I will do it.

I changed goodsToCollect into a vector when I worked on it anyway, because then numGoodTypesToCollect can be nicely contained only in the constructor. This makes it possible to very easily have an addon option to change the number in the future, if you want. I didn't want to implement such an addon option for this pull request though, because it's not very important and I didn't want to put more features into this pull request (uncharacteristical for me, I know xD).

JonathanSteinbuch: Oh, now I remember, another reason to use a vector here, was that SerializedGameData::PopContainer at the moment only seems to support dynamically sized containers.

Flamefire (approved):

PR #1318, opened 2020-12-18
Fix a click through bug for combobox lists in addon window

JonathanSteinbuch:
As the title says, the addon options window has a bug where you can click through a combo box list item and also select the option below. This commit fixes this by not only locking the list region for the parent of the list, but recursively for all parents.

### DIFF
--- a/libs/s25main/Window.cpp
+++ b/libs/s25main/Window.cpp
@@ -179,9 +179,7 @@ void Window::LockRegion(Window* window, const Rect& rect)
 
     // Also lock the region for all parents
     if(GetParent())
-    {
         GetParent()->LockRegion(this, rect);
-    }
 }
 
 /**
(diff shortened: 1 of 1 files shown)


Flamefire (changes requested):
Good catch! I'd just like the braces removed for the single line if as is done for all other places in the surrounding code.
Note: merging is going to take a bit as I'm currently replacing travis CI

Flamefire:
Can you rebase please so the new CI runs?

Flamefire:
Something is odd, close and reopen to retrigger CI

Flamefire:
@Flow86 Can you check where the `Linux_And_OSX ` CI stuff comes from?  That looks like some stuff I had in my branch but removed. To me the 3 PRs look good

Flamefire (approved):

Flow86:
> 
> 
> @Flow86 Can you check where the `Linux_And_OSX ` CI stuff comes from? That looks like some stuff I had in my branch but removed. To me the 2 PRs look good

which ones should be active?

Issue #1319, opened 2020-12-18
Game frame time wrong with vsync


JonathanSteinbuch:
Hello,

I mentioned this issue already in a comment for a pull request: https://github.com/Return-To-The-Roots/s25client/pull/1314#issuecomment-736555530

The issue is: The games can take longer than calculated by the game frame length (~20% on fast speed). I also found out the reason: Video frame limiting or vsync. ExecuteGameFrame only gets called every 1000/60 = ~16.7 ms at 60fps. In effect every game frame is on average 8.3 ms longer than it should be, which at a game frame length of 40ms is about 20%. Of course this effect can be reduced by turning off frame limiting but in my opinion the internal game speed should ideally not depend on the video settings.

The problem occurs because the execution of a game frame and drawing is coupled in the same loop. Hardware vsync makes this especially notorious, because when calling SwapBuffers the process sleeps until the vsync event.

@Flamefire commented on this: "I'm aware of the problem and actually we have to handle it and ideally in either way. So making multiple game steps per video frame should be possible too. For a similar purpose (video frame rate, aka software VSync) I created FrameTimer and FrameLimiter classes. Those should be used for that too, but needs checking if they are generic enough."

Today @Flamefire cautioned me to "discuss this in an issue and come up with a design first. I think this is very important but also easy to get wrong. So we should get it right in all circumstances which may be hard given we have to synchronize over multiple clients."

Over the last days I created three prototype-solutions. 

The first one is a minimal solution by keeping track of the little timing offsets and calling subsequent frames a little earlier, which works great when the game frame length is greater than the video frame length. When the two frame lengths get close, movements can be a little less smooth than normally, but if you want smooth you can increase the frame limit, that makes sense. Of course for a game frame length lower than the video frame length this method can't keep the time accurate (because we would have to do micro skips basically) but it still is a little closer than before.
Prototype: https://github.com/JonathanSteinbuch/s25client/tree/frametimingMinimal

The next solution solves the problem using multithreading by completely decoupling game frame and the rest. I think this is conceptually the nicest solution. It also offers opportunities for concurrent execution (although for simplicity I used a single mutex for all game data in my prototype, so effectively this isn't used). This solution has the drawback, that OpenGL wants all calls to be in the main thread. So as of now every time new land is explored or a planer works this segfaults. However, I think this could be solved by deferring all notifications that involve GL calls for the next drawing frame.
[...]

PR #1320, opened 2020-12-21
Fix client-side time synchronisation for simple cases

Flamefire:
Most simple implementation to avoid #1319 where actual simulation time will be lower than real time due to incrementing it essentially in chunks (similar to/of) VSync time.
This works for cases where the GF length is larger than the execution time for 1 GF + drawing and related.

This is only intermediate until we implement something better, see #1319, but improves the situation already. Similar to https://github.com/JonathanSteinbuch/s25client/commit/536a64fb9f83d55f1a4e5f0ce66ef052b3bbf5c1

### DIFF
--- a/libs/s25main/network/GameClient.cpp
+++ b/libs/s25main/network/GameClient.cpp
@@ -1217,11 +1217,21 @@ void GameClient::ExecuteGameFrame()
     }
 
     const unsigned curGF = GetGFNumber();
+    const bool isSkipping = skiptogf > curGF;
     // Is it time for the next GF? If we are skipping, it is always time for the next GF
-    if(skiptogf > curGF || (currentTime - framesinfo.lastTime) >= framesinfo.gf_length)
+    if(isSkipping || (currentTime - framesinfo.lastTime) >= framesinfo.gf_length)
     {
         try
         {
+            if(isSkipping)
+            {
+                // We are always in realtime
+                framesinfo.lastTime = currentTime;
+            } else
+            {
+                // Advance simulation time (lastTime) by 1 GF
+                framesinfo.lastTime += framesinfo.gf_length;
+            }
             if(replayMode)
             {
                 // In replay mode we have all commands in the file -> Execute them
@@ -1268,10 +1278,6 @@ void GameClient::ExecuteGameFrame()
                     replayinfo->replay.UpdateLastGF(curGF);
             }
 
-            // Store this timestamp
-            framesinfo.lastTime = currentTime;
-            // Reset frameTime
-            framesinfo.frameTime = FramesInfo::milliseconds32_t::zero();
         } catch(LuaExecutionError& e)
         {
             if(ci)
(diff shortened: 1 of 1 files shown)


Flamefire:
@JonathanSteinbuch If you want to take a look to check if you can find any issues with that. Should improve the situation although it is not complete, i.e. we need to handle the case where we have to run multiple GFs, which I leave to the issue for now.

JonathanSteinbuch:
Looks good to me!
I would guess that for almost all real world scenarios this is already good enough.

Flamefire:
Yes I think so too. However I've set a breakpoint at the "special case" if at the bottom and it got hit. Most frametimes were about 17-20ms as expected but some spiked to >100ms. So we could take that into account as well. Maybe keep track of average times for running a single GF and loop if that average is below the GF_length, i.e. we likely catch up. We can also think about whether to query the time again after running a GF or not (i.e. update `currentTime`), not doing so avoids the infinite loop issue but may lead to some lacks/delays (e.g.: Time for 1.99 GFs has passed -> 1 GF executed, now in real time the next GF is already due but when we don't update `currentTime` we won't notice until [e.g.] 16ms later `Run` is called again)

JonathanSteinbuch:
Now that you say it, it would probably already make it more resilient to do something similar as on the server side and only reset the lastTime when framesinfo.frameTime >= 4*framesinfo.gf_length. (Or some other factor, maybe 2*the number of game frames between network frames as that seems to be the maximum before we definitely lag for other players...)

If the frameTime is above maxFrameTime I wouldn't worry too much about the drawing accuracy either and just generally clamp it at maxFrameTime. Because this is a scenario where animations get a bit nonsmooth either way.

Personally I would not update currentTime during a single drawing. Not only does it avoid infinite loops, but it also might look a little more smooth as otherwise the time reference point inside a frame varies. Of course we are always one video frame behind in drawing (as compared to a little less than one video frame), but it's not an action game anyway.

Flamefire:
> Now that you say it, it would probably already make it more resilient to do something similar as on the server side and only reset the lastTime when framesinfo.frameTime >= 4*framesinfo.gf_length

This won't work: The drawing code for its interpolation has this assumption that frameTime is always the time into the next frame, i.e. less then gf_length and I'm a bit afraid to change that. Or did you mean clamping `frameTime` here as I already do but not touch `lastTime` i.e. let those be out-of-sync on purpose? This would likely work but introduce lags while keeping them in sync will be more smooth because the interpolation assumes that they are in sync, so we (currently) basically let the simulation lag to keep the animation smooth (If I got that right). Might be worth a shot to try as it may increase smoothness in multiplayer by compensating small lags on one players machine

JonathanSteinbuch:
> Or did you mean clamping `frameTime` here as I already do but not touch `lastTime` i.e. let those be out-of-sync on purpose?
Yes, that's exactly what I meant. And in this case animation is "wrong" if we touch lastTime or not.

I just tried it out with the following code on a large save file at gf_length 40ms with a breakpoint in both cases and it hit both but only once every few seconds. So some lags could be ameliorated by this, but about as many could not. When I changed the factor to 20 I didn't get a hit in the first case at all until I changed the gf_length to 20ms.

```using DurationType = decltype(framesinfo.gf_length);
const auto maxFrameTime = framesinfo.gf_length - DurationType(1);
    if(framesinfo.frameTime >= 4 * framesinfo.gf_length)
    {
        // This can happen, if we don't call this method in intervalls less than gf_length or gf_length has changed
        // TODO: Run multiple GFs per call.
        // For now just make sure it is less than gf_length by skipping some simulation time,
        // until we are only a bit less than 1 GF behind
        RTTR_Assert(framesinfo.gf_length > DurationType::zero());
        framesinfo.lastTime += framesinfo.frameTime - maxFrameTime;
        framesinfo.frameTime = maxFrameTime;
    } else if(framesinfo.frameTime >= framesinfo.gf_length)
        framesinfo.frameTime = maxFrameTime;
```

Flamefire:
Ok, done. Used 5 frames for now, we can make it more sophisticated later.

JonathanSteinbuch (approved):

Flow86:
so we "render" faster until we're on time again? right?

Flamefire:
> so we kind of render "faster" until we're on time again? right?

We must distinguish between "render" (i.e. draw on screen) and simulate (i.e. advance GFs). This simulates consistently and maybe faster, see my example at https://github.com/Return-To-The-Roots/s25client/issues/1319#issuecomment-748182978: At 80ms a GF should be run but was not due to time resetting instead of advancing. This is fixed now and in that scenario a GF will be run at 80ms. Additionally it can also catch up a bit

Flow86:
yes thats what I meant with "render" ;-)

PR #1321, opened 2020-12-22
Fix wrong produced ware for military buildings

Flamefire:
They should be none not GD_NOTHING which is an invalid good
This also fixes a long-standing bug where the donkey breeder productivity was ignored because it produces a job which is hard to represent in the current format


Flow86 (dismissed):

JonathanSteinbuch:
This seems to work again. :)

Although they don't break anything as the checked case never occurs (buildings that don't produce anything have waresNeeded.size() = 0), nobUsual.cpp lines 127,128, 380,381, 391,392 should probably also be removed or changed into an assert.

If the goal is to remove GD_NOTHING it should probably be renamed to GD_PLACEHOLDER or GD_DONKEY (for the use in iwBuilding.cpp line 67) and all other remaining occurences changed to work with boost::none

Flamefire:
Thanks, good catch.

> If the goal is to remove GD_NOTHING it should probably be renamed to GD_PLACEHOLDER or GD_DONKEY (for the use in iwBuilding.cpp line 67) and all other remaining occurences changed to work with boost::none

That's not so easy. It is used in other places to where none means "not recruitable" and NOTHING "no tool required", sic. Will need some more work, so let's get this in first.

JonathanSteinbuch (approved):

PR #1322, opened 2020-12-22
Refactor credits window and timer control

Flamefire:
This uses a `Timer` in the ctrlTimer and explicit chrono units.

Then refactors the credits screen to make use of timers instead of hand-rolled timing.

It also fixes a bug introduced in the resources PR where the "credits" resource wasn't loaded so the pictures were missing

### DIFF
--- a/libs/s25main/desktops/dskCredits.cpp
+++ b/libs/s25main/desktops/dskCredits.cpp
@@ -294,8 +294,7 @@ dskCredits::dskCredits() : Desktop(LOADER.GetImageN("setup013", 0)), itCurEntry(
     for(unsigned i = 0; i < NUM_NATIVE_NATIONS; i++)
         nations[i] = Nation(i);
 
-    if(!LOADER.Load(ResourceId("credits"))
-       || !LOADER.LoadFilesAtGame(worldDesc.get(DescIdx<LandscapeDesc>(0)).mapGfxPath, false, nations, {}))
+    if(!LOADER.LoadFilesAtGame(worldDesc.get(DescIdx<LandscapeDesc>(0)).mapGfxPath, false, nations, {}))
     {
         WINDOWMANAGER.Show(std::make_unique<iwMsgbox>(_("Error"), _("Failed to load game resources"), this, MSB_OK,
                                                       MSB_EXCLAMATIONRED, 0));
--- a/libs/s25main/desktops/dskCredits.cpp
+++ b/libs/s25main/desktops/dskCredits.cpp
@@ -19,6 +19,7 @@
 #include "Loader.h"
 #include "WindowManager.h"
 #include "controls/ctrlButton.h"
+#include "controls/ctrlTimer.h"
 #include "drivers/VideoDriverWrapper.h"
 #include "dskMainMenu.h"
 #include "ingameWindows/iwMsgbox.h"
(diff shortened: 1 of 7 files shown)


libs/s25main/desktops/dskCredits.cpp
```diff
@@ -34,21 +35,33 @@
...
 {
-    AddTextButton(0, DrawPoint(300, 550), Extent(200, 22), TC_RED1, _("Back"), NormalFont);
-    AddText(1, DrawPoint(400, 10), _("Return To The Roots"), COLOR_YELLOW, FontStyle::CENTER, LargeFont);
-    AddText(2, DrawPoint(400, 33), _("Credits"), COLOR_YELLOW, FontStyle::CENTER, LargeFont);
+    AddTextButton(ID_btBack, DrawPoint(300, 550), Extent(200, 22), TC_RED1, _("Back"), NormalFont);
+    AddText(ID_txtRttr, DrawPoint(400, 10), _("Return To The Roots"), COLOR_YELLOW, FontStyle::CENTER, LargeFont);
+    AddText(ID_txtCredits, DrawPoint(400, 33), _("Credits"), COLOR_YELLOW, FontStyle::CENTER, LargeFont);
+    AddTimer(ID_tmrBobAnim, 40); // Bob animation every 40ms
+    pageTimer = AddTimer(ID_tmrPage, duration_cast<std::chrono::milliseconds>(PAGE_TIME).count());
+    pageTimer->Stop();
```
Flow86, line 62: why not add a parameter to make it stopped by default?

libs/s25main/desktops/dskCredits.cpp
```diff
@@ -34,21 +35,33 @@
...
+
+    LOADER.Load(ResourceId("credits")); // Ignore failure
```
Flow86, line 64: why do we load it twice?

libs/s25main/desktops/dskCredits.cpp
```diff
@@ -454,11 +465,10 @@ void dskCredits::DrawBobs()
                     bob.direction = libsiedler2::ImgDir::E;
             }
         }
+    } else if(tmrId == ID_tmrPage)
+    {
+        GotoNextPage();
```
Flow86, line 470: so we can't go back on right click anymore? why?

Flamefire: We can, nothing changed there

Flow86: ah forget it

libs/s25main/desktops/dskCredits.cpp
Flamefire: Failed attempt I forgot to remove

libs/s25main/desktops/dskCredits.cpp
Flamefire: Because that would need to be passed through window and this is not worse, so I saved the effort

libs/s25main/desktops/dskCredits.cpp
```diff
@@ -308,38 +322,33 @@ void dskCredits::Draw_()
 
 void dskCredits::GotoNextPage()
 {
-    ++this->itCurEntry;
-    if(this->itCurEntry == entries.end())
-        this->itCurEntry = entries.begin();
-    timer.restart();
+    ++itCurEntry;
```
Flow86, line 324: we probably should snake-case that var here for consistency - like `it_cur_entry_`

libs/s25main/desktops/dskCredits.cpp
```diff
@@ -371,25 +380,22 @@ T randEnum()
...
-    int bobAnimStepsPerSec = 25;
-
-    int msSinceLastBobSpawn = VIDEODRIVER.GetTickCount() - bobSpawnTime;
-    int bob_spawnprosec = 5;
+    using namespace std::chrono_literals;
+    auto bobSpawnIntervall = 200ms;
 
     if(VIDEODRIVER.GetFPS() < 30)
-        bob_spawnprosec = 0;
+        bobSpawnIntervall = 0ms;
```
Flow86, line 386: one `l` for "interval"

libs/s25main/desktops/dskCredits.cpp
Flow86: or at least add the `_` to all members

Flamefire: Erm, everything is camelCase

libs/s25main/desktops/dskCredits.cpp
```diff
@@ -34,21 +35,33 @@
...
+const auto PAGE_TIME = std::chrono::milliseconds(12900);
 /// Duration for fading between pages
-const unsigned FADING_TIME = 2000;
+const auto FADING_TIME = std::chrono::seconds(2);
+using std::chrono::duration_cast;
+
+namespace {
+enum
+{
+    ID_btBack,
```
Flow86, line 47: do we have `bt` or `btn` everywhere else?

libs/s25main/desktops/dskCredits.cpp
Flow86: ah right, we have different styles everywhere :-D

libs/s25main/desktops/dskCredits.cpp
Flow86: ah yes we have different styles everywhere :-D

libs/s25main/desktops/dskCredits.cpp
Flamefire: ID_bt

libs/s25main/desktops/dskCredits.cpp
Flow86: should we at least log here that the resource failed to load? or is it automatically logged?

Flamefire: It is. There will be a status with timing on success or fail on error

Flow86 (approved):

Flow86:
we should replace appveyor with github actions too: https://github.com/microsoft/github-actions-for-desktop-apps

Flamefire:
> we should replace appveyor with github actions too: https://github.com/microsoft/github-actions-for-desktop-apps

IIRC appveyor has more configurations on which VS you want to use and such. But yes it is slow. Still a bonus to have different CIs I guess, see the travis incicent

Flow86:
yes it only has vs2019 available (but you could use it to run vs2017 toolchain) and you can even run clang on windows ;-)

see https://www.cazzulino.com/vs-preview-ci.html

PR #1324, opened 2020-12-27
Introduce linear interpolation function

Flamefire:
Also fix some CI failures (signedness and parameter name)

### DIFF
--- a/libs/s25main/desktops/dskLobby.cpp
+++ b/libs/s25main/desktops/dskLobby.cpp
@@ -1,3 +1,4 @@
+// Copyright (c) 2005 - 2020 Settlers Freaks (sf-team at siedler25.org)
 //
 // This file is part of Return To The Roots.
 //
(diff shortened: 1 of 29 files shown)


libs/s25main/desktops/dskLobby.cpp
```diff
@@ -73,7 +74,8 @@ dskLobby::dskLobby()
     // Chatfenster-Edit
     AddEdit(21, DrawPoint(20, 530), Extent(500, 22), TC_GREY, NormalFont);
 
-    AddTimer(30, 5000);
+    using namespace std::chrono_literals;
+    AddTimer(30, 5s);
```
Flow86, line 78: we should change the id to a define here too?

Flamefire: Yes we should do that for all desktops. Once someone got some time, not now. To much work...

PR #1325, opened 2020-12-28
Fix visual bug for some minimized windows

JonathanSteinbuch:
Fixes the following visual bug for some minimized windows.

minimizedBug


Flamefire (dismissed):
Good catch! Just checked: This is due to the window only being made for controls: I.e. buttons and such won't be drawn when minimized. But when overriding the method this logic is skipped

Flamefire:
Uff, can you rebase please? Jenkins seems to be picky -.-

@Flow86 Any chance to rebuild the apple boost with hidden visibility or suppress the linker warnings? I can't tell if there is any error actually hidden but doesn't look like

JonathanSteinbuch:
No problem. Did a rebase

PR #1326, opened 2020-12-28
Check and adjust position of ingame window for every call of SetPos and Resize

JonathanSteinbuch:
The following situation: When you 

1. Put an ingame window A at the very bottom and/or very right of the screen
2. Then close it (while it is at this position)
3. Then make the whole game screen smaller
4. Then reopen A.

The ingame window A will not be (fully) visible anymore.

This pull request fixes this. I did not see any reason why the position should not for every call of SetPos and Resize be adjusted so the ingame window is fully visible.


Flamefire (approved):
Looks good, thanks!

PR #1329, opened 2021-01-04
Fix message box window button id bug

JonathanSteinbuch:
Fixes a newly introduced bug where the results from a message box button click where wrongly registered.

ID_BT_0 was added twice as it is also added in AddButton itself.


Flamefire:
Ah gosh, my bad! This is why this was working at all. I got confused by that.
I would rather remove the `ID_BT_0 +` from `AddButton` because that is to close to `AddTextButton` and is what led me to thinking this was a bug before.

Flamefire (approved):
Thanks!
I'll add a test case soon to avoid this in the future

JonathanSteinbuch:
No problem! 

If I may say: Control ids seem to be confusion prone anyway. To change everything to an object oriented approach for handling mouse messages and addressing control elements would be a lot of work, though.

Issue #1290, opened 2020-08-16
Require Boost 1.69


Flamefire:
We currently require Boost 1.64+ but in #1289 I run into an issue fixed by Boost.Test 1.69: https://github.com/boostorg/test/commit/cc7796fbe953293ddb153f31bacac0a67622602d

Basically: For `BOOST_TEST(value)` the print_helper specialization is ignored making it fail for unprintable types.

Maybe it is time to upgrade Boost to 1.69.

In the same commit it makes sense to upgrade CMake to 3.14:
- We have a copy of FetchContent from 3.14 to make it work on earlier versions
- We have a copy of FindBoost from 3.11 to be able to find versions up to 1.67, hence we need a newer version for 1.69

Flamefire:
@Flow86 Hitting again an issue with Boost 1.64 which is fixed in 1.65 by https://github.com/boostorg/test/commit/aece067770203cdff62baaab494599bb1f2d4a3d

Can we upgrade? If so only a minor upgrade to 1.65 or 1.69 and CMake to 3.14?

Flow86:
didnt we update already?

Flow86:
ah, we did not, since debian:busters default boost is 1.69 and cmake 3.13

Flamefire:
So Boost 1.69 as the minimum and CMake 3.13 (or just keep the current one)?

I'd also like to optionally FetchContent https://github.com/boostorg/variant2 when on Boost < 1.73 as I find myself using variant more often and Variant2 is C++11 instead of (slow) MPL and on 1.73 it is even trivial for most usages meaning it is fast

Flamefire:
@Flow86 Could you update the linux runners on the build server to 1.69 please?

Flow86:
the problem is that we need to package the libs then, or do we use them statically anyhow?

Flamefire:
We have never packaged boost and always build statically: `-DRTTR_USE_STATIC_BOOST=ON`

Flow86:
I've changed the linux runner now to use the official backport package (boost 1.71)

PR #1331, opened 2021-01-04
Introduce new "smart" range-based team building options

Flamefire:
Continuation of #1310 original idea and implementation from @jdoerfert

The old team building options where either fixed (1,2,3,4) or purely random (?).
The new options (1-2,1-3,1-4) allow players to be placed in ranges of teams while they also ensure team sizes to be as even as
possible. That is, a player with the team 1-3 is placed in team 1, 2, or 3 in a way that minimizes the absolute difference in team size.

Something to be discussed which I'd put in here:
- Currently it is possible, that teams are not made as even as possible. Example: 4 players in Teams "1", "2", "1-3" and "1-4". In the current implementation this could lead to the last player put into team 3 (with a total 25% chance) and hence team 4 to be empty. The "ideal" distribution is to put the 3rd player into team 3 and the 4th into team 4, but that would not be random at all. Should this be done?
- Shall we keep our current random thingy as-is? This means
   - the "?" is evaluated before the "1-x"
   - we have constants `TM_RANDOMTEAM`-`TM_RANDOMTEAM4` which are all shown as "?". From a C++ perspective those are not required and we can use a single enum value and use the new code to assign a team on start. A single use case comes to mind: We expose those Lua, which means a Lua script could set a team without showing it to the user. This could be useful in rare circumstances. @Spikeone ? I'm inclined to remove those other enumerators though.

Edit: After discussion the following changes were made:
- TM_RANDOMTEAM constants were removed and only a single Team::Random was retained. Assignment of that happens in the same function as the smart team assignment prior to the smart teams. So a started game has either Team::None or Team1-4 as the teams for all players
- Smart teams now distribute players as even as possible. For this the most constrained (1-2) or assigned before the least constrained. I can't think of any case where this causes less randomness then the current approach if that was retried until the teams were as even as possible


Flamefire:
No opinions? @Spikeone @Flow86 maybe @vader1986  ?

Flow86:
I think the 2nd point I can agree on

vader1986:
> * Currently it is possible, that teams are not made as even as possible. Example: 4 players in Teams "1", "2", "1-3" and "1-4". In the current implementation this could lead to the last player put into team 3 (with a total 25% chance) and hence team 4 to be empty. The "ideal" distribution is to put the 3rd player into team 3 and the 4th into team 4, but that would not be random at all. Should this be done?

I would rather go for an option which tries to make teams as equal (in size & ideally in rating) as possible. Let's say all 4 players chooses `1-2`, I would expect both teams to have 2 players. I don't think anyone would expect that all 4 players can (theoretically) end up in the same team. 

> * Shall we keep our current random thingy as-is? This means
>   
>   * the "?" is evaluated before the "1-x"

Not sure if it matters to the user which one is evaluated before?

jdoerfert:
> > * Currently it is possible, that teams are not made as even as possible. Example: 4 players in Teams "1", "2", "1-3" and "1-4". In the current implementation this could lead to the last player put into team 3 (with a total 25% chance) and hence team 4 to be empty. The "ideal" distribution is to put the 3rd player into team 3 and the 4th into team 4, but that would not be random at all. Should this be done?
> 
> I would rather go for an option which tries to make teams as equal (in size & ideally in rating) as possible. Let's say all 4 players choose `1-2`, I would expect both teams to have 2 players. I don't think anyone would expect that all 4 players can (theoretically) end up in the same team.

That won't happen. 4 x `1-2` will result in 2 teams with 2 players each. IIRC, as long as you do not mix two `1-x` values, so `1-3` and `1-4`, you will always get "perfectly balanced" teams. Mixing them gives you the problem that you add "choice constraints" to the mix which limit your ability to place a random player in the team with the least number of players. The example by @Flamefire shows that. However, I'd argue that it is a misuse of the range based team assignment.

Flamefire:
> Not sure if it matters to the user which one is evaluated before?

This means that this random teams are assigned (fully randomly) and then the range-based teams are evaluated "smoothing" out the teams, hence you get (as far as possible) equal teams. But I think in the end the order doesn't matter to much. Maybe something weird like 2 players "1-2" and 2 using "?" where if the "?" would be evaluated first means that teams 1 and 2 have one player each and then the other 2 get randomly assigned to any of the 4 teams, so 1 player in each team is as likely as e.g. 3 players in team 1 and 1 in team 2

> However, I'd argue that it is a misuse of the range based team assignment.

Maybe. I'm just not sure if there isn't a valid use case for that (for the case with more players), e.g. 8 players, 4 with 1-2, 4 with 1-4 could be used to place the first 4 players randomly in the first 2 teams and the last 4 in teams 3 and 4 yielding a completely balanced distribution. Of course we could add support for options like "3-4", but I'm quite certain that this will lead to more problems of this kind.

Spikeone:
Just thinking about hiding teams, I'd rather add an option to hide them, as the game shows teams as well (or at least should). It's also possible to setup teams ingame so, I don't think one needs that.

I like the not random option, as even teams means even teams, but not sure what one would expect when using that option.

jdoerfert:
The way I read b66efa3 it eliminates some random choices that existed before (see below) but given the expect usage it doesn't matter anyway. As long as something provides reasonable random teams I'd be happy...

Team selection
```
  1    
  3    
  1-2  // Effectively team 2, no random choice.    
  1-3    
  1-3    
  1-3    
  1-3    
  1-3    
```

Flamefire:
You are right, I couldn't find that. Thanks a lot! So optimal is: Assuming the 1-2 was chosen last, then first 1-3 will be team 2, next 3 will be team 1-3 each, so all teams have 2 players, then the last 1-3 could be team 2 so the 1-2 will be team 1.

So we must distinguish this from the following case:
```
1
3
1-2
1-3
1-3
1-3
```
Again the first 1-3 will go into team 2. Now we must avoid the case that both 1-3 go into team 1 and 2 (so none in team 3) before the 1-2 is chosen which will make it imbalanced.

Then other approach:
- Get the teams with min amount of players and possible player choices
- From those select the ones with the least amount of possible players
- Select one team and one possible player randomly from that

This is basically the same approach you had, but step 2 fills the most constrained teams first.
I.e. after the first 2 assignments in the above example you will have:
```
1
3
1-2
1-3 -> 2
1-3 -> 1
1-3
Team sizes are 2, 1, 1
```
Before team 2 and 3 were both possible targets for the next player, but team 3 has only 1 possible player to choose for so the last 1-3 goes into 3 and 1-2 goes into 2. If the 2nd 1-3 went into team 2, then 1-2 would go into team 1. Hence you still have the randomness and all teams are even.

Can you think of any case where this algorithm yields a non-randomness or uneven teams although they'd be possible? I actually liked yours more as it is way more generic.

jdoerfert:
^ what you propose sounds good.

Flamefire:
Implemented, tested and found a bug there: 6 players with 1-3, 2 players with 1-4. Even teams are only possible, if the 2 1-4 are put in team 4. This is however not ensured by the above algorithm as it will put the first 1-4 into team 4 and then team 3 is the one with the least amount of players and hence chosen next. This may choose the second 1-4 and we have a failure. :(
The problem is harder than  thought...

jdoerfert:
> The problem is harder than thought...
1) Practically, any of the approaches was a *huge* improvement over the status quo and sufficient. The search for a perfect solution does cost more than it is worth (accumulated over the rest of the usage of the code).
2) Given up to 8 players and up to 4 teams, you could just as well brute force it. Always go with random, once balanced teams are found stick with them. If it never happens, pick one of the best balanced options by random.
3) You probably need something that is not greedy. The problem is that you look at the current state, not where you want to end up. If a choice prevents balanced teams, avoid it as long as possible. That also means you cannot only look at one team at a time.
4) Remember 1) and pick any implementation you had before.

Flamefire:
> The search for a perfect solution does cost more than it is worth (accumulated over the rest of the usage of the code).

I know... I see it as a challenge, some kind of puzzle to solve. And I want to learn something new through it. E.g. I found the term "assignment problem" for this. But seems to not fully apply.

My nest idea sounds simple but might just work: Gather the next team to fill by choosing the one where the sum of the current players and potential players is minimal. I have the feeling, this can be explained with the assignment problem but I can't quite find it. Probably because the assignment problem is a 1:1 mapping, while we have a n:1

I considered brute-force and then randomly choosing a solution, but there would be ~65k possibilities. To many...
I also thought about combining both algorithms: The optimal one (as current) which leads to perfectly sized teams including randomness when this is not possible and then using the old one which assigns players randomly using the target team size when choosing the next team.

I'll try the simple idea first and if that doesn't work I'd stick with the current one which guarantees optimal solutions at least.

jdoerfert:
> I know... I see it as a challenge, some kind of puzzle to solve. And I want to learn something new through it. 
While this is probably a big part of your motivation to work on such a project, it is arguably not the best for the project. That said, nobody argued you shouldn't spend as much time as you feel necessary. What I suggested is to go with a version that improves the existing solution and then tinker on your perfect one. I'm personally waiting for this to actually make it into the game so we can stop using random.org. Any version would have worked fine for >95% of realistic use cases. However, now I'm waiting, probably not alone, for you to solve your puzzle and learn something new. Is that really the best way of doing things?

Flamefire:
> Is that really the best way of doing things?

Obviously not and I would have preferred this to be merged already.
But I can't merge my own PRs but need someone else from the team to review and merge them.
This PR was ready after `Remove TM_RANDOMTEAM<N>` and has been waiting for 5 days now. I simply used the time while this is still waiting to improve upon the current state.

FWIW: I now have implemented an improved solution which is basically the same which I had prior but with an extra shuffle step at the end. So in the example from https://github.com/Return-To-The-Roots/s25client/pull/1331#issuecomment-770288592 the 1-2 player will still be assigned to team 2, but it may be swapped to team 1 later.
The probabilities might not be as even as they should be (i.e. there should be a 50% chance, that this player ends up in either team), but I don't feel like really calculating them and just rely on the random shuffle before that and the randomness in it and the team assignment in the first step.

Hence: Done and I hope @Flow86 will merge this soon.

jdoerfert:
> Hence: Done and I hope @Flow86 will merge this soon.

Thanks a lot. I know I sound critical but I am grateful for your work here.

Flamefire:
Damn, failure is due to another bug in Boost < 1.69, so this is now blocked by https://github.com/Return-To-The-Roots/s25client/issues/1290 and https://github.com/Return-To-The-Roots/s25client/pull/1351 (which upgrades the CI and requirements)

Flow86:
@Flamefire  needs fixing (conflict)

Flow86 (changes requested):
@Flamefire needs fixing (conflict)

Issue #1337, opened 2021-01-07
Add-on state change after save-load


RM87:
Add-on state is incorrectly loaded after resuming a saved game.

Demolish building after resource exhausted (was off during game setup and also in main menu options, is on after loading game).

Flamefire:
I can't see any obvious issue. Do you have a savegame/replay to reproduce?

RM87:
[2021-01-07_22-19-10.rpl.zip](https://github.com/Return-To-The-Roots/s25client/files/5788346/2021-01-07_22-19-10.rpl.zip)
Demolitions happen at GF 28076  and GF 28151 .

Flamefire:
That replay has the demolish setting enabled. Can't tell if that should be, but that's what the game file says.

What I also tried:
- create a new game
- in game lobby setup addon settings so that demolish is disabled
- start game
- save
- load that game
- demolish still disabled

Maybe you did accidentally enable it during game setup?

RM87:
A bit doubtful but as I am unable to reproduce this at the moment also then closing

PR #1339, opened 2021-01-08
Fix loading of Fish and crash on failed trade

Flamefire:
First was a `>=` vs `>` and the 2nd a missing return
Add tests for both cases and used more of push/popEnum for verification of gamedata.

Also added test case for #1337 but couldn't find anything

### DIFF
--- a/tests/s25Main/integration/testTrading.cpp
+++ b/tests/s25Main/integration/testTrading.cpp
@@ -15,6 +15,7 @@
 // You should have received a copy of the GNU General Public License
 // along with Return To The Roots. If not, see <http://www.gnu.org/licenses/>.
 
+#include "RttrForeachPt.h"
 #include "addons/const_addons.h"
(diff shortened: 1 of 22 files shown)


tests/s25Main/integration/testTrading.cpp
```diff
@@ -262,6 +263,35 @@ BOOST_FIXTURE_TEST_CASE(TradeFail, TradeFixture)
...
+    numDonkeys -= 2;
+    numHelpers -= 2;
+    numWoodcutters -= 2;
+    testExpectedWares();
+    testAfterLeaving(4);
+
+    // Let them walk a bit
+    RTTR_SKIP_GFS(40);
+
+    // No make everything enemy territory (e.g. enemy captured the building)
```
Flow86, line 283:
```suggestion
    // Now make everything enemy territory (e.g. enemy captured the building)
```

Flow86 (dismissed):

PR #1343, opened 2021-01-10
Performance boost for random map generation

vader1986:
By replacing std::vector to std::set to boost helper::contains (log(n) vs. log(O(n)) random maps of size 1024 x 1024 are now generated in less than 25% of the time (40s vs. 8s). Thanks @JonathanSteinbuch !

### DIFF
--- a/libs/s25main/mapGenerator/Algorithms.h
+++ b/libs/s25main/mapGenerator/Algorithms.h
@@ -148,6 +148,33 @@ namespace rttr { namespace mapGenerator {
     void UpdateDistances(NodeMapBase<unsigned>& distances, std::queue<MapPoint>& queue);
 
     /**
+     * Computes a map of distance values describing the distance of each grid position to the closest flagged point.
+     *
+     * @param flaggedPoints contains all flagged points
+     * @param size size of  the map
+     *
+     * @return distance of each grid position to closest flagged point.
+     */
+    template<class T_Container>
+    NodeMapBase<unsigned> DistancesTo(const T_Container& flaggedPoints, const MapExtent& size)
+    {
+        const unsigned maximumDistance = size.x * size.y;
+        std::queue<MapPoint> queue;
+        NodeMapBase<unsigned> distances;
+        distances.Resize(size, maximumDistance);
+
+        for(const MapPoint& pt : flaggedPoints)
+        {
+            distances[pt] = 0;
+            queue.push(pt);
+        }
+
+        UpdateDistances(distances, queue);
+
+        return distances;
+    }
+
+    /**
      * Computes a map of distance values describing the distance of each grid position to the closest position for
      * which the evaluator returned `true`. The computation takes place only within the specified area - points outside
      * the area are set to a maximum value of width + height of the map.
--- a/libs/s25main/mapGenerator/HeadQuarters.h
+++ b/libs/s25main/mapGenerator/HeadQuarters.h
@@ -47,14 +47,9 @@ namespace rttr { namespace mapGenerator {
     template<class T_Container>
     std::vector<MapPoint> FindHqPositions(const Map& map, const T_Container& area)
     {
-        std::vector<MapPoint> headQuarters;
-        for(const MapPoint& hq : map.headQuarters)
-        {
-            if(hq.isValid())
-            {
-                headQuarters.push_back(hq);
-            }
-        }
+        auto isHeadQuarter = [&map](const MapPoint& pt) {
+            return map.objectInfos[pt] == libsiedler2::OI_HeadquarterMask;
+        };
         auto isObstacle = [&map](MapPoint point) {
             return map.textureMap.Any(point, [](auto t) { return !t.Is(ETerrain::Buildable); });
         };
@@ -62,7 +57,7 @@ namespace rttr { namespace mapGenerator {
         // Quality of a map point as one player's HQ is a mix of:
         // 1. distance to other players' HQs (higher = better)
         // 2. distance to any other obstacle e.g. mountain & water (higher = better)
-        NodeMapBase<unsigned> potentialHeadQuarterQuality = DistancesTo(headQuarters, map.size);
+        NodeMapBase<unsigned> potentialHeadQuarterQuality = Distances(map.size, isHeadQuarter);
         const auto& obstacleDistance = Distances(map.size, isObstacle);
         const auto minDistance = helpers::clamp(GetMaximum(obstacleDistance, area), 2u, 4u);
 
--- a/libs/s25main/mapGenerator/HeadQuarters.h
+++ b/libs/s25main/mapGenerator/HeadQuarters.h
@@ -48,7 +48,7 @@ namespace rttr { namespace mapGenerator {
     std::vector<MapPoint> FindHqPositions(const Map& map, const T_Container& area)
     {
         std::vector<MapPoint> headQuarters;
-        for(const MapPoint& hq : map.headQuarters)
+        for(const MapPoint& hq : map.hqPositions)
         {
             if(hq.isValid())
             {
--- a/libs/s25main/mapGenerator/HeadQuarters.h
+++ b/libs/s25main/mapGenerator/HeadQuarters.h
@@ -47,9 +47,14 @@ namespace rttr { namespace mapGenerator {
     template<class T_Container>
     std::vector<MapPoint> FindHqPositions(const Map& map, const T_Container& area)
     {
-        auto isHeadQuarter = [&map](const MapPoint& pt) {
-            return map.objectInfos[pt] == libsiedler2::OI_HeadquarterMask;
-        };
+        std::vector<MapPoint> headQuarters;
+        for(const MapPoint& hq : map.headQuarters)
+        {
+            if(hq.isValid())
+            {
+                headQuarters.push_back(hq);
+            }
+        }
         auto isObstacle = [&map](MapPoint point) {
             return map.textureMap.Any(point, [](auto t) { return !t.Is(ETerrain::Buildable); });
         };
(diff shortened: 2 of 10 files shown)


vader1986:
Not entirely sure why the style & formatting check fails on a file I didn't change:
external/libutil/libs/network/src/Socket.cpp
Also didn't change any submodule reference, I think.

Flamefire:
>  external/libutil/libs/network/src/Socket.cpp

Ignore that. Will fix in another PR.

I just checked the code and I have a better approach for you. Basically what you do is:
- Create a vector (now a set, so a "collection") of points which you usually do by iterating over all points and taking a few based on a condition
- Pass this collection to `Restructure`
- This passes this to `Distances` as a `contains` lambda
- `Distances` iterates over all points and checks the collection if it contains the current point, if so it adds it to the queue
- `Restructure` then iterates over  all points again using the calculated distances

So basically there are 2 approaches to make this faster:
- Either `Restructure` receives the condition directly and just passes it to `Distances` which saves you the first iteration over all points and the repeated `contains`
- Pass the collection directly to `Distances` instead of a lambda checking for contains. You even already have an overload for that: `NodeMapBase<unsigned> Distances(const MapExtent& size, const T_Container& area, const unsigned defaultValue, T&& evaluator)`. You can (ab)use that by passing a defaultValue of `size.x * size.y` and an evaluator which just returns true (and hopefully gets removed by the optimizer). Alternatively add an overload which directly takes a container using some SFINAE (if the passed type is callable with a mappoint it is a evaluator, else a collection)

In any case I'd suggest to keep the vector as a set has worse memory behavior and will then be slower (after the above is improved).
And the same can be applied to `AddObjects`: Here you again pass a lambda with a simple contains to `Distances` (after the improvement you already made here) while you could pass the collection directly.

vader1986:
> And the same can be applied to AddObjects: Here you again pass a lambda with a simple contains to Distances (after the improvement you already made here) while you could pass the collection directly.

I'd rather leave it as a set to avoid adding a point multiple times when using `GetPointsInRadiusWithCenter`. But I changed the "focus" area part you suggested, makes the code simpler.

libs/s25main/mapGenerator/Resources.cpp
```diff
@@ -71,10 +72,22 @@ namespace rttr { namespace mapGenerator {
...
+                {
+                    forbiddenArea.insert(point);
+                }
+            } else if(map.textureMap.Any(pt, IsSnowOrLava))
+            {
+                forbiddenArea.insert(pt);
+            }
+        }
+
+        auto isForbiddenArea = [&forbiddenArea](const MapPoint& pt) { return helpers::contains(forbiddenArea, pt); };
```
Flamefire, line 90: As mentioned earlier: Why not pass this collection (can stay a set) into `Distances` directly? This would greatly improve performance as it saves a whole iteration and lot's of contains checks (although `set` makes them quite cheap)

Flamefire (changes requested):
> I'd rather leave it as a set to avoid adding a point multiple times when using GetPointsInRadiusWithCenter

Good point. Added a suggestion there: You can pass the collection (set, vector, ... doesn't matter. Just needs to be iterable) directly to `Distances`, see the 2nd point of https://github.com/Return-To-The-Roots/s25client/pull/1343#issuecomment-757519259 (the new approach for `Restructure` doesn't need to be touched, that's completely fine now)

Out of interest: Do you track the performance changes between the commits? Would be interesting to have some numbers on how much the changes affect it. I.e. I expect that last change to make it quite a bit faster than the 8s and the next to improve it a lot too.

BTW: I hope you don't mind me suggesting further improvements. You already identified the hotspots so that's why I checked them more thoroughly and I think making this faster is very valuable because (at least I) tend to create multiple random maps until I find one that I like most ;)

libs/s25main/mapGenerator/RandomMap.cpp
```diff
@@ -293,17 +272,7 @@ namespace rttr { namespace mapGenerator {
...
-        RTTR_FOREACH_PT(MapPoint, map_.size)
-        {
-            if(rnd_.ByChance(5))
-            {
-                focus.push_back(pt);
-            }
-        }
-
-        Restructure(map_, focus);
+        Restructure(map_, [this](const MapPoint& pt) { return rnd_.ByChance((pt.x + pt.y) % 6); });
```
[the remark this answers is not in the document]
vader1986: Good call, I didn't know how to ignore a parameter! 

There's a summary of the function in the header file. The name is not great, but it elevates landscape around "focus" points (defined through the "predicate" argument). The close a MapPoint is to a "focus" point, the higher the "z" value. I'm using this to define the general shape of the map.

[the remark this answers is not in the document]
vader1986: For performance measurement I'm using the Time Profiler from Xcode. It gives the entirely breakdown of function calls & time spent in a function, which is really useful. I can add some screenshots before & after.

vader1986:
> BTW: I hope you don't mind me suggesting further improvements. You already identified the hotspots so that's why I checked them more thoroughly and I think making this faster is very valuable because (at least I) tend to create multiple random maps until I find one that I like most ;)

No worries, absolutely appreciated! Just having lunch break & going through the comments.

libs/s25main/mapGenerator/RandomMap.cpp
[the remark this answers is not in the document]
Flamefire: Thanks!

No need for screenshots, numbers are fine.
The map is fully reproducible given a seed, isn't it? That would mean you could likely test the map generation time by temporarily fixing the seed and run this before and after each improvement and compare the times (maybe run a few times to average it). I guess your profiler could also give you a sum of times for calling the generation a few times (e.g. start the game and generate 10 maps, then ask the profiler for a sum and compare that)
Ok as you notice I'm already going overboard here ;D I have a kind of passion for performance optimizations...
Long story short: I'd be happy to have a couple numbers similiar to your OP here for the various stages of this PR if you don't mind.

BTW: I like your commits so far! :+1:

libs/s25main/mapGenerator/Algorithms.h
```diff
@@ -147,6 +147,33 @@ namespace rttr { namespace mapGenerator {
...
+        const unsigned maximumDistance = size.x * size.y;
+        std::queue<MapPoint> queue;
+        NodeMapBase<unsigned> distances;
+        distances.Resize(size, maximumDistance);
+
+        for(const MapPoint& pt : flaggedPoints)
+        {
+            distances[pt] = 0;
+            queue.push(pt);
+        }
```
Flamefire, line 170: If you want you can check if this is even faster. I'd expect a bit of a speed penalty by pushing the elements one by one to the queue which means reallocation.

```suggestion
        NodeMapBase<unsigned> distances;
        distances.Resize(size, maximumDistance);
        std::queue<MapPoint> queue(flaggedPoints.begin(), flaggedPoints.end());

        for(const MapPoint& pt : flaggedPoints)
            distances[pt] = 0;
```

Flamefire (approved):
Fine by me, one optional improvement possibility if you want

General remarks I've seen (no need to change)

- Maybe the Distance* functions can be partially unified to not have same/similar code multiple times if it doesn't hurt performance. See the idea with the `return true`-evaluator. For now its ok, but if we need a 4th version I'd revisit this
- While it is correct to do `std::vector<unsigned> x = {1u, 2u...}` I'd omit the `u` for readability (if the compilers accept it). Again no need to change, just for future additions.

vader1986:
`1024 x 1024 (2 players, mixed)

Commit: 1st commit of the PR
Runtime: 8.1s
47% SmoothHeightMap
22% PlaceHeadQuarters
 9% PlaceHarbors
 9% AddTextures
 7% AddObjects

Commit: “Define “focus” area through lambda instead of vector”
Runtime: 7.8s
47% SmoothHeightMap
23% PlaceHeadQuarters
 9% PlaceHarbors
 9% AddTextures
 7% AddObjects

Commit: “Pass container with flagged points to distance calc function”
Runtime: 7.7s
46% SmoothHeightMap
23% PlaceHeadQuarters
 9% PlaceHarbors
 9% AddTextures
 7% AddObjects

Commit: “Performance improvement for HQ position search”
(Latested)

Runtime: 7.6s
46% SmoothHeightMap
22% PlaceHeadQuarters
 9% PlaceHarbors
 9% AddTextures
 7% AddObjects
`

I didn't print the breakdown in % with whole accuracy. Also, this might result in different numbers, depending on the number of players, for example. I'll have a look at the other comment later!

libs/s25main/mapGenerator/Map.cpp
```diff
@@ -25,7 +25,7 @@ namespace rttr { namespace mapGenerator {
 
     Map::Map(const MapExtent& size, uint8_t players, const WorldDescription& worldDesc,
              DescIdx<LandscapeDesc> landscape, uint8_t maxHeight)
-        : hqPositions_(MAX_PLAYERS, MapPoint::Invalid()), textureMap(TextureMap(worldDesc, landscape, textures)),
+        : headQuarters(MAX_PLAYERS, MapPoint::Invalid()), textureMap(TextureMap(worldDesc, landscape, textures)),
```
Flamefire, line 28: Why this change? Seems unrelated and IMO the naming is worse: hqPositions was clear even without seeing the type. Oh and it would be `headquarters` IIRC.

Flamefire:
Thanks for the numbers. Intresting that the 2nd last commit didn't do much. I guess the suggested change then won't either.

Anyway it's 7.6s instead of 40s, so good enough. If you don't mind I'd like the renaming reverted as it IMO makes it less understandable and then this is good to go.

libs/s25main/mapGenerator/HeadQuarters.h
```diff
@@ -47,22 +47,17 @@ namespace rttr { namespace mapGenerator {
...
+        };
         auto isObstacle = [&map](MapPoint point) {
             return map.textureMap.Any(point, [](auto t) { return !t.Is(ETerrain::Buildable); });
         };
 
         // Quality of a map point as one player's HQ is a mix of:
         // 1. distance to other players' HQs (higher = better)
         // 2. distance to any other obstacle e.g. mountain & water (higher = better)
-        NodeMapBase<unsigned> potentialHeadQuarterQuality = DistancesTo(headQuarters, map.size);
+        NodeMapBase<unsigned> potentialHeadQuarterQuality = Distances(map.size, isHeadQuarter);
```
Flamefire: Was this revert on purpose? I was only referring to the renaming hqPositions_->headQuarters

Flamefire (approved):
Thanks!

libs/s25main/mapGenerator/Algorithms.h
Flow86: what about this?

Flamefire: I don't expect any significant improvements by this after  “Pass container with flagged points to distance calc function” brought only 0.1s and the current version mirrors the other impl of this function (and might be replaced in the future by that) so I see no benefit from this change

PR #1344, opened 2021-01-11
Refactored code for storing HQ positions

vader1986:
Just removed some useless code to slightly improve readability.

### DIFF
--- a/libs/s25main/mapGenerator/Map.cpp
+++ b/libs/s25main/mapGenerator/Map.cpp
@@ -85,7 +85,7 @@ namespace rttr { namespace mapGenerator {
 
         for(unsigned i = 0; i < 7; i++)
         {
-            if (hqPositions[i].isValid())
+            if(hqPositions[i].isValid())
             {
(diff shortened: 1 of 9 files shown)


libs/s25main/mapGenerator/Map.cpp
```diff
@@ -100,6 +83,15 @@ namespace rttr { namespace mapGenerator {
             resource[i] = this->resources[i];
         }
 
+        for(unsigned i = 0; i < 7; i++)
```
Flamefire, line 86: Note for the future: Please no magic numbers especially in loops iterating over collections. Use `size`, range-based-for or a named constant.

vader1986: Yeah, good point. I think I left the `7` there because the old S2 didn't support more than 7 players (just an explanation, no excuse).

Flamefire: Yep, I fixed that by introducing a named constant in libsiedler2, shame that there wasn't one already...

Issue #1346, opened 2021-01-12
Mapgenerator may generate invalid maps (missing players)


Flamefire:
The mapgenerator uses `PlaceHeadQuarter` (sidenote: typo, should be `PlaceHeadquarter`) which may fail. However this is not caught and a map is generated anyway.
This might lead to a real problem as I'm not sure how S2 would handle a HQ which is effectively set at `(-1,-1)`

I caught this problem in the improved test cases in https://github.com/Return-To-The-Roots/s25client/pull/1335 where a randomly sized water map which turns out small and with a lot of players fails to create enough islands (?) and therefore doesn't place all HQs.

@vader1986 Any idea how to handle this? It might be worth checking for valid points before writing those to the map header. (at least assert/check and throw). I'd say throwing an exception is acceptable if placing a HQ fails (you can decide if `PlaceHeadQuarter` should do that instead of returning false or maybe better the calling code as it could decide to use another location instead based on the return value). This can be caught and shown in the GUI (maybe simply: "Failed, try again or change settings"-style)

vader1986:
Yeah, this is a bit of an annoying edge case. There're already thing to avoid this case (re-try placing HQs in different positions), however, it can still happen. Maybe throwing an exception in this place (and 2 more places, same file):
https://github.com/Return-To-The-Roots/s25client/blob/master/libs/s25main/mapGenerator/RandomMap.cpp#L288

Something like this:
`if(!PlaceHeadQuarters(map_, rnd_, map_.players))`
`{`
`    throw exception;`
`}`

We should probably catch it somewhere and display an error message like:
"Failed to generate map, please retry!"

vader1986:
Sorry, I just read the last part of your message. Yes, makes sense what you suggest. Throwing the exception in `PlaceHeadQuarter` and catching it some more upper level to display a message in the GUI.

Flamefire:
> re-try placing HQs in different positions

This is what I mean: This could happen outside `PlaceHeadQuarters` then it should return a bool or inside it, then it should throw and not return anything. But in both cases this should be handled. I guess this will very likely occur on a 64x64 island map with 8 players (of course this is on the user, but currently it silently fails to generate a "correct" map)

vader1986:
Working on a PR for this.

vader1986:
I think this one can be closed?

Flamefire:
Yes. You can include a "Fixes #<nr>" in the PR or commit to auto-close such things :)

PR #1347, opened 2021-01-13
#1346 improve error handling on failed HQ placement

vader1986:
Probably better to create a custom exception type here, instead of a `std::runtime_error`?

### DIFF
--- a/libs/s25main/mapGenerator/HeadQuarters.cpp
+++ b/libs/s25main/mapGenerator/HeadQuarters.cpp
@@ -57,7 +57,7 @@ namespace rttr { namespace mapGenerator {
         return connectedArea;
     }
 
-    void PlaceHeadQuarters(Map& map, RandomUtility& rnd, int number, int retries)
+    void PlaceHeadquarters(Map& map, RandomUtility& rnd, int number, int retries)
     {
         auto maxRetries = retries;
         auto success = false;
@@ -92,7 +92,7 @@ namespace rttr { namespace mapGenerator {
 
         if(!success)
         {
-            throw std::runtime_error("could not find any valid HQ position");
+            throw std::runtime_error("Could not find any valid HQ position!");
         }
     }
 
--- a/libs/s25main/mapGenerator/HeadQuarters.cpp
+++ b/libs/s25main/mapGenerator/HeadQuarters.cpp
@@ -19,7 +19,7 @@
 #include "helpers/containerUtils.h"
 #include "mapGenerator/Algorithms.h"
 #include "mapGenerator/TextureHelper.h"
-#
+
 namespace rttr { namespace mapGenerator {
 
     std::vector<MapPoint> FindLargestConnectedArea(const Map& map)
@@ -57,7 +57,7 @@ namespace rttr { namespace mapGenerator {
         return connectedArea;
     }
 
-    bool PlaceHeadQuarters(Map& map, RandomUtility& rnd, int number, int retries)
+    void PlaceHeadQuarters(Map& map, RandomUtility& rnd, int number, int retries)
     {
         auto maxRetries = retries;
         auto success = false;
@@ -90,7 +90,10 @@ namespace rttr { namespace mapGenerator {
             retries--;
         }
 
-        return success;
+        if(!success)
+        {
+            throw std::runtime_error("could not find any valid HQ position");
+        }
     }
 
 }} // namespace rttr::mapGenerator
--- a/libs/s25main/mapGenerator/HeadQuarters.h
+++ b/libs/s25main/mapGenerator/HeadQuarters.h
@@ -103,7 +103,7 @@ namespace rttr { namespace mapGenerator {
     template<class T_Container>
     void PlaceHeadquarter(Map& map, int index, const T_Container& area)
     {
-        const auto& positions = FindHqPositions(map, area);
+        auto positions = FindHqPositions(map, area);
         if(positions.empty())
         {
             throw std::runtime_error("Could not find any valid HQ position!");
--- a/libs/s25main/mapGenerator/HeadQuarters.h
+++ b/libs/s25main/mapGenerator/HeadQuarters.h
@@ -20,6 +20,7 @@
 #include "helpers/mathFuncs.h"
 #include "mapGenerator/Map.h"
 #include "mapGenerator/RandomUtility.h"
+#include <stdexcept>
 
 namespace rttr { namespace mapGenerator {
 
@@ -91,19 +92,17 @@ namespace rttr { namespace mapGenerator {
      * @param index player index for the HQ
      * @param area area to place the HQ in
      *
-     * @returns true if a HQ was placed fro the specified player, false otherwise.
+     * @throw NoHqPositionFoundException
      */
     template<class T_Container>
-    bool PlaceHeadQuarter(Map& map, int index, const T_Container& area)
+    void PlaceHeadQuarter(Map& map, int index, const T_Container& area)
     {
         auto positions = FindHqPositions(map, area);
         if(positions.empty())
         {
-            return false;
+            throw std::runtime_error("could not find any valid HQ position");
         }
-
         map.hqPositions[index] = positions.front();
-        return true;
     }
 
     /**
@@ -114,8 +113,8 @@ namespace rttr { namespace mapGenerator {
      * @param number number of HQs to place - equal to the number of players
      * @param retries number of retries to place valid HQs on this map
      *
-     * @return false if no valid HQ position was found for at least one player, true otherwise
+     * @throw NoHqPositionFoundException
      */
-    bool PlaceHeadQuarters(Map& map, RandomUtility& rnd, int number, int retries = 10);
+    void PlaceHeadQuarters(Map& map, RandomUtility& rnd, int number, int retries = 10);
 
 }} // namespace rttr::mapGenerator
--- a/libs/s25main/mapGenerator/HeadQuarters.h
+++ b/libs/s25main/mapGenerator/HeadQuarters.h
@@ -44,6 +44,7 @@ namespace rttr { namespace mapGenerator {
      * @param area area within the HQ position should be
      *
      * @return all suitable HQ positions witihn the specified area (or the entire map if the area is empty).
+     * @throw runtime_error
      */
     template<class T_Container>
     std::vector<MapPoint> FindHqPositions(const Map& map, const T_Container& area)
@@ -97,7 +98,7 @@ namespace rttr { namespace mapGenerator {
      * @param index player index for the HQ
      * @param area area to place the HQ in
      *
-     * @throw NoHqPositionFoundException
+     * @throw runtime_error
      */
     template<class T_Container>
     void PlaceHeadQuarter(Map& map, int index, const T_Container& area)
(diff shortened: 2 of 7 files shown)


libs/s25main/mapGenerator/HeadQuarters.cpp
```diff
@@ -90,7 +90,10 @@ namespace rttr { namespace mapGenerator {
             retries--;
         }
 
-        return success;
+        if(!success)
+        {
+            throw std::runtime_error("could not find any valid HQ position");
```
vader1986, line 95: I don't know if there're any other edge cases for random map generation I'm not aware of, but maybe it is worth to create customs exceptions for this? Something like `class NoHqPositionFound : MapCreationError` and `MapCreationError : std::exception`? Wasn't sure about it, just took the `runtime_error` which was presumably matched already to handle IO errors when writing the file?

Flamefire: IMO it's fully fine. The caller may simply catch any runtime_error or even exception and report the "what", it doesn't matter much what type of error it was
However we could think about the error message. I guess it doesn't make sense to show it to the user, but maybe optionally log it (a potential range_error or so wouldn't be useful to the user, then we'd need a custom class). I'd make the messages regular sentences though, so starting with a capital letter.

libs/s25main/mapGenerator/Resources.cpp
```diff
@@ -60,31 +60,31 @@ namespace rttr { namespace mapGenerator {
...
-        std::set<MapPoint, MapPointLess> forbiddenArea;
+        std::set<MapPoint, MapPointLess> excludedArea;
         NodeMapBase<unsigned> probabilities;
         probabilities.Resize(map.size, 0u);
 
         // Do not allow to place trees/stone piles nearby head quarters or
         // harbor positions to avoid inaccessible harbors or invalid player positions.
 
-        auto isForbidden = [&map](const MapPoint& pt) {
+        auto harberOrHeadQuarter = [&map](const MapPoint& pt) {
```
Flamefire, line 70: harberOrHeadQuarter -> harborOrHeadquarter

Flamefire: BTW: Maybe check the other code(s) too while you are at it: It is either HQ or Headquarter. I'm not native english but I think a HeadQuarter i.e. head quarter would be a side of a coin or a sliced up head ;D (and yes I had to google this too)

tests/s25Main/mapGenerator/testHeadQuarters.cpp
```diff
@@ -165,7 +170,45 @@ BOOST_AUTO_TEST_CASE(PlaceHeadQuarter_places_hq_on_map_at_suitable_position)
...
+        std::vector<MapPoint> area;
+        bool exceptionThrown = false;
+        try
+        {
+            PlaceHeadQuarter(map, 0, area);
+        } catch(...)
+        {
+            exceptionThrown = true;
+        }
+        BOOST_REQUIRE(exceptionThrown);
```
Flamefire, line 188: There is a Boost.Test facility for this: BOOST_REQUIRE_THROW: https://www.boost.org/doc/libs/1_64_0/libs/test/doc/html/boost_test/testing_tools/exception_correctness.html

tests/s25Main/mapGenerator/testHeadQuarters.cpp
```diff
@@ -165,7 +170,45 @@ BOOST_AUTO_TEST_CASE(PlaceHeadQuarter_places_hq_on_map_at_suitable_position)
...
+        std::vector<MapPoint> area{MapPoint(0, 0)};
+        bool exceptionThrown = false;
+        try
+        {
+            PlaceHeadQuarter(map, 0, area);
+        } catch(...)
+        {
+            exceptionThrown = true;
+        }
+        BOOST_REQUIRE(exceptionThrown);
```
Flamefire, line 207: Same as above

tests/s25Main/mapGenerator/testHeadQuarters.cpp
```diff
@@ -174,10 +217,13 @@ BOOST_AUTO_TEST_CASE(PlaceHeadQuarters_returns_true_for_any_player_number_on_sui
...
-
-            auto success = PlaceHeadQuarters(map, rnd, players);
-
-            BOOST_REQUIRE(success);
+            try
+            {
+                PlaceHeadQuarters(map, rnd, players);
+            } catch(...)
+            {
+                BOOST_FAIL("Failed to place HQs");
```
Flamefire, line 225: No need for this, a test would fail if an exception is uncaught. If you want you can use `BOOST_REQUIRE_NO_THROW` to be explicit

tests/s25Main/mapGenerator/testHeadQuarters.cpp
```diff
@@ -202,4 +248,26 @@ BOOST_AUTO_TEST_CASE(PlaceHeadQuarters_places_hqs_for_any_player_number_on_suita
...
+            map.textures.Resize(size, textures.Find(IsWater));
+            bool exceptionThrown = false;
+            try
+            {
+                PlaceHeadQuarters(map, rnd, players);
+            } catch(...)
+            {
+                exceptionThrown = true;
+            }
+            BOOST_REQUIRE(exceptionThrown);
```
Flamefire, line 268: BOOST_REQUIRE_THROW

tests/s25Main/mapGenerator/testHeadQuarters.cpp
```diff
@@ -202,4 +248,26 @@ BOOST_AUTO_TEST_CASE(PlaceHeadQuarters_places_hqs_for_any_player_number_on_suita
     }
 }
 
+BOOST_AUTO_TEST_CASE(PlaceHeadQuarters_without_suitable_position_throws_exception)
+{
+    MapExtent size(32, 32);
+    RandomUtility rnd(0);
+
+    for(int players = 1; players < 8; players++)
```
Flamefire, line 256: Instead of this (which makes the test run much longer) use `rttr::test::randomNumber(1, 8)`, see the testMap.cpp (IIRC)
The semantic is a bit different: Using the loop means that for the number of players there might be different handling and all must be tested while the random number means "any number of player will trigger the behavior that should be tested" (and still, over time, tests all cases) while cutting the runtime to 1/8 :)

libs/s25main/mapGenerator/HeadQuarters.h
```diff
@@ -91,19 +98,17 @@ namespace rttr { namespace mapGenerator {
...
      *
-     * @returns true if a HQ was placed fro the specified player, false otherwise.
+     * @throw runtime_error
      */
     template<class T_Container>
-    bool PlaceHeadQuarter(Map& map, int index, const T_Container& area)
+    void PlaceHeadquarter(Map& map, int index, const T_Container& area)
     {
-        auto positions = FindHqPositions(map, area);
+        const auto& positions = FindHqPositions(map, area);
```
Flamefire, line 106: Huh? Why the const-ref on a return value by-value?

Flamefire (changes requested):
One new issue and the "harber" typo ;)

However I'd leave this open for bit the discussion in save intervals, but showing minutes and seconds there should be added anyway for UX.

libs/s25main/mapGenerator/HeadQuarters.h
vader1986: > One new issue and the "harber" typo ;)
> 
> However I'd leave this open for bit the discussion in save intervals, but showing minutes and seconds there should be added anyway for UX.

Displaying minutes & seconds how long it took to create the map until it failed?

Flamefire: Don't mind me, didn't have enough coffee. Wrong PR :D

Flamefire (approved):
LGTM, Thanks!

libs/s25main/mapGenerator/HeadQuarters.h
vader1986: > Don't mind me, didn't have enough coffee. Wrong PR :D

Dangerous! Cheers!

libs/s25main/mapGenerator/HeadQuarters.h
```diff
@@ -103,7 +103,7 @@ namespace rttr { namespace mapGenerator {
     template<class T_Container>
     void PlaceHeadquarter(Map& map, int index, const T_Container& area)
     {
-        const auto& positions = FindHqPositions(map, area);
+        auto positions = FindHqPositions(map, area);
```
Flamefire, line 106: Just for the future: I started developing the habit of making everything possible `const` and by now I like it as it avoids accidental misuse (e.g writing to the wrong variable)
Not worth changing now, but for future changes I wanted to suggest adopting this too (although we don't have a requirement for this)

vader1986: It does make sense. I wish there was a shorter way to mark sth as const, like `cauto` for `const auto` :D

Flamefire:
Sorry for the failing test, that seems to be a case where the water map is still to small for 8 players
Can you apply this patch (makes sense to include it here):
```
diff --git a/tests/s25Main/mapGenerator/testRandomMap.cpp b/tests/s25Main/mapGenerator/testRandomMap.cpp
index 7f826f279..89e021bcb 100644
--- a/tests/s25Main/mapGenerator/testRandomMap.cpp
+++ b/tests/s25Main/mapGenerator/testRandomMap.cpp
@@ -40,9 +40,9 @@ static void ValidateMap(const Map& map, const MapExtent& size, unsigned players)
     }
 }
 
-static MapExtent getRandomMapSize(const unsigned minSize = 32)
+static MapExtent getRandomMapSize(const unsigned minSize = 32, const unsigned maxSize = 80)
 {
-    auto size = rttr::test::randomPoint<MapExtent>(minSize, 80);
+    auto size = rttr::test::randomPoint<MapExtent>(minSize, maxSize);
     // Need even size
     size.x &= ~1u;
     size.y &= ~1u;
@@ -71,7 +71,7 @@ BOOST_AUTO_TEST_CASE(GenerateRandomMap_returns_valid_water_map)
     loadGameData(worldDesc);
     MapSettings settings;
 
-    settings.size = getRandomMapSize(50); // Need enough space for player islands
+    settings.size = getRandomMapSize(76, 90); // Need enough space for player islands
     settings.numPlayers = rttr::test::randomValue(2u, MAX_PLAYERS);
     settings.style = MapStyle::Water;
 

```

Flamefire (approved):

vader1986:
Something fishy going on here with the build (some check are not completing, but not failing either).

Flamefire (dismissed):

Flamefire:
@Flow86 Could you remove the requirement that branches are up-to-date? I find that inconvenient for contributors as every one needs to rebase when any PR is merged even when they don't touch each others files and hence are independent. This should be fully on the reviewer to decide if this needs a rebase, (force-)push and retest on all CIs (which is quite a lot of work) or not.

Flow86:
I can't change that, it will show up if there is a conflict.

Flamefire:
No I mean the status message "This branch is out of date... and must be updated", not the "conflicting files" message.
Seems you enabled "linear history" requirement: https://docs.github.com/en/github/administering-a-repository/about-protected-branches#require-linear-history
However I find this not that great. Having features grouped in branches as single commits and then merged as a single merge commit is fully fine and even haven a few (like 2-3) parallel, non-overlapping branches then independently (w/o rebase) merged still "looks good"
One problem is (similar to a recent critic) that it makes contributors do more work making them less likely to contribute more

Flamefire (approved):

Issue #1317, opened 2020-12-17
How to represent game time


JonathanSteinbuch:
Hello,

I would like to discuss how ingame time is supposed to be represented. As far as I can tell ingame time occurs explicitly in the statitistics window in the original.
![timeStatistics](https://user-images.githubusercontent.com/711115/102547750-f21c4b00-40b9-11eb-8373-ff887f769ebe.png)
Here the minutes displayed are the time it would take at 40ms per game frame (i.e. fast speed).

It occurs differently in tournament mode.
![timeTournament](https://user-images.githubusercontent.com/711115/102548626-38be7500-40bb-11eb-879a-2c7c2c1a5992.png)
Looking at the code for this, the minutes displayed are computed with the game frame length corresponding to the speed the game is actually run at (this means that under perfect conditions this should be the actual time the tournament game takes)

At other places actually game frames are displayed, though I feel those are more like debug features, so not as relevant:
![timeGameFrames](https://user-images.githubusercontent.com/711115/102548998-bc786180-40bb-11eb-8269-81a468b3ea09.png)

Those are all the instances I found where ingame time is represented.

In my economy mode implementationI went with the actual time under ideal conditions (as in the tournament length, after a question by @Flamefire). It is displayed in two places.
![timeEconomyLength](https://user-images.githubusercontent.com/711115/102549322-33adf580-40bc-11eb-999c-2371e44058b1.png)
![timeEconomic](https://user-images.githubusercontent.com/711115/102549338-36a8e600-40bc-11eb-9a7c-00d324483fbe.png)

So my question is how should time be displayed?
1. Assuming a reference speed (as in the statistics, although I guess normal speed as reference would be more intuitive). Pros: Simple. Makes game states after the same elapsed time more comparable. Was used in the statistics in the original Settlers 2. Cons: Can be confusing when you don't play at the reference speed (especially if you want to set a game length as in tournament or economy mode)
2. With actual time under ideal conditions (as in the tournament length). Pros: Is actually true (apart from the fact that currently frame times are wrong when you use hardware vsync, which I plan to fix) Cons: Not very comparable between different speeds
3. By game frame. Pros: Is definitely true. Comparable. Cons: The user doesn't know what a game frame is. Also quite unwieldy to write 22500 GF instead of 15min

Other options:
4. By a proxy for game frames (for example a "year" could be 45000 GFs or something)
5. Leave the differing representations as they are but write in the info texts how the mentioned time should be interpreted.
[...]

PR #1348, opened 2021-01-13
Base ingame times on normal speed

JonathanSteinbuch:
This addresses issue #1317 

Although there was no consensus on which speed to base ingame time there was a consensus that it should be consistent and without reference to game frames.

So as a preliminary solution I changed every time representation based on the assumption of normal speed. 

In the save dialog, the playReplay window and the post window in particular that meant changing GFs to time. 
timePostWindowNew

The only thing that maybe looks a bit silly is the autosave pulldown, because I didn't change the intervals (should I?):
autosave

Note: I didn't remove the display of GFs in the debug line at the top and for jumping/skipping as they are mostly debug/expert features

### DIFF
--- a/libs/s25main/Game.cpp
+++ b/libs/s25main/Game.cpp
@@ -49,9 +49,8 @@ void Game::Start(bool startFromSave)
         if(ggs_.objective == GameObjective::EconomyMode)
         {
             unsigned int selection = ggs_.getSelection(AddonId::ECONOMY_MODE_GAME_LENGTH);
-            world_.econHandler =
-              std::make_unique<EconomyModeHandler>(std::chrono::minutes(AddonEconomyModeGameLengthList[selection])
-                                                   / std::chrono::milliseconds(SPEED_GF_LENGTHS[referenceSpeed]));
+            world_.econHandler = std::make_unique<EconomyModeHandler>(
+              std::chrono::minutes(AddonEconomyModeGameLengthList[selection]) / SPEED_GF_LENGTHS[referenceSpeed]);
         }
         StatisticStep();
     }
@@ -97,7 +96,7 @@ void Game::RunGF()
     if(world_.HasLua())
         world_.GetLua().EventGameFrame(em_->GetCurrentGF());
     // Update statistic every 30 seconds
-    constexpr unsigned GFsIn30s = 30 * 1000 / SPEED_GF_LENGTHS[referenceSpeed];
+    constexpr unsigned GFsIn30s = 30s / SPEED_GF_LENGTHS[referenceSpeed];
     if(em_->GetCurrentGF() % GFsIn30s == 0)
         StatisticStep();
     // If some players got defeated check objective
--- a/libs/s25main/Game.cpp
+++ b/libs/s25main/Game.cpp
@@ -51,7 +51,7 @@ void Game::Start(bool startFromSave)
             unsigned int selection = ggs_.getSelection(AddonId::ECONOMY_MODE_GAME_LENGTH);
             world_.econHandler =
               std::make_unique<EconomyModeHandler>(std::chrono::minutes(AddonEconomyModeGameLengthList[selection])
-                                                   / (std::chrono::milliseconds)SPEED_GF_LENGTHS[referenceSpeed]);
+                                                   / std::chrono::milliseconds(SPEED_GF_LENGTHS[referenceSpeed]));
         }
         StatisticStep();
     }
--- a/libs/s25main/Game.cpp
+++ b/libs/s25main/Game.cpp
@@ -96,7 +96,7 @@ void Game::RunGF()
     if(world_.HasLua())
         world_.GetLua().EventGameFrame(em_->GetCurrentGF());
     // Update statistic every 30 seconds
-    constexpr unsigned GFsIn30s = 30s / SPEED_GF_LENGTHS[referenceSpeed];
+    constexpr unsigned GFsIn30s = std::chrono::duration<unsigned>(30) / SPEED_GF_LENGTHS[referenceSpeed];
     if(em_->GetCurrentGF() % GFsIn30s == 0)
         StatisticStep();
     // If some players got defeated check objective
--- a/libs/s25main/Game.cpp
+++ b/libs/s25main/Game.cpp
@@ -24,6 +24,7 @@
 #include "ai/AIPlayer.h"
 #include "lua/LuaInterfaceGame.h"
 #include "network/GameClient.h"
+#include "gameData/GameConsts.h"
 #include <boost/optional.hpp>
 
 Game::Game(const GlobalGameSettings& settings, unsigned startGF, const std::vector<PlayerInfo>& players)
@@ -48,8 +49,9 @@ void Game::Start(bool startFromSave)
         if(ggs_.objective == GameObjective::EconomyMode)
         {
             unsigned int selection = ggs_.getSelection(AddonId::ECONOMY_MODE_GAME_LENGTH);
-            world_.econHandler = std::make_unique<EconomyModeHandler>(
-              std::chrono::minutes(AddonEconomyModeGameLengthList[selection]) / GAMECLIENT.GetGFLength());
+            world_.econHandler =
+              std::make_unique<EconomyModeHandler>(std::chrono::minutes(AddonEconomyModeGameLengthList[selection])
+                                                   / (std::chrono::milliseconds)SPEED_GF_LENGTHS[referenceSpeed]);
         }
         StatisticStep();
     }
@@ -94,8 +96,9 @@ void Game::RunGF()
 
     if(world_.HasLua())
         world_.GetLua().EventGameFrame(em_->GetCurrentGF());
-    // Update statistic every 750 GFs (30 seconds on 'fast')
-    if(em_->GetCurrentGF() % 750 == 0)
+    // Update statistic every 30 seconds
+    constexpr unsigned GFsIn30s = 30 * 1000 / SPEED_GF_LENGTHS[referenceSpeed];
+    if(em_->GetCurrentGF() % GFsIn30s == 0)
         StatisticStep();
     // If some players got defeated check objective
     if(getNumAlivePlayers(world_) < numPlayersAlive)
--- a/libs/s25main/controls/ctrlTable.cpp
+++ b/libs/s25main/controls/ctrlTable.cpp
@@ -110,40 +110,26 @@ static int Compare(const std::string& a, const std::string& b, ctrlTable::SortTy
             s25util::ClassicImbuedStream<std::istringstream> ss_a(a);
             s25util::ClassicImbuedStream<std::istringstream> ss_b(b);
 
-            int val_a[3], val_b[3];
+            int seconds_a = 0;
+            int seconds_b = 0;
             char c;
 
             // "h:mm:ss" or "mm:ss"
-            for(int i = 0; i < 3; i++)
+            int tmp;
+            while(ss_a >> tmp)
             {
-                ss_a >> val_a[i];
-                ss_b >> val_b[i];
-                if(!ss_a)
-                {
-                    RTTR_Assert(i == 2);
-                    for(int j = i; j > 0; j--)
-                    {
-                        val_a[j] = val_a[j - 1];
-                    }
-                    val_a[0] = 0;
-                }
-                if(!ss_b)
-                {
-                    RTTR_Assert(i == 2);
-                    for(int j = i; j > 0; j--)
-                    {
-                        val_b[j] = val_b[j - 1];
-                    }
-                    val_b[0] = 0;
-                }
+                seconds_a *= 60;
+                seconds_a += tmp;
                 ss_a >> c;
+            }
+            while(ss_b >> tmp)
+            {
+                seconds_b *= 60;
+                seconds_b += tmp;
                 ss_b >> c;
             }
-            for(int i = 0; i < 3; i++)
-            {
-                if(val_a[i] != val_b[i] || i >= 2)
-                    return val_a[i] - val_b[i];
-            }
+
+            return seconds_a - seconds_b;
         }
         break;
     }
--- a/libs/s25main/controls/ctrlTable.cpp
+++ b/libs/s25main/controls/ctrlTable.cpp
@@ -104,6 +104,47 @@ static int Compare(const std::string& a, const std::string& b, ctrlTable::SortTy
                 return h_a - h_b;
             return min_a - min_b;
         }
+        case SRT::Time:
+        {
+            // Sort by time with format h:mm:ss or mm:ss
+            s25util::ClassicImbuedStream<std::istringstream> ss_a(a);
+            s25util::ClassicImbuedStream<std::istringstream> ss_b(b);
+
+            int val_a[3], val_b[3];
+            char c;
+
+            // "h:mm:ss" or "mm:ss"
+            for(int i = 0; i < 3; i++)
+            {
+                ss_a >> val_a[i];
+                ss_b >> val_b[i];
+                if(!ss_a)
+                {
+                    RTTR_Assert(i == 2);
+                    for(int j = i; j > 0; j--)
+                    {
+                        val_a[j] = val_a[j - 1];
+                    }
+                    val_a[0] = 0;
+                }
+                if(!ss_b)
+                {
+                    RTTR_Assert(i == 2);
+                    for(int j = i; j > 0; j--)
+                    {
+                        val_b[j] = val_b[j - 1];
+                    }
+                    val_b[0] = 0;
+                }
+                ss_a >> c;
+                ss_b >> c;
+            }
+            for(int i = 0; i < 3; i++)
+            {
+                if(val_a[i] != val_b[i] || i >= 2)
+                    return val_a[i] - val_b[i];
+            }
+        }
         break;
     }
     return 0;
--- a/libs/s25main/desktops/dskGameInterface.cpp
+++ b/libs/s25main/desktops/dskGameInterface.cpp
@@ -369,8 +369,8 @@ void dskGameInterface::Msg_PaintAfter()
     }
 
     // Draw speed indicator icon
-    const int startSpeed = SPEED_GF_LENGTHS[game_->ggs_.speed];
-    const int speedStep = startSpeed / 10 - static_cast<int>(GAMECLIENT.GetGFLength() / std::chrono::milliseconds(10));
+    const int speedStep =
+      static_cast<int>(SPEED_GF_LENGTHS[referenceSpeed] / 10ms) - static_cast<int>(GAMECLIENT.GetGFLength() / 10ms);
 
     if(speedStep != 0)
     {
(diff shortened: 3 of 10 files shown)


libs/s25main/Game.cpp
```diff
@@ -48,8 +49,9 @@ void Game::Start(bool startFromSave)
         if(ggs_.objective == GameObjective::EconomyMode)
         {
             unsigned int selection = ggs_.getSelection(AddonId::ECONOMY_MODE_GAME_LENGTH);
-            world_.econHandler = std::make_unique<EconomyModeHandler>(
-              std::chrono::minutes(AddonEconomyModeGameLengthList[selection]) / GAMECLIENT.GetGFLength());
+            world_.econHandler =
+              std::make_unique<EconomyModeHandler>(std::chrono::minutes(AddonEconomyModeGameLengthList[selection])
+                                                   / (std::chrono::milliseconds)SPEED_GF_LENGTHS[referenceSpeed]);
```
Flamefire, line 54:
```suggestion
                                                   / std::chrono::milliseconds(SPEED_GF_LENGTHS[referenceSpeed]));
```

Slightly better as it constructs instead of casts consistent with the other usages

libs/s25main/Game.cpp
```diff
@@ -94,8 +96,9 @@ void Game::RunGF()
 
     if(world_.HasLua())
         world_.GetLua().EventGameFrame(em_->GetCurrentGF());
-    // Update statistic every 750 GFs (30 seconds on 'fast')
-    if(em_->GetCurrentGF() % 750 == 0)
+    // Update statistic every 30 seconds
+    constexpr unsigned GFsIn30s = 30 * 1000 / SPEED_GF_LENGTHS[referenceSpeed];
```
Flamefire, line 100: @Flow86 I'm not sure if we should change that. Correct would be every 37,5s (on normal)
Or wait... Did we have our statics time (i.e. the "last 15min") always based on fast? Looks like it... O.o
Ok, so this comment just for reference that this (correctly) changes the reference time for the statistic

libs/s25main/controls/ctrlTable.cpp
```diff
@@ -104,6 +104,47 @@ static int Compare(const std::string& a, const std::string& b, ctrlTable::SortTy
...
+        {
+            // Sort by time with format h:mm:ss or mm:ss
+            s25util::ClassicImbuedStream<std::istringstream> ss_a(a);
+            s25util::ClassicImbuedStream<std::istringstream> ss_b(b);
+
+            int val_a[3], val_b[3];
+            char c;
+
+            // "h:mm:ss" or "mm:ss"
+            for(int i = 0; i < 3; i++)
```
Flamefire, line 117: That code is not to pretty and hard to read. How about getting the time in seconds which makes the comparison below trivial: Start with zero, read a number, if succeeded multiply the previous result by 60 and add the number, else ignore
Also note that you can directly do `if(ss >> number)` which IMO reads better  as "if number was read", but that's only my preference

libs/s25main/network/GameClient.cpp
```diff
@@ -1719,7 +1719,7 @@ std::string GameClient::FormatGFTime(const unsigned gf) const
     using std::chrono::duration_cast;
 
     // In Sekunden umrechnen
-    seconds numSeconds = duration_cast<seconds>(gf * framesinfo.gf_length);
+    seconds numSeconds = duration_cast<seconds>(gf * (std::chrono::milliseconds)SPEED_GF_LENGTHS[referenceSpeed]);
```
Flamefire, line 1722: Just for consistency: 
```suggestion
    seconds numSeconds = duration_cast<seconds>(gf * std::chrono::milliseconds(SPEED_GF_LENGTHS[referenceSpeed]));
```

Flamefire (changes requested):
> The only thing that maybe looks a bit silly is the autosave pulldown, because I didn't change the intervals (should I?):

They do indeed look odd. Could you at least make them in format "xx min yy s" (~likely needs to be translatable~) or "xx::yy" or so (not sure what is the most readable but I'd mention the minutes instead of e.g. 2500s. And maybe (@Flow86 @Spikeone ?) adjust the GFs so that on normal (and optional on fast) speed we get good numbers? Example (instead of the 5k+GFs): 6000, 12000, 18000, 36000, 72000, 108000 Gfs (5,10,15,30,60,90min on normal and a bit less on fast, but even minutes)
BTW: What is the official way of writing? I think it is "50s" not "50 s", isn't it? Doesn't matter much, but did you check for consistency if we already do that somewhere else?

> I didn't remove the display of GFs in the debug line at the top and for jumping/skipping as they are mostly debug/expert features

Correct, those need to stay GFs.

libs/s25main/Game.cpp
JonathanSteinbuch: Yes, it was based on fast before

libs/s25main/controls/ctrlTable.cpp
JonathanSteinbuch: Good point, changed it

JonathanSteinbuch:
> They do indeed look odd. Could you at least make them in format "xx min yy s" (likely needs to be translatable) or "xx::yy" or so (not sure what is the most readable but I'd mention the minutes instead of e.g. 2500s. And maybe (@Flow86 @Spikeone ?) adjust the GFs so that on normal (and optional on fast) speed we get good numbers? Example (instead of the 5k+GFs): 6000, 12000, 18000, 36000, 72000, 108000 Gfs (5,10,15,30,60,90min on normal and a bit less on fast, but even minutes)

If we change the intervals I would vote for whole minutes (starting at 1 min) and base the intervals on a time (and then internally convert it to # of GFs based on the reference speed). This way it will look good even if we change the reference speed later.

> BTW: What is the official way of writing? I think it is "50s" not "50 s", isn't it? Doesn't matter much, but did you check for consistency if we already do that somewhere else?

In the statistics window (in RTTR and in the original) it is written "15 m", "1 h", "4 h" and "16 h". Apart from that I'm not aware of any official source. There seem to be some places in RTTR where we write "30 minutes" or "30min" though, which potentially should be changed.

Flamefire:
> If we change the intervals I would vote for whole minutes (starting at 1 min) and base the intervals on a time (and then internally convert it to # of GFs based on the reference speed). This way it will look good even if we change the reference speed later.

Sounds good. Then maybe `1,5,10,15,30,60,90min`? ~The "look good" part is problematic. The times by me were chose because 5min normal is 4min fast, 3min very fast so that stays whole minutes. But e.g. 1min normal is 48s fast.~
Oh wait. That doesn't matter at all as the reference is fixed. So yes go for it.
Note that there is a debug option for saving every GF. I think it makes sense to move that out of the array then and add it manually maybe with "each GF" as the text (and handle it in the change listener)

> Apart from that I'm not aware of any official source. There seem to be some places in RTTR where we write "30 minutes" or "30min" though, which potentially should be changed.

Ok just checked: There should be a space. So the others could be changed, but don't need to. Just for new code we should use the space

JonathanSteinbuch:
> Sounds good. Then maybe `1,5,10,15,30,60,90min`? ~The "look good" part is problematic. The times by me were chose because 5min normal is 4min fast, 3min very fast so that stays whole minutes. But e.g. 1min normal is 48s fast.~
> Oh wait. That doesn't matter at all as the reference is fixed. So yes go for it.

Implemented this now

> Note that there is a debug option for saving every GF. I think it makes sense to move that out of the array then and add it manually maybe with "each GF" as the text (and handle it in the change listener)

By the way: I noticed that the debug setting is not set when i compile it with debug flag but can be regardless of how it is compiled set in the settings file. So there is basically two kinds of debug mode? Is that on purpose?

libs/s25main/ingameWindows/iwSave.cpp
```diff
@@ -27,18 +27,19 @@
...
 #include "liblobby/LobbyClient.h"
 #include "s25util/Log.h"
 #include <utility>
 
-const unsigned NUM_AUTO_SAVE_INTERVALS = 7;
-
-const std::array<unsigned, NUM_AUTO_SAVE_INTERVALS> AUTO_SAVE_INTERVALS = {500,   1000,   5000, 10000,
-                                                                           50000, 100000, 1 /*  */};
+// autosave intervals in minutes
+const auto AUTO_SAVE_INTERVALS = helpers::make_array(1u, 5u, 10u, 15u, 30u, 60u, 90u);
```
Flamefire, line 41: Could you use chrono units here? Makes it easier to extend.

```suggestion
using namespace std::chrono_literals;
const auto AUTO_SAVE_INTERVALS = helpers::make_array(1min, 5min, 10min, 15min, 30min, 60min, 90min);
```

libs/s25main/ingameWindows/iwSave.cpp
```diff
@@ -27,18 +27,19 @@
...
+constexpr unsigned NUM_AUTO_SAVE_INTERVALS = AUTO_SAVE_INTERVALS.size();
```
Flamefire, line 42: I think there is no need for this, just do a Search&Replace and use the `AUTO_SAVE_INTERVALS.size()` directly

libs/s25main/ingameWindows/iwSave.cpp
```diff
@@ -142,24 +143,33 @@ iwSave::iwSave() : iwSaveLoad(40, _("Save game!"))
...
     combo->AddString(_("Disabled")); // deaktiviert
 
-    // Last entry is only for debugging
-    const unsigned numIntervalls = SETTINGS.global.debugMode ? NUM_AUTO_SAVE_INTERVALS : NUM_AUTO_SAVE_INTERVALS - 1;
-
     // Die Intervalle
-    for(unsigned i = 0; i < numIntervalls; ++i)
-        combo->AddString(helpers::toString(AUTO_SAVE_INTERVALS[i]) + " GF");
+    for(unsigned i = 0; i < NUM_AUTO_SAVE_INTERVALS; ++i)
+        combo->AddString((boost::format(_("%1% m")) % AUTO_SAVE_INTERVALS[i]).str());
```
Flamefire, line 148:
```suggestion
    for(const std::chrono::minutes interval: AUTO_SAVE_INTERVALS)
        combo->AddString((boost::format(_("%1% min")) % interval.count()).str());
```

libs/s25main/ingameWindows/iwSave.cpp
```diff
@@ -142,24 +143,33 @@ iwSave::iwSave() : iwSaveLoad(40, _("Save game!"))
...
+    }
 
     // Richtigen Eintrag auswählen
     bool found = false;
-    for(unsigned i = 0; i < numIntervalls; ++i)
+    constexpr unsigned GFsPerMinute = 60 * 1000 / SPEED_GF_LENGTHS[referenceSpeed];
+    for(unsigned i = 0; i < NUM_AUTO_SAVE_INTERVALS; ++i)
     {
-        if(SETTINGS.interface.autosave_interval == AUTO_SAVE_INTERVALS[i])
+        if(SETTINGS.interface.autosave_interval == AUTO_SAVE_INTERVALS[i] * GFsPerMinute)
```
Flamefire, line 161: This can then be: 
```suggestion
        if(SETTINGS.interface.autosave_interval == AUTO_SAVE_INTERVALS[i] / std::chrono::milliseconds(SPEED_GF_LENGTHS[referenceSpeed])
```
Or introduce a constant `referenceGFLength` or so.
Reason is a) easier to read and b) avoid manual time computation via magic numbers

libs/s25main/ingameWindows/iwSave.cpp
```diff
@@ -174,9 +184,14 @@ void iwSave::Msg_ComboSelectItem(const unsigned /*ctrl_id*/, const unsigned sele
...
     if(selection == 0)
         SETTINGS.interface.autosave_interval = 0;
+    else if(selection >= NUM_AUTO_SAVE_INTERVALS)
+        SETTINGS.interface.autosave_interval = 1;
     else
+    {
         // ansonsten jeweilige GF-Zahl eintragen
-        SETTINGS.interface.autosave_interval = AUTO_SAVE_INTERVALS[selection - 1];
+        constexpr unsigned GFsPerMinute = 60 * 1000 / SPEED_GF_LENGTHS[referenceSpeed];
+        SETTINGS.interface.autosave_interval = AUTO_SAVE_INTERVALS[selection - 1] * GFsPerMinute;
```
Flamefire, line 193: See above

Flamefire (changes requested):
A few suggestions (mainly use chrono units). BTW: Big fan of the make_array :)

> I noticed that the debug setting is not set when i compile it with debug flag but can be set in the settings file regardless of how it is compiled. So there is basically two kinds of debug mode? Is that on purpose?

Kinda, yes. I think the reason is to be able to compile in release (with optimizations for speed) and still enable the auto-save every GF to debug an async or the like.

libs/s25main/ingameWindows/iwSave.cpp
JonathanSteinbuch: I wasn't aware of chrono literals. Useful :)

libs/s25main/ingameWindows/iwSave.cpp
JonathanSteinbuch: makes sense

Flamefire (approved):

Flamefire:
@Flow86 Did you change anything on Jenkins?

> ‘Jenkins’ doesn’t have label ‘docker’

Probably since today only.

libs/s25main/ingameWindows/iwSave.cpp
```diff
@@ -174,9 +183,14 @@ void iwSave::Msg_ComboSelectItem(const unsigned /*ctrl_id*/, const unsigned sele
...
     if(selection == 0)
         SETTINGS.interface.autosave_interval = 0;
+    else if(selection >= AUTO_SAVE_INTERVALS.size())
+        SETTINGS.interface.autosave_interval = 1;
     else
+    {
         // ansonsten jeweilige GF-Zahl eintragen
-        SETTINGS.interface.autosave_interval = AUTO_SAVE_INTERVALS[selection - 1];
+        SETTINGS.interface.autosave_interval =
+          AUTO_SAVE_INTERVALS[selection - 1] / std::chrono::milliseconds(SPEED_GF_LENGTHS[referenceSpeed]);
```
Flamefire, line 192: MSVC needs a `static_cast<unsigned>` here

libs/s25main/network/GameClient.cpp
```diff
@@ -1759,7 +1759,7 @@ unsigned GameClient::GetTournamentModeDuration() const
             < rttr::enum_cast(GameObjective::Tournament1) + NUM_TOURNAMENT_MODES)
     {
         const auto turnamentMode = rttr::enum_cast(game->ggs_.objective) - rttr::enum_cast(GameObjective::Tournament1);
-        return minutes(TOURNAMENT_MODES_DURATION[turnamentMode]) / framesinfo.gf_length;
+        return minutes(TOURNAMENT_MODES_DURATION[turnamentMode]) / milliseconds(SPEED_GF_LENGTHS[referenceSpeed]);
```
Flamefire, line 1762: MSVC needs a `static_cast<unsigned>` here (shame that a duration is signed, but whatever)

libs/s25main/Game.cpp
```diff
@@ -48,8 +49,9 @@ void Game::Start(bool startFromSave)
         if(ggs_.objective == GameObjective::EconomyMode)
         {
             unsigned int selection = ggs_.getSelection(AddonId::ECONOMY_MODE_GAME_LENGTH);
-            world_.econHandler = std::make_unique<EconomyModeHandler>(
-              std::chrono::minutes(AddonEconomyModeGameLengthList[selection]) / GAMECLIENT.GetGFLength());
+            world_.econHandler =
+              std::make_unique<EconomyModeHandler>(std::chrono::minutes(AddonEconomyModeGameLengthList[selection])
```
Flamefire, line 53: MSVC needs a `static_cast<unsigned>` here

Flamefire (changes requested):
Few appveyor failures (annoying...) about conversion
Maybe it is easier to just change the `SPEED_GF_LENGTHS` array to contain `std::chrono::duration<unsigned, std::milli>` which avoids the casts to `milliseconds` in all places and also the conversion warning and is easier to read. Just needs a quick check if this is save in all places to use an unsigned type but I'd guess so.

JonathanSteinbuch:
> Few appveyor failures (annoying...) about conversion

Earlier there were some test failures with the headquarters problem so I didn't check AppVeyor before. Sorry, and thanks for looking at it again.

> Maybe it is easier to just change the `SPEED_GF_LENGTHS` array to contain `std::chrono::duration<unsigned, std::milli>` which avoids the casts to `milliseconds` in all places and also the conversion warning and is easier to read. Just needs a quick check if this is save in all places to use an unsigned type but I'd guess so.

I had thought about that too, so I just did it. Runs fine for me. Will check if AppVeyour is okay with it, later.

It also made me realize that I had forgotten to change the speed indicator icon speed dependence. If everything is in reference to normal speed it only makes sense if any deviation from that speed is indicated.

libs/s25main/Game.cpp
```diff
@@ -94,8 +95,9 @@ void Game::RunGF()
 
     if(world_.HasLua())
         world_.GetLua().EventGameFrame(em_->GetCurrentGF());
-    // Update statistic every 750 GFs (30 seconds on 'fast')
-    if(em_->GetCurrentGF() % 750 == 0)
+    // Update statistic every 30 seconds
+    constexpr unsigned GFsIn30s = std::chrono::duration<unsigned>(30) / SPEED_GF_LENGTHS[referenceSpeed];
```
Flamefire, line 99: Shame here. The `30s` looked nicer, but well... I guess this is because `seconds` uses a 64bit type and hence is wider than `unsigned` while for `minutes` it is a 32 bit type. I'd probably use a cast at the end, but doesn't matter which way.

libs/s25main/desktops/dskGameInterface.cpp
```diff
@@ -369,8 +369,8 @@ void dskGameInterface::Msg_PaintAfter()
     }
 
     // Draw speed indicator icon
-    const int startSpeed = SPEED_GF_LENGTHS[game_->ggs_.speed];
-    const int speedStep = startSpeed / 10 - static_cast<int>(GAMECLIENT.GetGFLength() / std::chrono::milliseconds(10));
+    const int speedStep =
+      static_cast<int>(SPEED_GF_LENGTHS[referenceSpeed] / 10ms) - static_cast<int>(GAMECLIENT.GetGFLength() / 10ms);
```
Flamefire, line 373: Ah, this would have been the place for signed durations to allow that `/10ms` to be put outside. Anyway good enough, I'd say

libs/s25main/gameData/GameConsts.h
```diff
@@ -20,10 +20,16 @@
 #include "helpers/EnumArray.h"
 #include "helpers/make_array.h"
 #include "gameTypes/GameSettingTypes.h"
+#include <chrono>
 #include <limits>
 
+using namespace std::chrono_literals;
```
Flamefire, line 26: Just a word of caution here: `using namespace` in global namespace in a header is discouraged. However this one is fine by me.

Flamefire (approved):
> some test failures with the headquarters problem 

Yeah seen that too, will be fixed in #1347, sorry about that.

> It also made me realize that I had forgotten to change the speed indicator icon speed dependence.

Good catch. This might be one of the main reasons why I wanted to have the start speed as reference. Likely everyone will play on "fast" and now that icon will always be shown while before it was only shown very rarely (when the speed was changed mid-game), but it makes sense because all show times will be faster than displayed too so we need to indicate why.

Thanks for the changes (comments are information only), waiting for @Flow86 to either fix CI or just merge this.

PR #1351, opened 2021-01-18
Fix wrong Lua GD_SHIELDS constant and add tests

### LINKED ISSUE #1349: [LUA] goods broken?
looks like

```
rttr:GetPlayer(p):AddWares({
            [GD_BEER      ] = 12,
            [GD_TONGS     ] = 0,
            [GD_HAMMER    ] = 0,
            [GD_AXE       ] = 0,
            [GD_SAW       ] = 0,
            [GD_PICKAXE   ] = 0,
            [GD_SHOVEL    ] = 0,
            [GD_CRUCIBLE  ] = 0,
            [GD_RODANDLINE] = 0,
            [GD_SCYTHE    ] = 0,
            [GD_WATER     ] = 12,
            [GD_CLEAVER   ] = 0,
            [GD_ROLLINGPIN] = 0,
            [GD_BOW       ] = 0,
            [GD_BOAT      ] = 0,
            [GD_SWORD     ] = 0,
            [GD_IRON      ] = 0,
            [GD_FLOUR     ] = 0,
            [GD_FISH      ] = 2,
            [GD_BREAD     ] = 4,
            [GD_SHIELD    ] = 0,
            [GD_WOOD      ] = 0,
            [GD_BOARDS    ] = 15,
            [GD_STONES    ] = 12,
            [GD_GRAIN     ] = 0,
            [GD_COINS     ] = 0,
            [GD_GOLD      ] = 0,
            [GD_IRONORE   ] = 0,
            [GD_COAL      ] = 20,
            [GD_MEAT      ] = 2,
            [GD_HAM       ] = 0
        })
```

causes a crash - anything wrong with that?

Flamefire:
Fixes #1349

The tests are intentionally simple, validating all constants exist (C&P from the docs) and can be used at all (SafeEnum works)


tests/s25Main/lua/testLuaConstants.cpp
```diff
@@ -0,0 +1,229 @@
...
+#include <boost/test/unit_test.hpp>
+
+struct LuaConstsTestsFixture : public LuaTestsFixture
+{
+    LuaConstsTestsFixture()
+    {
+        initWorld();
+        // This adds an `elements` function which can iterate over a table generated by
+        // table.pack (important as nil entries caused by missing constants abort regular iteration)
+        executeLua(R"LUA(
```
Flow86, line 28: gcc-6 does not have support for raw strings yet?

Flow86 (changes requested):

tests/s25Main/lua/testLuaConstants.cpp
Flow86: or at least broken support in macros, could that be? see https://gcc.gnu.org/bugzilla/show_bug.cgi?id=55971

Flamefire: Strange... Can't reproduce on godbolt: https://godbolt.org/z/ddKxKo
Doing a quick check, the we can decide if we simply remove support for GCC 6 (which I can't even install anymore on Ubuntu 20) for lack of proper C++11 support or just remove the macro (not that great, but a comment will suffice)

Flow86: perhaps its 6.0 on the pipeline?

Flamefire: Cmake says 6.5 (which is odd as Godbolt has only 6.4, oh and GCC x.0 doesn't exist. They start at x.1 and 6.1 is also working on godbolt). Anyway, removed the macro call.

Flow86:
still failing?
```
CMakeFiles/Test_integration.dir/testAttacking.cpp.o: In function `AttackSuite::ConquerWithMultipleWalkingIn::test_method()':
testAttacking.cpp:(.text+0xcc2d): undefined reference to `boost::test_tools::tt_detail::print_log_value<decltype(nullptr)>::operator()(std::ostream&, decltype(nullptr))'
testAttacking.cpp:(.text+0xce1b): undefined reference to `boost::test_tools::tt_detail::print_log_value<decltype(nullptr)>::operator()(std::ostream&, decltype(nullptr))'
testAttacking.cpp:(.text+0xd4fe): undefined reference to `boost::test_tools::tt_detail::print_log_value<decltype(nullptr)>::operator()(std::ostream&, decltype(nullptr))'
```

Flamefire:
Bug in Boost.Test fixed in 1.67: https://github.com/boostorg/test/commit/c767b8790aa3f8d84c75cd1e57416b2cacf18056

Shall I add a workaround or shall we raise the requirement to 1.67? See also https://github.com/Return-To-The-Roots/s25client/issues/1290

Flamefire:
@Flow86 Any preference?

Flow86:
we probably want to push to 1.69? since thats the "stable" debian buster version

Flamefire:
> we probably want to push to 1.69? since thats the "stable" debian buster version

Hm, seems like Ubuntu 18 still has 1.65.1. So some users would need to install a newer version manually. To me this is fine, what do you say?
Also something seems to be different on Linux jenkins. How/where is Boost installed there?

Flow86:
yes probably we stay at where we were then?

Flamefire:
Hm, Jenkins runs 1.67 (which is where this bug was fixed) on linux, 1.69 on the others. Ubuntu 20 has 1.71 but 18 does only have 1.65. So if it isn't to much trouble I'd say update the jenkins linux boost to 1.69 too

@vader1986 @JonathanSteinbuch What is your dev environment? Would it be a hassle to use Boost 1.69 or are you using that already anyway? I'd like to gauge the impact for devs and you are the most current contributors, so that's all the data we have for now ;) I don't mind reverting to 1.65 but if it doesn't hurt anyone then 1.69 would be better. There are quite a few changes/fixes which will/may become relevant.

vader1986:
I'm on Mac OS X 11.1 / Xcode 12.3. Don't think it's a problem to go for 1.69

JonathanSteinbuch:
@Flamefire Most recently I used Boost 1.75 (on Ubuntu 20.04). No hassle there for me.

Flamefire:
Ok thanks, that sounds good! Then I'd say we go to 1.69. (and I'd likely FetchContent boost::variant2 from 1.73 for the better performance)
@Flow86 Could you update the (linux) build server image to boost 1.69?

Flow86 (approved):

PR #1352, opened 2021-01-19
Algorithm refactoring

vader1986:


vader1986:
Just a note - I checked performance between both commits. None of them has any obvious impact, seemed to be a small code style improvement anyway.

libs/s25main/mapGenerator/Algorithms.h
```diff
@@ -225,47 +225,36 @@ namespace rttr { namespace mapGenerator {
...
-        unsigned valuesInRange = 0;
-        RTTR_FOREACH_PT(MapPoint, values.GetSize())
-        {
-            if(values[pt] >= minimum && values[pt] <= maximum)
-            {
-                valuesInRange++;
-            }
-        }
-        return valuesInRange;
+        std::function<unsigned(const std::vector<T>&)> countBetween = [minimum, maximum](const std::vector<T>& values) {
```
Flamefire, line 251:
```suggestion
        auto countBetween = [minimum, maximum](const std::vector<T>& values) {
```
https://www.reddit.com/r/cpp/comments/kzvjgn/c_performance_trap_2_unnecessary_stdfunction/
Keep this a lambda and then we can change the `nodes.Map` later to also take a lambda. `std::function` has some heavy overhead: https://www.reddit.com/r/cpp/comments/kzvjgn/c_performance_trap_2_unnecessary_stdfunction/ (found recently). Note that the rest does not need to change but it avoids changing this again

Flamefire (changes requested):
Thanks, this indeed looks cleaner. Minor thing for future proofing (and GHA seems to have flake and needs to restart anyway)

libs/s25main/mapGenerator/Algorithms.h
vader1986: When I try to pass it in as a lambda, I get:
`No matching member function for call to 'Map'`

I guess it doesn't like the template return type `T_Result` of `Map` in combination with a lambda?

vader1986: The idea of the `NodeMapBase.Map` function was to have a more efficient way to perform operations on the stored nodes by operating directly on the `std::vector<Node>` instead of iterating over all `MapPoint`s. Also, that way I could use standard functions like `count_if` which makes the code more readable. The obvious disadvantage that I'm exposing the underlying storage of the `NodeMapBase` (but only as `const ref`). 

I also thought about splitting up the `Map` function into more concrete ones, like `unsigned Count(lambda_describing_which_values_to_count)`. Maybe there's a nicer way to achieve this?

Flamefire: Ah yes, good point, sorry I missed that.
The idea is good, but maybe for this a getter is better or even adding (proxy) begin/end functions so you could even do `for(const auto& node: map)`, but I guess `for(const auto& node: map.getNodes())` is good enough (side note: I want to get to having method names lowerCase for consistency and distinguishing from types)
The reason is that for `map` I expected that the lambda is called for each node, which is the usual meaning of "map". We could add `accumulate(lambda, startVal)` and a "real" `map(lambda)`, but that depends on what we need. Judging from the current usages I'd say `getNodes` or `begin/end` is the best and removing the `Map` function. The latter does not fully expose the storage class, so maybe that is better and it allows exposing non-const iterators (exposing a non-const vector is impossible as changing its size would break the class)

And maybe add an inverse `getIdx`: `MapPoint getPosition(unsigned index)`(name?) (you used that functionality in one of the Map-calls)

What do you say? Would you be up for those changes? (Maybe even in this PR as it looks related)

> I guess it doesn't like the template return type T_Result of Map in combination with a lambda?

Yeah it can't deduce it. So `map.Map<unsigned>(..)` would need to be used or something like `template<typename T> auto Map(T&& func) const;`

vader1986: Yes, definitely like your suggestions. I'll take a look at this after work!

vader1986:
> Thanks, this indeed looks cleaner. Minor thing for future proofing (and GHA seems to have flake and needs to restart anyway)

Looks like GHA are still failing. I wonder if something changed with the docker image it is running in.

Flamefire:
> Looks like GHA are still failing. I wonder if something changed with the docker image it is running in.

Can you try adding `sudo apt update` before the `sudo apt install` in .github/workflows/unit-tests.yml?
Probably break it in 2 statements/lines with the pipe (see the OSX step).

And I'd use a typedef/using for `std::vector<Node>::const_iterator` (member type(def) `const_iterator`) to avoid the repetition and don't tie it to std::vector (explicitly at least, of course in the typedef it has to). But I'm also fine with the current state.

libs/s25main/world/NodeMapBase.h
```diff
@@ -31,6 +31,7 @@ class NodeMapBase final : public MapBase
 
 public:
     using Node = T_Node;
+    typedef typename std::vector<Node>::const_iterator const_iterator;
```
Flamefire, line 34: For consistency. Will apply this now as the rest looks ok.
```suggestion
    using const_iterator = typename std::vector<Node>::const_iterator;
```

Flamefire (approved):

libs/s25main/mapGenerator/Algorithms.h
```diff
@@ -324,7 +292,9 @@ namespace rttr { namespace mapGenerator {
         while(currentNodes < expectedNodes && limit <= maximum)
         {
             previousNodes = currentNodes;
-            currentNodes = Count(values, minimum, limit);
+            currentNodes = std::count_if(values.begin(), values.end(), [minimum, limit](const T value) {
```
Flamefire, line 295: MSVC needs a static_cast here (do a `git pull` first to get my change)
Can't do that on GH as it likely needs a reformat.

Flamefire (changes requested):

Flamefire (dismissed):

vader1986:
Removing lines of code unfortunately lead to less test coverage (at least in %).

Flamefire:
> Removing lines of code unfortunately lead to less test coverage (at least in %).

That's fine. The coverage CI is informative only and should encourage good tests. Covering new code or changed code (one of the 2 checks) is the goal, though not always (easily) possible.

PR #1353, opened 2021-01-22
Mountain river transitions

vader1986:
image

I think this looks a bit more interesting. I planned to add the option to configure rivers in the UI later. Could do it now or in the same PR. What would you prefer?

### DIFF
--- a/libs/s25main/ingameWindows/iwMapGenerator.cpp
+++ b/libs/s25main/ingameWindows/iwMapGenerator.cpp
@@ -30,7 +30,7 @@ using namespace rttr::mapGenerator;
 
 enum
 {
-    CTRL_LAST_ID = 6, // last UI control ID used before enum controls
+    CTRL_LAST_ID = 7, // last UI control ID used before enum controls
     CTRL_PLAYER_NUMBER,
     CTRL_MAP_STYLE,
     CTRL_MAP_SIZE,
@@ -38,11 +38,12 @@ enum
     CTRL_RATIO_GOLD,
     CTRL_RATIO_IRON,
     CTRL_RATIO_COAL,
-    CTRL_RATIO_GRANITE
+    CTRL_RATIO_GRANITE,
+    CTRL_RIVERS
 };
 
 iwMapGenerator::iwMapGenerator(MapSettings& settings)
-    : IngameWindow(CGI_MAP_GENERATOR, IngameWindow::posLastOrCenter, Extent(250, 350), _("Map Generator"),
+    : IngameWindow(CGI_MAP_GENERATOR, IngameWindow::posLastOrCenter, Extent(250, 380), _("Map Generator"),
                    LOADER.GetImageN("resource", 41), true),
       mapSettings(settings)
 {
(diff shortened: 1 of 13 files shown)


libs/s25main/mapGenerator/Triangles.h
```diff
@@ -35,22 +35,6 @@ namespace rttr { namespace mapGenerator {
...
-        Triangle Inverse() const { return Triangle(!rsu, position); }
-
-        /**
-         * Creates a new, inverted triangle with the specified position.
-         *
-         * @param position position of the triangle
-         *
-         * @returns a new, inverted triangle (RSU -> LSD, LSD -> RSU).
-         */
-        Triangle Inverse(const MapPoint& position) const { return Triangle(!rsu, position); }
```
vader1986, line 53: Those functions seemed fairly useless. Also "flip" would be a more appropriate name than "inverse".

libs/s25main/mapGenerator/Textures.cpp
```diff
@@ -190,13 +198,21 @@ namespace rttr { namespace mapGenerator {
...
+        const auto mountainFoot = SelectPoints(
           [this](const MapPoint& pt) {
               return this->textureMap_.Any(pt, IsMinableMountain) && !this->textureMap_.All(pt, IsMountainOrSnowOrLava);
           },
           this->textures_.GetSize());
 
         ApplyMountainTransitions(mountainFoot);
+
+        const auto moutainWaterTransition = SelectPoints(
+          [this](const MapPoint& pt) {
```
Flamefire, line 210: That was a lot of lambdas to follow ;)

Related: I've seen that you take the functor by instance, instead of by universal reference: https://github.com/Return-To-The-Roots/s25client/blob/7db186c6e58f72ca2ab2ecd080f325e264644a2c/libs/s25main/mapGenerator/Textures.h#L128 I'd usually suggest to use `T_Predicate&&` and forward as required or by const ref. Doesn't make a difference here as you are passing function pointers, but when you do benchmarking again you could check that.
You could also check, how the compiler would handle it, if you convert `SelectPoints` and `IsMinableMountain` & friends to inline functions instead of defining them in the cpp file. Maybe the compiler can inline all the lambdas/functions making this way more efficient. But of course should be measured first if this is worth it...

Just some ideas for the future I got while reviewing and wanted to mention them :)

Flamefire (approved):

Flamefire:
> I planned to add the option to configure rivers in the UI later. Could do it now or in the same PR. What would you prefer?

It is basically an option for the new feature, so I'd add it here. Likely also less work for you, but no preference from my side: Small PRs are fast to review but being "more complete" makes the (other) change easier to understand.

libs/s25main/mapGenerator/Textures.cpp
vader1986: I only used those helper functions to avoid confusing stuff, e.g. by accidentally using two different ways to checking whether a texture is water or not. That was the original thought behind it. 

Just tried to move everything into header and put an "inline" upfront, but there was no real performance boost, unfortunately.

vader1986:
Added to the UI to control number of rivers:
image

Flamefire (approved):

libs/s25main/ingameWindows/iwMapGenerator.cpp
```diff
@@ -30,19 +30,20 @@ using namespace rttr::mapGenerator;
 
 enum
 {
-    CTRL_LAST_ID = 6, // last UI control ID used before enum controls
+    CTRL_LAST_ID = 7, // last UI control ID used before enum controls
```
Flamefire, line 33: Just a hint: To avoid such changes, simply give all IDs an enum value, even the texts.

PR #1354, opened 2021-01-24
Removed redundant Distances function

vader1986:


libs/s25main/mapGenerator/Terrain.cpp
```diff
@@ -28,7 +28,7 @@ namespace rttr { namespace mapGenerator {
     {
         const MapExtent& size = map.size;
         auto& z = map.z;
-        auto distances = Distances(size, predicate);
+        auto distances = DistancesTo(SelectPoints(predicate, size), size);
```
Flamefire, line 31: I'm not sure about this. If I look at this it looks worse: the `size` gets duplicated and it has to create an intermediate vector which incurs a memory allocation which may be avoided (depending on how `Distances` is implemented, which is transparent from this PoV)
I fear that if this becomes a bottleneck (or is it maybe already?) changing it is made more difficult.

What was the reason for this? Have you considered using `SelectPoints` in `Distances` instead? This still means the function needs to be tested but it avoids the `size` duplication and allows more freedom in optimizing that function later.

Or maybe passing in the size is a mistake: It needs to match the size of whatever predicate checks which can not really be validated automatically. Maybe those functions should be part of `map` and the predicate gets passed the map (i.e. `*this`) and the point?

Just thinking loud here, I'd be happy to hear your opinion, I'm not settled on anything yet.

vader1986: So the `Distances` function was also creating a vector before (see the removed code in `Algorithm.h`. I agree with the duplicated size, that's a bit odd. The only reason I pass the size into `DistancesTo` is to calculate the max distance, which could be passed in instead:
`const unsigned maximumDistance = size.x * size.y;`

Initially I just changed this inside of the `Distances` function to call `SelectPoints` instead of duplicating the logic. But generally, it seemed to be more readable this way:
`DistancesTo(SelectedPoints(predicate,size),size)` compared to `Distances(predicate, size)`.

Flamefire: > So the Distances function was also creating a vector before 

Yeah, what I mean: If that creation of the vector turns out to be slow one would only need to change the `Distances` function. By passing in a vector directly you are loosing some information/flexibility there.

>  const unsigned maximumDistance = size.x * size.y;

Hm, it seems the only reason this is used is so the `UpdateDistances` knows which nodes to visit, correct? Because the maximum real distance would be `(size.x + size.y) / 4` or so. So if you only need a marker here, why not use `std::numeric_limits<unsigned>::max()` or `unsigned(-1)` (which IIRC is valid)?   
Oh, just seen you need `size` for creating the distanceMap too. Anyway it might be worth using a fixed constant here.

>  But generally, it seemed to be more readable this way: `DistancesTo(SelectPoints(predicate,size),size)` compared to `Distances(predicate, size)`.

I'm not sure. Duplicating logic is always itching me although I can see the "beauty" of this approach.
But the `size), size)` confuses me when reading it. Maybe just renaming the old function helps? E.g. `DistancesToMatching(predicate, size)`? As mentioned, this has the benefit of being able to change that function to not create a vector/use `SelectPoints` to directly creating the queue instead of copying the vector into the queue.

If you got another idea especially about the size issue, let me know. If not, we could leave this open for a couple days in case someone has a new idea. Currently I'd prefer calling `SelectPoints` inside a function and maybe replace this later

vader1986: Replaced the`maximumDistance` by using `unsigned(-1)` and added the `DistancesTo` function with evaluator again (this times just calls `SelectPoints` & the other overload of `DistancesTo`).

Flamefire: Thanks. CCing @Flow86 if he wants to chime in the discussion

But merging for now 👍

Flamefire (approved):

Issue #933, opened 2018-11-11
AI does (again) not build fisheries


Xermio:
Seems like this old bug is back: https://github.com/Return-To-The-Roots/s25client/issues/578

Tested on the map "X - Das letzte Tor".

With version 0.8.2: AI builds lots of fisheries.
With the current version: AI does not build a single fishery.

Tested with "Inexhaustible fish" enabled and disabled, makes no difference.

RM87:
Might that have something to do with it? Both large numbers(positive and negative)  do grow away from zero every 400GF)

![image](https://user-images.githubusercontent.com/859636/104109497-31177800-52d7-11eb-9d76-3f91b5fd2b47.png)

Flamefire:
I just checked that a bit: Functions involved are `FindBestPositionDiminishingResource` and `CalcResourceValue`

They seem to be calculating wrong values, i.e. there shouldn't be negative values.
Hidden in `FindBestPositionDiminishingResource` there seems to be a function which updates the resource map in the given radius.
To solve this (and greatly enhance the AI) one would need to extract that into smallish, testable functions and test them thoroughly. The code is so complex and confusing that I'm unable to tell if it does the right thing (it obviously doesn't judging from the results)

So a good, black-box testcase would be: Call the `UpdateResourceMapInRadius` function (extracted from above) for different points and radius values without changing the underlying data. The value per point should be equal no matter where the update was started from (and above zero). With the current implementation one could get the expected value per point with `CalcResourceValue(const MapPoint pt, AIResource res)`

However the problem seems to be much bigger: The resource maps seem to have been abused at some point so there is a mixture of functions (potentially) reading and writing to it with inconsistent value calculations.
So this needs to be checked and refactored first so that there is only 1 true, valid and tested usage.

@jhkl who originally wrote the code is gone AFAIK so its gonna be hard...

desofity:
So as I understood correctly AIResourceMap's is used to find best place for the building to get desired resource. In case of Fish it was in my opinion implemented wrong: in debug window numbers were only on the water. That's way even if the algorithm confirmed some place is suited for fishery, AI still can't build, because it is a water.

So in `void AIResourceMap::init()` I switched code
```
if(res == AIResource::Fish)
{
    isValid =
      aii.gwb.IsOfTerrain(pt, [](const TerrainDesc& desc) { return desc.kind == TerrainKind::Water; });
}
```
to
```
if(res == AIResource::Fish)
{
[...]

PR #1355, opened 2021-01-24
Refactor AI and teach it about unlimitted resources

Flamefire:
The current AI didn't know about inexhaustible granite mines and fish

I also started changing the resourceMaps array to an EnumArray which lead to a bunch of related changes by introducing separate enums for surface, subsurface, and node resources which makes the intention clearer:
- AIResource is what the AI cares about
- Surface and SubSurface resources are the ones from AIResource which are above/below the surface and includes `Nothing` (the enums aren't contiguous so can't use OptionalEnum)
- AINodeResource is all AIResources + a few special ones like "blocked" and "multiple". I kept this during refactoring but might remove later as the aiMap.res entries aren't updated so they become out of date fast. Might fix that later or remove

Most importantly I merged the resource rating calculation and position search into the resource maps and removed what was in there. It was almost the same but unused so now we have only a single algorithm and rating mechanism. Check the commit message for details

I know that the current resource rating calculation is broken, see #933. I'll fix that in a follow-up PR together with tests but I want to get this refactoring in first. It has no functional changes except for a few trivial bugfixes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,26 +8,6 @@ endif()
 
 project(s25client)
 
-if(NOT MSVC)
-    #Check for ccache
-    if(APPLE)
-        find_program(CCACHE_PROGRAM NAMES ccache-lipo ccache)
-    else()
-        find_program(CCACHE_PROGRAM ccache)
-    endif()
-    mark_as_advanced(CCACHE_PROGRAM)
-    if(CCACHE_PROGRAM)
-        foreach(lang C CXX)
-            # Do only set ccache compiler if we don't already use a ccache wrapper
-            # Example: /usr/lib/ccache/g++
-            if(NOT CMAKE_${lang}_COMPILER MATCHES "ccache")
-                message(STATUS "Using ccache(${CCACHE_PROGRAM}) for ${lang} to speed up builds")
-                set(CMAKE_${lang}_COMPILER_LAUNCHER "${CCACHE_PROGRAM}")
-            endif()
-        endforeach()
-    endif()
-endif()
-
 message(STATUS "Using CMake ${CMAKE_VERSION}")
 
 if(DEFINED CMAKE_TOOLCHAIN_FILE)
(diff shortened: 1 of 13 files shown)


CMakeLists.txt
```diff
@@ -5,26 +5,6 @@ endif()
 
 project(s25client)
 
-if(NOT MSVC)
-    #Check for ccache
-    if(APPLE)
-        find_program(CCACHE_PROGRAM NAMES ccache-lipo ccache)
```
Flow86, line 11: we should add "buildcache" to the list (i.e usable for VS builds)

libs/s25main/ai/aijh/AIPlayerJH.cpp
```diff
@@ -1105,14 +947,14 @@ MapPoint AIPlayerJH::FindPositionForBuildingAround(BuildingType type, const MapP
     {
         case BuildingType::Woodcutter:
         {
-            foundPos = FindBestPosition(around, AIResource::Wood, BUILDING_SIZE[type], 20, 11);
+            foundPos = FindBestPosition(around, AIResource::Wood, BUILDING_SIZE[type], 11, 20);
```
Flow86, line 950: can we name those magic constants somehow useful?

CMakeLists.txt
Flamefire: Got to be done in libutil now :)
However I'm not sure there is a benefit. First time I hear about buildcache, so who uses it? Also it seems it (max) disable parallel compiler invocation for MSVC: https://github.com/mbitsnbites/buildcache/blob/master/doc/usage.md#using-with-visual-studio--msbuild

libs/s25main/ai/aijh/AIPlayerJH.cpp
Flamefire: I have no idea where they are coming from and why. Probably experimenting.
However the "11" as radius is used often, so we could name this at least.

CMakeLists.txt
Flow86: Well yes, but its quiet nice to use (i.e I do use it at work ;-))

libs/s25main/ai/aijh/AIPlayerJH.cpp
Flow86: ok that would be great

Flamefire: Done

Flow86 (changes requested):
@Flamefire needs fixing (conflict)

Flow86 (approved):

PR #1356, opened 2021-01-24
HQ positions on random maps consider mountain distance

vader1986:
Hi guys, 

This is basically a new feature for random maps to consider the distance of HQs to mountains for HQ placement. That should generate more competitive maps with fairer HQs. Also added a configuration option to specify how fair away players should be positioned. Also, I hope the coed & comment changes make the workflow of `FindHqPositions` easier to follow!

image

Example map - mountain distance "close":
image

Example map - mountain distance "far":
image

Let me know if you have any suggestions for improvement!

### DIFF
--- a/libs/s25main/ingameWindows/iwMapGenerator.cpp
+++ b/libs/s25main/ingameWindows/iwMapGenerator.cpp
@@ -98,7 +98,7 @@ iwMapGenerator::iwMapGenerator(MapSettings& settings)
         combo->AddString(_(desc.get(DescIdx<LandscapeDesc>(i)).name));
 
     curPos.y += 30;
-    AddText(CTRL_TXT_MOUNTAIN_DIST, curPos, _("Mountain Distance"), COLOR_YELLOW, FontStyle{}, NormalFont);
+    AddText(CTRL_TXT_MOUNTAIN_DIST, curPos, _("HQ distance to mountain"), COLOR_YELLOW, FontStyle{}, NormalFont);
     curPos.y += 20;
     combo = AddComboBox(CTRL_MOUNTAIN_DIST, curPos, comboSize, TextureColor::Grey, NormalFont, 100);
     combo->AddString(_("Close"));
--- a/libs/s25main/ingameWindows/iwMapGenerator.cpp
+++ b/libs/s25main/ingameWindows/iwMapGenerator.cpp
@@ -71,8 +71,7 @@ iwMapGenerator::iwMapGenerator(MapSettings& settings)
     const Extent buttonSize(100, 20);
 
     curPos.y += 30;
-    ctrlComboBox* combo =
-      AddComboBox(CTRL_PLAYER_NUMBER, curPos, comboSize, TextureColor::Grey, NormalFont, 100);
+    ctrlComboBox* combo = AddComboBox(CTRL_PLAYER_NUMBER, curPos, comboSize, TextureColor::Grey, NormalFont, 100);
     for(unsigned n = 2; n <= MAX_PLAYERS; n++)
         combo->AddString(boost::str(boost::format(_("%1% players")) % n));
 
--- a/libs/s25main/ingameWindows/iwMapGenerator.cpp
+++ b/libs/s25main/ingameWindows/iwMapGenerator.cpp
@@ -64,49 +64,67 @@ iwMapGenerator::iwMapGenerator(MapSettings& settings)
         return;
     }
 
-    AddTextButton(CTRL_BTN_BACK, DrawPoint(20, 370), Extent(100, 20), TextureColor::Red1, _("Back"), NormalFont);
-    AddTextButton(CTRL_BTN_APPLY, DrawPoint(130, 370), Extent(100, 20), TextureColor::Green2, _("Apply"), NormalFont);
+    DrawPoint curPos(20, 0);
 
+    const Extent comboSize(210, 20);
+    const Extent progressSize(130, 20);
+    const Extent buttonSize(100, 20);
+
+    curPos.y += 30;
     ctrlComboBox* combo =
-      AddComboBox(CTRL_PLAYER_NUMBER, DrawPoint(20, 30), Extent(210, 20), TextureColor::Grey, NormalFont, 100);
+      AddComboBox(CTRL_PLAYER_NUMBER, curPos, comboSize, TextureColor::Grey, NormalFont, 100);
     for(unsigned n = 2; n <= MAX_PLAYERS; n++)
         combo->AddString(boost::str(boost::format(_("%1% players")) % n));
 
-    combo = AddComboBox(CTRL_MAP_STYLE, DrawPoint(20, 60), Extent(210, 20), TextureColor::Grey, NormalFont, 100);
+    curPos.y += 30;
+    combo = AddComboBox(CTRL_MAP_STYLE, curPos, comboSize, TextureColor::Grey, NormalFont, 100);
     combo->AddString(_("Water"));
     combo->AddString(_("Land"));
     combo->AddString(_("Mixed"));
 
-    combo = AddComboBox(CTRL_MAP_SIZE, DrawPoint(20, 90), Extent(210, 20), TextureColor::Grey, NormalFont, 100);
+    curPos.y += 30;
+    combo = AddComboBox(CTRL_MAP_SIZE, curPos, comboSize, TextureColor::Grey, NormalFont, 100);
     combo->AddString("64 x 64");
     combo->AddString("128 x 128");
     combo->AddString("256 x 256");
     combo->AddString("512 x 512");
     combo->AddString("1024 x 1024");
 
-    AddText(CTRL_TXT_LANDSCAPE, DrawPoint(20, 120), _("Landscape"), COLOR_YELLOW, FontStyle{}, NormalFont);
-    combo = AddComboBox(CTRL_MAP_TYPE, DrawPoint(20, 140), Extent(210, 20), TextureColor::Grey, NormalFont, 100);
+    curPos.y += 30;
+    AddText(CTRL_TXT_LANDSCAPE, curPos, _("Landscape"), COLOR_YELLOW, FontStyle{}, NormalFont);
+    curPos.y += 20;
+    combo = AddComboBox(CTRL_MAP_TYPE, curPos, comboSize, TextureColor::Grey, NormalFont, 100);
     for(unsigned i = 0; i < desc.landscapes.size(); i++)
         combo->AddString(_(desc.get(DescIdx<LandscapeDesc>(i)).name));
 
-    AddText(CTRL_TXT_MOUNTAIN_DIST, DrawPoint(20, 170), _("Mountain Distance"), COLOR_YELLOW, FontStyle{}, NormalFont);
-    combo = AddComboBox(CTRL_MOUNTAIN_DIST, DrawPoint(20, 190), Extent(210, 20), TextureColor::Grey, NormalFont, 100);
+    curPos.y += 30;
+    AddText(CTRL_TXT_MOUNTAIN_DIST, curPos, _("Mountain Distance"), COLOR_YELLOW, FontStyle{}, NormalFont);
+    curPos.y += 20;
+    combo = AddComboBox(CTRL_MOUNTAIN_DIST, curPos, comboSize, TextureColor::Grey, NormalFont, 100);
     combo->AddString(_("Close"));
     combo->AddString(_("Normal"));
     combo->AddString(_("Far"));
     combo->AddString(_("Very far"));
 
-    AddText(CTRL_TXT_GOAL, DrawPoint(20, 225), _("Gold:"), COLOR_YELLOW, FontStyle{}, NormalFont);
-    AddProgress(CTRL_RATIO_GOLD, DrawPoint(100, 220), Extent(130, 20), TextureColor::Grey, 139, 138, 100);
-    AddText(CTRL_TXT_IRON, DrawPoint(20, 255), _("Iron:"), COLOR_YELLOW, FontStyle{}, NormalFont);
-    AddProgress(CTRL_RATIO_IRON, DrawPoint(100, 250), Extent(130, 20), TextureColor::Grey, 139, 138, 100);
-    AddText(CTRL_TXT_COAL, DrawPoint(20, 285), _("Coal:"), COLOR_YELLOW, FontStyle{}, NormalFont);
-    AddProgress(CTRL_RATIO_COAL, DrawPoint(100, 280), Extent(130, 20), TextureColor::Grey, 139, 138, 100);
-    AddText(CTRL_TXT_GRANITE, DrawPoint(20, 315), _("Granite:"), COLOR_YELLOW, FontStyle{}, NormalFont);
-    AddProgress(CTRL_RATIO_GRANITE, DrawPoint(100, 310), Extent(130, 20), TextureColor::Grey, 139, 138, 100);
+    curPos.y += 35;
+    AddText(CTRL_TXT_GOAL, curPos, _("Gold:"), COLOR_YELLOW, FontStyle{}, NormalFont);
+    AddProgress(CTRL_RATIO_GOLD, DrawPoint(100, curPos.y - 5), progressSize, TextureColor::Grey, 139, 138, 100);
+    curPos.y += 30;
+    AddText(CTRL_TXT_IRON, curPos, _("Iron:"), COLOR_YELLOW, FontStyle{}, NormalFont);
+    AddProgress(CTRL_RATIO_IRON, DrawPoint(100, curPos.y - 5), progressSize, TextureColor::Grey, 139, 138, 100);
+    curPos.y += 30;
+    AddText(CTRL_TXT_COAL, curPos, _("Coal:"), COLOR_YELLOW, FontStyle{}, NormalFont);
+    AddProgress(CTRL_RATIO_COAL, DrawPoint(100, curPos.y - 5), progressSize, TextureColor::Grey, 139, 138, 100);
+    curPos.y += 30;
+    AddText(CTRL_TXT_GRANITE, curPos, _("Granite:"), COLOR_YELLOW, FontStyle{}, NormalFont);
+    AddProgress(CTRL_RATIO_GRANITE, DrawPoint(100, curPos.y - 5), progressSize, TextureColor::Grey, 139, 138, 100);
+    curPos.y += 30;
+    AddText(CTRL_TXT_RIVERS, curPos, _("Rivers:"), COLOR_YELLOW, FontStyle{}, NormalFont);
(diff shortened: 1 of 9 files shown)


vader1986:
@Flamefire let me know if the frequency of PRs is too high. I appreciate the reviews either way!

Flow86:
> @Flamefire let me know if the frequency of PRs is too high. I appreciate the reviews either way!

Thank you very much for your effort. Keep up coding :-) Even though you have a high frequency of turning in PRs, its definitly a lot of value to the game!

libs/s25main/mapGenerator/HeadQuarters.h
```diff
@@ -42,45 +43,71 @@ namespace rttr { namespace mapGenerator {
...
+        const auto mountain = [&map](const MapPoint& pt) { return map.textureMap.Any(pt, IsMinableMountain); };
+        const auto mountainDistance = DistancesTo(map.size, mountain);
+        const auto desiredDistance =
+          std::max(GetMinimum(mountainDistance, possiblePositions), playerDistanceToMountains);
+        const auto maxDistance = static_cast<unsigned>(map.size.x + map.size.y) / 4;
 
         std::vector<MapPoint> positions;
-        for(const MapPoint& pt : area)
+        auto allowedMountainDistance = desiredDistance;
+        while(positions.empty() && allowedMountainDistance < maxDistance)
```
Flamefire, line 86: I fear that this becomes quite expensive. For a 512x512 map this may run 256 times over possiblePositions

An idea I have: Sort possiblePositions by mountainDistance. Then you can determine the range in that sequence and move that, this way you basically go over possiblePositions (after the sort) at most once.

However thinking about this: Why does the range only increase to upper bound and moves away from lower? Is that intended? E.g. for `allowedMountainDistance=10` it will check points with distance in '(5, 15)` (exclusive), then '(6 , 16)`, ... If this is intended, then the above approach becomes even easier: If the first range did not find anything, you simply need to check if there are points left and if so take the next one and all with equal distance.

libs/s25main/mapGenerator/HeadQuarters.h
```diff
@@ -42,45 +43,71 @@ namespace rttr { namespace mapGenerator {
...
+        if(possiblePositions.empty())
+        {
+            return possiblePositions;
+        }
+
+        // Ensure HQ positions keep desired distance to mountains
+        const auto mountain = [&map](const MapPoint& pt) { return map.textureMap.Any(pt, IsMinableMountain); };
+        const auto mountainDistance = DistancesTo(map.size, mountain);
+        const auto desiredDistance =
+          std::max(GetMinimum(mountainDistance, possiblePositions), playerDistanceToMountains);
```
Flamefire, line 81: Not sure about this, what is the intention? Without this and a higher min mountainDistance it will select the points which are as close as allowed by playerDistanceToMountains. With this it may choose ones which are even further. Is that what is intended?

BTW: Please don't overuse `auto`. My rule of thumb: If the type is on the RHS or complicated, use `auto` but in general prefer explicit types. Especially with mountainDistance and desiredDistance next to each other I'd suggest to rename the former `mountainDistances` (then you don't need the type to know it's a collection) and showing the type of desiredDistance as `unsigned`(?). Reason: You even assign this to `auto allowedMountainDistance` and then subtract 5 from that. To see if this is valid or underflows an unsigned you need to go to allowedMountainDistance, then desiredDistance, then playerDistanceToMountains and find out what GetMinimum returns.
I really like `auto` but there are situations (like this) where it makes reasoning and review hard

Flamefire (changes requested):
> let me know if the frequency of PRs is too high.

As @Flow86 said ;)
Smaller PRs are good, even when there are many. Makes review faster. Also note that I'm usually very much up for discussion for anything I mention in the reviews, which IMO is what they are for.
Also don't be worried if it takes a while until I review, might be busy but will come back to it. Just keep on working on another branch/feature :)

libs/s25main/ingameWindows/iwMapGenerator.cpp
```diff
@@ -87,17 +89,24 @@ iwMapGenerator::iwMapGenerator(MapSettings& settings)
...
-    AddText(CTRL_TXT_IRON, DrawPoint(20, 205), _("Iron:"), COLOR_YELLOW, FontStyle{}, NormalFont);
-    AddProgress(CTRL_RATIO_IRON, DrawPoint(100, 200), Extent(130, 20), TextureColor::Grey, 139, 138, 100);
-    AddText(CTRL_TXT_COAL, DrawPoint(20, 235), _("Coal:"), COLOR_YELLOW, FontStyle{}, NormalFont);
-    AddProgress(CTRL_RATIO_COAL, DrawPoint(100, 230), Extent(130, 20), TextureColor::Grey, 139, 138, 100);
-    AddText(CTRL_TXT_GRANITE, DrawPoint(20, 265), _("Granite:"), COLOR_YELLOW, FontStyle{}, NormalFont);
-    AddProgress(CTRL_RATIO_GRANITE, DrawPoint(100, 260), Extent(130, 20), TextureColor::Grey, 139, 138, 100);
-
-    AddText(CTRL_TXT_RIVERS, DrawPoint(20, 295), _("Rivers:"), COLOR_YELLOW, FontStyle{}, NormalFont);
-    AddProgress(CTRL_RIVERS, DrawPoint(100, 290), Extent(130, 20), TextureColor::Grey, 139, 138, 100);
+    AddText(CTRL_TXT_MOUNTAIN_DIST, DrawPoint(20, 170), _("Mountain Distance"), COLOR_YELLOW, FontStyle{}, NormalFont);
```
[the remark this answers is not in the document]
vader1986: Yeah, I'll change it. I guess at some point I need to add a tooltip text (like for the add-ons) if the parameters become too complex. But `HQ distance to mountain` seems still self-explaining.

libs/s25main/mapGenerator/HeadQuarters.h
```diff
@@ -42,45 +43,71 @@ namespace rttr { namespace mapGenerator {
...
+        while(positions.empty() && allowedMountainDistance < maxDistance)
         {
-            if(obstacleDistance[pt] >= minDistance)
+            for(const MapPoint& pt : possiblePositions)
             {
-                positions.push_back(pt);
-            } else
-            {
-                potentialHqQuality[pt] = 0;
+                if(obstacleDistance[pt] >= minObstacleDistance && mountainDistance[pt] < allowedMountainDistance + 5
```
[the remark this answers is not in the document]
vader1986: Yupp, sorry forgot to remove it here.

libs/s25main/mapGenerator/HeadQuarters.h
vader1986: Absolutely right. I changed the whole thing a bit now to widen the search radius rather than changing the `allowedMountainDistance`. Also I didn't consider distance between multiple HQs at all, so I added a minimum distance as well.

libs/s25main/mapGenerator/HeadQuarters.h
vader1986: Updated the code as mentioned above. I had a quick look through `autos`, will keep an eye on it in the future as well.

libs/s25main/mapGenerator/HeadQuarters.h
```diff
@@ -42,45 +44,66 @@ namespace rttr { namespace mapGenerator {
...
+        // Keep minimum distance to other HQs and desired distance to mountains
+        const auto mountain = [&map](const MapPoint& pt) { return map.textureMap.Any(pt, IsMinableMountain); };
+        const auto mountainDistances = DistancesTo(map.size, mountain);
+        const auto desiredDistance = static_cast<unsigned>(distance);
+        const auto distanceToOtherHqs = DistancesTo(map.hqPositions, map.size);
+        const unsigned minimumHqDistance = (map.size.x + map.size.y) / 16;
+        const int maximumRadius = (map.size.x + map.size.y) / 4;
         std::vector<MapPoint> positions;
-        for(const MapPoint& pt : area)
+        for(int radius = 1; positions.empty() && radius < maximumRadius; radius *= 2)
```
vader1986, line 86: So I changed the logic here to widen the radius to look for points with the desired mountain distance. 
I increment the radius by `*2` to avoid potential performance issues on larger maps.

libs/s25main/mapGenerator/MapSettings.h
```diff
@@ -31,25 +31,30 @@ namespace rttr { namespace mapGenerator {
...
+    enum class MountainDistance : uint8_t
     {
-        MapSettings()
-            : numPlayers(2), size(MapExtent::all(128)), ratioGold(9), ratioIron(36), ratioCoal(40), ratioGranite(15),
-              rivers(15), type(0), style(MapStyle::Mixed)
-        {}
+        Close = 5,
+        Normal = 15,
+        Far = 25,
+        VeryFar = 30
```
vader1986, line 39: I'm not sure if it is obvious why I chose those numbers. In theory, you might think you would like this to depend on the map size. In practice, it can get really annoying if you play a 1024x1024 map and you're 200 nodes away from any mountain. You might never reach it. 

So I thought, even on larger maps it's still good to be in reach of mountains within a reasonable amount of time. The "reasonable amount of time" would be equal, independent of the map size.

libs/s25main/mapGenerator/HeadQuarters.h
```diff
@@ -42,45 +44,66 @@ namespace rttr { namespace mapGenerator {
...
         {
-            if(obstacleDistance[pt] >= minDistance)
-            {
-                positions.push_back(pt);
-            } else
+            for(const MapPoint& pt : possiblePositions)
             {
-                potentialHqQuality[pt] = 0;
+                if(std::abs(static_cast<int>(mountainDistances[pt] - desiredDistance)) < radius
+                   && distanceToOtherHqs[pt] > minimumHqDistance)
```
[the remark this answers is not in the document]
Flamefire: Makes sense. Then 2 new questions:

1. Is it intentional that the minimumHqDistance is ignored in the fallback below?
2. Would it make sense to remove those non-matching points first, before the radius-loop? If 1. is not intentional, then I'd certainly do so, otherwise it might still be a good idea but is not that crucial.

Also the cast is misplaced: If `desiredDistance > mountainDistances[pt]` then you get UB because the value is very large and hence not representable in int. Only the int->unsigned conversion is valid. I know this is fine in practice but I wouldn't want to introduce UB here.

And I still think the algorithm is not the right choice here, yes it works and is simple enough but the worst case behavior could be bad. See my previous idea about sorting `possiblePositions`. Seeing the current change I'd even go further: Sort by the difference (the `std::abs`-result) and then simply take the first points with the same radius. Sorting could also be done in a vector of `pair<MapPoint, distanceDifference>`.
Or, maybe still better: It is possible in a single pass:

- Start with `minDistanceDifference=<maxint>`
- Go exactly once over possiblePositions skipping invalid ones (minimumHqDistance)
- If that points distanceDifference is equal to minDistanceDifference add to possiblePoints
- if it is lower, clear possiblePoints, reset minDistanceDifference and add to possiblePoints

Because that's what you are trying here, aren't you: Selecting the points with the least difference of mountainDistances to desiredDistance

Then the only way that `positions` is empty afterwards is for the case when all points are less than minimumHqDistance (BTW: If it's a minimum I'd expect that this distance is valid, hence `distanceToOtherHqs[pt] >= minimumHqDistance`, oversight or intentional?)

To me this sounds good, what do you think? Am I missing anything or is there a advantage of the current implementation?

Flamefire (changes requested):

libs/s25main/mapGenerator/HeadQuarters.h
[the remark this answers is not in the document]
vader1986: I think you're right, I'll take a look at this probably tomorrow!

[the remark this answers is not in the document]
vader1986: Just updated the PR. To answer your question:
> Is it intentional that the minimumHqDistance is ignored in the fallback below?
Yes, that was intentional. In case we can't find any suitable position with mountains in range & minimum HQ distance, then we just take the position with most distance to other HQs.

Flamefire:
Ah, you changed the HQ placement decision again to something different. This becomes hard to review...
Could you describe (in very short), what the `FindHqPositions` should do? E.g. which positions it should or should not return (e.g. based on obstacles, other HQs, mountain distance) and how the ranking/order of the returned, valid positions should be determined.

Sorry for the bother, I'd usually want to avoid having to write "specifications", but there were multiple instances where I wasn't sure if something is intentional or not (hence my questions). So it would be easier to be able to talk about a "specification" (even just short and simple) and then compare the implementation against that.

vader1986:
> Ah, you changed the HQ placement decision again to something different. This becomes hard to review...
> Could you describe (in very short), what the `FindHqPositions` should do? E.g. which positions it should or should not return (e.g. based on obstacles, other HQs, mountain distance) and how the ranking/order of the returned, valid positions should be determined.
> 
> Sorry for the bother, I'd usually want to avoid having to write "specifications", but there were multiple instances where I wasn't sure if something is intentional or not (hence my questions). So it would be easier to be able to talk about a "specification" (even just short and simple) and then compare the implementation against that.

Tried to simplify the workflow in `FindHqPositions` and improve the function description in the code.

libs/s25main/mapGenerator/HeadQuarters.h
```diff
@@ -38,9 +38,14 @@ namespace rttr { namespace mapGenerator {
...
-     * for the entire map just leave the area empty. The resulting HQ positions are sorted by quality (highest quality
-     * first). Good HQ positions are positions which are far away from other HQs and in a widely buildable area.
+     * Finds suitable positions for a HQ in the specified area of the map. The resulting HQ positions are sorted by quality.
+     * To find suitable positions, this function does:
+     * 1. look for points within a buildable area of radius 2 (min. req. of HQ)
+     * 2. if no such point exists return empty
+     * 3. sort all those points by their distance to existing HQs
+     * 4. filter out points within minimum distance to existing HQs
+     * 5. if no remaining points just return result of 3.
+     * 6. sort remaining points by how close they're to desired mountain distance
```
Flamefire, line 48: Ok, prior this (point 6) was (good case only): 

- Choose points which closest match desired mountain distance
- Sort by distance to other HQs (descending)

Correct? Asking for comparison and also as the `minObstacleDistance` is gone which (I think) usually placed the HQs in radius 4 (not 2), but not sure if that mattered in practice.
Anyway the new algorithm is good too, just 2 issues:

1. The sorting in point 3 (distance to other HQs) now only matters for the fallback point 5 (min distance to other HQs not possible), so it should be moved into that to avoid the work
1. I expect that there are many points which are "same close" to desired mountain distance, correct?
Would it make sense to sort those by HQ distance? Basically similar to what it was before, where quality was always related to HQ distance. For this you can either include the HQ distance as a tie breaker in the sort functor or do the sorting in point 3 as you currently do and use `std::stable_sort` (which allocates memory)

Minor notes for the future:
- I would not _require_ to repeat the points in the code, but it's nice to have.
- Point 2 could be removed from the description without changing the others, I'd treat that as an optimization because the sorting and filtering also works for "no points". Makes the description shorter (only reason I mention that)

Flamefire (changes requested):
> Tried to simplify the workflow in FindHqPositions and improve the function description in the code.

Thanks, much appreciated! :+1:

BTW: Judging from the formatting commits I assume you don't use IDE formatting. Which IDE do you use? Most allow to clang-format the code on every save. Or `make clangFormat` (or what it is on your build system) do automatically do this for all changed files (only, just the first run will do all and it requires CMake to have found clang-format)

libs/s25main/mapGenerator/HeadQuarters.h
vader1986: > Ok, prior this (point 6) was (good case only):
> Choose points which closest match desired mountain distance
> Sort by distance to other HQs (descending)
> Correct? 

Yes, that's right. I removed the whole thing around `minObstacleDistance` which happened to be a left-over before this logic changed. 2 is the clamped minimum anyway, so that wouldn't change much. 

> The sorting in point 3 (distance to other HQs) now only matters for the fallback point 5 (min distance to other HQs not possible), so it should be moved into that to avoid the work

I assumed that would keep the order in case objects have the same value when sorting for mountain distance afterwards. Obviously, I've been wrong, thanks for mentioning `std::stable_sort`, I think that's the way to go in this case. 

> I would not require to repeat the points in the code, but it's nice to have.

I felt like those points would help a dev looking through the code. Probably better if the code reflects those points directly, rather than repeating the comments. 

> Point 2 could be removed from the description without changing the others, I'd treat that as an optimization because the sorting and filtering also works for "no points". Makes the description shorter (only reason I mention that)

That was intended as an optimization to avoid calculating `distanceToOtherHqs` if that makes sense. Happy to remove it from the description anyway as it's kind of obvious.

vader1986:
> BTW: Judging from the formatting commits I assume you don't use IDE formatting. Which IDE do you use? Most allow to clang-format the code on every save. Or `make clangFormat` (or what it is on your build system) do automatically do this for all changed files (only, just the first run will do all and it requires CMake to have found clang-format)

Couldn't find a way to make clangFormat work for me in Xcode. Probably worth giving it another try, just got used to those extra commits.

Flamefire (approved):
> I assumed that would keep the order in case objects have the same value when sorting for mountain distance afterwards.

The problem is that keeping the order is (way) less efficient hence that's not done (by default)

> I felt like those points would help a dev looking through the code. Probably better if the code reflects those points directly, rather than repeating the comments.

Agree with both points ;) Just wanted to save you some work in the future in case you felt obliged to add those.

> That was intended as an optimization to avoid calculating distanceToOtherHqs if that makes sense.

That's what I meant: It is an optimization. The calculation of distanceToOtherHqs doesn't appear in the description so that doesn't need to be mentioned too. The description is rather meant as a specification of the input->output relation not a description of how you do that. Ideally it could be turned into unit tests, but at least it should give an idea what to expect for input-output and be a base for discussing what should be done and compare to what is done (e.g. the missing description that final quality is distanceDifference, then HQDistance which lead to the stable_sort question)

> Couldn't find a way to make clangFormat work for me in Xcode. Probably worth giving it another try, just got used to those extra commits.

I see, that seems to be difficult. Xcode has "projects" or "targets" too, doesn't it? If CMake found clang-format (or you point it to it) you should have a target there which you can "build". Might be the easiest, at least that's what I do in VS: Before a build do a clangFormat build, or at least when all is green.

Anyway, looks good to me now. Thanks for your efforts!

vader1986:
I used to have a clangformat target in Xcode, but at some point it disappeared (probably because cmake can't find clang-format, or at least not the right version). Was just too lazy to investigate it so far.

Flamefire:
That's the reason. We currently use clang-format-10, so you'd need to install that too. At least the binaries are versioned so that should work.
Reason is that the formatting varies  with the major version so we have to choose one. 10 should be widely available, 11 not so much yet

PR #1357, opened 2021-01-26
Improve island generation on random water maps

vader1986:
**Current state**
The center of each player island is calculated by finding the `MapPoint` with the greatest distance to any land. Therefore, the first island depends only on the distance to the center island (which is not populated by any HQ). There are usually multiple points with equal distance to the center island. Therefore the algorithm chooses the first one starting from `x=0/y=0` and iterating through the points with the maximum distance like a standard iteration over the map (`x=0,1,2,...width-1` etc.). That usually leads to the first island (blue player) being placed in the top-right corner of the map. Other players' islands are similar predictable. 

**Suggested improvement**
Choose a minimum distance of every island-center to any land. Then randomly choose one of the `MapPoints` keeping this distance. If there's no such point available just fall back to using the point with maximum distance to land. 

As a separate (fairly Independent) change, I decided to make player island for larger maps a bit smaller. On tiny maps each player should have a reasonably sized island to have at list a tiny mountain available on the home island. On larger maps, currently the islands become huge that it becomes even challenging to find a harbor position to get away from the home island. Therefore, I think it makes sense to decrease the size a little bit. Maybe instead of the current approach I could just add an upper boundary for home islands, happy to discuss!

### DIFF
--- a/libs/s25main/mapGenerator/Islands.cpp
+++ b/libs/s25main/mapGenerator/Islands.cpp
@@ -28,16 +28,12 @@ namespace rttr { namespace mapGenerator {
 
         const auto land = [&map](const MapPoint& pt) { return map.z[pt] > map.height.minimum; };
         const auto distances = DistancesTo(map.size, land);
-
-        const auto findSuitableIslandCenter = [&rnd, &distances, distanceToLand]() {
-            const auto possibleCenters = SelectPoints(
-              [&distances, distanceToLand](const MapPoint& pt) {
-                  return distances[pt] > distanceToLand * 10; // keep proper distance to land
-              },
-              distances.GetSize());
-            return possibleCenters.empty() ? GetMaximumPoint(distances) : rnd.RandomItem(possibleCenters);
-        };
-        const MapPoint center = findSuitableIslandCenter();
+        const unsigned maxDistance = *std::max_element(distances.begin(), distances.end());
+        const unsigned minDistance = std::min(maxDistance, distanceToLand * 10);
+        const auto possibleCenters = SelectPoints([&distances, minDistance](const MapPoint& pt) {
+            return distances[pt] >= minDistance;
+        }, distances.GetSize());
+        const MapPoint center = rnd.RandomItem(possibleCenters);
 
         const auto compare = [&distances, &center](const MapPoint& rhs, const MapPoint& lhs) {
             // computes prefered extension points for the island by considering distance to
--- a/libs/s25main/mapGenerator/Islands.cpp
+++ b/libs/s25main/mapGenerator/Islands.cpp
@@ -30,9 +30,8 @@ namespace rttr { namespace mapGenerator {
         const auto distances = DistancesTo(map.size, land);
         const unsigned maxDistance = *std::max_element(distances.begin(), distances.end());
         const unsigned minDistance = std::min(maxDistance, distanceToLand * 10);
-        const auto possibleCenters = SelectPoints([&distances, minDistance](const MapPoint& pt) {
-            return distances[pt] >= minDistance;
-        }, distances.GetSize());
+        const auto possibleCenters = SelectPoints(
+          [&distances, minDistance](const MapPoint& pt) { return distances[pt] >= minDistance; }, distances.GetSize());
         const MapPoint center = rnd.RandomItem(possibleCenters);
 
         const auto compare = [&distances, &center](const MapPoint& rhs, const MapPoint& lhs) {
--- a/libs/s25main/mapGenerator/Islands.cpp
+++ b/libs/s25main/mapGenerator/Islands.cpp
@@ -27,10 +27,13 @@ namespace rttr { namespace mapGenerator {
         Island island;
 
         const auto isLand = [&map](const MapPoint& pt) { return map.z[pt] > map.height.minimum; };
-        const auto distances = DistancesTo(map.size, isLand);
-        const auto center = GetMaximumPoint(distances);
-
-        auto compare = [&distances, &center](const MapPoint& rhs, const MapPoint& lhs) {
+        const auto distances = DistancesTo(SelectPoints(isLand, map.size), map.size);
+        const auto possibleCenters = SelectPoints([&distances, distanceToLand](const MapPoint& pt) {
+            return distances[pt] > distanceToLand * 10;
+        }, map.size);
+        const auto center = possibleCenters.empty() ? GetMaximumPoint(distances) : rnd.RandomItem(possibleCenters);
+        
+        const auto compare = [&distances, &center](const MapPoint& rhs, const MapPoint& lhs) {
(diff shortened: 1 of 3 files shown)


Flamefire:
This change seems complex, so a question first (haven't looked at the code yet and I think it is easiest to talk about the goal first, so no need to change any code just yet)

> There are usually multiple points with equal [maximum] distance to the center island [...] -->
> Choose a minimum distance of every island-center to any land. Then randomly choose one of the MapPoints keeping this distance. If there's no such point available just fall back to using the point with maximum distance to land.

So the problem was, that there is basically a "ring" (using this as an example with the simplification that there was a single, perfectly round center island) of equally "good" points and the current algorithm chooses always the first. So obvious solution would be to choose a random point of those.
But seems that islands being placed on a "ring", even though randomly, isn't random enough. Correct so far?
So instead you changed the single-point-wide ring, on which islands can be placed, to a 0-n points wide ring by "expanding" the ring to the middle. Example in 2D: `xOOIIOOx` -> `xxOIIOxx` (I=Island, O=impossible, x=possible) Also correct?

The new fallback is basically what the current algorithm did, right? I'm a bit concerned about the "the point with maximum distance", as the current problem was, that there were multiple such points, so I guess you'd want a random one of such points here too, to avoid this, don't you?
Given that, I think it makes sense to clamp the minimum distance (you didn't say how you choose that, so I'd assume a fixed value?) to the "maximum distance to land" that would be used in the fallback. This way you don't need the fallback at all which should simplify the algorithm (description) and the code.

> Maybe instead of the current approach I could just add an upper boundary for home islands, happy to discuss!

So the size was (and still is) proportional to the map size? I'd say this depends on what the map should look like. Is it that each player has an island and there is a center island and that's it? Then it makes sense the way it is. Or should there be many islands where players can hop around? Then there should be a bounded size (upper and lower). Lower boundary would be something fixed so it doesn't get too small, but might increase with map size (e.g. on a very large maps, very small (start) islands don't make much sense). The upper bound follows the same logic: On small maps the upper bound mustn't be too big or the islands won't fit or be too close to each other. On larger maps they shouldn't grow to large either or you won't have many islands (this is what you changed here, IIUC)
I could imagine a user-adjustable size and/or amount of islands. Probably only 1 as they depend on each other and only as a course adjustment, i.e. most control is still on the generator. E.g. the generator determines a upper and lower bound, divides that in 3 equal(ish) parts and the user can choose which of those 3 parts to choose. It might as well be (in the extreme case) that all 3 parts are the same (e.g. upperbound==lowerbound+-2)

vader1986:
Fair questions, I generally wonder if it makes sense to create a GitHub issue for new features instead of a PR (which is rather draft in this case anyway). Just for the future I mean

Flamefire:
It depends, as usual. ;) 
I do both: Discuss changes in an issue first, so the general idea is clear, then create a PR implementing that. Or start implementing something clear (or when the idea is vague) to verify it works, then discuss in the PR. Sometimes it is even easier to discuss code instead of an idea. Best example is the AI stuff I'm currently doing. Lot's of refactoring, removing unused/old stuff, changing and fixing what is there. There are some functional changes here but the existing code is so complex and "difficult" that I wouldn't even know what to discuss first as I didn't understand half of it ;)
Also the enum stuff I did: It was a new feature (for the engineering side) but it was straight forward and details could easily be changed.

In the end it is your choice and your time. If it is fast to implement/change then likely a summary and PR is fine, in the worst case that you missed something big and have to revert not much work is lost but at least you can experiment with it already.

BTW: Just to be sure: When asking questions I (usually) don't want to convince you of a way, just make sure that everything was thought of and is as intended. If I go overboard with perfectionism, let me know ;)

vader1986:
> So the problem was, that there is basically a "ring" (using this as an example with the simplification that there was a single, perfectly round center island) of equally "good" points and the current algorithm chooses always the first. So obvious solution would be to choose a random point of those.
> But seems that islands being placed on a "ring", even though randomly, isn't random enough. Correct so far?

Yes, that is correct, even though it would actually be more fair for each player to have the same/similar distance to the center island. 

> So instead you changed the single-point-wide ring, on which islands can be placed, to a 0-n points wide ring by "expanding" the ring to the middle. Example in 2D: xOOIIOOx -> xxOIIOxx (I=Island, O=impossible, x=possible) Also correct?

Yes, that would be the intention, even though the implementation doesn't fully match it. The current implementation would rather be:
1. calculate distance of each map point to land
2. filter for points with a minimum distance to land
3. choose random point of the filtered points

Saying that, your description makes more sense to me and wouldn't require any "fallback". By the way, the "minimum distance" is currently: 
`minimum distance = island radius * 10` with `island radius` being a map size dependent value. More accurately, island radius is some sort of "brush size" used a "paint" the island. That's probably another idea to re-think (but not part of this PR at the moment).

> The new fallback is basically what the current algorithm did, right? I'm a bit concerned about the "the point with maximum distance", as the current problem was, that there were multiple such points, so I guess you'd want a random one of such points here too, to avoid this, don't you?

The fallback case would only come into play if the "minimum distance" from step 2 (see above) is larger than the maximum distance to land. Which really doesn't make much sense, thinking about it. 

I'll respond to the "island size" stuff later. 

Actually, after looking through the whole island code again, would you mind if I close this PR and open a GitHub issue instead? Then we could probably discuss generally what a "good" water map would look like. Then, based on that I could design an algorithm rather than trying to fix stuff within the function.

Flamefire:
> even though the implementation doesn't fully match it. The current implementation would rather be:

Not sure if I missed anything but I think we are talking about the same, so all good.

> would you mind if I close this PR and open a GitHub issue instead? 

That is up to you. I think it already is an improvement as you state in the OP. So only change required would be that in the calculation of `possibleCenters` the `10` is not hardcoded but determined as `min(distanceToLand * 10, *max_element(distances...))`
If you change the comparison to `>=` instead of `>` this is guaranteed to find at least one point, in the worst case this will be `GetMaximumPoint`, but usually the same as currently or at least a random max point. You also wouldn't need the lambda for that, I'm not sure how good the optimizers are and if they can optimize through all of them....

If you want to describe everything in an issue first, that's fine by me too. I can check that for issues or suggest minor improvements but I likely won't have enough time to help designing it in the first place.

vader1986:
Ah okay, it is some improvement yes. I'll implement your suggested changes, but remove the changes of the island size for now (because they're breaking harbor placement). 

Then, I would like to create a GitHub issue to discuss further improvements. I think it makes sense to rethink a few things about water maps, like:
* what could possible maps look like? Currently it will generate only one center island + "home" island for each player. It would not be difficult to add more variety, but it should be all consistent in the end.
* also the harbor placement is not ideal. Right now it will only place harbors on islands of a "minimum size" which is depends on the map size. So on a 1024x1024 map a lot of islands which are in theory suitable to populate are not reachable.
* ... and a few more things ...

Flamefire (approved):

PR #1359, opened 2021-01-27
Configurable stone piles and trees for random maps

vader1986:
Should be a fairly simple change, just making some "magic constants" configurable in the UI. Couldn't really see any drawback for this.

### DIFF
--- a/libs/s25main/mapGenerator/Resources.cpp
+++ b/libs/s25main/mapGenerator/Resources.cpp
@@ -58,7 +58,7 @@ namespace rttr { namespace mapGenerator {
         throw std::invalid_argument("invalid landscape type");
     }
 
-    void AddObjects(Map& map, RandomUtility& rnd)
+    void AddObjects(Map& map, RandomUtility& rnd, const MapSettings& settings)
     {
         std::set<MapPoint, MapPointLess> excludedArea;
         NodeMapBase<unsigned> probabilities;
@@ -112,9 +112,9 @@ namespace rttr { namespace mapGenerator {
             }
         }
 
-        auto range = GetRange(waterDistance);
-        auto probRange = ValueRange<unsigned>(15, 40);
-        auto probDiff = probRange.GetDifference();
+        const auto range = GetRange(waterDistance);
+        const auto probRange = ValueRange<unsigned>(settings.trees / 2, settings.trees);
+        const unsigned probDiff = probRange.GetDifference();
 
         const auto mountainFoot = [&map](const MapPoint& pt) {
             return !map.textureMap.All(pt, IsMountainOrSnowOrLava);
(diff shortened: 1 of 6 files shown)


libs/s25main/mapGenerator/Resources.cpp
```diff
@@ -164,9 +164,7 @@ namespace rttr { namespace mapGenerator {
                 map.objectTypes[pt] = tree.index + rnd.RandomValue(0, 7);
             }
 
-            auto graniteProb = probabilities[pt] / 4;
-
-            if(graniteProb > 0 && rnd.ByChance(graniteProb))
+            if(probabilities[pt] > 0 && rnd.ByChance(settings.stonePiles))
```
Flamefire, line 167: 2 questions here: this code may replace a tree that was placed above. Could you change it so that either-or is placed? To keep the current logic this means moving the granite-if before the tree-if and use an else. Simple enough but reads better to have only a single assign to `map.object*`
But main question: Now that the stone probability is mostly independent from `probabilities`, doesn't this mean that stone gets placed almost randomly (especially in small instances instead of chunks)? 
But I'm also concerned about the `probabilities[pt] > 0` (which could reuse `treeProb` from above): If the tree-slider is set to zero then, if I understand the generation of `probabilities` correctly, you will have all zeroes there. So no stone either. IMO that's a (new) bug, because prior the granite and tree prob wasn't configurable. So I'd say for the new setting you need another loop/prop calculation, which is independent from the tree probability. Or am I missing anything?

vader1986: >  this code may replace a tree that was placed above. Could you change it so that either-or is placed? To keep the current logic this means moving the granite-if before the tree-if and use an else.

Just updated that (only using continue instead of if/else). 

> But main question: Now that the stone probability is mostly independent from probabilities, doesn't this mean that stone gets placed almost randomly (especially in small instances instead of chunks)?

Yes, that is right. But granite piles wouldn't really depend on the same parameters as trees. Ideally that would be humidity of the terrain, which is currently implied through the distance to mountain & water. 

I think it makes sense to separate the logic for stone pile generation. And long term I can come up with a more sensible logic (than poor randomness) where to place which amount of stone piles. 

> But I'm also concerned about the probabilities[pt] > 0 (which could reuse treeProb from above): ...

Yes, that's a bug.

Flamefire (approved):
I'm generally fine with the change. Have you checked a few maps how stones are placed now? I.e. it makes more sense to have clusters of stone in a few places than a single piece of stone every few nodes, which would look odd and if it becomes to much there won't be space left to build etc.
Of course this can be improved later so no rush, but I'd like to make sure that maps don't become more or less unplayable by this. Would be sad for the map gen.
Also is the stone probability bounded? Or does 0 mean no stone and 100 mean every possible node is stone (and hence no more wood, which might even be ok, nice challenge ;) )?

vader1986:
> Also is the stone probability bounded? Or does 0 mean no stone and 100 mean every possible node is stone (and hence no more wood, which might even be ok, nice challenge ;) )?

100% stone automatically mean 0% trees. 

The distribution of stone piles hasn't changed much (tested it on many maps). Even before those changes there were rarely any clusters of stone. I could actually change it to do something like taking a (small) random radius around a point which has been chosen (`rnd.ByChance(settings.stonePiles)`) for stone and populate it with stone piles. But I hoped I find some time to think about a better approach than that.

Issue #1358, opened 2021-01-27
current git fails to compile 


stefson:
I just ran into this, x86_64 linux with gcc-9: 

```

/var/tmp/portage/games-strategy/s25rttr-9999-r5/work/s25rttr-9999/libs/s25main/desktops/dskSelectMap.cpp:71:108: error: noexcept-expression evaluates to ‘false’ because of a call to ‘static constexpr Point<T> Point<T>::all(const T&) [with T = short unsigned int]’ [-Werror=noexcept]
   71 | p015", 0)), csi(std::move(csi)), mapGenThread(nullptr), waitWnd(nullptr)
      |                                                                        ^

In file included from /var/tmp/portage/games-strategy/s25rttr-9999-r5/work/s25rttr-9999/libs/s25main/DrawPoint.h:20,
                 from /var/tmp/portage/games-strategy/s25rttr-9999-r5/work/s25rttr-9999/libs/s25main/Window.h:20,
                 from /var/tmp/portage/games-strategy/s25rttr-9999-r5/work/s25rttr-9999/libs/s25main/desktops/Desktop.h:20,
                 from /var/tmp/portage/games-strategy/s25rttr-9999-r5/work/s25rttr-9999/libs/s25main/desktops/dskSelectMap.h:20,
                 from /var/tmp/portage/games-strategy/s25rttr-9999-r5/work/s25rttr-9999/libs/s25main/desktops/dskSelectMap.cpp:18:
/var/tmp/portage/games-strategy/s25rttr-9999-r5/work/s25rttr-9999/libs/common/include/Point.h:67:20: error: but ‘static constexpr Point<T> Point<T>::all(const T&) [with T = short unsigned int]’ does not throw; perhaps it should be declared ‘noexcept’ [-Werror=noexcept]
   67 | constexpr Point<T> Point<T>::all(const T& val)
      |                    ^~~~~~~~
/var/tmp/portage/games-strategy/s25rttr-9999-r5/work/s25rttr-9999/libs/s25main/desktops/dskSelectMap.cpp:71:108: error: noexcept-expression evaluates to ‘false’ because of a call to ‘constexpr DescIdx<T>::DescIdx(uint8_t) [with T = LandscapeDesc; uint8_t = unsigned char]’ [-Werror=noexcept]
   71 | p015", 0)), csi(std::move(csi)), mapGenThread(nullptr), waitWnd(nullptr)
      |                                                                        ^

[...]

PR #1361, opened 2021-01-28
Add noexcept specifications to Point and DescIdx and modernize Point.h 

Flamefire:
I reduced the repetition of the Point class by using the preprocessor to stamp out the arithmetic functions. Adding of `noexcept` should also help and avoid the compiler warning from #1358

I also validated the codegen on godbolt for GCC, Clang and MSVC to be optimal (as-if handwritten)

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -17,6 +17,8 @@ jobs:
           - { compiler: clang-7,  os: ubuntu-18.04, buildType: Release }
           - { compiler: clang-7,  os: ubuntu-18.04, buildType: Debug }
           - { compiler: clang  ,  os: macos-10.15,  buildType: Debug, boostVersion: 1.74.0 } # Multiple bugs with recent OSX until 1.74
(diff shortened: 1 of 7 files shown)


.github/workflows/unit-tests.yml
```diff
@@ -17,6 +17,8 @@ jobs:
           - { compiler: clang-7,  os: ubuntu-18.04, buildType: Release }
           - { compiler: clang-7,  os: ubuntu-18.04, buildType: Debug }
           - { compiler: clang  ,  os: macos-10.15,  buildType: Debug, boostVersion: 1.74.0 } # Multiple bugs with recent OSX until 1.74
+          # GCC 9 is know to show a few warnings that GCC 10 has "fixed", make sure this doesn't happen for us
```
vader1986, line 20: `know`->`known` :D

Flamefire: Thanks! :)

Flow86 (approved):

Flow86 (approved):

PR #1362, opened 2021-02-01
Random map harbor placement

vader1986:
* little bit of refactoring of harbor positioning
* choose random center for water & mixed maps

Let me know what you think, maybe there's more to improve around harbor positioning.

### DIFF
--- a/libs/s25main/mapGenerator/Algorithms.h
+++ b/libs/s25main/mapGenerator/Algorithms.h
@@ -34,7 +34,7 @@ namespace rttr { namespace mapGenerator {
         std::vector<E> joined;
         for(const auto& el : container)
         {
-            std::copy(el.begin(), el.end(), std::back_inserter(joined));
+            joined.insert(joined.end(), el.begin(), el.end());
         }
         return joined;
     }
--- a/libs/s25main/mapGenerator/Algorithms.h
+++ b/libs/s25main/mapGenerator/Algorithms.h
@@ -29,7 +29,8 @@
 namespace rttr { namespace mapGenerator {
 
     template<typename T>
-    auto join(const T& container){
+    auto join(const T& container)
+    {
         using E = typename T::value_type::value_type;
         std::vector<E> joined;
         for(const auto& el : container)
--- a/libs/s25main/mapGenerator/Algorithms.h
+++ b/libs/s25main/mapGenerator/Algorithms.h
@@ -28,9 +28,9 @@
 
 namespace rttr { namespace mapGenerator {
 
-    template<template<typename...> class C1, typename... T1, template<typename...> class C2, typename... T2, typename E>
-    std::vector<E> join(const C1<C2<E, T2...>, T1...>& container)
-    {
+    template<typename T>
+    auto join(const T& container){
+        using E = typename T::value_type::value_type;
         std::vector<E> joined;
         for(const auto& el : container)
         {
--- a/libs/s25main/mapGenerator/Algorithms.h
+++ b/libs/s25main/mapGenerator/Algorithms.h
@@ -28,6 +28,17 @@
 
 namespace rttr { namespace mapGenerator {
 
+    template<template<typename...> class C1, typename... T1, template<typename...> class C2, typename... T2, typename E>
+    std::vector<E> join(const C1<C2<E, T2...>, T1...>& container)
+    {
+        std::vector<E> joined;
+        for(const auto& el : container)
+        {
+            std::copy(el.begin(), el.end(), std::back_inserter(joined));
+        }
+        return joined;
+    }
+
     /**
      * Smoothes the specified nodes with a smoothing kernel of the specified extent (radius).
      *
--- a/libs/s25main/mapGenerator/Harbors.cpp
+++ b/libs/s25main/mapGenerator/Harbors.cpp
@@ -69,7 +69,9 @@ namespace rttr { namespace mapGenerator {
         std::set<MapPoint, MapPointLess> visited;
 
         const auto isCoast = [&map](const MapPoint& pt) {
-            return map.textureMap.Any(pt, IsLand) && map.textureMap.Any(pt, IsWater);
+            const auto allWater = [&map](const MapPoint& p) { return map.textureMap.All(p, IsWater); };
+            return map.textureMap.Any(pt, IsLand) && map.textureMap.Any(pt, IsWater) &&
+              helpers::contains_if(map.textures.GetNeighbours(pt), allWater);
         };
 
         RTTR_FOREACH_PT(MapPoint, map.size)
--- a/libs/s25main/mapGenerator/Harbors.cpp
+++ b/libs/s25main/mapGenerator/Harbors.cpp
@@ -90,7 +90,7 @@ namespace rttr { namespace mapGenerator {
 
         for(std::vector<MapPoint>& coast : coasts)
         {
-            coast.erase(std::remove_if(coast.begin(), coast.end(), closeToRiver), coast.end());
+            helpers::remove_if(coast, closeToRiver);
         }
 
         return coasts;
@@ -100,7 +100,7 @@ namespace rttr { namespace mapGenerator {
     {
         const auto coasts = FindCoastlines(map, rivers);
         std::vector<MapPoint> harbors;
-        for(const auto& coast : coasts)
+        for(const std::vector<MapPoint>& coast : coasts)
         {
             if(coast.size() < static_cast<unsigned>(coastSize))
             {
--- a/libs/s25main/mapGenerator/Harbors.cpp
+++ b/libs/s25main/mapGenerator/Harbors.cpp
@@ -70,8 +70,8 @@ namespace rttr { namespace mapGenerator {
 
         const auto isCoast = [&map](const MapPoint& pt) {
             const auto allWater = [&map](const MapPoint& p) { return map.textureMap.All(p, IsWater); };
-            return map.textureMap.Any(pt, IsLand) && map.textureMap.Any(pt, IsWater) &&
-              helpers::contains_if(map.textures.GetNeighbours(pt), allWater);
+            return map.textureMap.Any(pt, IsLand) && map.textureMap.Any(pt, IsWater)
+                   && helpers::contains_if(map.textures.GetNeighbours(pt), allWater);
         };
 
         RTTR_FOREACH_PT(MapPoint, map.size)
--- a/libs/s25main/mapGenerator/Harbors.cpp
+++ b/libs/s25main/mapGenerator/Harbors.cpp
@@ -68,16 +68,8 @@ namespace rttr { namespace mapGenerator {
         std::vector<std::vector<MapPoint>> coasts;
         std::set<MapPoint, MapPointLess> visited;
 
-        const auto partOfRiver = [&rivers](const MapPoint& pt) {
-            return helpers::contains_if(rivers, [&pt](const River& river) { return river.find(pt) != river.end(); });
-        };
-
-        const auto distanceToRiver = DistancesTo(map.size, partOfRiver);
-        const auto allWater = [&map](const MapPoint& pt) { return map.textureMap.All(pt, IsWater); };
-
-        const auto isCoast = [&map, allWater, &distanceToRiver](const MapPoint& pt) {
-            return map.textureMap.Any(pt, IsLand) && map.textureMap.Any(pt, IsWater)
-                   && helpers::contains_if(map.textures.GetNeighbours(pt), allWater) && distanceToRiver[pt] >= 5;
+        const auto isCoast = [&map](const MapPoint& pt) {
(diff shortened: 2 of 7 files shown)


libs/s25main/mapGenerator/Algorithms.h
```diff
@@ -28,6 +28,17 @@
 
 namespace rttr { namespace mapGenerator {
 
+    template<template<typename...> class C1, typename... T1, template<typename...> class C2, typename... T2, typename E>
+    std::vector<E> join(const C1<C2<E, T2...>, T1...>& container)
```
vader1986, line 32: There might be a better place for crazy template functions.

libs/s25main/mapGenerator/Harbors.h
```diff
@@ -48,6 +48,6 @@ namespace rttr { namespace mapGenerator {
      * @param coastSize minimum number of nodes a coastline requires to be considered for harbor positions
      * @param nodesPerHarbor number of coast nodes required per harbor position
      */
-    void PlaceHarbors(Map& map, const std::vector<River>& rivers, int coastSize, int nodesPerHarbor = 50);
+    void PlaceHarbors(Map& map, const std::vector<River>& rivers, int coastSize = 32, int nodesPerHarbor = 64);
```
vader1986, line 51: The `coastSize` is the minimum number of nodes a coast has to contain to be considered for harbor placement. `nodesPerHarbor` defines the number of nodes for a coast: 
`#harbors = max(1, nodes(coast) / nodesPerHarbor)`

libs/s25main/mapGenerator/Algorithms.h
Flamefire: This is indeed a bit crazy... Let's see, you have a container of containers here and need the type of the innermost element. So:
```
template<typename T>
auto join(const T& container){
  using E = typename T::value_type::value_type;
```

Is this function performance relevant? If so a sum of the sizes passed to reserve would help.

libs/s25main/mapGenerator/Harbors.cpp
```diff
@@ -68,16 +68,8 @@ namespace rttr { namespace mapGenerator {
...
-        };
-
-        const auto distanceToRiver = DistancesTo(map.size, partOfRiver);
-        const auto allWater = [&map](const MapPoint& pt) { return map.textureMap.All(pt, IsWater); };
-
-        const auto isCoast = [&map, allWater, &distanceToRiver](const MapPoint& pt) {
-            return map.textureMap.Any(pt, IsLand) && map.textureMap.Any(pt, IsWater)
-                   && helpers::contains_if(map.textures.GetNeighbours(pt), allWater) && distanceToRiver[pt] >= 5;
+        const auto isCoast = [&map](const MapPoint& pt) {
+            return map.textureMap.Any(pt, IsLand) && map.textureMap.Any(pt, IsWater);
```
Flamefire, line 72: I'm unsure about the logic change here:
- prior: One of the neighbor points must be fully surrounded by water.
- now: any neighboring texture of this point must be water

Does this really work? I mean great if it does, as it's much faster, but can't this lead to problems?

libs/s25main/mapGenerator/Harbors.cpp
```diff
@@ -86,19 +78,29 @@ namespace rttr { namespace mapGenerator {
...
         return coasts;
     }
 
     void PlaceHarbors(Map& map, const std::vector<River>& rivers, int coastSize, int nodesPerHarbor)
     {
-        auto coasts = FindCoastlines(map, rivers);
-        for(auto coast : coasts)
+        const auto coasts = FindCoastlines(map, rivers);
+        std::vector<MapPoint> harbors;
+        for(const auto& coast : coasts)
```
Flamefire, line 103: Can you include the type (instead of auto) either here or in line 101? It's hard to follow otherwise, thanks! :)

tests/s25Main/mapGenerator/testAlgorithms.cpp
```diff
@@ -24,6 +24,36 @@ using namespace rttr::mapGenerator;
 
 BOOST_AUTO_TEST_SUITE(AlgorithmsTests)
 
+BOOST_AUTO_TEST_CASE(join_vector_of_simple_vectors)
+{
+    std::vector<std::vector<int>> multiVector{{1, 2, 3}, {2, 3, 4}, {5, 6}};
+    std::vector<int> result = join(multiVector);
```
Flamefire, line 30: You can directly do `BOOST_TEST(join(multiVector) == std::vector<int>{1, 2, 3, 2, 3, 4, 5, 6}, boost::test_tools::per_element())` (Check others tests, might misremember exact usage)
Or keep the `result` (const of course) to also check that the return type is as expected

libs/s25main/world/MapBase.cpp
```diff
@@ -138,6 +138,11 @@ unsigned MapBase::CalcDistance(const Position& p1, const Position& p2) const
     return ((dy + (dx > 0 ? dx : 0)) / 2);
 }
 
+unsigned MapBase::CalcMaxDistance() const
+{
+    return CalcDistance(MapPoint::all(0), MapPoint(size_.x / 2, size_.y / 2));
```
Flamefire, line 143: Out of curiosity I resolved this to `return (size.x > size.y/2) ? size.x/2 + size.y/4 : size.y/2` which isn't intuitive but I'm pretty sure all is correct (i.e. `CalcDistance`), so leave this as-is
However I omitted one part: If size.y is not a multiple of 4 then the maxdistance is one less.
Got to check that later...

libs/s25main/mapGenerator/Algorithms.h
```diff
@@ -28,6 +28,17 @@
...
+    {
+        std::vector<E> joined;
+        for(const auto& el : container)
+        {
+            std::copy(el.begin(), el.end(), std::back_inserter(joined));
```
Flamefire, line 37:
```suggestion
            joined.insert(joined.end(), el.begin(), el.end());
```
Might be more efficient.

Flamefire (changes requested):
Thanks! A few remarks inline and a general one:
Could you please use `BOOST_TEST*` macros in the test? That has automatic value decomposition, i.e. shows what was wrong if it fails.
Use `BOOST_TEST(x == y)` in general and `BOOST_TEST_REQUIRE(x == y)` only when the rest of the test won't be useful/possible when this fails, e.g. `BOOST_TEST_REQUIRE(ptr); BOOST_TEST(*ptr == 42)`
Note that `BOOST_TEST_REQUIRE` is not  the same as `BOOST_REQUIRE` (the latter only says "failed" on failure)

I'm step by step converting the tests to use the "new" macros and the usual `BOOST_TEST` is even shorter than `BOOST_REQUIRE` ;)

libs/s25main/mapGenerator/Harbors.cpp
Flow86: probably better in 101

Flamefire: I wouldn't mind. They are close and the only important thing to know here is that `coasts` is a collection, the exact and bulky type isn't important, but for `coast` it is (more) important and the type is shorter.

libs/s25main/mapGenerator/Algorithms.h
vader1986: It's not really performance relevant, at least not the way it is currently used (but I get your point). 

Just a nice challenge to join all elements in a container of containers I thought. When used it only saves a few lines of code in another place, was just proud I got it compiled lol. 

Might be a bit simpler to understand with more meaningful template type names I guess.

libs/s25main/mapGenerator/Harbors.cpp
```diff
@@ -86,19 +78,29 @@ namespace rttr { namespace mapGenerator {
...
                 }
             }
         }
+
+        const auto distanceToRiver = DistancesTo(join(rivers), map.size);
+        const auto closeToRiver = [&distanceToRiver](const MapPoint& pt) { return distanceToRiver[pt] < 5; };
+
+        for(std::vector<MapPoint>& coast : coasts)
+        {
+            coast.erase(std::remove_if(coast.begin(), coast.end(), closeToRiver), coast.end());
```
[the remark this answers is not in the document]
vader1986: Ah, I thought it was somewhere, just didn't remember where to find it! Cheers!

vader1986:
> Thanks! A few remarks inline and a general one:
> Could you please use `BOOST_TEST*` macros in the test? That has automatic value decomposition, i.e. shows what was wrong if it fails.
> Use `BOOST_TEST(x == y)` in general and `BOOST_TEST_REQUIRE(x == y)` only when the rest of the test won't be useful/possible when this fails, e.g. `BOOST_TEST_REQUIRE(ptr); BOOST_TEST(*ptr == 42)`
> Note that `BOOST_TEST_REQUIRE` is not the same as `BOOST_REQUIRE` (the latter only says "failed" on failure)
> 
> I'm step by step converting the tests to use the "new" macros and the usual `BOOST_TEST` is even shorter than `BOOST_REQUIRE` ;)

For some reasons  I get a compiler error using `BOOST_TEST(result[i] == expectedResult[i]);` (both vectors of `MapPoint`). The error says `Static_assert failed due to requirement boost::has_left_shift...`. Is there a documentation for those functions?

libs/s25main/mapGenerator/Harbors.cpp
vader1986: sure, makes sense

vader1986:
Weird, suddenly the `BOOST_TEST` worked. Must have missed something ...

libs/s25main/mapGenerator/Harbors.cpp
vader1986: The "one of the neighbor points must be fully surrounded by water" was there to  ignore harbors being placed on 1-texture-rivers. Later I passed all rivers to the function anyway, so I could have removed that code before I suppose. Unless I'm missing some edge case I just forgot about ...

vader1986: There might be 2 islands very close to each other I guess. I wonder if there's a simpler solution to include that case.

libs/s25main/world/MapBase.cpp
vader1986: The settler coordinate system gave me quite a headache in the past. But yeah, your solution should work as well. Since it's not performance critical at all, I guess it's just about what's more intuitive. On the other hand, I think as a dev I'd just  expect those functions (like `GetNeighbor`, `CalcDistance`, etc.) just to work without looking into the code if not absolutely necessary. So I wouldn't mind changing it, but either way is fine ...

libs/s25main/mapGenerator/Harbors.cpp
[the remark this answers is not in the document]
Flamefire: containerUtils.h is the header, so for similar stuff look there. I'm open for adding new, commonly used operations which are...unwieldy there or at least "rangify" existing ones (i.e. accept a container instead of iterators)

BTW: I noticed that C++20 introduced `std::erase_if` as a free function, so I guess we should update the names in containerUtils to match that and avoid confusion. Just FYI, I'll likely do that later sometime

libs/s25main/mapGenerator/Harbors.cpp
Flamefire: I guess this won't be trivial, but if you have an idea when this could break, why not add a testcase for it? "Just" prepare a map as-if it was random-generated so far with the situation you describe and call this function.
Then you can either really BOOST_TEST it, split the function so you can or at least debug it and verify that way that it works as expected.
Up to you, we can postpone this to when an actual bug shows up and if it doesn't then we can save the work.

libs/s25main/world/MapBase.cpp
Flamefire: Fully agreed, the hexagonal map is difficult... But I don't see how to make it easier to work with so it is what it is.
No, don't change, as mentioned there is the corner case of non-multiple-of-4 which I omitted and I'd say we would need a test for this function if we change it to that (iterate over all points and check that no distance is bigger than this. but that would be quite expensive...)

Flamefire (approved):
> For some reasons I get a compiler error using BOOST_TEST(result[i] == expectedResult[i]); (both vectors of MapPoint). The error says Static_assert failed due to requirement boost::has_left_shift.... Is there a documentation for those functions?

`BOOST_TEST(lhs == rhs)` prints the values of lhs and rhs on failure using (basically) `std::cout << rhs << "!=" << lhs`, so the type must be printable/streamable. This is what the error says: It was not.

For points there is the header `PointOutput.h` and for pretty much everything else `GameTypesOutput.h`, otherwise you may need to write a custom stream op or use BOOST_TEST_TOOLS_DONT_PRINT (or so), see other tests or the boost docu: https://www.boost.org/doc/libs/1_69_0/libs/test/doc/html/index.html

PR #1365, opened 2021-02-03
Fix and improve coverage on CI

Flamefire:
A change in the coverage providers disabled our coverage reports.
This fixes this and enhances it: There are now scripts to collect coverage data only for local execution and a new step to enforce 100% code coverage with meaningful reporting and documentation.

It also fixes an issue in the test code which I found with that `IsMountainOrSnowOrLava` was used instead of `IsMineableMountain` (used later). @vader1986 Please verify that the change is as intended

I also slightly refactored the tests to replace the `RunTest` functions by Fixtures, which they basically were as there was some weird coverage bug caused by that.


Flamefire:
What I'd like to suggest: As test coverage is now enforced to be 100% it might be useful to exclude the test folder on codecov/coveralls.
According to https://codecov.io/gh/Return-To-The-Roots/s25client we have 1/5 lines in the tests folder. This means any new test which increases the useful coverage will not increase the percentage that much, i.e. you (currently) would need to cover 100 lines more to get 1% increase then you would if the test folder was excluded.

Downside: You can't view the coverage of the test folder that easily anymore if that would be below 100%. However the CI at least tells you the files and lines as text output, so I'd say that is enough

For comparison the current numbers: With tests 57.9% coverage, without 47.5%.

tests/s25Main/mapGenerator/mapGenFixtures.h
```diff
@@ -0,0 +1,36 @@
...
+    WorldDescription worldDesc;
+    using Map = rttr::mapGenerator::Map;
+
+    MapGenFixture() { loadGameData(worldDesc); }
+
+    Map createMap(const MapExtent size, unsigned numPlayers = 1) const
+    {
+        return Map(size, numPlayers, worldDesc, landscape);
+    }
+};
```
vader1986, line 36: Nice one!

tests/s25Main/mapGenerator/testResources.cpp
```diff
@@ -16,191 +16,177 @@
...
-     */
-
-    for(unsigned y = 0; y < map.size.y; y++)
+    MapExtent size = MapExtent(16, 16);
+    Map map;
+
+    ResourceTestFixture() : map(createMap(size))
     {
-        for(unsigned x = 0; x < 2; x++)
+        auto mountain = map.textureMap.Find(IsMinableMountain);
```
vader1986, line 34: That's correct - `IsMountainOrSnowOrLava` was not intended.

Flow86 (dismissed):

Flow86:
I would say we remove the test from coverage reporting, as long as you some output somewhere of it.

PR #1367, opened 2021-02-05
Add more generic serialization functions

Flamefire:
Some improvements for the (serialization) tests.

Add functions for serializing containers, including fixed size ones with best possible speed by potentially reading/writing the whole container at once where possible. This often is as we have many arrays and vectors of uint8_t values or enums.
Also adding functions for serializing points.

This now allows to use those in e.g. the GameCommands.

The most crucial commit is "Use helpers push/popPoint and push/popContainer", I checked with a savegame but some manual checking could be a good idea.

### DIFF
--- a/libs/common/include/helpers/serializeContainers.h
+++ b/libs/common/include/helpers/serializeContainers.h
@@ -67,14 +67,14 @@ namespace detail {
         container.resize(ser.PopVarSize());
     }
     template<typename T>
-    std::enable_if_t<!isResizableContainer<T>::value> maybePopSizeAndResize(Serializer& ser, T& container)
+    std::enable_if_t<!isResizableContainer<T>::value> maybePopSizeAndResize(Serializer&, T&)
     {}
 
     template<typename T>
-    void popContainer(Serializer& ser, T& container, long)
+    void popContainer(Serializer& ser, T&& container, long)
     {
-        using Type = typename T::value_type;
-        for(auto& el : container)
+        using Type = typename std::remove_reference_t<T>::value_type;
+        for(auto&& el : container) // auto&& required for vector<bool> proxy object
             el = popEnumOrIntegral<Type>(ser);
     }
     template<typename T>
@@ -99,9 +99,9 @@ void pushContainer(Serializer& ser, const T& container, bool ignoreSize = false)
 }
 
 template<typename T>
-void popContainer(Serializer& ser, T& result, bool ignoreSize = false)
+void popContainer(Serializer& ser, T&& result, bool ignoreSize = false)
 {
-    using Type = typename T::value_type;
+    using Type = typename std::remove_reference_t<T>::value_type;
     static_assert(std::is_integral<Type>::value || std::is_enum<Type>::value,
                   "Only integral types and enums are possible");
 
--- a/libs/common/include/helpers/serializeContainers.h
+++ b/libs/common/include/helpers/serializeContainers.h
@@ -1,0 +1,140 @@
+// Copyright (c) 2021 - 2021 Settlers Freaks (sf-team at siedler25.org)
+//
+// This file is part of Return To The Roots.
+//
+// Return To The Roots is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 2 of the License, or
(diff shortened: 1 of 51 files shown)


libs/common/include/helpers/serializeContainers.h
```diff
@@ -0,0 +1,140 @@
...
+            }
+        }
+    };
+    template<typename T>
+    struct PopContainer<T, false, true, std::enable_if_t<std::is_pointer<decltype(std::declval<T>().data())>::value>>
+    {
+        static void pop(Serializer& ser, T& container)
+        {
+            container.clear();
+            container.resize(ser.PopVarSize());
```
Flow86, line 87: is the size pushed with pushrawdata?

libs/s25main/GamePlayer.cpp
```diff
@@ -74,10 +74,7 @@ GamePlayer::GamePlayer(unsigned playerId, const PlayerInfo& playerInfo, GameWorl
     global_inventory.clear();
 
     // Statistiken mit 0en füllen
-    memset(&statistic[StatisticTime::T15Minutes], 0, sizeof(statistic[StatisticTime::T15Minutes]));
-    memset(&statistic[StatisticTime::T1Hour], 0, sizeof(statistic[StatisticTime::T1Hour]));
-    memset(&statistic[StatisticTime::T4Hours], 0, sizeof(statistic[StatisticTime::T4Hours]));
-    memset(&statistic[StatisticTime::T16Hours], 0, sizeof(statistic[StatisticTime::T16Hours]));
+    memset(&statistic, 0, sizeof(statistic));
```
Flow86, line 77: can we use some methods to clear the statistics somehow instead of memset?

libs/common/include/helpers/serializeContainers.h
Flamefire: No, see the push method

libs/s25main/GamePlayer.cpp
Flamefire: It is an array of arrays of PODs, so memset is fine here. Can't add a method there, since it's an array. We could use `statistic = {}` which has the same effect, but IMO memset is more clear.

Flow86: `statistic = {}` would be less error prone

libs/common/include/helpers/serializeContainers.h
```diff
@@ -0,0 +1,120 @@
...
+} // namespace detail
+
+template<typename T>
+void pushContainer(Serializer& ser, const T& container, bool ignoreSize = false)
+{
+    using Type = typename T::value_type;
+    static_assert(std::is_integral<Type>::value || std::is_enum<Type>::value,
+                  "Only integral types and enums are possible");
+
+    if(detail::isResizableContainer<T>::value && !ignoreSize)
```
Flow86, line 96: probably we should still always transfer the sizes, would still make it less error prone

Flamefire: There are multiple reasons not to. For fixed size containers (std::array) it is not required at all and would just make the messages/savegames larger. I introduced the VarSize format to save some size. And sometimes we really don't want to, e.g. for the compatibility code where size is serialized differently in older versions
As for error prone: We have multiple safeguards in place already: After every Event and GameObject there is a marker as a kind of "checksum". That was enough to find almost all issues I introduced when implementing this. The only issue that was slightly harder (but still found by that) was a confusion of push and pop (we should really have 2 classes... but: later)

Flow86: I think the size is not the problem here, the question is if its recoverable/reconstructable if we change the sizes of the arrays or between platforms.

Flamefire: >  if we change the sizes of the arrays 

This is the same as changing the types of the data or adding/removing any other field: That makes the serialization fail (latest at the next checksum) and needs to be handled explicitely at the call site as is currently done for e.g. the different gameDataVersions

> between platforms.

There is no platform dependency here, so this isn't an issue :)

Flow86: ok as long we're not running into any deserialization problems (alignement) when you move your savegame to a different platform

PR #1368, opened 2021-02-09
Improve notification system

Flamefire:
I stumbled over something and wanted to test if the notification system and the unique type ids work across TU boundaries.
While writing the tests I found that not being able to unsubscribe in publish is a hassle and basically impossible to test: The unsubscribe happens in the deleter of the `shared_ptr`, so if that throws the pointer won't be deleted and/or the application will crash.

I hence implemented the missing functionality and checked where it can be improved. Using the flat_map doesn't help much for speed but increases complexity a lot. I hence reverted it but kept the commit for the future in case this becomes a bottleneck.


libs/s25main/notifications/NotificationManager.h
```diff
@@ -29,36 +29,34 @@ class NotificationManager
...
 
 public:
-    NotificationManager();
     ~NotificationManager();
 
     /// Subscribe to a specific notification.
     /// Unsubscribes when the subscription has no references left
     template<class T_Note>
-    Subscription subscribe(std::function<void(T_Note)> callback);
+    Subscription subscribe(std::function<void(const T_Note&)> callback) noexcept;
```
Flow86, line 39: should we probably use some existing library for observer pattern instead of implementing it ourselfs? (reactivex, or similar?)

Flamefire: I was thinking about this too. But that would add another dependency and likely be slow. We could also use Boost.Signals2 somehow.
This solution has a few advantages:
- It is very fast and needs to (shown by changing list to deque having a measurable impact)
- Pretty small with small compiletime overhead (Boost.Signals2 is header only and as you can imagine large because it needs to be very general)
- Allows passing of arbitrary data and that is the type on which to subscribe which makes it very easy to use and type safe

If we disallow subscribe, unsubscribe and publish during publishing (again) the class would even be simpler.

And TBH: I don't feel like rethinking the usages, subscriptions/callback handling etc. and learning a new framework. This works, has worked and is now even more flexible, so why change it? (CI failures were some spurious, unrelated issues and minor oversights introduced during optimization [e.g deque invalidates iterators but not references, which I missed])

Flow86: what about https://github.com/ddinu/observable ?

Flamefire: Looks better (has the possibility to send data) but still: We would still need the "registry" with the mapping of types to signals and the accompanying type erasure (`void*`) or another approach for the signals. Major problem though: Use of std::thread-stuff (mutex and so on) which doesn't work on MinGW
Also as mentioned: The current implementation is tuned to our use case and hence faster than the alternatives.

libs/s25main/notifications/NotificationManager_impl.h
```diff
@@ -58,62 +40,94 @@ struct NotificationManager::NoteCallbackBase
...
+        *itEl = nullptr;
     else
-        RTTR_Assert(false);
+        callbacks.erase(itEl);
     // Set only the flag. Actual deletion of the callback must be handled by the shared_ptr deleter
     callback->SetUnsubscribed();
 }
 
 template<class T_Note>
 void NotificationManager::publish(const T_Note& notification)
```
Flow86, line 103: normally this is named `notify` ?

libs/s25main/notifications/NotificationManager_impl.h
```diff
@@ -58,62 +40,94 @@ struct NotificationManager::NoteCallbackBase
...
+            if(cb)
+                cb->execute(notification);
+            else
+                hasEmpty = true;
+        }
+        RTTR_Assert(subs.isPublishing);
+        --subs.isPublishing;
+        if(!subs.isPublishing && hasEmpty)
+            helpers::erase(callbacks, nullptr);
+    } catch(...)
```
Flow86, line 127: can you add a test which throws an exception inside execute? than this code should be triggered

tests/s25Main/notificationSystem/TestNote1.h
```diff
@@ -0,0 +1,27 @@
...
+
+#pragma once
+
+#include "notifications/notifications.h"
+
+struct TestNote1
+{
+    ENABLE_NOTIFICATION(TestNote1)
+
+    int value = 0;
```
Flow86, line 26: shouldn't we use something different for the values?

libs/s25main/notifications/NotificationManager_impl.h
Flamefire: Not in this case. You **notify** a subscriber but you **publish** a notification. In this case it is a "broadcast to whoever listens" so "publish" makes more sense. See e.g. the callsite: `manager.publish(MyNotification{foo});`

tests/s25Main/notificationSystem/TestNote1.h
Flamefire: This is only the default value so to avoid a ctor. Actual values are set via usage `TestNote1{42}`

libs/s25main/notifications/NotificationManager_impl.h
Flamefire: Good catch! Done.

tests/s25Main/notificationSystem/TestNote1.h
Flow86: ah okay

libs/s25main/notifications/NotificationManager_impl.h
Flow86: yes coveralls or so catched that :-D

PR #1369, opened 2021-02-11
Configurable number of islands for water & mixed maps

vader1986:
First, I fixed the size of players' home islands to a value depending on the map size (200-1200 nodes). Also, I added configurable "extra" islands (few/medium/many) of random size (range between 200 & 1200 nodes). This makes water & mixed maps slightly more interesting to play I think.

At the moment, I did not bind the central island to a specific size. I liked the idea of having a big island where multiple players can meet eventually. But up to discussion, could also make it optional/configurable as well. 

Also, I have to increase the test coverage, I'll look at that this weekend.

### DIFF
--- a/libs/s25main/ingameWindows/iwMapGenerator.cpp
+++ b/libs/s25main/ingameWindows/iwMapGenerator.cpp
@@ -28,35 +28,6 @@
 
 using namespace rttr::mapGenerator;
 
-enum
-{
-    CTRL_BTN_BACK = 0,
-    CTRL_BTN_APPLY,
-    CTRL_TXT_LANDSCAPE,
-    CTRL_TXT_GOAL,
-    CTRL_TXT_IRON,
-    CTRL_TXT_COAL,
-    CTRL_TXT_GRANITE,
-    CTRL_TXT_RIVERS,
-    CTRL_TXT_MOUNTAIN_DIST,
-    CTRL_TXT_TREES,
-    CTRL_TXT_STONE_PILES,
-    CTRL_TXT_ISLANDS,
-    CTRL_PLAYER_NUMBER,
-    CTRL_MAP_STYLE,
-    CTRL_MAP_SIZE,
-    CTRL_MAP_TYPE,
-    CTRL_RATIO_GOLD,
-    CTRL_RATIO_IRON,
-    CTRL_RATIO_COAL,
-    CTRL_RATIO_GRANITE,
-    CTRL_RIVERS,
-    CTRL_MOUNTAIN_DIST,
-    CTRL_TREES,
-    CTRL_STONE_PILES,
-    CTRL_ISLANDS
-};
-
 iwMapGenerator::iwMapGenerator(MapSettings& settings)
     : IngameWindow(CGI_MAP_GENERATOR, IngameWindow::posLastOrCenter, Extent(270, 520), _("Map Generator"),
                    LOADER.GetImageN("resource", 41), true),
--- a/libs/s25main/ingameWindows/iwMapGenerator.cpp
+++ b/libs/s25main/ingameWindows/iwMapGenerator.cpp
@@ -41,6 +41,7 @@ enum
     CTRL_TXT_MOUNTAIN_DIST,
     CTRL_TXT_TREES,
     CTRL_TXT_STONE_PILES,
+    CTRL_TXT_ISLANDS,
     CTRL_PLAYER_NUMBER,
     CTRL_MAP_STYLE,
     CTRL_MAP_SIZE,
@@ -52,11 +53,12 @@ enum
     CTRL_RIVERS,
     CTRL_MOUNTAIN_DIST,
     CTRL_TREES,
-    CTRL_STONE_PILES
+    CTRL_STONE_PILES,
+    CTRL_ISLANDS
 };
 
 iwMapGenerator::iwMapGenerator(MapSettings& settings)
-    : IngameWindow(CGI_MAP_GENERATOR, IngameWindow::posLastOrCenter, Extent(270, 470), _("Map Generator"),
+    : IngameWindow(CGI_MAP_GENERATOR, IngameWindow::posLastOrCenter, Extent(270, 520), _("Map Generator"),
                    LOADER.GetImageN("resource", 41), true),
       mapSettings(settings)
 {
@@ -108,6 +110,14 @@ iwMapGenerator::iwMapGenerator(MapSettings& settings)
     combo->AddString(_("Normal"));
     combo->AddString(_("Far"));
     combo->AddString(_("Very far"));
+
+    curPos.y += 30;
+    AddText(CTRL_TXT_ISLANDS, curPos, _("Islands"), COLOR_YELLOW, FontStyle{}, NormalFont);
+    curPos.y += 20;
+    combo = AddComboBox(CTRL_ISLANDS, curPos, comboSize, TextureColor::Grey, NormalFont, 100);
+    combo->AddString(_("Few"));
+    combo->AddString(_("Medium"));
+    combo->AddString(_("Many"));
 
     const int pgrOffset = 120;
     curPos.y += 35;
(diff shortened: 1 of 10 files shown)


libs/s25main/mapGenerator/RandomMap.cpp
```diff
@@ -98,6 +97,28 @@ namespace rttr { namespace mapGenerator {
         return 6;
     }
 
+    unsigned GetIslandSize(const MapExtent& size)
+    {
+        const unsigned combinedSize = size.x + size.y;
+        if(combinedSize <= 128)
```
Flamefire, line 103: I think it makes more sense to use the number of nodes of the map for deciding (I.e. `prodOfComponents(size)`)
Because a 100 * 28 map and a 64*64 differ by a factor of 2 in nodes.

Also minor style: How about using `if-else if-else` chain without braces? That would move the cases more together and although the else is technically redundant I think it reads easier here as you don't need to read the indented line to understand the structure.

libs/s25main/ingameWindows/iwMapGenerator.cpp
```diff
@@ -52,11 +53,12 @@ enum
...
     CTRL_MOUNTAIN_DIST,
     CTRL_TREES,
-    CTRL_STONE_PILES
+    CTRL_STONE_PILES,
+    CTRL_ISLANDS
 };
 
 iwMapGenerator::iwMapGenerator(MapSettings& settings)
-    : IngameWindow(CGI_MAP_GENERATOR, IngameWindow::posLastOrCenter, Extent(270, 470), _("Map Generator"),
+    : IngameWindow(CGI_MAP_GENERATOR, IngameWindow::posLastOrCenter, Extent(270, 520), _("Map Generator"),
```
vader1986, line 61: Is there a nice way to test the `IngameWindow` sub classes (like simulating change of combo selections)? I looked into `Test_UI/testIngameWindow` but didn't manage to call any members due to access modifiers (private/protected).

Flamefire: What exactly do you want to do? You can use `GetCtrl` or `GetCtrls` to get a specific one and change that and/or call `Msg_ButtonClick` and friends directly. They are public (at least in `Window`)

libs/s25main/mapGenerator/RandomMap.cpp
```diff
@@ -251,16 +244,19 @@ namespace rttr { namespace mapGenerator {
         const auto waterNodes = std::count(map_.z.begin(), map_.z.end(), map_.height.minimum);
```
Flamefire, line 244: Maybe cast this here to unsigned to avoid the warning below:

```
/source/libs/s25main/mapGenerator/RandomMap.cpp: In member function 'void rttr::mapGenerator::RandomMap::CreateWaterMap()':
/source/libs/s25main/mapGenerator/RandomMap.cpp:257:23: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
         if(waterNodes > map_.players * islandSize)
```

libs/s25main/ingameWindows/iwMapGenerator.cpp
vader1986: When I do something like this:
```
uiHelper::initGUITests();
rttr::mapGenerator::MapSettings settings;
iwMapGenerator wnd(settings);    
auto* combo = wnd.GetCtrl<ctrlComboBox>(CTRL_ISLANDS);
combo->SetSelection(2);
```

I get an error for the last line of code saying:
`Member access into incomplete type 'ctrlComboBox'`

Flamefire: "incomplete type" means you are missing the include. --> `#include "controls/ctrlComboBox.h"` in the test will solve this

tests/s25Main/UI/testIngameWindow.cpp
```diff
@@ -36,3 +39,67 @@ BOOST_AUTO_TEST_CASE(IngameWnd)
...
+    uiHelper::initGUITests();
+    rttr::mapGenerator::MapSettings settings;
+    iwMapGenerator wnd(settings);
+    wnd.GetCtrl<ctrlComboBox>(CTRL_PLAYER_NUMBER)->SetSelection(2); // 4 players
+    wnd.GetCtrl<ctrlComboBox>(CTRL_MAP_TYPE)->SetSelection(2); // LandscapeDesc(2)
+    wnd.GetCtrl<ctrlComboBox>(CTRL_MAP_STYLE)->SetSelection(1); // MapStyle::Land
+    wnd.GetCtrl<ctrlComboBox>(CTRL_MAP_SIZE)->SetSelection(4); // 1024x1024
+    wnd.GetCtrl<ctrlComboBox>(CTRL_MOUNTAIN_DIST)->SetSelection(3); // MountainDistance::VeryFar
+    wnd.GetCtrl<ctrlComboBox>(CTRL_ISLANDS)->SetSelection(2); // IslandAmount::Many
+    wnd.GetCtrl<ctrlProgress>(CTRL_RATIO_GOLD)->SetPosition(25);
```
Flamefire, line 83: Can you use different values? Bonus points for random ones, but likely overkill here. Reason is that the current test won't catch confusion of sliders

libs/s25main/ingameWindows/iwMapGenerator.cpp
vader1986: Still having a problem with the `BOOST_TEST`. Somehow it doesn't compile (not even on CI server). `BOOST_REQUIRE` seems to work, that's why I used it in the past I believe.

Flamefire:
Compilation fails because of missing output operators (I.e. `operator<<`). You may include GameTypesOutput.h for most of the types and/or use code like:
```
// LCOV_EXCL_START
static std::ostream& operator<<(std::ostream& out, const YOURENUM e)
{
    return out << static_cast<unsigned>(e);
}
// LCOV_EXCL_STOP
```
or `BOOST_TEST_DONT_PRINT_LOG_VALUE(YOURENUM)`

tests/s25Main/mapGenerator/testRandomMap.cpp
```diff
@@ -72,7 +72,7 @@ BOOST_AUTO_TEST_CASE(GenerateRandomMap_returns_valid_water_map)
     loadGameData(worldDesc);
     MapSettings settings;
 
-    settings.size = getRandomMapSize(83, 100); // Need enough space for player islands
+    settings.size = getRandomMapSize(84, 100); // Need enough space for player islands
```
Flamefire, line 75: Seems this is still not enough. Have you tried it locally with the minimum size and MAX_PLAYERS? Maybe the change here makes the player islands to small? Because previously this test worked with those sizes, didn't it?

vader1986: I think that fixed minimum island size of 200 might be too small. I'll try playing around locally.

vader1986: Or 200 is too large which would lead to a similar result (e.g. 6 players get a large island and there's not enough space for the remaining player).

libs/s25main/ingameWindows/iwMapGenerator.cpp
Flamefire: I guess you already found my comment at https://github.com/Return-To-The-Roots/s25client/pull/1369#issuecomment-779374470 ?
The difference between those 2 macros is that the latter will print the values on failure. I.e. you can write `BOOST_TEST(foo == 2)` and it will tell you the value of `foo` when the assertion fails, which is very valuable when checking a failure. It is equivalent to `BOOST_CHECK_EQUAL(foo, 2)` which does the same. `BOOST_REQUIRE` does only check, if the assertion is true or false and doesn't tell you anything about the values. This makes it hard, when a test fails.

For that the types must be printable, that's what the compile error says. Especially strong enums/enum classes are not printable by default, hence you need the custom output op or exclude those from being printed. This was the reason I added the GameTypesOutput header

vader1986: Yeah got that from your previous comment! Was too lazy to override the `<<` operator for those 3 types, but I should probably do that.

libs/s25main/mapGenerator/RandomMap.cpp
```diff
@@ -220,6 +211,8 @@ namespace rttr { namespace mapGenerator {
         const auto mountainLevel = LimitFor(map_.z, land, static_cast<uint8_t>(map_.height.minimum + 1)) + 1;
         const auto rivers = CreateRivers(center);
 
+        CreateFreeIslands(std::count(map_.z.begin(), map_.z.end(), map_.height.minimum));
```
Flamefire, line 214: This is where appveyor is warning. I admit that this here is a pointless warning and I've never understood why std::count returns a signed value. But introducing a variable and a cast here does make sense. It isn't immediately obvious what this counts, so a named variable helps

Flamefire (changes requested):
Looks good so far. A few minor notes to polish it up. Last ones here, I promise ;)

Flamefire (approved):
Thanks!

PR #1371, opened 2021-02-14
Remove online ranking for good

### LINKED ISSUE #124: Remove Online Ranking
Über ein Online Ranking wurde bereits hier schon einmal diskutiert:
http://www.siedler25.org/index.php?com=forum&mod=forum&action=thread&id=805

https://bugs.launchpad.net/s25rttr/+bug/899114

Flamefire:
Closes #124


Flow86:
the ranking button is still inside the lobby.

What I suggest to keep is the liblobby stuff, since if we want to implement it sometime for real, this is already in there correctly. I only would remove the access to that information (ranking button which needs to be removed and points information you already removed)

Flamefire:
> the ranking button is still inside the lobby.

Which one and where?

> What I suggest to keep is the liblobby stuff, since if we want to implement it sometime for real, this is already in there correctly

It's still in the protocol at least and in the git history. And IMO that is a big **if**. I think there are other things to do and as @Spikeone already mentioned a ranking is very hard to do as you'd need many games and a good algorithm so it is meaningful. So I'd guess it would be a major rework or a full rewrite. Hence we can keep the dead code removed for now. The commit messages are clear enough so you can find and revert or copy the code if (big if) that is required in some future ;)

Flow86 (approved):

Issue #1229, opened 2020-05-18
Feature Request:  Allow Building While on Pause


akuenzi:
One of the most annoying things to face in-game (at least for me :) ), is when I'm in the middle of placing where I want buildings to go... and then I get a notification that my settlement is under attack.  In the early stages of the settlement, I like to monitor closely what's going on in the battle... which then hinders my building placement.  Of course, none of this matters to the AI, who can continue their building projects AND do battle against his mere human... 

As such, it would be nice, upon receiving such a notification, to be able to hit 'pause,' and then be allowed to finish placing where I want buildings to go, and then go watch the battle.  And of course, I don't mean that the literal building of them would continue; I'm only talking about having the ability to place them and link up roads to them, so that my settlers can get going on them as soon as I hit 'unpause.'

Thank you for considering this. :)

Spikeone:
As this is only for singleplayer, isn't it enough to simply slow down the game (hit + / -)? At 1 GF = 70ms the game is quite slow^^.

Also the AI only do actions every 1000 / 500 / 100 GF (not sure if the numbers are correct) so in fact they also do wait^^

Flamefire:
I would deny this. Not being able to actually do multiple things at once is part of the RTS genre and what makes it challenging. Also RttR is as well for multi-player and there PAUSE means pause. It would not be fair if others could build during that time.

Issue #350, opened 2016-01-04
Remove visual settings


Flamefire:
I'd remove the "visual" settings of the warehouses, as they are possibly wrong (probably the reason for #349)

The "visual" showing of the settings before it is actually changed is a good idea, but they should be kept in sync with the actual setting. Assume e.g. a GameCommand is dropped (can happen during pause, maybe we should remove this restriction as GCs are not send during pause) then the visual setting will always be the opposite of the actual setting.

So 2 ideas:
1. Remove the visual setting altogether and just toggle the image in the ingame window on a click -> Reopening the window will always show the actual status, but that might be outdated a few ms later if one happens to open-reopen it fast -> wh might notify the window so it can update the overlays
2. Add an event on every change of the visual settings with a duration of 2.5 times the NWF length and set the visual settings equal to real settings on event. The event is replaced by another change of the visual settings and changes to the real setting also change the visual setting and remove the event --> window needs to be notified, but settings are most of the times correct (GC received at right time, GC ignored, GC received later)

I'd opt for option 1. What do you thing? @jhkl @MarcusSt @Flow86

jhkl:
I might be wrong, but aren't the visual settings also used for some ingame (not in a window) elements like the "Do not supply with coins" sign? This would not work with option 1 I guess?

Flamefire:
Right. For those cases we would need option 2. I'd also replace all toggle game commands by explicit enable/disable GCs so that on execution the actual state is correctly updated. This would already solve some problems: Disabling a coin supply that is disabled but visually enabled would do nothing but update the visual state.

Spikeone:
Just asking, this is not about removing them from the game, right?
Sounded title sounds like you want to remove the icons indicating the current state.

Flamefire:
No. The point is: You have visual and actual settings. Once you set e.g. coins stop, then visually the coins are disabled, a command is send to disable them, and when that is executed they are actually disabled. Problem is: Coins are not enabled or disabled but toggled: On<->Off. So assume this: Stop coins-> Show disabled sign + send command. Command never gets executed -> Coins show as stopped but are actually enabled. If you try to enable them, they are actually disabled and so on...

Flamefire:
Improved to a good state for now, final fix can be done later

PR #1372, opened 2021-02-14
Fix some problems with unexpected behaviour during and after pauses

JonathanSteinbuch:
Hello there,

Pauses have some weird problems. Most notably when you are in the process of building a road when you pause the game then you can click to finish the road and this leaves an unremovable ghost road as seen below.

pauseProblems

Another problem is that for many ingame settings windows if you change something during a pause the changes very often are not registered, but still displayed. In the above picture this is visible for the military window, where the number of soldiers in barracks, even after the pause, does not reflect the choice made during the pause.

To address these issues I propose the changes in this pull request.

The guiding principles that I used are as follows (I tried to stay close to the ideas used already in the code and the things decided in some issues):

1. Most importantly at any time the displayed settings should reflect the actual settings or settings that will be changed in the next timer cycle.
2. Since adding game commands in a pause fails (and rightly so, as decided in #1229), any function that issues a game command potentially during a pause has to check whether it was successful and react accordingly.
3. Still a player can click through the windows in pause (to look at stats or see where a road could be built or whatever) and the windows don't have to be aware of the pause. Particularly, I think it's rude to just close the road building window for other players during a pause, so it stays open but they can't finish the road or any other road until the pause ends.
4. If control elements (progress bars and such) can be directly changed during a pause, then if the window is still open when the pause ends, the game commands for these changes will be issued after the pause ends. If the window is closed during a pause however, the noncommited changes made to control elements are discarded.

Especially because of 4 these changes reduce the importance of the visual settings in the game client. So this is related to #350.

During this I removed that for the host the pause starts immediately (as opposed to when the pause command is executed for all players) as it's potentially problematic (for example it's the direct reason for the ghost road problem mentioned at the beginning). If there was a particular reason for the immediate pause for the host we should discuss that.

I hope you find these changes helpful. I think they should make the behaviour during and after pauses much more predictable.

Best Regards
Jonathan

### DIFF
--- a/libs/s25main/desktops/dskGameInterface.cpp
+++ b/libs/s25main/desktops/dskGameInterface.cpp
@@ -930,8 +930,8 @@ void dskGameInterface::Run()
 
 void dskGameInterface::GI_StartRoadBuilding(const MapPoint startPt, bool waterRoad)
 {
-    // Im Replay und in der Pause keine Straßen bauen
-    if(GAMECLIENT.IsReplayModeOn() || GAMECLIENT.IsPaused())
+    // Im Replay keine Straßen bauen
+    if(GAMECLIENT.IsReplayModeOn())
         return;
 
     road.mode = waterRoad ? RoadBuildMode::Boat : RoadBuildMode::Normal;
@@ -1079,9 +1079,11 @@ void dskGameInterface::OnChatCommand(const std::string& cmd)
 
 void dskGameInterface::GI_BuildRoad()
 {
-    GAMECLIENT.BuildRoad(road.start, road.mode == RoadBuildMode::Boat, road.route);
-    road.mode = RoadBuildMode::Disabled;
-    WINDOWMANAGER.SetCursor(Cursor::Hand);
+    if(GAMECLIENT.BuildRoad(road.start, road.mode == RoadBuildMode::Boat, road.route))
+    {
+        road.mode = RoadBuildMode::Disabled;
+        WINDOWMANAGER.SetCursor(Cursor::Hand);
+    }
 }
 
 void dskGameInterface::GI_WindowClosed(Window* wnd)
@@ -1155,16 +1157,16 @@ void dskGameInterface::CI_GamePaused()
     messenger.AddMessage(_("SYSTEM"), COLOR_GREY, ChatDestination::System, _("Game was paused."));
 
     /// Straßenbau ggf. abbrechen, wenn aktiviert
-    if(road.mode != RoadBuildMode::Disabled)
-    {
-        // Fenster schließen
-        if(roadwindow)
-        {
-            roadwindow->Close();
-            roadwindow = nullptr;
-        }
-        GI_CancelRoadBuilding();
-    }
+    /*   if(road.mode != RoadBuildMode::Disabled)
+       {
+           // Fenster schließen
+           if(roadwindow)
+           {
+               roadwindow->Close();
+               roadwindow = nullptr;
+           }
+           GI_CancelRoadBuilding();
+       }*/
 }
 
 void dskGameInterface::CI_GameResumed()
--- a/libs/s25main/desktops/dskGameInterface.cpp
+++ b/libs/s25main/desktops/dskGameInterface.cpp
@@ -1155,18 +1155,6 @@ void dskGameInterface::CI_ReplayEndReached(const std::string& msg)
 void dskGameInterface::CI_GamePaused()
 {
     messenger.AddMessage(_("SYSTEM"), COLOR_GREY, ChatDestination::System, _("Game was paused."));
-
-    /// Straßenbau ggf. abbrechen, wenn aktiviert
-    /*   if(road.mode != RoadBuildMode::Disabled)
-       {
-           // Fenster schließen
-           if(roadwindow)
-           {
-               roadwindow->Close();
-               roadwindow = nullptr;
-           }
-           GI_CancelRoadBuilding();
-       }*/
 }
 
 void dskGameInterface::CI_GameResumed()
--- a/libs/s25main/ingameWindows/TransmittingSettingsWindow.cpp
+++ b/libs/s25main/ingameWindows/TransmittingSettingsWindow.cpp
@@ -1,0 +1,70 @@
+// Copyright (c) 2005 - 2021 Settlers Freaks (sf-team at siedler25.org)
+//
+// This file is part of Return To The Roots.
+//
+// Return To The Roots is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 2 of the License, or
+// (at your option) any later version.
+//
+// Return To The Roots is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Return To The Roots. If not, see <http://www.gnu.org/licenses/>.
+
+#include "TransmittingSettingsWindow.h"
+#include "WindowManager.h"
+#include "iwMsgbox.h"
+#include "network/GameClient.h"
+#include <mygettext/mygettext.h>
+
+TransmittingSettingsWindow::TransmittingSettingsWindow(unsigned id, const DrawPoint& pos, const Extent& size,
+                                                       std::string title, glArchivItem_Bitmap* background, bool modal,
+                                                       bool closeOnRightClick, Window* parent)
+    : IngameWindow(id, pos, size, title, background, modal, closeOnRightClick, parent), settings_changed(false)
+{
+    // Timer for transmitting changes every 2 seconds
+    using namespace std::chrono_literals;
+    AddTimer(1001, 2s);
+}
+
+void TransmittingSettingsWindow::Close()
+{
+    TransmitSettings();
+    if(!settings_changed)
+    {
+        IngameWindow::Close();
+    } else
+    {
+        WINDOWMANAGER.Show(std::make_unique<iwMsgbox>(GetTitle(),
+                                                      _("The changes could not be applied and will be discarded. A "
+                                                        "potential reason for this is that the game is paused."),
+                                                      this, MsgboxButton::OkCancel, MsgboxIcon::ExclamationRed, 1000));
+    }
+}
+
+void TransmittingSettingsWindow::Msg_MsgBoxResult(const unsigned msgbox_id, const MsgboxResult mbr)
+{
+    switch(msgbox_id)
+    {
+        case 1000: // Close?
+        {
+            if(mbr == MsgboxResult::Ok)
+            {
+                IngameWindow::Close();
+            }
+        }
+        break;
+    }
+}
+
+void TransmittingSettingsWindow::Msg_Timer(const unsigned /*ctrl_id*/)
+{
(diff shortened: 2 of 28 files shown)


Flamefire (changes requested):
Great work!
Few notes inline and to your points:
> Most importantly at any time the displayed settings should reflect the actual settings or settings that will be changed in the next timer cycle.

Take care that this might even be later: E.g. changing a setting, that is transmitted and on the way BACK from the server. Then changing it again so the current visual setting will be for the 2nd next NWF tick. From what I've seen this should work ATM.

> Particularly, I think it's rude to just close the road building window for other players during a pause, so it stays open but they can't finish the road or any other road until the pause ends.

See inline comment. Maybe it's better to do so.

> If the window is closed during a pause however, the noncommited changes made to control elements are discarded.

Great that this is defined now. Maybe it makes sense to warn/ask the player about such things? IIRC we have a close function which can prevent closing which could be used to show a message. Sending stuff in the dtor might not have been the best idea anyway...

> During this I removed that for the host the pause starts immediately (as opposed to when the pause command is executed for all players) as it's potentially problematic (for example it's the direct reason for the ghost road problem mentioned at the beginning). If there was a particular reason for the immediate pause for the host we should discuss that.

IMO we should keep and fix that. Especially in a single player game you want immediate pause, it would look strange otherwise. This is very important for replays but maybe also for a single player or during debugging, e.g. pause right before a soldier enters a building. I **think** we could simply call the OnMsgPause function directly instead of setting `framesinfo.isPaused = true;` Could you check that?

libs/s25main/desktops/dskGameInterface.cpp
```diff
@@ -930,8 +930,8 @@ void dskGameInterface::Run()
 
 void dskGameInterface::GI_StartRoadBuilding(const MapPoint startPt, bool waterRoad)
 {
-    // Im Replay und in der Pause keine Straßen bauen
-    if(GAMECLIENT.IsReplayModeOn() || GAMECLIENT.IsPaused())
+    // Im Replay keine Straßen bauen
+    if(GAMECLIENT.IsReplayModeOn())
```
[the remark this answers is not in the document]
JonathanSteinbuch: I changed this because I felt it fits better with the behaviour of the other controls. (i.e. you can also look at which buildings to build or at options windows, etc.) And since I added a check that you can only finish building a road outside of a pause this didn't feel necessary anymore.

But if you prefer to continue disallowing this specifically I can reinstate the old version of this line instead

libs/s25main/desktops/dskGameInterface.cpp
```diff
@@ -1153,18 +1155,6 @@ void dskGameInterface::CI_ReplayEndReached(const std::string& msg)
 void dskGameInterface::CI_GamePaused()
 {
     messenger.AddMessage(_("SYSTEM"), COLOR_GREY, ChatDestination::System, _("Game was paused."));
-
-    /// Straßenbau ggf. abbrechen, wenn aktiviert
-    if(road.mode != RoadBuildMode::Disabled)
```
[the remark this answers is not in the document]
JonathanSteinbuch: You will not be stuck in road mode since you can still manually cancel road building.

That said, my reasoning for removing this is the same as above and if you prefer to have it, no real problem with me.

libs/s25main/desktops/dskGameInterface.cpp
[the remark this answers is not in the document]
Flamefire: Agreed, ok

libs/s25main/desktops/dskGameInterface.cpp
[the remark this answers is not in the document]
Flamefire: Ah, ok, then that is fine too.

So only the braces please (in general I avoid braces if the `if` and statement below or short (single line), but here we have a comment and a statement and a very long if so the braces help with readability) -> https://github.com/Return-To-The-Roots/s25client/pull/1372#discussion_r575804883

JonathanSteinbuch:
Thank you, @Flamefire, for your helpful review.

> > If the window is closed during a pause however, the noncommited changes made to control elements are discarded.
> 
> Great that this is defined now. Maybe it makes sense to warn/ask the player about such things? IIRC we have a close function which can prevent closing which could be used to show a message. Sending stuff in the dtor might not have been the best idea anyway...

Good idea! I added a dialog box in the case that settings would be discarded. Since it's the same for all of these windows I implemented it in the form of another class TransmittingSettingsWindow with base class IngameWindow.

> > During this I removed that for the host the pause starts immediately (as opposed to when the pause command is executed for all players) as it's potentially problematic (for example it's the direct reason for the ghost road problem mentioned at the beginning). If there was a particular reason for the immediate pause for the host we should discuss that.
> 
> IMO we should keep and fix that. Especially in a single player game you want immediate pause, it would look strange otherwise. This is very important for replays but maybe also for a single player or during debugging, e.g. pause right before a soldier enters a building. I **think** we could simply call the OnMsgPause function directly instead of setting `framesinfo.isPaused = true;` Could you check that?

Yes, this should work. It feels a bit wrong to call OnMsgPause directly, but I did it anyway.

libs/s25main/ingameWindows/TransmittingSettingsWindow.cpp
```diff
@@ -0,0 +1,70 @@
...
+#include <mygettext/mygettext.h>
+
+TransmittingSettingsWindow::TransmittingSettingsWindow(unsigned id, const DrawPoint& pos, const Extent& size,
+                                                       std::string title, glArchivItem_Bitmap* background, bool modal,
+                                                       bool closeOnRightClick, Window* parent)
+    : IngameWindow(id, pos, size, title, background, modal, closeOnRightClick, parent), settings_changed(false)
+{
+    // Timer for transmitting changes every 2 seconds
+    using namespace std::chrono_literals;
+    AddTimer(1001, 2s);
```
Flamefire, line 31: Not sure here. Some windows assign special values to the IDs e.g. to map goods to ids and back. It should at least be a protected static constexpr value. Best would likely be to give this an ID of 1 or 0(?) and use the enum-id approach with IDs starting after that. But I'd guess that is overkill for now, so at least make it a named constant...

libs/s25main/ingameWindows/TransmittingSettingsWindow.h
```diff
@@ -0,0 +1,44 @@
...
+protected:
+    /// whether any settings where changed after the last successful transmission
+    bool settings_changed;
+
+public:
+    TransmittingSettingsWindow(unsigned id, const DrawPoint& pos, const Extent& size, std::string title,
+                               glArchivItem_Bitmap* background, bool modal = false, bool closeOnRightClick = true,
+                               Window* parent = nullptr);
+
+    virtual ~TransmittingSettingsWindow() {}
```
Flamefire, line 33:
```suggestion
```

No need, IngameWindow and Window is already virtual

libs/s25main/ingameWindows/TransmittingSettingsWindow.h
```diff
@@ -0,0 +1,44 @@
...
+
+    /// Updates the control elements with values from visual settings
+    virtual void UpdateSettings() {}
+    /// sends potential changes to the client
+    virtual void TransmitSettings() {}
```
Flamefire, line 38:
```suggestion
    virtual void UpdateSettings() = 0;
    /// sends potential changes to the client
    virtual void TransmitSettings() = 0;
```

I think it doesn't make sense to subclass this without implementing those 2

libs/s25main/ingameWindows/TransmittingSettingsWindow.h
```diff
@@ -0,0 +1,44 @@
...
+
+    void Close() override;
+
+    virtual void Msg_Timer(unsigned ctrl_id) override;
```
Flamefire, line 42:
```suggestion
    void Msg_Timer(unsigned ctrl_id) override;
```

Per convention: `virtual` introduces a method and `override [final]` overrides it. So always use either or for consistence

libs/s25main/ingameWindows/TransmittingSettingsWindow.h
```diff
@@ -0,0 +1,44 @@
...
+    void Msg_MsgBoxResult(const unsigned msgbox_id, const MsgboxResult mbr) override;
```
Flamefire, line 43:
```suggestion
    void Msg_MsgBoxResult(unsigned msgbox_id, MsgboxResult mbr) override;
```

`const` only in impl. clang-tidy will complain anyway

libs/s25main/ingameWindows/iwAction.cpp
```diff
@@ -338,12 +338,14 @@ void iwAction::AddUpgradeRoad(ctrlGroup* group, unsigned& /*x*/, unsigned& width
...
-void iwAction::DoUpgradeRoad()
+bool iwAction::DoUpgradeRoad()
 {
     Direction flag_dir;
     const noFlag* flag = gwv.GetWorld().GetRoadFlag(selectedPt, flag_dir);
     if(flag)
-        GAMECLIENT.UpgradeRoad(flag->GetPos(), flag_dir);
+        return GAMECLIENT.UpgradeRoad(flag->GetPos(), flag_dir);
+    else
+        return false;
```
Flamefire, line 348: **Maybe**:

```suggestion
        return true;
```

If there is no flag anymore, the task is done and the window can be closed, or what do you think?

Flamefire (changes requested):
I like the approach, I think it makes sense. If you want a better name, IIRC classes like that were called "Adapter", so maybe `TransmitSettingsIgwAdapter`? (I think we use Igw in other places) Just so it is clear that this is not a real Window yet. But up to you

libs/s25main/ingameWindows/TransmittingSettingsWindow.cpp
```diff
@@ -0,0 +1,70 @@
...
+        WINDOWMANAGER.Show(std::make_unique<iwMsgbox>(GetTitle(),
+                                                      _("The changes could not be applied and will be discarded. A "
+                                                        "potential reason for this is that the game is paused."),
+                                                      this, MsgboxButton::OkCancel, MsgboxIcon::ExclamationRed, 1000));
+    }
+}
+
+void TransmittingSettingsWindow::Msg_MsgBoxResult(const unsigned msgbox_id, const MsgboxResult mbr)
+{
+    switch(msgbox_id)
```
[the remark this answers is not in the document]
JonathanSteinbuch: True :D

libs/s25main/ingameWindows/iwAction.cpp
JonathanSteinbuch: I think you are exactly right.

libs/s25main/network/GameClient.cpp
```diff
@@ -1660,9 +1660,10 @@ void GameClient::SetPause(bool pause)
     } else if(IsHost())
     {
         // Pause instantly
+        auto msg = new GameMessage_Pause(pause);
```
JonathanSteinbuch, line 1663:
```suggestion
        auto *msg = new GameMessage_Pause(pause);
```

libs/s25main/network/GameClient.cpp
```diff
@@ -1660,9 +1660,10 @@ void GameClient::SetPause(bool pause)
     } else if(IsHost())
     {
         // Pause instantly
+        auto *msg = new GameMessage_Pause(pause);
```
JonathanSteinbuch, line 1663:
```suggestion
        auto* msg = new GameMessage_Pause(pause);
```

Flamefire (approved):

PR #1373, opened 2021-02-14
Improve speed of GetPointsInRadius

Flamefire:
- Preallocate memory (test map load time decreased from 18s to 14s in debug mode)
- Auto-deduce result_type and remove that typedef (now allows lambdas)


libs/s25main/figures/nofScout_Free.cpp
```diff
@@ -156,7 +156,7 @@ struct IsScoutable
 void nofScout_Free::GoToNewNode()
 {
     std::vector<MapPoint> available_points =
-      gwg->GetPointsInRadius<-1>(flag->GetPos(), SCOUT_RANGE, Identity<MapPoint>(), IsScoutable(player, *gwg));
+      gwg->GetMatchingPointsInRadius<-1>(flag->GetPos(), SCOUT_RANGE, IsScoutable(player, *gwg));
```
Flow86, line 159: no test for this?

libs/s25main/figures/nofWorkman.cpp
```diff
@@ -119,8 +119,8 @@ struct NodeHasResource
 MapPoint nofWorkman::FindPointWithResource(ResourceType type) const
 {
     // Alle Punkte durchgehen, bis man einen findet, wo man graben kann
-    std::vector<MapPoint> pts =
-      gwg->GetPointsInRadius<1>(pos, MINER_RADIUS, Identity<MapPoint>(), NodeHasResource(*gwg, type), true);
+    const std::vector<MapPoint> pts =
```
Flow86, line 122: no test for this?

libs/s25main/figures/nofScout_Free.cpp
Flamefire: No not yet. And IMO out of scope for this PR. Adding tests for the figures/jobs is quite complicated as you need to set up a whole world etc. Might do later but not in this PR. I think it is clear from the change that it doesn't change behavior if the function works as expected and that at least is tested

libs/s25main/figures/nofWorkman.cpp
Flamefire: see above

Flow86 (approved):

libs/s25main/figures/nofScout_Free.cpp
Flamefire: Test added

libs/s25main/figures/nofWorkman.cpp
Flamefire: Test added

PR #1377, opened 2021-02-19
Fix and refactor RANDOM

Flamefire:
Found a bug where RANDOM_ENUM would not return the last enum value due to confusing argument name "max". I replaced that my maxExcl to avoid such issues in the future

After that I factored out the output functions to make including Random.h more lightweight.
Finally I used RANDOM_RAND everywhere where possible and refactored the usages a bit so all is consistent. Naturally a RANDOM_ELEMENT was added to simplify usage even more. This helps in detecting/avoiding bugs


Flamefire:
@Flow86 this is important especially as it makes replays go async (again) because it behaves correctly again now. So please merge soon

PR #1379, opened 2021-02-21
[NFC] Add getters useful for AIs that want to limit knowledge to "human levels"

jdoerfert:


Flow86:
Is there a way to add unittests for these getters?

Flamefire:
I don't mind the changes in general but IMO we should first have a use case for this. And while a more "human" AI is good in general it is very hard to achieve and implement with reasonable performance and behavior.
The granite size might be useful already but the other 2 less so ATM(!): Implementing properly using geologists in the AI is hard and possibly slow.
So I would prefer this changes as part of an AI change PR where they are used for something.

jdoerfert:
Do you have any rule set that explains how your process is supposed to work? I have a hard time finding the logic.

Flamefire:
Sorry, I didn't mean to close this. Must have accidentally hit the wrong button.

> Do you have any rule set that explains how your process is supposed to work? I have a hard time finding the logic.

The process of adding such functionality?
My thinking here is, that I don't see yet if and how this new code is going to be used. E.g. `GetAggressors`: What would you do with that? Why would anything need a list of modifiable references to the attackers? This sounds like someone later might come around and use that to do "things", e.g. change the attackers which is very likely not how it should be done. So how would an AI use that? Having an implementation of a use case first would answer that question and allows to judge how that function needs to look like. E.g. I see a use case for `HasAggressors` (already exists) or `NumAggressors` in a distant future to decide whether to counter-attack. But I don't see that to be implemented anywhere soon. (Might be wrong of course)
Similar to the other 2: Here I can see how it **could** be used. But until it is, it is basically dead code.

So I think it makes more sense to implement an enhancement of the AI, e.g. counting the sizes of stone piles instead of only the distance to any, and add the required functions for that to the classes used.

I'm basically afraid to make the current code longer without an immediate gain.

Does that make sense?

jdoerfert:
> Sorry, I didn't mean to close this. Must have accidentally hit the wrong button.
> 
> > Do you have any rule set that explains how your process is supposed to work? I have a hard time finding the logic.
> 
> The process of adding such functionality?
> My thinking here is, that I don't see yet if and how this new code is going to be used. E.g. `GetAggressors`: What would you do with that? Why would anything need a list of modifiable references to the attackers? This sounds like someone later might come around and use that to do "things", e.g. change the attackers which is very likely not how it should be done. So how would an AI use that? Having an implementation of a use case first would answer that question and allows to judge how that function needs to look like. E.g. I see a use case for `HasAggressors` (already exists) or `NumAggressors` in a distant future to decide whether to counter-attack. But I don't see that to be implemented anywhere soon. (Might be wrong of course)

I guess I should have mentioned my AI exists and I did not create getters without reason. Unsure in which state the github version is but at some point it consistently beat the existing AI and it scales. Not to say there are not major design issues that now force me to do a partial rewrite. The "modifiable reference" comment is a nit in my book, like: Looks good, please make the aggressors const too. (Not that you could not const_cast things but anyway.)

> Similar to the other 2: Here I can see how it **could** be used. But until it is, it is basically dead code.
> 
> So I think it makes more sense to implement an enhancement of the AI, e.g. counting the sizes of stone piles instead of only the distance to any, and add the required functions for that to the classes used.
> 
> I'm basically afraid to make the current code longer without an immediate gain.
> 
> Does that make sense?

Sure. The opposing view is that these getters (and more) should exist (as others do) such that you don't have to duplicate and carry code downstream. My AI prototype uses all of them and it seems natural to share as much code as early on as possible. Others benefit from it, and the code I carry downstream is less and more localized. 

FWIW, how the AI uses `GetAggressors` (feel free to look at my fork):
When I get attacked I try to counter right away to force the fight in the middle. To do this properly you should send to the attacking military buildings as that basically guarantees to intercept their troops. You don't want to randomly attack places but specifically the ones you see enemy units come out of. If you know your aggressors you can check if you see their home base, if so, you can counter attack them, if you only know that you are attacked the AI has no option but to counter attack randomly, wasting pressure soldiers on suicide missions.

Flamefire:
> I guess I should have mentioned my AI exists and I did not create getters without reason. 

Ah, wow! I wasn't aware anyone ever planned that because of the amount of work involved.

>  The "modifiable reference" comment is a nit in my book,

Yeah, I see no option at the moment. Maybe using a span later helps. How about simply adding a TODO comment that the attackers should be const so we don't forget?

> feel free to look at my fork

Did now. Looks impressive!
One thing I've seen while scanning over it: `helpers/containerUtils.h` contains `remove_if` which is the same as your `discard_if`. Maybe there are more functions you can use. I recently also implemented more printing support for game types, see `GameTypesOutput.h`
Also try to avoid using GAMECLIENT or the visual stuff from there. The former should go at some point from more and more places so we can have e.g. dedicated servers and stuff like that. The visual stuff is only for hiding network latency from the user. The AI shouldn't need this or implement similar things itself. Same here: That should go and be moved to e.g. the windows alone. It has always bugged me because its hard to keep in sync if done for longer.

Looking forward to your future PRs! :)

Flamefire (approved):

libs/s25main/nodeObjs/noSign.h
```diff
@@ -40,6 +40,10 @@ class noSign : public noDisappearingEnvObject
     /// An x,y zeichnen.
     void Draw(DrawPoint drawPt) override;
 
+    /// Return the amount of resources found, as drawn on the sign.
+    uint8_t GetSignAmount() const { return resource.getAmount(); }
```
Flamefire, line 44: FYI: I combined those 2 functions into 1 which is more general: https://github.com/Return-To-The-Roots/s25client/pull/1373/files#diff-121534ca71eb140b4b21a437798817be9d09a6b8c8b1acdd824d465be1440ab0R44

PR #1382, opened 2021-02-22
Add test to check that a precreated replay does not async and remove GameWorld class

Flamefire:
To avoid issues such as https://github.com/Return-To-The-Roots/s25client/pull/1377 I thought we should include a test which runs a large replay and tests for asyncs.

In implementing this I noticed that it was a bit difficult to load the replay and went over the GameWorld class which shouldn't really exist upon checking. So I moved that functionality into MapLoader and MapSerializer which also required some more changes so all accesses are sane. Especially the trade path cache was a bit tricky, so I refactored that and the econ handler access was fully public with no need.
I also went over instances of `const shared_ptr&` which isn't exactly useful. Replacing this by instances and moving has better semantics and at other places not passing the ownership but the object contained also makes usage more sane and allows to NOT create a shared_ptr just to pass that to a function expecting that for no other reason than simplicity.

As the replay is long (~30s in release, much longer in debug) I only added it to release configs (the test is always built though, just not run) and to not create false coverage I excluded it from coverage builds.

The replay was created on master, so all changes done here are already "tested".


libs/s25main/buildings/noBaseBuilding.h
```diff
@@ -79,7 +79,7 @@ class noBaseBuilding : public noRoadNode
 
     /// Ermittelt die Flagge, die vor dem Gebäude steht
     noFlag* GetFlag() const;
-    /// Same as GetFlag()->GetPos()
+    /// Same as GetFlagPos()
```
Flow86, line 82: weird comment change?

libs/s25main/Game.cpp
```diff
@@ -49,8 +49,8 @@ void Game::Start(bool startFromSave)
         if(ggs_.objective == GameObjective::EconomyMode)
         {
             unsigned int selection = ggs_.getSelection(AddonId::ECONOMY_MODE_GAME_LENGTH);
-            world_.econHandler = std::make_unique<EconomyModeHandler>(
-              std::chrono::minutes(AddonEconomyModeGameLengthList[selection]) / SPEED_GF_LENGTHS[referenceSpeed]);
+            world_.setEconHandler(std::make_unique<EconomyModeHandler>(
```
Flow86, line 52: should we probably  hand over the calculation to the econ-handler? What I think where we have a problem: If we pause the game, the time will still count on for this kind of goal, but I would expect that paused time does not count

libs/s25main/buildings/noBaseBuilding.h
Flamefire: Good catch, fixed

libs/s25main/Game.cpp
Flamefire: > but I would expect that paused time does not count

It does not. The EconomyModeHandler is passed a number of GFs here

Flow86: thats what I meant, so why not simply pass `AddonEconomyModeGameLengthList[selection]` inside instead of std::chrono::minutes?

tests/common/CMakeLists.txt
```diff
@@ -1,3 +1,4 @@
 add_testcase(NAME common
     LIBS s25Common testHelpers
+    COST 1
```
Flow86, line 3: why not use labels for the tests?

https://stackoverflow.com/questions/24495412/ctest-using-labels-for-different-tests-ctesttestfile-cmake

Flamefire: I thought about that and discarded it. It doesn't seem to play nice with with "make test" and I have found other reports of people wanting something like this and this was the best solution
As mentioned in the description: this test is still build. You can run it manually in debug e.g. Through VS. No problem there so I kept this

libs/s25main/Game.cpp
Flamefire: Because this selection is a gui choice and should be independent. This makes tests simpler 
And again: this does not pass minutes here, it passes GFs. Check the constructor. :)

Flamefire: I moved the "minutes" into `AddonEconomyModeGameLengthList` which makes more sense and might avoid the confusion here

Flamefire:
Found a small issue with a format string which I had to fix. All done and tested locally now.

Flow86 (changes requested):
conflict, needs rebase

PR #1384, opened 2021-02-27
Add functions so AI can send and receive chat messages

### LINKED ISSUE #1381: [AI] Add ChatNote to allow AI to receive (and act on) chat messages

Flamefire:
Closes #1381 

@jdoerfert Does this interface suffice?

@Flow86 Testing the code in the GameClient is still too hard to do in a unit test. I tested the changes in AIPlayerJH manually which writes a message shortly after start and that works.

As a note: The chat message isn't send immediately but with the next NWF. I think that is fair given that they happen often enough and it keeps the "fetch" code close by

### DIFF
--- a/libs/s25main/ai/AIInterface.cpp
+++ b/libs/s25main/ai/AIInterface.cpp
@@ -344,9 +344,9 @@ bool AIInterface::CallSpecialist(const noFlag* flag, Job job)
     return CallSpecialist(flag->GetPos(), job);
 }
 
-void AIInterface::Chat(const std::string& message)
+void AIInterface::Chat(const std::string& message, ChatDestination destination)
 {
-    pendingChatMsgs_.push_back(std::make_unique<GameMessage_Chat>(playerID_, ChatDestination::All, message));
+    pendingChatMsgs_.push_back(std::make_unique<GameMessage_Chat>(playerID_, destination, message));
 }
 
 std::vector<std::unique_ptr<GameMessage_Chat>> AIInterface::FetchChatMessages()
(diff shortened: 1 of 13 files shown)


Flamefire:
We already have `FetchGameCommands` and it is "fetch" not "get", so IMO that already makes it clear the messages/cmds are removed when they are returned. I'd keep that terminology as we use "get" for all getters so the "fetch" stands out here

libs/s25main/ai/AIInterface.cpp
```diff
@@ -340,3 +343,15 @@ bool AIInterface::CallSpecialist(const noFlag* flag, Job job)
 {
     return CallSpecialist(flag->GetPos(), job);
 }
+
+void AIInterface::Chat(const std::string& message)
```
jdoerfert, line 347: Can we add the chat destination here too, otherwise I'll still need to use the old/existing interface for team communication.

jdoerfert (commented):
Cool! Thanks for the quick turn around.
Once landed I might teach the existing bot simple commands, though for that one it's probably harder to do some things I have in mind, e.g., "expand north", "expand slowly", "build XYZ".


Interface to act on messages looks good, one nit on the send interface in the AI interface but that's minor. LGTM otherwise.

libs/s25main/ai/AIInterface.cpp
Flamefire: Good point. Done

Flow86 (changes requested):
conflicted files, please rebase :-/

Flow86 (approved):

Spikeone:
@Flamefire OnChatMessage is missing in the docs, isn't it?

PR #1385, opened 2021-03-01
Cluster mineable resources in random maps

jdoerfert:
The random map generator placed resources purely random in a mountain. This is
not great. Manual maps always cluster resources and we should do so too.
However, having all resources in every mountain is also a good thing for
the balance of the game so we want to keep the clusters small. This is a
try to build small clusters while keeping the ratios the same.

With this patch, mountains (default distribution) look like this:

### DIFF
--- a/libs/s25main/mapGenerator/Resources.cpp
+++ b/libs/s25main/mapGenerator/Resources.cpp
@@ -217,7 +217,7 @@ namespace rttr { namespace mapGenerator {
                 if(randomMRIindex < 0)
                     continue;
 
-                auto &mRI = mRIs[randomMRIindex];
+                auto& mRI = mRIs[randomMRIindex];
 
                 // Adjust and check the budget, if we are over we will not place the resource because we have already in
                 // the past.
--- a/libs/s25main/mapGenerator/Resources.cpp
+++ b/libs/s25main/mapGenerator/Resources.cpp
@@ -217,7 +217,7 @@ namespace rttr { namespace mapGenerator {
                 if(total == 0)
                     continue;
 
-                // Pick a random resource, -1 indicates none.
+                // Pick a random resource
                 unsigned randomMRIindex = resourcePicker();
                 auto& mRI = mRIs[randomMRIindex];
 
--- a/libs/s25main/mapGenerator/Resources.cpp
+++ b/libs/s25main/mapGenerator/Resources.cpp
@@ -21,6 +21,7 @@
 #include "mapGenerator/Algorithms.h"
 #include "mapGenerator/Terrain.h"
 #include "mapGenerator/TextureHelper.h"
+#include "world/MapGeometry.h"
 
 namespace rttr { namespace mapGenerator {
 
@@ -193,49 +194,57 @@ namespace rttr { namespace mapGenerator {
                                         mineableResourceInfo(settings.ratioGranite, libsiedler2::R_Granite)};
         int total = settings.ratioCoal + settings.ratioGold + settings.ratioIron + settings.ratioGranite;
 
+        // Helper to pick an index [0,4) into the mRIs array that identifies a randomly selected resource we should
+        // place. Note that we do not perform budget adjustments and checks here.
+        auto resourcePicker = [&]() {
+            int ratio = 0;
+            int randomNumber = rnd.RandomValue(1, std::max(1, total));
+            for(int i = 0; i < 4; ++i)
+            {
+                ratio += mRIs[i].ratio;
+                if(randomNumber < ratio)
+                    return i;
+            }
+            return -1;
+        };
+
         RTTR_FOREACH_PT(MapPoint, map.size)
         {
             if(textures.All(pt, IsMinableMountain))
             {
-                int ratio = 0;
-                int randomNumber = rnd.RandomValue(1, total);
-                for(auto mRI : mRIs)
+                // Pick a random resource, -1 indicates none.
+                int randomMRIindex = resourcePicker();
+                if(randomMRIindex < 0)
+                    continue;
+
+                auto mRI = mRIs[randomMRIindex];
+
+                // Adjust and check the budget, if we are over we will not place the resource because we have already in
+                // the past.
+                ++mRI.budget;
+                if(mRI.budget <= 0)
+                    continue;
+
+                // Budget is positive so we want to place resources, however, we avoid overwriting existing clusters as
+                // this results in sprinkles. If we have a resource on this field we keep the budget and place a cluster
+                // later.
+                if(resources[pt])
+                    continue;
+
+                // Cluster the resource around the map point.
+                for(int xd = -4; xd <= 4; ++xd)
                 {
-                    ratio += mRI.ratio;
-
-                    // Wrong resource, look further.
-                    if(randomNumber > ratio)
-                        continue;
-
-                    // Right resource, adjust and check the budget.
-                    ++mRI.budget;
-                    if(mRI.budget <= 0)
-                        break;
-
-                    // Budget is positive, avoid overwriting existing clusters though
-                    if(resources[pt])
-                        continue;
-
-                    // Cluster the resource around the map point.
-                    for(int xd = -4; xd <= 4; ++xd)
+                    for(int yd = -4; yd <= 4; ++yd)
                     {
-                        for(int yd = -4; yd <= 4; ++yd)
+                        MapPoint pt_d = MakeMapPoint(Position(pt.x + xd, pt.y + yd), map.size);
+                        // Only place it on mines that have no resource yet, adjust the budget for each placed
+                        // resource.
+                        if(textures.All(pt_d, IsMinableMountain) && !resources[pt_d])
                         {
-                            MapPoint pt_d(pt.x + xd, pt.y + yd);
-                            // Avoid out of bound accesses.
-                            if(pt_d.x >= map.size.x || pt_d.y >= map.size.y)
-                                continue;
-
-                            // Only place it on mines that have no resource yet, adjust the budget for each placed
-                            // resource.
-                            if(textures.All(pt_d, IsMinableMountain) && !resources[pt_d])
-                            {
-                                --mRI.budget;
-                                resources[pt_d] = mRI.resource + rnd.RandomValue(1, 8);
-                            }
+                            --mRI.budget;
+                            resources[pt_d] = mRI.resource + rnd.RandomValue(1, 8);
                         }
                     }
-                    break;
                 }
             } else if(textures.All(pt, IsWater))
             {
--- a/libs/s25main/mapGenerator/Resources.cpp
+++ b/libs/s25main/mapGenerator/Resources.cpp
@@ -188,10 +188,10 @@ namespace rttr { namespace mapGenerator {
             libsiedler2::Resource resource;
             MineableResourceInfo(int ratio, libsiedler2::Resource resource) : ratio(ratio), resource(resource) {}
         };
-        MineableResourceInfo mRIs[4] = {MineableResourceInfo(settings.ratioCoal, libsiedler2::R_Coal),
-                                        MineableResourceInfo(settings.ratioGold, libsiedler2::R_Gold),
-                                        MineableResourceInfo(settings.ratioIron, libsiedler2::R_Iron),
-                                        MineableResourceInfo(settings.ratioGranite, libsiedler2::R_Granite)};
+        auto mRIs = helpers::make_array(MineableResourceInfo(settings.ratioCoal, libsiedler2::R_Coal),
+                                                           MineableResourceInfo(settings.ratioGold, libsiedler2::R_Gold),
+                                                           MineableResourceInfo(settings.ratioIron, libsiedler2::R_Iron),
+                                                           MineableResourceInfo(settings.ratioGranite, libsiedler2::R_Granite));
         int total = settings.ratioCoal + settings.ratioGold + settings.ratioIron + settings.ratioGranite;
 
         // Helper to pick an index [0,4) into the mRIs array that identifies a randomly selected resource we should
--- a/libs/s25main/mapGenerator/Resources.cpp
+++ b/libs/s25main/mapGenerator/Resources.cpp
@@ -239,7 +239,7 @@ namespace rttr { namespace mapGenerator {
                         const MapPoint pt_d = resources.MakeMapPoint(pt + Position(xd, yd));
                         // Only place it on mines that have no resource yet, adjust the budget for each placed
                         // resource.
-                        if(textures.All(pt_d, IsMinableMountain) && !resources[pt_d])
+                        if(!resources[pt_d] && textures.All(pt_d, IsMinableMountain))
                         {
                             --mRI.budget;
                             resources[pt_d] = mRI.resource + rnd.RandomValue(1, 8);
--- a/libs/s25main/mapGenerator/Resources.cpp
+++ b/libs/s25main/mapGenerator/Resources.cpp
@@ -21,7 +21,6 @@
 #include "mapGenerator/Algorithms.h"
 #include "mapGenerator/Terrain.h"
 #include "mapGenerator/TextureHelper.h"
-#include "world/MapGeometry.h"
 
 namespace rttr { namespace mapGenerator {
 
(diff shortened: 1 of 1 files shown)


libs/s25main/mapGenerator/Resources.cpp
```diff
@@ -177,39 +177,65 @@ namespace rttr { namespace mapGenerator {
     {
         const auto& textures = map.textureMap;
         auto& resources = map.resources;
+        struct mineableResourceInfo
```
Flamefire, line 180: Type names should be UpperCamelCase
```suggestion
        struct MineableResourceInfo
```

libs/s25main/mapGenerator/Resources.cpp
```diff
@@ -177,39 +177,65 @@ namespace rttr { namespace mapGenerator {
...
-
-                ratio += settings.ratioGranite;
-                if(randomNumber < ratio)
+                for(auto mRI : mRIs)
                 {
-                    resources[pt] = libsiedler2::R_Granite + rnd.RandomValue(0, 8);
+                    ratio += mRI.ratio;
+
+                    // Wrong resource, look further.
+                    if(randomNumber > ratio)
```
Flamefire, line 207: I'm having a bit of trouble understanding this. What exactly is the "right"/"wrong" resource? To me this looks like it favors e.g. iron over coal. I feel it should pick 1 resource with probability given by the ratios. I understand there is a problem, when that resource is "over budget", but what is supposed to happen then?

libs/s25main/mapGenerator/Resources.cpp
```diff
@@ -177,39 +177,65 @@ namespace rttr { namespace mapGenerator {
...
+                        continue;
+
+                    // Right resource, adjust and check the budget.
+                    ++mRI.budget;
+                    if(mRI.budget <= 0)
+                        break;
```
Flamefire, line 213: This line not being covered looks like there is a bug in the algorithm, I'd expect this to happen quite often, but then again: If you got the "right resource" but it is over budget, isn't it possible, that NO resource will be placed here?

libs/s25main/mapGenerator/Resources.cpp
```diff
@@ -177,39 +177,65 @@ namespace rttr { namespace mapGenerator {
...
+
+                    // Budget is positive, avoid overwriting existing clusters though
+                    if(resources[pt])
+                        continue;
```
Flamefire, line 217: Why the continue here? What is there to do if there already is a resource?

libs/s25main/mapGenerator/Resources.cpp
```diff
@@ -177,39 +177,65 @@ namespace rttr { namespace mapGenerator {
...
+
+                    // Cluster the resource around the map point.
+                    for(int xd = -4; xd <= 4; ++xd)
+                    {
+                        for(int yd = -4; yd <= 4; ++yd)
+                        {
+                            MapPoint pt_d(pt.x + xd, pt.y + yd);
```
Flamefire, line 224: I'd suggest to use either MakeMapPoint or `GetPointsInRadiusWithCenter` (maybe base on https://github.com/Return-To-The-Roots/s25client/pull/1373 which changed that a bit) Currently you are basically ignoring the wrap-around. The latter will make these patches round instead of square, your choice which you like better.
I'm also curious about the `-4`: Wouldn't that always yield a point where resources have been placed already?

libs/s25main/mapGenerator/Resources.cpp
Flamefire: If this is supposed to pick exactly 1 resource: Maybe extract a function for that? This avoids mixing the code with placing resources and budget checking and makes it much easier to understand.

Flamefire (changes requested):
1 naming issue and a few questions. The goal is great and the approach might work. I just want to double-check it does as I don't fully understand all of it.

libs/s25main/mapGenerator/Resources.cpp
jdoerfert: The same logic to "pick" a resource is used as was in the original code.
randomNumber in [0 - total]
[0, ratioCoal] -> coal
[ratioCoal, ratioCoal+ratioGold] -> goal
[ratioCoal+ratioGold, ratioCoal+ratioGold+ratioIron] -> iron
...

the only difference is that I packed it in a loop and didn't unroll it manually as it was before.

libs/s25main/mapGenerator/Resources.cpp
jdoerfert: yes it is. and that is how -4 later comes into play. The original also had empty fields, which I don't find to be problematic.

jdoerfert: the not being covered part might just be the test?

libs/s25main/mapGenerator/Resources.cpp
jdoerfert: This avoids that we try to build a cluster around an existing resource. So if a field is not free, a different cluster used it. Without this we would use surrounding fields that are still free which results in more noise usually.

libs/s25main/mapGenerator/Resources.cpp
jdoerfert: round is fine, -4 is explained above, will change this.

libs/s25main/mapGenerator/Resources.cpp
Flamefire: Ok, but then this should be a break here shouldn't it?

libs/s25main/mapGenerator/Resources.cpp
Flamefire: Just thinking: you pick Randomly the most common res. Place 16 fields of great below, then the next time this resource is picked it should be over budget. But according to coverage it is not. Our am I missing anything?

libs/s25main/mapGenerator/Resources.cpp
Flamefire: I see and should be correct.
Could you create a little function or lambda which (randomly) returns the resource (or index) to use? This would make this code much clearer and avoid the confusion with break VS continue mentioned below.

libs/s25main/mapGenerator/Resources.cpp
jdoerfert: what coverage? 
A resource is randomly picked, placed in a cluster and for each placement budget is decremented by one. So in the end it might be -5. The next 5 times the resource is picked we will not place it but just even out the budget. After that, a new cluster is placed for the resource.

libs/s25main/mapGenerator/Resources.cpp
jdoerfert: correct.

libs/s25main/mapGenerator/Resources.cpp
Flamefire: The break here should be hit if the budget is ever negative. But there was a note here that this line is not covered (check the files view once ci ran through) so it looks like this situation which should occur does not. 
Does that explain what I mean? 

Maybe check in the debugger to make sure this is hit or check back here after ci passed

jdoerfert: Last version was missing a reference, now it hits the continue if you create a random map (so budget goes below 0).

libs/s25main/mapGenerator/Resources.cpp
```diff
@@ -177,39 +178,73 @@ namespace rttr { namespace mapGenerator {
...
+            // The ratio is the user defined ratio for this resource.
+            int ratio;
+            // The resource encoding
+            libsiedler2::Resource resource;
+            MineableResourceInfo(int ratio, libsiedler2::Resource resource) : ratio(ratio), resource(resource) {}
+        };
+        MineableResourceInfo mRIs[4] = {MineableResourceInfo(settings.ratioCoal, libsiedler2::R_Coal),
+                                        MineableResourceInfo(settings.ratioGold, libsiedler2::R_Gold),
+                                        MineableResourceInfo(settings.ratioIron, libsiedler2::R_Iron),
+                                        MineableResourceInfo(settings.ratioGranite, libsiedler2::R_Granite)};
```
Flamefire, line 194: Use a std::array here to avoid the magic number "4" in the resource picker function
```suggestion
        auto mRIs = helpers::make_array(MineableResourceInfo(settings.ratioCoal, libsiedler2::R_Coal),
                                                           MineableResourceInfo(settings.ratioGold, libsiedler2::R_Gold),
                                                           MineableResourceInfo(settings.ratioIron, libsiedler2::R_Iron),
                                                           MineableResourceInfo(settings.ratioGranite, libsiedler2::R_Granite));
```

libs/s25main/mapGenerator/Resources.cpp
```diff
@@ -177,39 +178,73 @@ namespace rttr { namespace mapGenerator {
...
         int total = settings.ratioCoal + settings.ratioGold + settings.ratioIron + settings.ratioGranite;
```
Flamefire, line 195:
```suggestion
        const int total = settings.ratioCoal + settings.ratioGold + settings.ratioIron + settings.ratioGranite;
```

libs/s25main/mapGenerator/Resources.cpp
```diff
@@ -177,39 +178,73 @@ namespace rttr { namespace mapGenerator {
...
 
+        // Helper to pick an index [0,4) into the mRIs array that identifies a randomly selected resource we should
+        // place. Note that we do not perform budget adjustments and checks here.
+        auto resourcePicker = [&]() {
+            int ratio = 0;
+            int randomNumber = rnd.RandomValue(1, std::max(1, total));
+            for(int i = 0; i < 4; ++i)
```
Flamefire, line 202:
```suggestion
            for(int i = 0; i < mRIs.size(); ++i)
```

libs/s25main/mapGenerator/Resources.cpp
```diff
@@ -177,39 +178,73 @@ namespace rttr { namespace mapGenerator {
...
+            {
+                ratio += mRIs[i].ratio;
+                if(randomNumber < ratio)
```
Flamefire, line 205: I think this should be
```suggestion
                if(randomNumber <= ratio)
```

Example: All ratios are 20, total is 100 --> 1-20 is first, 21-40 is 2nd, ... 81 to 100 is 4th. So 20 values per index
Currently if randomNumber is 100 no value will be picked and (at least) the first would only have 19 "chances" while the 2nd has 20

libs/s25main/mapGenerator/Resources.cpp
```diff
@@ -177,39 +178,73 @@ namespace rttr { namespace mapGenerator {
...
+                    return i;
+            }
+            return -1;
```
Flamefire, line 208: This should never happen, should it? But it makes me think: What if all ratios are 0? Then no resources should be placed, right? Maybe simply assert that here and never call the function for that (place the `if(total==0)` in the loop below) or check that first in the function and return -1 right away. Either way you can then also remove  the `max` call.

As you have to return something, the loop can go to `mRIs.size() - 1` and you return that here, no need to add and check the ratio for the last element anyway as it will always be true after the change to `<=` above

libs/s25main/mapGenerator/Resources.cpp
```diff
@@ -177,39 +178,73 @@ namespace rttr { namespace mapGenerator {
...
 
-                ratio += settings.ratioGranite;
-                if(randomNumber < ratio)
+                // Cluster the resource around the map point.
+                for(int xd = -4; xd <= 4; ++xd)
                 {
-                    resources[pt] = libsiedler2::R_Granite + rnd.RandomValue(0, 8);
+                    for(int yd = -4; yd <= 4; ++yd)
+                    {
+                        MapPoint pt_d = MakeMapPoint(Position(pt.x + xd, pt.y + yd), map.size);
```
Flamefire, line 239:
```suggestion
                        const MapPoint pt_d = resources.MakeMapPoint(pt + Position(xd, yd));
```
reads a bit simpler :)

libs/s25main/mapGenerator/Resources.cpp
```diff
@@ -177,39 +178,73 @@ namespace rttr { namespace mapGenerator {
...
+                        // Only place it on mines that have no resource yet, adjust the budget for each placed
+                        // resource.
+                        if(textures.All(pt_d, IsMinableMountain) && !resources[pt_d])
```
Flamefire, line 242: Switching the conditions should make this quite a bit faster:
```suggestion
                        if(!resources[pt_d] && textures.All(pt_d, IsMinableMountain))
```

libs/s25main/mapGenerator/Resources.cpp
```diff
@@ -177,39 +178,73 @@ namespace rttr { namespace mapGenerator {
...
+                int randomMRIindex = resourcePicker();
+                if(randomMRIindex < 0)
                     continue;
-                }
 
-                ratio += settings.ratioCoal;
-                if(randomNumber < ratio)
-                {
-                    resources[pt] = libsiedler2::R_Coal + rnd.RandomValue(0, 8);
+                auto &mRI = mRIs[randomMRIindex];
```
Flamefire, line 220: Clang-format:
```suggestion
                auto& mRI = mRIs[randomMRIindex];
```

libs/s25main/mapGenerator/Resources.cpp
```diff
@@ -21,6 +21,7 @@
 #include "mapGenerator/Algorithms.h"
 #include "mapGenerator/Terrain.h"
 #include "mapGenerator/TextureHelper.h"
+#include "world/MapGeometry.h"
```
Flamefire, line 24:
```suggestion
```

No need. `NodeMapBase` already has a MakeMapPoint member

Flamefire (changes requested):
Looking good so far.
A bit polishing, see inline suggestions (typed here, so might be slightly wrong or misformatted)
Only actual change required additionally would be to the `resourcePicker` function to explicitly handle the case for "no resources"

While on it: We could think about what should happen, if the resource ratios don't add up to 100. Say all ratios are 10%, so (maybe) half the mountain spots should be empty (sum is 50%). Not sure if this is useful and it would require quite some work thinking about how to handle those cases and also when the sum is over 100% etc. but maybe some players want that? @Spikeone Have you ever designed such maps (manually)?
This should be done in a follow-up though, to get this in first :)

libs/s25main/mapGenerator/Resources.cpp
jdoerfert: This is for the 0 case. I can rewrite it obviously...

libs/s25main/mapGenerator/Resources.cpp
```diff
@@ -177,39 +178,75 @@ namespace rttr { namespace mapGenerator {
...
+                // No mineable resources wanted.
+                if(total == 0)
                     continue;
-                }
 
-                ratio += settings.ratioCoal;
-                if(randomNumber < ratio)
-                {
-                    resources[pt] = libsiedler2::R_Coal + rnd.RandomValue(0, 8);
+                // Pick a random resource, -1 indicates none.
```
Flamefire, line 220:
```suggestion
                // Pick a random resource
```

Or alternatively rename the lambda to `pickRandomResourceIdx` then the comment can be removed completely

Flamefire (changes requested):
1 outdated comment and a test failure:

`c:/projects/s25client/tests/s25main/mapgenerator/testresources.cpp(162): fatal error: in "ResourcesTests/AddResources_updates_resources_according_to_textures": critical check map.resources[pt] != R_None has failed [0 == 0]`

So the test needs to be improved as "no resource" is possible. Maybe check that there are just a few? Can we estimate how many are "ok"?
I expect that for very imbalanced distributions, say 80-5-5-5-5 there could be many free spots.

Flamefire:
@Flow86 Why do you merge PRs that have test failures?

Flow86:
it had?

Flamefire:
Yes, check the status for https://github.com/Return-To-The-Roots/s25client/pull/1385/commits/7d47998c03c82df5bd7121eb75fbd66680399c45 which shows a red mark for the commit, on appveyor and I had written that in my review at https://github.com/Return-To-The-Roots/s25client/pull/1385#pullrequestreview-602274504 . I was waiting for an answer there

Flamefire:
@jdoerfert Any chance you could change/fix the test or code in the near future? You could either completely avoid "no resources" somewhow, or update the test to be less strict. Not sure here.

jdoerfert:
> @jdoerfert Any chance you could change/fix the test or code in the near future? You could either completely avoid "no resources" somewhow, or update the test to be less strict. Not sure here.

Unsure when I get to it :(

PR #1386, opened 2021-03-01
Simplify Gameobject

Flamefire:
This mainly removes Serialize_* and Destroy_* functions and makes GetGOT final where possible.

I also found a bug for noNothing and removed the problematic enumerator completely.
Finally I double-checked the ownership and introduced unique_ptr in the SGD to make it clear.


libs/s25main/buildings/noBaseBuilding.cpp
```diff
@@ -82,7 +82,7 @@ noBaseBuilding::noBaseBuilding(const NodalObjectType nop, const BuildingType typ
 
 noBaseBuilding::~noBaseBuilding() = default;
 
-void noBaseBuilding::Destroy_noBaseBuilding()
+void noBaseBuilding::Destroy()
```
Flow86, line 85: shouldn't this override?

libs/s25main/buildings/noBuildingSite.h
```diff
@@ -69,27 +69,16 @@ class noBuildingSite : public noBaseBuilding
...
-
-    /// Serialisierungsfunktionen
-protected:
-    void Serialize_noBuildingSite(SerializedGameData& sgd) const;
-
-public:
-    void Serialize(SerializedGameData& sgd) const override { Serialize_noBuildingSite(sgd); }
-
-    GO_Type GetGOT() const override { return GO_Type::Buildingsite; }
+    GO_Type GetGOT() const final { return GO_Type::Buildingsite; }
```
Flow86, line 75: we could also make a template out of the base class and give it that enum on definition?

Flamefire: That would mean we need a template which increases compile times.
Also it is not really useful here: This is exactly what a pure virtual method is for, so what would we gain? Even worse: We have this quite some way down the hierarchy chain

PR #1391, opened 2021-03-07
Fix memory leaks

Flamefire:
- [x] Based on #1387

I noticed memory leaks due to the harbor building not deleting its figures and wares.
To eliminate that once and for all I introduced a strong ownership model via unique_ptr and also replaced (some) instances of pointers by references in arguments (doing all at once is way to much work...)


Flow86 (changes requested):
conflicts

Flamefire:
Rebased

Flamefire:
Test failure due to merge of https://github.com/Return-To-The-Roots/s25client/pull/1385 not by this PR

Issue #1383, opened 2021-02-25
Compress savegames and replays


Flamefire:
I think it makes sense to compress the data part of savegames and replays. The header should be stored uncompressed for fast access in the list but the game data (e.g. savegame, commands, ...) should be compressed.
For instance a normal replay of about 200k GFs is about 800kB. It can be easily compressed to about half the size.
For a replay starting from a savegame the size is about 4MB and can be compressed to about 400kB

I noticed that when creating the PR for testing replays for asyncs, because we really don't want large binary data in the repo.

Flow86:
you mean for transfer between players (aka on game start?)? On disk I don't care to be honest?

Flamefire:
No I meant for on disk. I know disk space is cheap and everything but the savegames and replays can pile up. And being able to "simply" reduce the size by a factor of 2-10 makes me feel guilty for wasting space. (Again, I have no problem with *using* space, just wasting it)
We already have a more or less simple way for compressing data so we could implement this. The only issue is, that this requires increasing the format version as we need to store a flag if the data is compressed. This means old savegames and replays can't be loaded anymore but at least we can drop all the compatibility workarounds, especially after #1367 is merged.

For replays it is a bit tricky because we need to have a compressed and uncompressed format. The latter is used for directly writing to disk so we have the replay even if the program is force-terminated/crashed. My idea was that on closing the file we simply copy the header and the compress the rest after setting the flag, that the data is now compressed.

Flow86:
hmm its making everything so much more complicated, I don't like that

PR #1393, opened 2021-03-13
Improve speed of RoadPathFinder

Flamefire:
I discovered 2 main hotspots in the roadpatfinder:
- Calling GetNeighbour due to multiple, duplicate checks and pointer dereferences -> Introduce GetRoutes to get all routes at once
- `Flag::GetPunishmentPoints` iterated over an array of 8 wares, that is usually empty and can be skipped. Using a fixed capacity vector avoids that but breaks replays

As replays were already broken I decided to introduce the compression I mentioned in  #1383 which shrinks savegames by 90% and replays starting at GF 0 by 25-50% which IMO is worth it, especially as the tests contain replays so the repo size gets reduced

- [x] Based on #1391


libs/s25main/Replay.cpp
```diff
@@ -32,28 +33,76 @@ uint16_t Replay::GetVersion() const
...
+        if(data.size() >= uncompressedSize)
+            return true;
+        compressedReplay.WriteUnsignedInt(uncompressedSize);
+        compressedReplay.WriteUnsignedInt(data.size());
+        compressedReplay.WriteRawData(data.data(), data.size());
+
+        // All done. Replace uncompressed replay
+        compressedReplay.Close();
+        file.Close();
+        boost::filesystem::rename(tmpReplayFile.filePath, filepath_);
```
Flow86, line 99: does this overwrite existing files?

libs/s25main/gameTypes/CompressedData.cpp
```diff
@@ -33,61 +34,82 @@ bool CompressedData::DecompressToFile(const boost::filesystem::path& filePath, u
...
-    unsigned compressedLen = data.size();
-    int err = BZ2_bzBuffToBuffCompress(&data[0], &compressedLen, uncompressedData.get(), length, 9, 0, 250);
-    if(err != BZ_OK)
+    try
+    {
+        data = compress(uncompressedData);
+    } catch(const std::runtime_error& err)
     {
-        LOG.write("FATAL ERROR: BZ2_bzBuffToBuffCompress failed with error: %d\n") % err;
+        LOG.write("FATAL ERROR: 1%\n") % err.what();
```
Flow86, line 76: isnt there a % missing before the 1?

libs/s25main/Replay.cpp
Flamefire: Yes and it should as it is meant to overwrite the just created replay: https://www.boost.org/doc/libs/1_45_0/libs/filesystem/v3/doc/reference.html#rename

libs/s25main/gameTypes/CompressedData.cpp
Flamefire: Yes, good catch!
It is practically impossible to trigger this codepath though, so can't be tested: http://linux.math.tifr.res.in/manuals/html/manual_3.html#SEC36

Flow86 (changes requested):
conflict, rebase necessary

PR #1394, opened 2021-03-14
Rename GameWorldGame to GameWorld

Flamefire:
We now have GameWorldBase and GameWorld. The (IMO) confusing GameWorldGame was renamed after the loader class GameWorld was removed earlier

### DIFF
--- a/libs/s25main/GameObject.cpp
+++ b/libs/s25main/GameObject.cpp
@@ -19,7 +19,7 @@
 #include "EventManager.h"
 #include "SerializedGameData.h"
 #include "postSystem/PostMsg.h"
-#include "world/GameWorldGame.h"
+#include "world/GameWorld.h"
 #include <iostream>
 
 /**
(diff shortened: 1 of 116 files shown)


libs/s25main/GameObject.cpp
```diff
@@ -28,7 +28,7 @@
 unsigned GameObject::objIdCounter_ = 0;
 unsigned GameObject::objCounter_ = 0;
 
-GameWorldGame* GameObject::gwg = nullptr;
+GameWorld* GameObject::world = nullptr;
```
Flow86, line 31: we can't really get rid of this static world efficiently, do we?

Flamefire: Every gameobject needs access to it somehow. With a lot of refactoring we could pass that into the HandleEvent function which then passes it down to where it is needed. Or pass it through all ctors and store in GameObject. Likely easier.

PR #1397, opened 2021-03-19
Fix test failures in mapGenerator TestResources

Flamefire:


tests/common/testContainerUtils.cpp
```diff
@@ -158,4 +158,20 @@ BOOST_AUTO_TEST_CASE(RemoveIf)
     BOOST_TEST(vecIn == vecExp, boost::test_tools::per_element());
 }
 
+BOOST_AUTO_TEST_CASE(Count)
+{
+    std::vector<int> values = {1, 2, 2, 4, 4, 4};
+    BOOST_TEST(helpers::count(values, 1) == 1u);
+    BOOST_TEST(helpers::count(values, 2) == 2u);
+    BOOST_TEST(helpers::count(values, 3) == 0u);
+    BOOST_TEST(helpers::count(values, 4) == 3u);
```
Flow86, line 167: you should probably add a negative test too:
```suggestion
BOOST_TEST(helpers::count(values, 0) == 0u);
```

Flamefire: What does that test, that isn't already tested? Also keep in  mind, that this is a simple wrapper function, hence testing the algorithm is already done, so the tests can be minimal to check if it works at all.

PR #1399, opened 2021-03-21
Update Lua version in build documentation 

Mailaender:
Followup of https://github.com/Return-To-The-Roots/s25client/commit/99207edef86830f66193d312e68e25cf441fe8cf.


Flamefire:
I think we should still recommend 5.2, but I'd appreciate a comment instead about using 5.1 or 5.3 as an alternative

Mailaender:
Updated.

Flamefire (approved):

PR #1402, opened 2021-03-25
Fix a double-free when a defender is chosen out of the leave list

Flamefire:
See https://www.siedler25.org/index.php?com=forum&mod=forum&action=thread&id=1677&start=-1#last

### DIFF
--- a/libs/s25main/GameManager.cpp
+++ b/libs/s25main/GameManager.cpp
@@ -33,9 +33,9 @@
 #include "ogl/glArchivItem_Bitmap.h"
 #include "liblobby/LobbyClient.h"
 #include "libsiedler2/Archiv.h"
-#include "s25util//dynamicUniqueCast.h"
 #include "s25util/Log.h"
 #include "s25util/error.h"
+#include <boost/pointer_cast.hpp>
 
 GameManager::GameManager(Log& log, Settings& settings, VideoDriverWrapper& videoDriver, AudioDriverWrapper& audioDriver,
                          WindowManager& windowManager)
@@ -174,7 +174,7 @@ bool GameManager::ShowSplashscreen()
     libsiedler2::Archiv arSplash;
     if(!LOADER.Load(arSplash, RTTRCONFIG.ExpandPath(s25::files::splash)))
         return false;
-    auto image = libutil::dynamicUniqueCast<glArchivItem_Bitmap>(arSplash.release(0));
+    auto image = boost::dynamic_pointer_cast<glArchivItem_Bitmap>(arSplash.release(0));
     if(!image)
     {
         s25util::error(_("Splash screen couldn't be loaded!\n"));
--- a/libs/s25main/Loader.cpp
+++ b/libs/s25main/Loader.cpp
@@ -56,9 +56,9 @@
 #include "s25util/Log.h"
 #include "s25util/StringConversion.h"
 #include "s25util/System.h"
-#include "s25util/dynamicUniqueCast.h"
 #include "s25util/strAlgos.h"
 #include <boost/filesystem.hpp>
+#include <boost/pointer_cast.hpp>
 #include <boost/range/adaptor/map.hpp>
 #include <algorithm>
 #include <chrono>
@@ -257,7 +257,7 @@ bool Loader::LoadSounds()
         try
         {
             libsiedler2::Archiv sng = archiveLoader_->loadFileOrDir(oggFile);
-            auto music = libutil::dynamicUniqueCast<MusicItem>(sng.release(0));
+            auto music = boost::dynamic_pointer_cast<MusicItem>(sng.release(0));
             if(music)
                 sng_lst.emplace_back(std::move(music));
             else
--- a/libs/s25main/Loader.cpp
+++ b/libs/s25main/Loader.cpp
@@ -833,7 +833,7 @@ void Loader::fillCaches()
                 bitmap->setNx(image->getNx());
                 bitmap->setNy(image->getNy());
 
-                bmp.add(bitmap.release(), true);
+                bmp.add(std::move(bitmap));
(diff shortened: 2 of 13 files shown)


libs/s25main/buildings/nobBaseWarehouse.cpp
```diff
@@ -1026,15 +1026,14 @@ std::unique_ptr<nofDefender> nobBaseWarehouse::ProvideDefender(nofAttacker& atta
...
-        nofSoldier* soldier;
+        std::unique_ptr<nofSoldier> soldier;
         // Soldat?
         if((*it)->GetGOT() == GO_Type::NofAggressivedefender)
         {
-            auto* aggDefender = static_cast<nofAggressiveDefender*>(it->get());
-            aggDefender->NeedForHomeDefence();
-            soldier = aggDefender;
+            soldier.reset(static_cast<nofSoldier*>(it->release()));
+            static_cast<nofAggressiveDefender&>(*soldier).NeedForHomeDefence();
```
Flow86, line 1034: why not use `dynamic_pointer_cast` ?

libs/s25main/buildings/nobBaseWarehouse.cpp
```diff
@@ -1026,15 +1026,14 @@ std::unique_ptr<nofDefender> nobBaseWarehouse::ProvideDefender(nofAttacker& atta
...
         } else if((*it)->GetGOT() == GO_Type::NofPassivesoldier)
-            soldier = static_cast<nofPassiveSoldier*>(it->get());
+            soldier.reset(static_cast<nofSoldier*>(it->release()));
```
Flow86, line 1036: and here couldn't we simply write something like
```suggestion
soldier = std::dynamic_pointer_cast<nofSoldier>(*it));
```

libs/s25main/buildings/nobBaseWarehouse.cpp
Flow86: oh right, thats only for shared_ptrs. didnt we have something like `dynamic_unique_cast` already somewhere? I thought we have

libs/s25main/buildings/nobBaseWarehouse.cpp
Flamefire: Because a dynamic cast is to much here. We already pay the virtual cost by the GetGOT call above, so no need to pay it again.
However I'm thinking about whether we should add our own methods for dynamic_pointer_cast of unique ptrs (yes we have one already, but that is destructive in case of a type mismatch) and static_pointer_cast (that can be destructive in all cases)

Flamefire: Done

libs/s25main/buildings/nobBaseWarehouse.cpp
Flamefire: Found https://www.boost.org/doc/libs/1_64_0/libs/smart_ptr/pointer_cast.html and removed ours

Flow86: true I always forget that we have boost to our command

libs/s25main/GameManager.cpp
```diff
@@ -174,7 +174,7 @@ bool GameManager::ShowSplashscreen()
     libsiedler2::Archiv arSplash;
     if(!LOADER.Load(arSplash, RTTRCONFIG.ExpandPath(s25::files::splash)))
         return false;
-    auto image = libutil::dynamicUniqueCast<glArchivItem_Bitmap>(arSplash.release(0));
+    auto image = boost::dynamic_pointer_cast<glArchivItem_Bitmap>(arSplash.release(0));
```
Flow86, line 177: do we still need to call `.release(...)`?

libs/s25main/Loader.cpp
```diff
@@ -257,7 +257,7 @@ bool Loader::LoadSounds()
         try
         {
             libsiedler2::Archiv sng = archiveLoader_->loadFileOrDir(oggFile);
-            auto music = libutil::dynamicUniqueCast<MusicItem>(sng.release(0));
+            auto music = boost::dynamic_pointer_cast<MusicItem>(sng.release(0));
```
Flow86, line 260: do we still need to call `.release(...)`?

libs/s25main/TerrainRenderer.cpp
```diff
@@ -145,7 +144,7 @@ void TerrainRenderer::LoadTextures(const WorldDescription& desc)
                 for(unsigned i = 0; i < textures->size(); i++)
                 {
                     terrainTextures[curIdx.value].textures.push_back(
-                      libutil::dynamicUniqueCast<glArchivItem_Bitmap>(textures->release(i)).release());
+                      dynamic_cast<glArchivItem_Bitmap*>(textures->release(i).release()));
```
Flow86, line 147: do we still need to call `.release(...)`?

libs/s25main/GameManager.cpp
Flow86: ah stupid me, thats a libsiedler archive.

libs/s25main/Loader.cpp
Flow86: ah stupid me, thats a libsiedler archive.

libs/s25main/TerrainRenderer.cpp
Flow86: isnt this a potential memory leak? if the dynamic cast returns null but we still release the pointer?

Flamefire: In theory yes, however the ptr_vector would throw (or assert?) if a nullptr was added, so we assume already, that the cast is successful, which it must be.
Still found, that we can put unique_ptr directly into those vectors, so even the possible leak is now fixed.

PR #1403, opened 2021-03-28
Remove version from window title

### LINKED ISSUE #1378: Remove version from window title
Issue from Discord:

> can you change the title of the game when running? So that my friends can see when I play it here on discord?  Now I can add the game, but the title is just the version
> so it shows like:
> AdamPodstavka
> Playing v0.8.2-0383c0c

I guess this only applies to the stable in that extend but IMO we don't need the version in the window title. I'd vote to remove it as it is already in the lower left corner of the menu

CCing @Flow86

Flamefire:
Closes #1378


Flow86:
I think we added it once so people directly see their version even on host game screen (to check that they have the same version, etc)

PR #1415, opened 2021-04-28
Fix AI AdjustSettings to create correct tools

### LINKED ISSUE #1084: KI stops functioning correctly after some time
Hello Dev-Team, 
i have a strange problem, i try to explain it as good as i can.

I like to play Maps with ships, but somehow the KI doesn't work properly with it. 
After some time they stop producing tools except pickaxes, i looked it up by changing the player to the KI. In the Tools-Order-Menu every tool was on 0% except pickaxes. 

Yeah you probably see the problem already.

This leads to the issue, that in a further gameplay for example the shipyard doesn't get a worker and the KI stops expanding. Which leads to winning because you can do your thing without interruption. Boring.

I attached a savegame and replay of my latest game, i tried a really big map sorry for the size.
[bug2.zip](https://github.com/Return-To-The-Roots/s25client/files/3454883/bug2.zip)

Keep up the great work on this. :)

crhbetz:
### Disclaimer: This is the first time I ever touched C++ code. I'm a python hobbyist with no professional background. I hope I can make a humble contribution to this great project.

As mentioned on the forums and in an issue (fixes #1084), AI tended to build wrong tools. To investigate, I privately added logging corresponding to the commented tools in `AIPlayerJH.cpp` `AdjustSettings`:

https://github.com/Return-To-The-Roots/s25client/blob/85a6b994185edbe1479462efc640abcc9a048fb9/libs/s25main/ai/aijh/AIPlayerJH.cpp#L2364-L2368

Running a game observing this logging, then saving the game and returning as one of the AIs, I saw they set up their tools wrong.

I found the order in `ToolConsts.h` matched the order of the ingame interface, but the indexes used in `AIPlayerJH.cpp` did not match this - as you can see, `GoodType::Saw` would be index 3 here, while above `toolsettings[2]` is commented as `Saw`.

https://github.com/Return-To-The-Roots/s25client/blob/85a6b994185edbe1479462efc640abcc9a048fb9/libs/s25main/gameData/ToolConsts.h#L24-L37

After matching the indexes in `AIPlayerJH.cpp` to `ToolConsts.h`, which is what this PR does, local testing shows AI producing expected tools again.

In case direct contact would be required, I joined the discord server as `Spezus`.

### DIFF
--- a/libs/s25main/ai/aijh/AIPlayerJH.cpp
+++ b/libs/s25main/ai/aijh/AIPlayerJH.cpp
@@ -2362,20 +2362,21 @@ void AIPlayerJH::AdjustSettings()
     ToolSettings toolsettings;
     const Inventory& inventory = aii.GetInventory();
     // Saw
-    toolsettings[2] = (inventory.goods[GoodType::Saw] + inventory.people[Job::Carpenter] < 2) ?
+    toolsettings[3] = (inventory.goods[GoodType::Saw] + inventory.people[Job::Carpenter] < 2) ?
                         4 :
                         inventory.goods[GoodType::Saw] < 1 ? 1 : 0;
+
     // Pickaxe
-    toolsettings[3] = (inventory.goods[GoodType::PickAxe] < 1) ? 1 : 0;
+    toolsettings[4] = (inventory.goods[GoodType::PickAxe] < 1) ? 1 : 0;
     // Hammer
-    toolsettings[4] = (inventory.goods[GoodType::Hammer] < 1) ? 1 : 0;
+    toolsettings[1] = (inventory.goods[GoodType::Hammer] < 1) ? 1 : 0;
     // Crucible
     toolsettings[6] = (inventory.goods[GoodType::Crucible] + inventory.people[Job::IronFounder]
                        < bldPlanner->GetNumBuildings(BuildingType::Ironsmelter) + 1) ?
                         1 :
                         0;
     // Scythe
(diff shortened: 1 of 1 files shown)


Flow86:
We probably should simply change the magic number indexes to the GoodType indexes? (like instead of 
```
toolsettings[2]
```
we simply use
```
toolsettings[GoodType::Saw]
```

Flamefire:
@crhbetz Thanks for your PR, this is greatly appreciated!
Quick question: Have you verified, that each index corresponds to the expected Tool in the TOOLS array? If so I'd accept this PR

@Flow86 We should certainly do so. But it isn't that easy: The tools are non-contiguous. So either we need a new enum Tools, which IMO makes sense and allows using the enum-foreach-loop and enum-array, or ToolSettings should map between goods and tool-indices by searching the TOOLS array.
I'd go for the former, as the confusion of having `GoodType::Axe` and `Tool::Axe` is not to bad compared to having to map between stuff repeatadly and only be able to catch mistakes at runtime.

crhbetz:
> Quick question: Have you verified, that each index corresponds to the expected Tool in the TOOLS array? If so I'd accept this PR

@Flamefire I just watched AIs play with my added logging and re-loaded the game as the respective AI when they had intended to change their tool setups, to check if the settings showed up in the UI accordingly. Repeated this until I had each individual tool confirmed multiple times. The UI always matched my log output. I'm thus fairly confident the tool setup now works as intended in the code.

Or did you have another form of verification in mind? :sweat_smile:

Flamefire (approved):
Thanks!

PR #1422, opened 2021-05-22
Introduce Tool enum

Flamefire:
Followup to #1415 to avoid magic numbers which has lead to the bug(s) fixed in that PR by using a strong-enum.

### DIFF
--- a/libs/s25main/ai/aijh/AIPlayerJH.cpp
+++ b/libs/s25main/ai/aijh/AIPlayerJH.cpp
@@ -2346,59 +2346,45 @@ unsigned AIPlayerJH::GetNumAIRelevantSeaIds() const
 void AIPlayerJH::AdjustSettings()
 {
     // update tool creation settings
-    ToolSettings toolsettings;
+    ToolSettings toolsettings{};
     const Inventory& inventory = aii.GetInventory();
-    // Saw
-    toolsettings[3] = (inventory.goods[GoodType::Saw] + inventory.people[Job::Carpenter] < 2) ?
-                        4 :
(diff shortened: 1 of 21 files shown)


libs/s25main/ai/aijh/AIPlayerJH.cpp
```diff
@@ -2346,59 +2346,45 @@ unsigned AIPlayerJH::GetNumAIRelevantSeaIds() const
...
+                                      < bldPlanner->GetNumBuildings(BuildingType::Bakery) + 1) ?
+                                       1 :
+                                       0;
+    toolsettings[Tool::Shovel] = (!hasBasicToolsOrder && inventory.goods[GoodType::Shovel] < 1) ? 1 : 0;
+    toolsettings[Tool::Axe] =
+      (!hasBasicToolsOrder && inventory.goods[GoodType::Axe] + inventory.people[Job::Woodcutter] < 12
+       && inventory.goods[GoodType::Axe] < 1) ?
+        1 :
+        0;
+    toolsettings[Tool::Tongs] = 0; //(!hasBasicToolsOrder&&(aii.GetInventory().goods[GoodType::Tongs]<1))?1:0;
```
Flow86, line 2374: so it never orders these? could it be that this is the "bug" why the AI does not produce any gold or weapons anymore at some point?

Flow86 (approved):

libs/s25main/ai/aijh/AIPlayerJH.cpp
Flamefire: Might be. I improved this by using a smart algorithm taking the buildings into account which need those tools

Flow86 (approved):

PR #1423, opened 2021-05-23
Add Arm optimization flags for Raspberry Pi 4

eltorrero:
Hi guys,

nothing big, but as I added this for myself I thought I send these changes to you as well.

KR,
Robert


Flamefire:
Looks correct (don't have a Raspi to test, so assuming it is), thanks!

Flamefire (approved):

PR #1431, opened 2021-08-19
Fix build with Boost 1.77 (missing <algorithm> include)

thesamesam:
Fix build with Boost 1.77 (missing <algorithm> include)

Needed for std::transform which is defined within <algorithm>. The
build fails with newer versions	of Boost, presumably because of
a missing indirect include.

Closes: https://bugs.gentoo.org/808767


Flow86 (approved):

Flamefire:
Thanks a lot!

PR #1432, opened 2021-08-21
keep window positions

### LINKED ISSUE #1380: In game window position should remembered
Either explicitly or implicitly we should remember the in game window positions even across restarts.

JonathanSteinbuch:
This makes ingame window positions persistent between game sessions. When an ingame window class is destructed the position and whether it was still open is written to the settings. At the opening of the GameInterface for the next game each status is checked and relevant windows opened.

Also the options whether to show build-quality, building names and productivity are saved.

All these ingame settings are stored in a new ini file to keep the main config file clutter free and so main settings are not reset when more ingame options are added.

Closes #1380. 

PS: After moving halfway across the country and starting a new job I finally found time and peace of mind to turn this feature into a state in which I think it could be merged into master... that's why it took so long


libs/s25main/Game.cpp
```diff
@@ -32,15 +30,7 @@ void Game::Start(bool startFromSave)
     if(startFromSave)
         CheckObjective();
     else
-    {
-        if(ggs_.objective == GameObjective::EconomyMode)
```
Flow86, line 36: we probably do not want that the addon is removed?

libs/s25main/desktops/dskGameInterface.cpp
```diff
@@ -203,6 +183,47 @@ void dskGameInterface::StartScrolling(const Position& mousePos)
...
+        WINDOWMANAGER.ShowAfterSwitch(std::make_unique<iwMusicPlayer>());
+    if(windows[CGI_STATISTICS].isOpen)
+        WINDOWMANAGER.ShowAfterSwitch(std::make_unique<iwStatistics>(gwv.GetViewer()));
+    if(windows[CGI_ECONOMICPROGRESS].isOpen && gwv.GetWorld().getEconHandler())
+        WINDOWMANAGER.ShowAfterSwitch(std::make_unique<iwEconomicProgress>(gwv.GetViewer()));
+    if(windows[CGI_DIPLOMACY].isOpen)
+        WINDOWMANAGER.ShowAfterSwitch(std::make_unique<iwDiplomacy>(gwv.GetViewer(), GAMECLIENT));
+    if(windows[CGI_SHIP].isOpen)
+        WINDOWMANAGER.ShowAfterSwitch(
+          std::make_unique<iwShip>(gwv, GAMECLIENT, gwv.GetViewer().GetPlayer().GetShipByID(0)));
```
Flow86, line 222: does this really work? what if he opens a game where no ship is there?

libs/s25main/ingameWindows/IngameWindow.cpp
```diff
@@ -45,8 +47,9 @@ IngameWindow::IngameWindow(unsigned id, const DrawPoint& pos, const Extent& size
     // Load last position or center the window
     if(pos == posLastOrCenter)
     {
-        if(id < last_pos.size() && last_pos[id].isValid())
-            SetPos(last_pos[id]);
+        auto& settings = SETTINGS.windows.persistentSettings;
+        if(helpers::contains(persistentWindows, GetID()) && settings[GetID()].lastPos.isValid())
+            SetPos(settings[GetID()].lastPos);
```
Flow86, line 52: does this check if the window location is "out of bounds"?

libs/s25main/ingameWindows/IngameWindow.h
```diff
@@ -75,6 +73,8 @@ class IngameWindow : public Window
     void MouseLeftUp(const MouseCoords& mc);
     void MouseMove(const MouseCoords& mc);
 
+    GUI_ID GetID() const { return static_cast<GUI_ID>(Window::GetID()); }
```
Flow86, line 76: what if its not a GUI_ID?

libs/s25main/ingameWindows/IngameWindows.h
```diff
@@ -0,0 +1,86 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Return To The Roots. If not, see <http://www.gnu.org/licenses/>.
+
+#pragma once
+
+#include "iwAIDebug.h"
```
Flow86, line 20: if you only need these in the cpp, include it there, otherwise you include it everywhere you include this file

libs/s25main/world/GameWorldView.h
```diff
@@ -38,11 +38,11 @@ class GameWorldView
     std::vector<IDrawNodeCallback*> drawNodeCallbacks;
 
     /// Show building quality icons
-    bool show_bq;
+    bool& show_bq;
```
Flow86, line 41: this is so extremly implicit/intransparent what it changes. Can we have a better way? i.e directly use the SETTINGS then?

libs/s25main/ingameWindows/iwMinimap.cpp
```diff
@@ -41,7 +44,16 @@ iwMinimap::iwMinimap(IngameMinimap& minimap, GameWorldView& gwv)
     // Fenster vergrößern/verkleinern
     AddImageButton(4, DrawPoint(0, 0), BUTTON_SIZE, TextureColor::Grey, LOADER.GetImageN("io", 109));
 
-    Resize(GetSize());
+    Resize(extended ? MINIMAP_SIZE_BIG : MINIMAP_SIZE);
+}
+
+iwMinimap::~iwMinimap()
```
Flamefire, line 50: Missing the parent call which saves the position

libs/s25main/ingameWindows/iwMinimap.cpp
```diff
@@ -41,7 +44,16 @@ iwMinimap::iwMinimap(IngameMinimap& minimap, GameWorldView& gwv)
...
+{
+    // Possibly save our old position
```
Flamefire, line 52: That's not what this code does, is it?

libs/s25main/ingameWindows/iwMinimap.cpp
```diff
@@ -41,7 +44,16 @@ iwMinimap::iwMinimap(IngameMinimap& minimap, GameWorldView& gwv)
...
+    if(helpers::contains(persistentWindows, GetID()))
+    {
+        SETTINGS.windows.persistentSettings[GetID()].option = (unsigned)extended;
```
Flamefire, line 55:
```suggestion
        SETTINGS.windows.persistentSettings[GetID()].option = static_cast<unsigned>(extended);
```

libs/s25main/Game.cpp
Flamefire: It was moved to GameClient, but it should stay here

libs/s25main/world/GameWorldView.h
Flamefire: Agreed. Maybe even better: Where this is created, also set the options after creation from outside and when removed save it to SETTINGS?
I imagine having multiple of these instances for split-screen somewhere in the future.

tests/s25Main/integration/testEconomyMode.cpp
```diff
@@ -50,6 +50,7 @@ class EconModeFixture : WorldFixture<CreateEmptyWorld, 3>
 
 BOOST_FIXTURE_TEST_CASE(EconomyMode3Players, EconModeFixture)
 {
+    world.setEconHandler(std::make_unique<EconomyModeHandler>(1000));
```
Flamefire, line 53: Remove after moving this back

tests/s25Main/integration/testEconomyMode.cpp
```diff
@@ -108,6 +109,8 @@ BOOST_FIXTURE_TEST_CASE(EconomyModeSerialization, EconModeFixture)
 {
     ggs.setSelection(AddonId::ECONOMY_MODE_GAME_LENGTH,
                      rttr::test::randomValue<unsigned>(1, AddonEconomyModeGameLengthList.size() - 1));
+    world.setEconHandler(
```
Flamefire, line 112: Remove after moving this back

Flamefire (changes requested):
Thanks a lot, this looks good so far. Just a bit polishing :)

libs/s25main/Game.cpp
JonathanSteinbuch: First of all thank you for looking at it thoroughly and for all your suggestions.

I moved the Econhandler initialization to the GameClient because Game::Start() is called too late for the dskGameInterface to construct the economic progress window when the game interface is constructed. 
I also think it makes more sense to initialize this with the gameclient already as that is also where the gameworld setup is being done and where we read savegame snapshots which can also lead to creation of the econhandler.

libs/s25main/desktops/dskGameInterface.cpp
JonathanSteinbuch: It seems weird, I agree. But it works and is in fact the same way the iwShip window is constructed when opened from iwMainMenu.

libs/s25main/ingameWindows/IngameWindow.cpp
JonathanSteinbuch: SetPos does the out of bounds check, yes

libs/s25main/world/GameWorldView.h
JonathanSteinbuch: Makes sense, I will change this, with possible future splitscreen in mind

libs/s25main/ingameWindows/IngameWindows.h
```diff
@@ -0,0 +1,86 @@
...
+#include "iwTransport.h"
+#include "iwVictory.h"
+#include "gameData/const_gui_ids.h"
+#include <map>
+#include <string>
+
+struct PersistentWindowData
+{
+    std::string name;
+    std::string optionName;
```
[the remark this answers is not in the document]
JonathanSteinbuch: I want to also save whether the minimap is extended or not. Since it's only used there, maybe I should just make a separate option for minimap extendedness in the ingameOptions category of the settings file.

libs/s25main/desktops/dskGameInterface.cpp
```diff
@@ -26,29 +26,7 @@
...
-#include "ingameWindows/iwOptionsWindow.h"
-#include "ingameWindows/iwPostWindow.h"
-#include "ingameWindows/iwRoadWindow.h"
-#include "ingameWindows/iwSave.h"
-#include "ingameWindows/iwShip.h"
-#include "ingameWindows/iwSkipGFs.h"
-#include "ingameWindows/iwTextfile.h"
-#include "ingameWindows/iwTrade.h"
-#include "ingameWindows/iwVictory.h"
+#include "ingameWindows/IngameWindows.h"
```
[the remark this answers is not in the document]
JonathanSteinbuch: Ok, I will try to reorganize this with less includes

libs/s25main/ingameWindows/iwMinimap.cpp
JonathanSteinbuch: True, sorry

libs/s25main/ingameWindows/iwMinimap.cpp
JonathanSteinbuch: I disagree. Parent destructors are always called automatically in c++.

libs/s25main/ingameWindows/IngameWindow.h
JonathanSteinbuch: I renamed it to GetGUIID, good suggestion.

libs/s25main/ingameWindows/IngameWindows.h
JonathanSteinbuch: After consideration, I removed the IngameWindows.h header altogether. The includes are moved back to dskGameInterface.cpp and the map of persistent windows I moved to Settings.cpp.

libs/s25main/world/GameWorldView.h
JonathanSteinbuch: I changed this to retrieve the values from SETTINGS in the constructor of the GameWorldView and write them back to SETTINGS in the destructor.

libs/rttrConfig/src/files.h
```diff
@@ -55,7 +55,8 @@ namespace resources {
     constexpr auto boat = "<RTTR_GAME>/DATA/BOBS/BOAT.LST";
     constexpr auto boot_z = "<RTTR_GAME>/DATA/BOOT_Z.LST";
     constexpr auto carrier = "<RTTR_GAME>/DATA/BOBS/CARRIER.BOB";
-    constexpr auto config = "<RTTR_CONFIG>/CONFIG.INI"; // main config file
+    constexpr auto config = "<RTTR_CONFIG>/CONFIG.INI";               // main config file
+    constexpr auto ingameOptions = "<RTTR_CONFIG>/INGAMEOPTIONS.INI"; // ingame options file
```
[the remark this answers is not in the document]
JonathanSteinbuch: Changed this. 👍

libs/s25main/Settings.cpp
```diff
@@ -293,13 +304,61 @@ void Settings::Load()
...
+    {
+        if(libsiedler2::Load(settingsPathIngame, settingsIngame) != 0)
+            throw std::runtime_error("File missing");
+
+        const libsiedler2::ArchivItem_Ini* iniIngame =
+          static_cast<libsiedler2::ArchivItem_Ini*>(settingsIngame.find("ingame"));
+        if(!iniIngame)
+        {
+            throw std::runtime_error("Missing section");
+        }
```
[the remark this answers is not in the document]
JonathanSteinbuch: I removed these braces.

libs/s25main/Settings.cpp
```diff
@@ -293,13 +304,61 @@ void Settings::Load()
...
 // Routine zum Speichern der Konfiguration
 void Settings::Save()
 {
     libsiedler2::Archiv settings;
-    settings.alloc(SECTION_NAMES.size());
-    for(unsigned i = 0; i < SECTION_NAMES.size(); ++i)
+    unsigned totalSize = SECTION_NAMES.size();
+    settings.alloc(totalSize);
+    unsigned i;
+    for(i = 0; i < SECTION_NAMES.size(); ++i)
```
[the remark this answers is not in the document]
JonathanSteinbuch: Undid this

libs/s25main/desktops/dskGameInterface.cpp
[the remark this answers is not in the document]
JonathanSteinbuch: I undid this and removed the header IngameWindows.h altogether

libs/s25main/ingameWindows/IngameWindow.cpp
```diff
@@ -45,8 +47,9 @@ IngameWindow::IngameWindow(unsigned id, const DrawPoint& pos, const Extent& size
     // Load last position or center the window
     if(pos == posLastOrCenter)
     {
-        if(id < last_pos.size() && last_pos[id].isValid())
-            SetPos(last_pos[id]);
+        auto& settings = SETTINGS.windows.persistentSettings;
```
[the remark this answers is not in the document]
JonathanSteinbuch: It doesn't work exactly the way you suggested, because the []-operator for maps is not const. But I now changed it to use the method std::map::find() and then using iterators, which allows this to be const.

libs/s25main/ingameWindows/iwMinimap.cpp
JonathanSteinbuch: I now changed the minimap extension to an explicit bool option in the ingame section of the SETTINGS instead, which makes this cast unneccesary

libs/s25main/ingameWindows/IngameWindows.h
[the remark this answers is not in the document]
JonathanSteinbuch: I changed this as in my suggestion in the previous comment.

JonathanSteinbuch:
I now finished taking care of your kind suggestions. I didn't change the moving of the setup of the EconHandler. My reasoning for that change is found [in my comment above](https://github.com/Return-To-The-Roots/s25client/pull/1432#discussion_r693524421).

Flamefire (changes requested):
On the econ Handler:

> I moved the Econhandler initialization to the GameClient because Game::Start() is called too late

I see you call `ShowPersistentWindowsAfterSwitch` in the ctor. This is actually not the right place, because the action order is: createWindow, switch, i.e. 2 separate actions. This is why Game.Start is called on activate. I'd suggest you move the call to `ShowPersistentWindowsAfterSwitch` into the same if-condition as the game->Start call.

> I also think it makes more sense to initialize this with the gameclient already as that is also where the gameworld setup is being done and where we read savegame snapshots which can also lead to creation of the econhandler.

The GameClient does to much already. I lately moved more and more out of it into better places. The changes you needed to do to the tests IMO show that Game::Start is the better place for this. That deserialization leads to creation of the econhandler is IMO independent as due to the design of the econhandler it needs to be able to receive events and hence is part of the game state.

Bottomline: Checking for victory conditions is part of the Game, not the GameClient (which should in an ideal world just relay messages from network and windowing system), hence I'd rather see that in Game::Start and move that new function so that it works :)

libs/s25main/ingameWindows/IngameWindow.cpp
```diff
@@ -59,12 +60,17 @@ IngameWindow::IngameWindow(unsigned id, const DrawPoint& pos, const Extent& size
...
-        settings->second.lastPos = GetPos();
-        settings->second.isOpen = !closeme;
+        // Possibly save our old position
+        auto& settings = SETTINGS.windows.persistentSettings.at(GetGUIID());
+        settings.lastPos = GetPos();
+        settings.isOpen = !closeme;
+    } catch(const std::out_of_range& oor)
+    { // Window is not persistent, no problem
+    } catch(std::runtime_error& err)
+    { // SETTINGS was probably destroyed already, don't save but print a warning
```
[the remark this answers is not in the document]
JonathanSteinbuch: I added the try catch block because clang-tidy doesn't like it if destructors can throw exceptions. I don't know if it's possible but I guess it doesn't hurt to catch it here

libs/s25main/ingameWindows/IngameWindow.cpp
```diff
@@ -59,12 +60,17 @@ IngameWindow::IngameWindow(unsigned id, const DrawPoint& pos, const Extent& size
...
-    if(settings != SETTINGS.windows.persistentSettings.end())
+    try
     {
-        settings->second.lastPos = GetPos();
-        settings->second.isOpen = !closeme;
+        // Possibly save our old position
+        auto& settings = SETTINGS.windows.persistentSettings.at(GetGUIID());
+        settings.lastPos = GetPos();
+        settings.isOpen = !closeme;
+    } catch(const std::out_of_range& oor)
```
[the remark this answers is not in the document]
JonathanSteinbuch: Ok. I think I agree. I changed it to use find

libs/s25main/ingameWindows/IngameWindow.cpp
```diff
@@ -58,15 +61,19 @@ IngameWindow::IngameWindow(unsigned id, const DrawPoint& pos, const Extent& size
...
+        SETTINGS.windows.persistentSettings[GetID()].isOpen = !closeme;
+    }
 }
 
 void IngameWindow::Resize(const Extent& newSize)
 {
     DrawPoint iSize(newSize);
     iSize = elMax(DrawPoint(0, 0), iSize - DrawPoint(contentOffset + contentOffsetEnd));
     SetIwSize(Extent(iSize));
+    SetPos(GetPos());
```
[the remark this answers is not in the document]
JonathanSteinbuch: It isn't, you are right

libs/s25main/Game.cpp
JonathanSteinbuch: I now moved this back here and construct the windows after game start as suggested by @Flamefire [here](https://github.com/Return-To-The-Roots/s25client/pull/1432#pullrequestreview-737034812).

libs/s25main/ingameWindows/IngameWindow.cpp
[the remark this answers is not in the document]
Flamefire: I see. However in that case the exception should really be thrown, as it means we need to rethink that case and fully avoid it instead of just logging it (which likely no one will ever notice...).
Thinking about alternatives...
We need to set the position and the open flag. So I guess it should work to set the position in the SetPos function and isOpen in the ctor and reset in Close. Yes it means a few more calls to this code (maybe make it a function?) but I think doing bookkeeping in the destructor is not "clean" as shown by the need to handle this exception.

What do you think?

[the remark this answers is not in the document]
JonathanSteinbuch: I'm not the biggest fan, because it's easier to forget if someone adds another function that changes those values in the future. But I see your point and don't see a better alternative without using SETTINGS from the destructor, so I just changed it the way you suggested anyway.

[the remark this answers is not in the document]
Flamefire: I know... But it's a good compromise IMO and in theory there should be tests for that new function which should also tests this... But yeah.
Thanks for the changes, LGTM

Flamefire (changes requested):
Accidental downgrade of mygettext submodule. Update that to latest master again and it should be good to go :)

Flamefire (approved):

Flamefire:
@Flow86 Some issue with the runner on jenkins?

Flow86:
@Flamefire yes somethings weird with the jenkins runner. It closes the child container suddenly sometimes. Don't know why yet

PR #1433, opened 2021-09-08
Fix test broken by change in libsamplerate 0.2.0

chewi:
Since libsndfile/libsamplerate@26d92c369394bcd0b0cea488890edce1a0d757d5, initialising libsamplerate with 0 channels or less causes a SIGABRT.

### DIFF
--- a/libs/libsamplerate/include/samplerate.hpp
+++ b/libs/libsamplerate/include/samplerate.hpp
@@ -152,6 +152,8 @@ class State : public detail::StateBase
 {
     static SRC_STATE* createOrThrow(Converter converter, int channels)
     {
+        if (channels <= 0)
+            throw std::runtime_error("Channel count must be >= 1.");
         int error;
         auto* state = src_new(static_cast<int>(converter), channels, &error);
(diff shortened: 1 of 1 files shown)


Flow86:
is there a way to check on the version and skip these tests if necessary?

chewi:
Maybe, via some preprocessor defines derived from `${Samplerate_VERSION_STRING}`. I'm not going to do that for you though. :stuck_out_tongue: Frankly, I don't see the point. These tests are there to ensure the project code doesn't do something it shouldn't. You already know it doesn't do it right now. Before long, you'll probably use a newer version of libsamplerate and the very few people who do run the tests probably will too.

chewi:
Sorry, I didn't word that very well. You're testing the result of undefined behaviour in libsamplerate for a case that I'm guessing would crash the game regardless? It occurs to me that you could just handle this in `samplerate.hpp` and keep the test as it is. I'll update the PR later.

chewi:
Pushed. I hope you'll agree this is better.

PR #1437, opened 2021-09-11
Set C++ standard (to C++14) using CMake

chewi:
Not adding the `-std` flag broke builds with GCC 11 until the code was fixed up recently. This should prevent this sort of thing from happening in future. It also ensures that the compiler is actually new enough.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,9 @@ if(DEFINED CMAKE_TOOLCHAIN_FILE)
     message(STATUS "Used Toolchain definition file '${CMAKE_TOOLCHAIN_FILE}'")
 endif()
 
+set(CMAKE_CXX_STANDARD 14)
+set(CXX_STANDARD_REQUIRED ON)
+
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/Modules" "${CMAKE_SOURCE_DIR}/external/libutil/cmake")
 if(CMAKE_VERSION VERSION_LESS 3.14)
     list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/external/libutil/cmake/cmake_3.14")


Flamefire:
Instead of doing this we use target-specific requirements, i.e. the targets require C++14 or newer.
Why is C++14 exactly required? Can you remind me of the problem you mentioned?

chewi:
Here is the issue I was referring to, where [it failed with GCC 11](https://bugs.gentoo.org/787299#c9) because C++17 is the new default. I also based this on the fact that your README said you need a "C++14 compatible compiler".

Do you mean build targets like videoSDL2? I don't see the C++ standard being set anywhere at the moment. If you actually mean a target platform like Visual Studio on Windows then this change will do the right thing there too.

Flow86:
do we set it manually (i.e by adding -std=...) somewhere?

Flow86:
okay. since the appveyor test fails due to removed boost 1.71 - can you please change this in the appveyor.yml too? then we can do it in one step :-)

Flamefire:
We set that for each target: https://github.com/Return-To-The-Roots/s25client/blob/master/libs/common/CMakeLists.txt#L19-L20

This allows users/maintainers to overwrite the default with `cmake -DCMAKE_CXX_STANDARD=14` or `cmake -DCMAKE_CXX_STANDARD=2a` as they like or might be even required.

This PR will fix the used standard to 14 while IMO we should be 14+ compatible and fix bugs as they occur. So I'd rather say we should enhance CI to test with newer standards.

> I also based this on the fact that your README said you need a "C++14 compatible compiler".

This is true and as mentioned above (and done similarly by the Gentoo guys) you may configure your compiler to use C++14 if that specific compiler turns out to be incompatible with C++14 when running in C++17 mode.
But again: I'd rather like RTTR to be compatible with C++17 than fixing it at C++14 which is the reason for the target usage requirements

chewi:
Ah, my grepping failed me. Fair enough though, it's good to keep looking forwards. I'll let you know if I hit any issues with newer compilers.

Flamefire:
Opened #1439 to address the CI part

PR #1439, opened 2021-09-14
Enhance CI: Update compilers and tested standards

Flamefire:


Flamefire:
@Flow86 All done, but it seems the required tests must be adjusted. Maybe only make the ones with the minimal compiler versions required, so we can easily update the others

Flow86:
okay which ones do you want to be required?

Flamefire:
> okay which ones do you want to be required?

Static analysis, gcc-6, clang-7 (2 each), clang-macos, jenkins
All others should pass but I don't think we have to enforce this as those with merge rights know not to merge when any fails. Appveyor should also be required but is sometimes flaky, so manual inspection and ignoring that is ok

Flow86:
ok. new list will be
grafik
then?

Flamefire:
Looks good, yes

Flow86 (approved):

Issue #1435, opened 2021-09-09
0.9.1 Can't join games


Spikeone:
It looks like a Linux player can not join a game hosted by a windows player.

Both players ran rttr.bat to make sure they have the same version - no luck, the game still greets with version missmatch

Flamefire:
Could you post the begin of the log, i.e. the part where the version is stated? Should be something like 
> Return To The Roots v0.9.1-.... Compiled with ...

Flamefire:
Looks like the linux version has the revision truncated:
Linux: 7273dd8c3fa98f3892af28d415aca77a
Windows: 7273dd8c3fa98f3892af28d415aca77a5d58fdd4

We need to double check that

Spikeone:
yeah sorry, don't have linux and the player I played with didn't answer since

stefson:
here is the output from my self compiled v0.9.1 binary:

s25client 
Return To The Roots v0.9.1-7273dd8c3fa98f3892af28d415aca77a5d58fdd4
Compiled with GNU C++ version 9.4.0 for linux 64 Bit

hope that helps

Flamefire:
@Flow86 Still not sure where the shorter revision comes from.
If I do `git checkout v0.9.1 && cat .git/HEAD` I get the correct one on my Linux system. Something wrong with the build images?

Flow86:
when I run it manually on the build images I also get the right version?

stefson:
so have you ironed this one out? If not, I'm on linux and am willing to help you push out a stable 0.9.2 release. Just ping me with @

Flamefire:
@stefson Feel free to try and build a stable linux version from current sources. If that works too, I'd say we go and try to tag a 0.9.2 and build it on Jenkins. The new check I added should catch the issue if it still occurs.

stefson:
It's multiplayer, I can't test it alone man :-) 

(joining the lobby works)

Flamefire:
Just check the revision on startup that it is indeed long enough. But if it compiles it will work due to the new check

stefson:
Well, that part was never a problem with my checkout:

```
Return To The Roots v20210924-df0fe3ae7ef062245914360f6275fcadbb08d4b3
Compiled with Clang version 12.0.1  for linux 64 Bit
```

It was a problem for the binaries provided by your buildserver. Hence the fix is only in the updater, jenkins, and workflows, right?

Spikeone:
[...]

PR #1450, opened 2021-10-29
Fix crash when flag is captured while having active flag workers and enhance connect fail message for version mismatch

### LINKED ISSUE #1449: s25client crashing: pure virtual method called
Hello! 

Not sure how do I report the issue properly, terribly sorry.
I am running the latest s25client.exe build on Windows 10.
The game is crashing, and all I see in the console log is this message:

>pure virtual method called
>terminate called without an active exception

The log file has no messages related to the crash.
The game crashes on the "Good old times" map by BlueByte.
I've got a replay file that shows the last minute before the crash.

The crash happens randomly when I do something with the game objects or when I save the game.
The game also crashes when trying to save the game after the replay finishes.
Here is a screencast on Youtube:

[s25client crash](http://www.youtube.com/watch?v=tlJGHLUmy4I "s25client crash")

The replay file, the savefile and the map is in this zip archive: [crash_save_replay.zip](https://github.com/Return-To-The-Roots/s25client/files/7428438/crash_save_replay.zip).
Please let me know if I can somehow improve the bug report.

Thanks :)

Flamefire:
See https://github.com/Return-To-The-Roots/s25client/issues/1449 and help a bit for https://github.com/Return-To-The-Roots/s25client/issues/1435

Fixes #1449

### DIFF
--- a/libs/s25main/GamePlayer.cpp
+++ b/libs/s25main/GamePlayer.cpp
@@ -1256,7 +1256,7 @@ void GamePlayer::CallFlagWorker(const MapPoint pt, const Job job)
         wh->OrderJob(job, flag, true);
 }
 
-bool GamePlayer::IsFlagWorker(nofFlagWorker* flagworker)
+bool GamePlayer::IsFlagWorker(const nofFlagWorker* flagworker)
 {
     return helpers::contains(flagworkers, flagworker);
 }
--- a/libs/s25main/GamePlayer.h
+++ b/libs/s25main/GamePlayer.h
@@ -222,7 +222,7 @@ public:
         RTTR_Assert(IsFlagWorker(flagworker));
         flagworkers.remove(flagworker);
     }
-    bool IsFlagWorker(nofFlagWorker* flagworker);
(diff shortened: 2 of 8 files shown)


Flamefire:
IMO this is not possible. You can only connect to servers with the same version, so the protocol version is the same as the game version. IIRC we don't have a protocol version, do we?
If somehow an old and new client/server communicate with this it will either result in ignoring the version (->all fine) or a runtime error leading in a close of the connection (->also fine) but as mentioned above: As this is after the part where the server information is already exchanged (i.e. the version) and you cannot connect to another version this should never happen anyway

Flow86:
but what if the client is connecting to an old version, thats what the message is for? then that server does not respond with the string.
I think we have something like this, but its only for LobbyProtocol I think

Flow86:
the revision was also never the check criteria at least for the stable versions.

Flamefire:
It was. After the connection was created: https://github.com/Return-To-The-Roots/s25client/blob/master/libs/s25main/network/GameServer.cpp#L918-L919

PR #1476, opened 2021-12-03
Allow all building types on Harbor spots 

### LINKED ISSUE #1473: Only harbor building on harbor spots possible
As the title says, it's currently not possible to place anything but a harbor building on a harbor spot which is incorrect.

Flamefire:
A harbor spot is like a castle spot with additional harbor building
Fixes #1473

Also adds tests for that

### DIFF
--- a/libs/s25main/world/GameWorldView.cpp
+++ b/libs/s25main/world/GameWorldView.cpp
@@ -480,7 +480,7 @@ void GameWorldView::DrawConstructionAid(const MapPoint& pt, const DrawPoint& cur
     if(bq != BuildingQuality::Nothing)
     {
         constexpr auto bqImgs = getBqImgs();
-        glArchivItem_Bitmap* bm = LOADER.GetMapTexture(bqImgs[bq]);
+        auto* bm = LOADER.GetMapTexture(bqImgs[bq]);
         // Draw building quality icon
         bm->DrawFull(curPos);
         // Show ability to construct military buildings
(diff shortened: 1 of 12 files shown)


libs/s25main/world/GameWorldView.cpp
```diff
@@ -462,12 +462,25 @@ void GameWorldView::DrawMovingFiguresFromBelow(const TerrainRenderer& terrainRen
...
 }
 
+constexpr auto getBqImgs()
+{
+    helpers::EnumArray<unsigned, BuildingQuality> imgs{};
+    imgs[BuildingQuality::Flag] = 50;
+    imgs[BuildingQuality::Hut] = 51;
+    imgs[BuildingQuality::House] = 52;
+    imgs[BuildingQuality::Castle] = 53;
+    imgs[BuildingQuality::Mine] = 54;
```
Flow86, line 472: shouldn't we set these in the order of the enum instead?

Flamefire: I don't see a need for this. I changed the order of the enum to fit our needs: Better performance: https://github.com/Return-To-The-Roots/s25client/pull/1476/commits/0c474e86fc172214f95f3d621dca3b1dd88e9eed

Due to the `constexpr` this whole array is a constant and the function makes it very clear which BQ is which image. I find that better than the kinda mysterious `49 + rttr::enum_cast(bq)`

Flamefire:
@Flow86 rebased and resolved merge conflicts

Flow86 (approved):

PR #1478, opened 2021-12-06
Add locked icon to show locked addons

### LINKED ISSUE #1470: Show locked Addons
When enabling some addons for change, it's quite hard to see which addons can be changed. One could display a lock, indicating the user can't change those.

Lock Icon can be enhanced if functionality is added.
locked
S2_LOCK

Flamefire:
Display lock icon to show which addons cannot be changed.
Closes #1470

Actual render image: 
Unbenannt


Flamefire:
No lock and disabled. I just changed the code to always show the lock to check positioning etc.
If you want to improve the lock, do so before merging.
Btw: need to enhance bmp support. This icon can be saved with 4bbp, but in rttr we need 24 bbp making the file size 6 times larger...
And with 8bpp (possible) there is an issue with transparency which I got to check. Paint.net is used  for saving and iirc rttr uses the first palette color for transparency
Any ideas?

Spikeone:
But thats the wrong way around then, isn't it?
Locked: Can't change => show lock + disable controls
Open: Can change => no lock + enable controls

In the screenshot you posted, I'd expect to not be able to change any of the listed addons, due to their lock, but your explanation says the opposite, doesn't it? Sorry if I'm confusing something here but the lock should be shown in front of addons that can't be changed

Flamefire:
@Spikeone it is as you want it ;)

To quote:
> Actual render image with forced lock show:

I.e. as I didn't have a script which locks the addons I simply removed the condition which prevents the lock to be shown: https://github.com/Return-To-The-Roots/s25client/pull/1478/files#diff-aa023d1c37483680e2b5db8d9ad11def4c5b9c19db95a710327bd2ec8305f54dR28

Just seen that my wording was a bit confusing. Sorry for that, I was on my phone.

Spikeone:
Sorry, I might be confused a bit, but your screenshot (https://user-images.githubusercontent.com/309017/144905953-c382080f-7094-4ad1-a575-44a777262fe1.png) shows addons that are locked but the controls aren't disabled. If they were, the dropdown buttons and checkboxes should be gone, shouldn't they? So in that screenshot it looks like locked icon + non locked addon - that causes my confusion, or I'm drunk, not sure anymore 🍡

Flamefire:
> shows addons that are locked but the controls aren't disabled.

No, as I didn't use a script to lock ANY addon I just made the code to show the lock icon (and only that). So yes "locked icon + non locked addon" for all just for the purpose of this screenshot by removing the line which would make the lock NOT show if the addon can be modified.

Spikeone:
S2_LOCK_2

@Flamefire you may try out this one - and take whichever looks better, if you find both awfull, I'll give it another shot

Flamefire:
Updated the PR with some changes:
- New picture
- Support for 4BPP and 32BPP BMP files (your image can be stored using only 4BPP making it require less space)

New rendering with all addons disabled (now for real):
Unbenannt

What do you say?

Flamefire:
Rebased the PR and fixed the conflicts.
@Spikeone Your comment on the new picture?

Spikeone:
@Flamefire I'm a bit confused because the letter colors and lock colors do not match, although I thought I used the same colors. But in general it's looking good I think

Flamefire:
> @Flamefire I'm a bit confused because the letter colors and lock colors do not match, although I thought I used the same colors. But in general it's looking good I think

I double checked and it is the same color(s). But the the font uses more of the darkest color while the lock has a gradient

Flow86:
I do not understand the PR fully, for what is the lock for?

Flamefire:
> I do not understand the PR fully, for what is the lock for?

See the related issue #1470
> When enabling some addons for change, it's quite hard to see which addons can be changed. One could display a lock, indicating the user can't change those.

Flamefire:
@Flow86 Can this be merged please?

Flow86:
@Flamefire conflicts :-/

Flamefire:
@Flow86 Rebased. Git was even able to auto-resolve the conflicts.

PR #1485, opened 2022-01-21
Update Ingame-Readmes and display user-data folder

Flamefire:
- Remove version and year from header avoiding the need to update them
- Add release notes for 0.9.x
- Update copyright year to 2022

- Show user data folder in the console window and the savegame folder in the load game window:

Additionally I removed the redundant `RTTR_CONFIG` (same as `RTTR_USERDATA`) and added an option to the text-control to have a max-length to avoid overlong user names etc. from breaking the window. Not really great yet, but does its job.

### DIFF
--- a/data/RTTR/texte/de/keyboardlayout.txt
+++ b/data/RTTR/texte/de/keyboardlayout.txt
@@ -1,4 +1,4 @@
-                  SIEDLER II.5 - RETURN TO THE ROOTS
+                          RETURN TO THE ROOTS
 ------------------------------------------------------------------------
 Globale Tastaturbelegung im Spiel:
 
@@ -47,6 +47,6 @@ P:.................... Pause/Fortsetzen (Auch nach GF Sprung)
 1-7:.................. Spielerwechsel
 
 ------------------------------------------------------------------------
-http://www.siedler25.org    Copyright (C) 2005-2022 Settlers Freaks
+https://www.siedler25.org        Copyright (C) 2005-2022 Settlers Freaks
 ------------------------------------------------------------------------
 
--- a/data/RTTR/texte/de/keyboardlayout.txt
+++ b/data/RTTR/texte/de/keyboardlayout.txt
@@ -1,4 +1,4 @@
-2016.06.03        SIEDLER II.5 - RETURN TO THE ROOTS           0.8.1-dev
+                  SIEDLER II.5 - RETURN TO THE ROOTS
 ------------------------------------------------------------------------
 Globale Tastaturbelegung im Spiel:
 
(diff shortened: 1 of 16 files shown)


data/RTTR/texte/de/keyboardlayout.txt
```diff
@@ -47,6 +47,6 @@ P:.................... Pause/Fortsetzen (Auch nach GF Sprung)
 1-7:.................. Spielerwechsel
 
 ------------------------------------------------------------------------
-http://www.siedler25.org    Copyright (C) 2005-2016 Settlers Freaks
+http://www.siedler25.org    Copyright (C) 2005-2022 Settlers Freaks
```
Flow86, line 50: we should make it https

data/RTTR/texte/de/readme.txt
```diff
@@ -382,5 +393,5 @@ E. Übersicht: Updates und Änderungen
   - alles ;-)
 
 --------------------------------------------------------------------------------
-http://www.siedler25.org                 Copyright (C) 2005-2017 Settlers Freaks
+http://www.siedler25.org                 Copyright (C) 2005-2022 Settlers Freaks
```
Flow86, line 396: same here

data/RTTR/texte/keyboardlayout.txt
```diff
@@ -46,5 +46,5 @@ J:.................... Go to specific GF
 P:.................... Pause/Resume (also after GF jump)
 1-7:.................. Change to other player
 ------------------------------------------------------------------------
-http://www.siedler25.org    Copyright (C) 2005-2016 Settlers Freaks
+http://www.siedler25.org    Copyright (C) 2005-2022 Settlers Freaks
```
Flow86, line 49: and here

data/RTTR/texte/de/keyboardlayout.txt
```diff
@@ -1,4 +1,4 @@
-2016.06.03        SIEDLER II.5 - RETURN TO THE ROOTS           0.8.1-dev
+                  SIEDLER II.5 - RETURN TO THE ROOTS
```
Flow86, line 1: we should write Return to the roots only

data/RTTR/texte/keyboardlayout.txt
```diff
@@ -1,4 +1,4 @@
-2016.06.03        SIEDLER II.5 - RETURN TO THE ROOTS           0.8.1-dev
+                  SIEDLER II.5 - RETURN TO THE ROOTS
```
Flow86, line 1: same here

Flamefire:
As the tests needed to be restarted anyway I included a patch for our SDL2 driver using a wrong union member (although only technically wrong as the union types have the same layout) and tests for ctrlText to increase coverage. The changes in s25Client can't really be tested as that would write actual user data with no good way to prevent this...

Flamefire:
As the tests needed to be restarted anyway I included a patch for our SDL2 driver using a wrong union member (although only technically wrong as the union types have the same layout) and tests for ctrlText to increase coverage. The changes in s25Client can't really be tested as that would write actual user data with no good way to prevent this...

All ready now. @Flow86

Flow86 (approved):

PR #1487, opened 2022-01-22
Fix meaning of playlist repeats and make playlists infinite

### LINKED ISSUE #1480: Tracks seem to always play twice
I don't quite understand how the music player is supposed to work but there is a more/less repeats button. If I set it to 1, each track plays twice, but if I set it to 0, nothing plays at all! Surely this is not how it's supposed to work?

Flamefire:
The repeats now actually mean repeats, i.e. a value of 1 repeats the song once, playing it twice.
On ordered playlists this will play each song N+1 times before starting the next.
For random playlists it doesn't matter much, but a higher value basically increases the chance of playing the same song again, up to N+1 times (in total).
After all songs of the playlist have been player N+1 times the whole playlist is reshuffled (if set) and restarted.

Fixes #1480

Basically uses "Option 5" which should please everyone.

### DIFF
--- a/libs/s25main/Playlist.cpp
+++ b/libs/s25main/Playlist.cpp
@@ -3,6 +3,7 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 
 #include "Playlist.h"
+#include "helpers/containerUtils.h"
 #include "mygettext/mygettext.h"
 #include "s25util/Log.h"
 #include "s25util/StringConversion.h"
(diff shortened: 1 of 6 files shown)


libs/s25main/Playlist.cpp
```diff
@@ -12,81 +13,64 @@
...
+#include <stdexcept>
 
 namespace bnw = boost::nowide;
 
 Playlist::Playlist(std::vector<std::string> songs, unsigned numRepeats, bool random)
     : songs_(std::move(songs)), numRepeats_(numRepeats), random_(random)
 {
-    Prepare();
+    if(helpers::contains(songs_, std::string{}))
+        throw std::invalid_argument("Empty song passed");
```
Flow86, line 24: does this get catched somewhere? (and it should be `song**s**`

Flamefire:
@Flow86 *ping*

Flow86 (approved):

PR #1488, opened 2022-01-22
Fix drawing glitch on start

### LINKED ISSUE #1456: Wrongly drawn terrain when starting a game
When starting a game, one has a strange black border at the bottom. By moving the camera just a bit this is fixed but strange anyway:

grafik

Flamefire:
Move maxNodeAltitude from TerrainRenderer to GameWorldViewer

This field was initialized to late: In GenerateOpenGL
It is used by the GameWorldView constructor: Through the `Move` function which recalculates the first and last map points to draw based on the max height.
This moves makes the value available very early.

Also adds tests and a little refactoring of the `MoveTo` function (move the relative move into new `MoveBy`)

Fixes #1456


Flamefire:
@Flow86 This needs merging fast as it also contains the fix for the failing test on master

Issue #1481, opened 2021-12-29
SWAP File is getting bigger and bigger


Runshak:
Hi, I´m playing on a new MacBook Air with M1 CPU. I am playing a quite large MAP with 7 Players.

So far so good. But during playing the Game is building a SWAP-File. At the beginning it was just 50GB but now its already over 100GB. Is this normal? 
The Software is at the newest Version (0.9.1 )

Performance is until now acceptable. Only during saving the FPS are dropping down.

Spikeone:
@Runshak what does "quite large" exactly mean? What sizes are we talking about here?

stefson:
can you please provide us with the map you're playing?

Runshak:
Just need to know how to find out from a running game how big the size is. Or which files I have to attach

Runshak:
Hi,

The big swap-file I could solve after restarting my MacBook. I guess it was not an initial problem from RTTR. Maybe something went wrong because of MacOS. After restarting the whole System the SWAP file didn’t grow again and was just 5 GB.

Other issues during the game are:

- the in game stats of the players are empty. Just can watch myself
- big map can’t see details in the map overview in the game
- ressources are not always used from the storage building -> buildings around it don’t get their needs but the storage building is full of it
	so the resources are often transported very far from other places.
- in the Terminal you can see when an enemy (KI) give up the game. But ingame you don’t get a message about this.

The newest Replay, the newest Savegame, config.ini and ingameoptions.ini are attached to this mail （follow the Link).


Mit freundlichen Gruessen / Kind regards / 诚 挚 的 问 候

Markus Musterle

Spikeone:
> The newest Replay, the newest Savegame, config.ini and ingameoptions.ini are attached to this mail （follow the Link).

There is no link - at least I can't see any, please attach a zip file

Runshak:
Hi, the files are too big. Here the link again

https://www.icloud.com/iclouddrive/0a90ZYwSXSEXQxvgU0AEOkdhA#RTTR

Spikeone:
@Runshak requires a login

![grafik](https://user-images.githubusercontent.com/1276342/150757883-724a3bcc-8712-4483-adf9-1f439e43a4cd.png)

Runshak:
Hallo, 

 Markus Musterle ***@***.***) hat Sie eingeladen, den Ordner „ RTTR “ in Dropbox anzusehen. 

 Markus Musterle schreibt dazu:
[...]

PR #1491, opened 2022-01-30
Connect improvements: Show progress of map transfer and fix leaking network messages

Flamefire:
Major contribution here is: Add window showing the connection (and map transfer) progress
Oh yes: And an abort button to abort the connection while it is in progress.

This replaces the previous "Please wait" window with one showing actual progress.
Very useful for larger maps/savegames where transferring the map takes a considerable amount of time.
Example (note: Changed the background to grey and the status text color to yellow now but didn't update the screenshot): 

This required a change to the connection process to report this status as a callback. During that I also reworked the connection stages and enforced them via checks, so it is verified the messages arrive in the correct order and a server may not send anything unexpected, i.e. enforce the connection protocol.

A few drive-by changes:
- Hints for clang-format to understand the MOCK macros
- Style/safety changes (enums as IDs for controls, maximum map size enforcement instead of blindly allocating possible all available memory)
- Fix memory leak after handling network messages, see https://github.com/Return-To-The-Roots/s25client/blob/e68c6a248e5c7e196085f62d7ff66606b3f7fca5/libs/s25main/network/NetworkPlayer.cpp#L45
https://github.com/Return-To-The-Roots/libutil/blob/423c6e343cf953e2c08aed2d93c865e2a087b2e3/libs/network/src/MessageQueue.cpp#L39 So maybe this was the cause of #1481
- (Finally) rename "HostGame" desktop to "GameLobby" to avoid the confusion about its purpose
- Copy the map via filesystem when starting a server: Much faster game hosting now especially for big maps

I also needed to generalize closeOnRightClick to userClosable:
The only reason to disallow the close on right click is to have a dedicated button to press, e.g. to force a choice of "Save" or "Discard". E.g. for the connection and message box windows this could be a real issue as the caller might not get one of the expected results if the users closes the window via ESC (or the icon). @Spikeone Can you imagine any situation where we might want to allow closing a window via ESC or the icon (in the title bar) but not via right-click? IMO it is only consistent that a window closable via ESC is also closable via rightclick or the icon and the other way round.

The current major source of a lack while creating a game is the time it takes to compress the map while preparing the server. Can't avoid that (for now).


Flamefire:
New screenshot (note that the color of the progressbar changes from red over yellow to green as with others)

Spikeone:
@Flamefire I'm loving it - does the moon cursor change to a hand when overing over abort? Is the window right click closeble (or however thats written :D). If yes, I think it shouldn't

Flow86:
@Flamefire great thing. I like it. Let me know as soon as you have finished it, so we can get this in fast :-)

Flamefire:
@Spikeone 
> does the moon cursor change to a hand when overing over abort? 

No. However the button is highlighted and changing the cursor is imo to hard to implement correctly for this use case.

> Is the window right click closeble (or however thats written :D). If yes, I think it shouldn't

Nope. Also not closable by ESC. See the paragraph about 
> I also needed to generalize closeOnRightClick to userClosable:

Also the question still stands as I want to make sure not to introduce another bug like the unclosable action window :D
> Can you imagine any situation where we might want to allow closing a window via ESC or the icon (in the title bar) but not via right-click?

Flamefire:
@Flow86 Should be done now. Added a test for the new window to ensure a few things and increase coverage. So when CI passes, this is good to go

Spikeone:
> Can you imagine any situation where we might want to allow closing a window via ESC or the icon (in the title bar) but not via right-click?

I can't. I think they are the quite same actions. Although that'd mean that any window that is right click to closable, should have the corresponding icon which is not the case I think - mhhh

libs/s25main/ingameWindows/iwConnecting.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+iwConnecting::iwConnecting(ServerType serverType, std::unique_ptr<ILobbyClient> lobbyClient)
+    : IngameWindow(CGI_CONNECTING, IngameWindow::posCenter, Extent(300, 120), _("Connecting"),
+                   LOADER.GetImageN("resource", 41), true, false),
+      serverType_(serverType), lobbyClient_(std::move(lobbyClient))
+{
+    WINDOWMANAGER.SetCursor(Cursor::Moon);
+
+    AddText(ID_txtStatus, DrawPoint(150, 30), _("Please wait..."), COLOR_YELLOW, FontStyle::CENTER, NormalFont);
+    AddPercent(ID_prgLoad, DrawPoint(20, 50), Extent(260, 26), TextureColor::Grey, COLOR_YELLOW, SmallFont,
```
Flow86, line 34: using downloadProgress via the pointer is against SOLID principles (since ctrlPercent does not have control over it) but I think this is okay for this case.

Flow86 (approved):

PR #1501, opened 2022-02-19
Spike/addons/settlers iv mining

Spikeone:
What should this addon (actually 4, for each type of mine) do:
- Alter mining, depending on the ground resources, the chance to actually mine the resource decreases
- All mines use up deposits up to 1 ore left
- Chance is determined by the radius and remaining ore compared to max ore (so, combinedRemainingOre / possibleSpots), that way if we add an addon that increases/decreases mining radius this one will still work
- Depending on the chance, there will be either not resource or the mine's resource be produced (there is an another addon to alter this)
- Settings can be changed for all types of mines indivdually (since it makes sense to e.g. have gold finite, but make granite infinite)
- There is also a setting to make ore always available, so ignoring the mountain's resource (makes most sense for granite only)

Currently (as far as I remember) the following is missing:
AI:
- should honor possible settings (i.e. if it's allowed to mine gold everywhere, the AI should do that instead of searching)
- should honor combinations (i.e. granite mines always but all other mines just as they are)
- should somewhat honor the current accumulated productivity of mines (i.e. 10 coal mines at 10% are just as good as 1 coal mine at 100%)

Missing Addon: Alternative production cycle product:
This one alters whats produced when an alternate production cycle runs. The addon should have the following settings:
- produce nothing (default)
- produce granite (25%)
- produce granite (50%)
- produce granite (100%)
- produce lower grade (e.g. gold mine produces iron, iron produce coal, coal produce granite, and granite just nothing)
Although the last setting doesn't make sense logical, it does make sense for the gameplay. This addon is compatible to other addon settings, since it is only taken into account when setting the alternate cycle resource

Missing Addon: Show real productivity:
This one shows a productivity rate for mines. If a mine type is inexhaustible, 100% is shown anyway. Maybe some color could indicate if this mine is already at it's lowest rate or not (e.g. >50% = "good green", >min% = "okay yellow", == min% = "bad red")
settings:
- disable (default)
- enable
This addon is compatible to other settings, since it should show 100% if original mining behavior is enabled or inexhaustible. I think it is a nice to have

TODO:
- [ ] Resolve merge conflicts
- [ ] Resolve remaining change requests
- [ ] AI implementation
- [ ] Alternative production cylce addon

Nice To Have:
- [ ] Show real productivity addon


libs/s25main/figures/nofMiner.cpp
```diff
@@ -71,20 +72,100 @@ helpers::OptionalEnum<GoodType> nofMiner::ProduceWare()
...
+        case BuildingType::IronMine: return settings.getSelection(AddonId::MINES_IRON);
+        case BuildingType::CoalMine: return settings.getSelection(AddonId::MINES_COAL);
+        case BuildingType::GraniteMine: return settings.getSelection(AddonId::MINES_GRANITE);
+        default: return 0;
+    }
+}
+
+bool nofMiner::StartWorking()
+{
+    workplace->is_emptyCycle = false;
```
Flamefire, line 98: Why is this a property of the workplace? Shouldn't this be part of the miner? You can use this in `ProduceWare` to return `boost::none` aka "nothing"

libs/s25main/figures/nofMiner.cpp
```diff
@@ -71,20 +72,100 @@ helpers::OptionalEnum<GoodType> nofMiner::ProduceWare()
...
+                    useResPt = curPt;
+            }
+
+            // no resources left (mine was built on invalid spot)
+            if(sumResAmount == 0)
+                return false;
+
+            // 19 = amount of nodes a mine can reach
+            // 7  = maximum resource amount a node possibly has
+            if(RANDOM.Rand(RANDOM_CONTEXT(), 19 * 7) > sumResAmount)
```
Flamefire, line 134: To many magic numbers for my liking. Maybe put the contents of the `FindAllPointsWithResource` here and first get all points in the given radius then iterate over them. Then you can use the size of the vector of all points here (I guess that is the 19?)
This also avoids calling `world->GetNode(curPt).resources` 2 times per node (one in `HasResource` one above)

And for the `7`: There should be an (enforced and named) constant for that. If not, the `Resource` class should get one.
And there is a shortcut for this line:
```suggestion
            if(RANDOM_RAND(19 * 7) > sumResAmount)
```
See https://github.com/Return-To-The-Roots/s25client/blob/fd16c75ecc85ef814885ce9b1fd41d1e701fa6cd/libs/s25main/random/Random.h#L99

libs/s25main/figures/nofMiner.cpp
```diff
@@ -71,20 +72,100 @@ helpers::OptionalEnum<GoodType> nofMiner::ProduceWare()
...
+            {
+                // failed, use food and start working - but produce nothing
+                workplace->is_emptyCycle = true;
+            } else
+            {
+                // if success, use 1 quantity if any
+                if(!useResPt.isValid())
```
Flamefire, line 141: This doesn't make sense. On success this must be set, so the `if` can be removed. Maybe even use this check instead of `if(sumResAmount == 0)`?

libs/s25main/figures/nofMiner.cpp
```diff
@@ -71,20 +72,100 @@ helpers::OptionalEnum<GoodType> nofMiner::ProduceWare()
...
+                    world->ReduceResource(useResPt);
+            }
+        }
+        break;
+        case 2: // inexhaustible
+        {
+            MapPoint resPt = FindPointWithResource(GetRequiredResType());
+            if(!resPt.isValid())
```
Flamefire, line 149: Do we even need to check this here? This is already done in `AreWaresAvailable` and for inexhaustible it cannot disappear, can it?
If this is required I'd prefer: 
```suggestion
            if(!FindPointWithResource(GetRequiredResType()).isValid())
```
Again: No need for a named entity (variable) if it isn't (really) used.

libs/s25main/figures/nofMiner.cpp
```diff
@@ -71,20 +72,100 @@ helpers::OptionalEnum<GoodType> nofMiner::ProduceWare()
...
+        }
+        break;
+        case 3: // inexhaustible, everywhere
+        {
+        }
+        break;
+        case 0: // original behavior
+        default:
+        {
+            MapPoint resPt = FindPointWithResource(GetRequiredResType());
```
Flamefire, line 160:
```suggestion
            const MapPoint resPt = FindPointWithResource(GetRequiredResType());
```

libs/s25main/figures/nofWorkman.cpp
```diff
@@ -55,7 +55,7 @@ void nofWorkman::HandleStateWaiting2()
 {
     current_ev = nullptr;
     // Ware erzeugen... (noch nicht "richtig"!, sondern nur viruell erstmal)
-    if(!(ware = ProduceWare()).has_value())
+    if(!(ware = ProduceWare()).has_value() || workplace->is_emptyCycle)
```
Flamefire, line 58: As explained above: I think it makes more sense to put the emptyCycle member into the miner class and handle it in `ProduceWare`

libs/s25main/figures/nofWorkman.cpp
```diff
@@ -110,3 +110,13 @@ MapPoint nofWorkman::FindPointWithResource(ResourceType type) const
 
     return MapPoint::Invalid();
 }
+
+std::vector<MapPoint> nofWorkman::FindAllPointsWithResource(ResourceType type) const
```
Flamefire, line 114: See above: This is incredibly miner specific especially due to the `MINER_RADIUS` used below. So better put this into the miner class or even the function directly.

libs/s25main/GlobalGameSettings.cpp
```diff
@@ -66,6 +66,10 @@ void GlobalGameSettings::registerAllAddons()
         AddonBurnDuration,
         AddonCatapultGraphics,
         AddonChangeGoldDeposits,
+        AddonMinesGranite,
```
Flamefire, line 69: Can you please keep those sorted? Makes it easier to find missing ones.

Flamefire (changes requested):
General: Good idea I'd say. Although I'm not sure it is worth (i.e. will it be used?) having 4 addons instead of 1-2 (all mines or granite and others). Also I'm not really a fan of the "Everywhere" setting. This makes geologists worthless and IMO changes the game to much... Do people really want this? Personal opinion though and it is an addon so things like this are possible.

As for the implementation I added some comments. Some (very) specific some more general, so don't panic ;-)

General advise (kind of a TLDR):
- If you C&P something think again about alternatives it often becomes an issue later on
- If you need comments think again. Here: Mostly the "magic numbers" which can be replaced by enums, named constants, functions...
- `const` on anything which doesn't need to be changed
- No need for a variable if you simply return it, return it directly.

These are rules of thumb, so conditions apply ;-)

libs/s25main/figures/nofMiner.cpp
Spikeone: As far as I currently understand the logic, the ```AreWaresAvailable``` function is called only for any work cycle **except** the initial one. Maybe I missunderstood the logic at one point (or rather my testing, as first my mines ran exactly 1 time and then failed in the ```FindPointWithResource``` function in ```AreWaresAvailable```)
So yeah, going to use your suggestion :)

libs/s25main/figures/nofMiner.cpp
Spikeone: In the Settlers IV like version resources are used up to 1 remaining ore - therefore if all nodes are used up, then no node with > 1 ore is left - so nothing to use and minimum chance for that mine at that spot. Hope that makes sense in a way

Spikeone:
Well, while I was thinking about the "InexhaustibleGraniteMine" addon which had a different idea initially than it currently implements, I felt like people might use a mix of the settings, or at least they made sense for me:
- Granit Mines => Everywhere => Enough building material for longer games (maps often do not provide enough with low/very low goods)
- Coal/Iron => Settlers IV like => There is always some production left and the game never ends in a lock
- Gold => Original => Still reason to go for the mountains containing gold

At least thats the setting I'm currently thinking about as default for myself. On the other hand I agree, that having 4 addons (which I need to find a good name for, so they are next to each other) isn't that good either. But as I really think the "Granite Mines Everywhere" Addon is a good idea, I'd go with 2 anyway - and then 4 isn't that much more. Mhh well

libs/s25main/figures/nofMiner.cpp
Flamefire: The `AreWaresAvailable` is called for every work cycle, as that is the precondition for starting to work. See `TryToWork`.

The current work cycle is:
- If wares available: Wait_1 else Stop
- Work
- Wait_2
- Carry out ware (if any)
- Once back in the building (or if never left as no ware was produced): Restart

Flamefire:
> At least thats the setting I'm currently thinking about as default for myself. On the other hand I agree, that having 4 addons (which I need to find a good name for, so they are next to each other) isn't that good either. But as I really think the "Granite Mines Everywhere" Addon is a good idea, I'd go with 2 anyway - and then 4 isn't that much more.

Agreed.   
As for the sorting: Sorting is done by the name. So instead of `Change coal mine behavior` I'd suggest: `"Mine behavior: Coal`, i.e. especially get rid of the word "Change" because "Change coal mine behavior: Default" doesn't make sense ;-) So the shorter version even reads better.    
Instead of "Mine" one could use "Mining", not sure what is better. E.g. `"Mining behavior (Coal)"` is also possible.

libs/s25main/figures/nofMiner.cpp
Flamefire: Ah my bad. Haven't seen that "useResPt" is set to the first node containing **more** than 1 resource.
Maybe improve the comment: I got confused by "use 1 quantity if any" which reads as if: "use 1 res if there is any res anywhere". Suggestion: "Success: Use 1 quantity if any point with more than 1 was found (never fully exhaust the resource)"

The "if success" also made me think it refers to the following condition

Spikeone: Ah, you were right, this didn't make sense at alle - if a valid point has been found and the "roll" was successfull, the resource should be used - it was the other way round, so you were right

Spikeone:
@Flamefire hope I implemented your suggestions correctly. I have 3 ToDos left, 2 of them are simply for skipping checks which are possibly redundant (in some cases). But saving the current workcycle state is required, so the savegame version increases - and I have to find out on how to correctly push and pop the workcycle

libs/s25main/figures/nofMiner.cpp
```diff
@@ -71,20 +76,100 @@ helpers::OptionalEnum<GoodType> nofMiner::ProduceWare()
...
+
+    MiningBehavior addonSettings = GetMiningBehavior();
+
+    switch (addonSettings)
+    {
+        case MiningBehavior::S4Like:
+        {
+            int sumResAmount = 0;
+
+            MapPoint nonMinumResPt;
```
Flamefire, line 120: Did you mean `nonMinimumResPt`? Also no need for the empty line between those 2 variable initializations.

libs/s25main/figures/nofMiner.cpp
```diff
@@ -71,20 +76,100 @@ helpers::OptionalEnum<GoodType> nofMiner::ProduceWare()
...
+            {
+                isAlteredWorkcycle = true;
+            } else
+            {
+                if(nonMinumResPt.isValid())
+                    world->ReduceResource(nonMinumResPt);
+            }
+        }
+        break;
+        // TODO: Is this required? FindPointWithResource checks if there is at least one
```
Flamefire, line 152: I don't understand this TODO. It simply describes `FindPointWithResource` (almost... It doesn't check if there is at least one, it returns the first it finds so it can be used to check if there is at least one)

libs/s25main/figures/nofMiner.cpp
```diff
@@ -71,20 +76,100 @@ helpers::OptionalEnum<GoodType> nofMiner::ProduceWare()
...
+        case MiningBehavior::Inexhaustible:
+        {
+            const MapPoint resPt = FindPointWithResource(GetRequiredResType());
+            if(!resPt.isValid())
```
Flamefire, line 156: See previous change:
```suggestion
            if(!FindPointWithResource(GetRequiredResType()).isValid())
```

libs/s25main/figures/nofMiner.cpp
```diff
@@ -71,20 +76,100 @@ helpers::OptionalEnum<GoodType> nofMiner::ProduceWare()
...
+                return false;
+        }
+        break;
+        case MiningBehavior::AlwaysAvailable: break;
+        case MiningBehavior::Normal:
+        default:
```
Flamefire, line 162: I'd remove the "default" here. All cases should be handled already.

libs/s25main/figures/nofMiner.h
```diff
@@ -5,11 +5,12 @@
 #pragma once
 
 #include "nofWorkman.h"
+#include "addons/AddonMiningOverhaul.h"
+#include <gameData/GameConsts.h>
```
Flamefire, line 9: Every include slows down compilation, and includes in headers even more --> Transitive includes: Everyone including this header now also includes those headers so any change in e.g. GameConsts.h now requires recompiling this class and every file including this header although this header does not depend on `GameConsts.h`, does it?

TLDR: Be cautious about adding includes, especially to headers. See if it is enough to add them to the cpp files (it is here for the GameConsts.h, but not for the addons file)

libs/s25main/figures/nofMiner.h
```diff
@@ -19,10 +20,13 @@ class nofMiner : public nofWorkman
     unsigned short GetCarryID() const override;
     /// Der Arbeiter erzeugt eine Ware
     helpers::OptionalEnum<GoodType> ProduceWare() override;
+    /// alter workcycle (addon)
+    bool isAlteredWorkcycle;
```
Flamefire, line 24: How is the work cycle "altered" (i.e. "changed", I guess)? I mean: It is the same, it simply produces nothing.
So suggestions: `isFailedMiningAttempt` or `produceNothing` (with a comment: "... in this work cycle")

See also the usage in `ProduceWare` at https://github.com/Return-To-The-Roots/s25client/pull/1501/files#diff-a34581854bed9a5f4ca267f64c0e919f48ffd4b9c09f732351db4c127fb2e932R65-R66 and how the name makes the code more readable: `if(produceNothing) return boost::none`

libs/s25main/figures/nofWorkman.cpp
```diff
@@ -100,13 +100,16 @@ struct NodeHasResource
...
+}
+
+MapPoint nofWorkman::FindPointWithResource(ResourceType type, unsigned radius) const {
     const std::vector<MapPoint> pts =
-      world->GetMatchingPointsInRadius<1>(pos, MINER_RADIUS, NodeHasResource(*world, type), true);
+      world->GetMatchingPointsInRadius<1>(pos, radius, NodeHasResource(*world, type), true);
     if(!pts.empty())
         return pts.front();
 
     workplace->OnOutOfResources();
```
Flamefire, line 112: As you noticed earlier: This likely shouldn't be done here, but at the call site. You can (don't need to) change this, when touching the code already, especially if it bites you in the a** while working on your stuff ;-)

I often do such drive-by-improvements myself, but it is hard to not do too many changes once you start...

Just as a note, no need to change this (now).

libs/s25main/figures/nofWorkman.h
```diff
@@ -8,6 +8,7 @@
 #include "nofBuildingWorker.h"
 #include "gameTypes/GoodTypes.h"
 #include "gameTypes/Resource.h"
+#include <vector>
```
Flamefire, line 11: Unused:
```suggestion
```

libs/s25main/gameData/GameConsts.h
```diff
@@ -20,6 +20,10 @@ constexpr GameSpeed referenceSpeed = GameSpeed::Normal;
 
 /// Reichweite der Bergarbeiter
 constexpr unsigned MINER_RADIUS = 2;
+constexpr unsigned MINER_RADIUS_SETTLERSIV = 1u;
+
+/// maximum quantity for ores
+constexpr unsigned MINER_MAX_QUANTITY = 7u;
```
Flamefire, line 26: This looks so arbitrary here. Especially as the documentation in the `Resource` class states
> Maximum number of resource types and amount is 15!

However it is correct in that the MapLoader uses `7` as the max value for ores, by design of the S2 map format, and also for water (by choice), while it uses 4 for fish.

We should probably also even check/assert this during loading.

BTW: The comment is a hint for an unfit name. How about `MAX_ORE_QUANTITY`? The current name sounds, like this is the maximum number of miners you can have and requires the comment to make it clear ;-)

It might still make sense to keep the comment but as something like "Maximum amount of the (mineable) ore resources`.  This adds 2 keywords: "mineable" and "resource" making it easier to know where this might be used.   
I hope you get my way to think about this.

libs/s25main/world/MapBase.cpp
```diff
@@ -182,3 +182,9 @@ MapPoint MapBase::MakeMapPoint(Position pt) const
 {
     return ::MakeMapPoint(pt, size_);
 }
+
+unsigned MapBase::numPointsRadius(unsigned radius, bool includePt) const {
```
Flamefire, line 186: Ah I already had a formula for that :D

I'd suggest to make this a `constexpr` function in the header to allow for better optimizations and usage in template arguments (as you tried earlier). Simply replace the semicolon at https://github.com/Return-To-The-Roots/s25client/pull/1501/files#diff-61d8eac8fa8b6b4b316044fff0023aeb2eccf6858afb2e54ee4a6faa90558d1dR98 by the part between and including the braces and put `constexpr` in front of the return type. Done, no other changes required.

libs/s25main/world/MapBase.cpp
```diff
@@ -182,3 +182,9 @@ MapPoint MapBase::MakeMapPoint(Position pt) const
...
+    // For every additional radius we get 6 * curRadius more points. Hence we have 6 * sum(1..radius) points + the
+    // center point if requested This can be reduced via the gauss formula to the following:
+    return (radius * radius + radius) * 3u + (includePt ? 1u : 0u);
+}
```
Flamefire, line 190: Watch out for this: Github puts a symbol here to show "No newline at end of file".

So make sure each file ends with an empty line. Otherwise if you add e.g. a new function here (best example: what you did here) it would show this line as changed (added newline at the end of this line) instead of nicely adding (here 6) new lines (as it does now because the file had an empty line at the end)

Flamefire (changes requested):
> I have 3 ToDos left, 2 of them are simply for skipping checks which are possibly redundant (in some cases). But saving the current workcycle state is required, so the savegame version increases 

I don't think it is really possible to skip the checks, or worth it... IMO ok like it is.
As for saving/loading: You have a FIFO buffer here, i.e. what goes in first, comes out first.

As the base class needs to be constructed first, it means it needs to be saved first. I.e. your miner-save function starts with calling the workman save function. Then you push your own states: `sgd.pushBool(produceNothing)` (IIRC)

The loading is then the other way round: First the base class, done by `: nofWorkman(sgd, obj_id)` in the constructor call. Then you pop your own states: `, produceNothing(sgd.popBool())` (instead of the `, isAlteredWorkcycle(false)` in the other constructor)

However as we add stuff here we could keep compatibility with the old version: If the sgd version is smaller than the new version, init the field with `false` (we didn't have empty cycles before), else pop the value. This way you can load old and new save games :) Search for "sgd.GetGameDataVersion()" to see some examples.

libs/s25main/gameData/GameConsts.h
Spikeone: Maybe I also failed a bit to correctly distinguish, as I've never seen a well as a mine - which makes the name ```MINER_RADIUS_SETTLERSIV``` also wrong.

libs/s25main/figures/nofMiner.h
Spikeone: I knew you'd stumble across this :)

I called it altered for two reasons, I may also add an addon which lets you switch between "Produce nothing" and "Produce granite" when using settlers IV mining. This also allows other addons (if one likes) like gold mines have a chance to produce a gem. So in my special case, I could name the variable like you suggested, but as this can be a more general variable I named it ```isAlteredWorkcycle```. Are you good with that, if not, I will change it until this variable becomes another use.

libs/s25main/figures/nofMiner.h
Spikeone: Allright, never put that into consideration. Don't know why, but I had in mind to always add the includes to the ```*.h``` file. Which makes no sense at all I guess, or rather has no use.

libs/s25main/addons/AddonMiningOverhaul.h
```diff
@@ -0,0 +1,52 @@
+// Copyright (C) 2005 - 2021 Settlers Freaks (sf-team at siedler25.org)
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#pragma once
+
+#include "AddonList.h"
+#include "mygettext/mygettext.h"
+
+class AddonMiningOverhaulBase : public AddonList
```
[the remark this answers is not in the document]
Spikeone: > Why this lengthy name (here and in the name)? How would you translate it (to German)?

AddonZumAnpassenDerBergwerklogik :D 

I'm not so happy with the behavior anymore. This  somewhat more feels like something visible, possibly a worker. Like how does a soldier behave when his home building is lost, does he return, does he wander around. Overhaul implied some change of logic, which is often invisible and just like "physics", you feel the result but can't grab the reason itself - it's just the way it is. Can change the name if you don't agree to that though :)

> "Mining behavior (coal): No change" vs "Mining overhaul: Coal: No Change"

I don't see where this Representation comes from, as there is no textual representation that way as values are highlighted by the right column. So I rather see ```Mining behavior (coal)``` vs ```Mining overhaul: coal``` where I liked the second version more, besides the discussion if it should be overhaul or behavior. Anyway, I don't care that much for the name in the window (except lenghty names like "Number of scouts required fore exploration expedition"). So same here, can change the name if you'd like me to :)

Flamefire (changes requested):
Some more comments.

BTW: Do you use Visual Studio?
Download this [static-build of clangformat](https://github.com/Return-To-The-Roots/s25client/files/8153980/clang-format-10.zip) Just put this anywhere in your PATH, or into a new folder which you add to your PATH environment variable. Then rerun CMake which will find this (or add it manually under `ClangFormat_BINARY`)   
This will add another target which you can rightclick->build/erzeugen and that will reformat all changed(!) files (on first time, all files as it doesn't know which were changed)

libs/s25main/addons/AddonMiningOverhaul.h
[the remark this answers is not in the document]
falbrechtskirchinger: If I may add my 2 cents:

First, on "mine" vs "mining":
- "Mining overhaul", as the act of mining is being overhauled.
- "Mine behavior", as behavior applies to the exhaustion of the mine.

Personally, I'd use "Mine exhaustion (Coal)"/"Minenerschöpfung (Kohle)". Also, note the parentheses. Otherwise, you'd end up with an awkward double-colon when writing out settings: "Mine exhaustion: Coal: Default"

An alternative might be "Mine productivity"/"Minenergiebigkeit". I don't like it but maybe it sparks someone else's idea.

Spikeone:
Currently (as far as I remember) the following is missing:
AI:
- should honor possible settings (i.e. if it's allowed to mine gold everywhere, the AI should do that instead of searching)
- should honor combinations (i.e. granite mines always but all other mines just as they are)
- should somewhat honor the current accumulated productivity of mines (i.e. 10 coal mines at 10% are just as good as 1 coal mine at 100%)

Alternative production cycle product:
This one alters whats produced when an alternate production cycle runs. The addon should have the following settings:
- produce nothing (default)
- produce granite (25%)
- produce granite (50%)
- produce granite (100%)
- produce lower grade (e.g. gold mine produces iron, iron produce coal, coal produce granite, and granite just nothing)
Although the last setting doesn't make sense logical, it does make sense for the gameplay.
This addon is compatible to other addon settings, since it is only taken into account when setting the alternate cycle resource

Show productivity:
This one shows a productivity rate for mines. If a mine type is inexhaustible, 100% is shown anyway. Maybe some color could indicate if this mine is already at it's lowest rate or not (e.g. >50% = "good green", >min% = "okay yellow", == min% = "bad red")
settings:
- enable
- disable
This addon is compatible to other settings, since it should show 100% if original mining behavior is enabled or inexhaustible.

FYI @falbrechtskirchinger

Spikeone:
@Flamefire mayber we can get this working as well. I don't know if there was any TODO left - but I was missing an addional helper addon for this one, which should do the following:
Settlers IV mining on nothing:
- empty run
- 50% chance stone
- 100% chance stone
- lower grade resource (e.g. gold mine finds iron, iron mine finds coal, coal mine finds stone, granite mine finds nothing)

Flamefire:
@Spikeone Can you edit the initial description of this PR to describe what exactly this should do and how it looks like. 4yrs later I don't really remember this anymore. Especially with your last comment I don't really know which addons there will be, which added, which removed (any?) and what is missing.
You can use TODO lists with `- [ ]` here

Spikeone:
@Flamefire updated the initial comment, described the addon some more

Flamefire:
This is currently continued and will be done by https://github.com/Return-To-The-Roots/s25client/pull/1942 which uses the ideas from here
So closing this :)

PR #1502, opened 2022-02-26
Improve random map generator config window 

Flamefire:
- Reduce max map size to 384 (> 256 already causes lags, so 1024 is way to much)
- Array of selectable map sizes allows easy adding/changing.
- Don't close on right-click, only via apply/close buttons

@Spikeone I used the sizes suggested via Discord. Did I got them correctly? Any other sizes worth adding? See https://github.com/Return-To-The-Roots/s25client/compare/master...Flamefire:mapGenWindow?expand=1#diff-f4365764480163081761e529ad52c318a8f0abf3b61e25cf5b9f3869c9eddd19R23-R24


Spikeone:
> @Spikeone I used the sizes suggested via Discord. Did I got them correctly? Any other sizes worth adding? See https://github.com/Return-To-The-Roots/s25client/compare/master...Flamefire:mapGenWindow?expand=1#diff-f4365764480163081761e529ad52c318a8f0abf3b61e25cf5b9f3869c9eddd19R23-R24

I'm not sure why we do have fixed sizes. It'd be better to have a dropdown for width and height. The intervall may be like in the original editor n * 16, with 2 <= n <= 16 (so 32 to 256), if we do also support maps up to 384, this intervall needs adjustment.

Flamefire:
> I'm not sure why we do have fixed sizes. It'd be better to have a dropdown for width and height. The intervall may be like in the original editor n * 16, with 2 <= n <= 16 (so 32 to 256), if we do also support maps up to 384, this intervall needs adjustment.

Great idea!

Well, "support" is hard of a word... We actually "support" Maps up to 2048x2048, i.e. that is the maximum the code can handle. It is only that the bigger the map gets, the more needs to be processed in each GF which causes those lags. But this is also dependent on the processing power, i.e. CPU, RAM, ...

Flamefire:
Added sizes from 32 to 320 in steps of 32 and separate for X & Y:

Spikeone:
> Well, "support" is hard of a word... We actually "support" Maps up to 2048x2048, i.e. that is the maximum the code can handle. It is only that the bigger the map gets, the more needs to be processed in each GF which causes those lags. But this is also dependent on the processing power, i.e. CPU, RAM, ...

Yeah agreed, in most games such settings are then marked in some way or inform the user, that he migh experience lags or can't properly save/load games. Something like all sizes larger than 256x256 are marked red in dialogues

Flamefire:
> Something like all sizes larger than 256x256 are marked red in dialogues

Good idea. However I'm not even sure we can always handle savegames of 256x256 maps :D

And we need a good way to implement such a highlight. Here we could have a red text below "Größe" saying "Non-standard size: > 256x256", to keep it short.
(Currently) we can't color a row in the maps table, but we could either show a confirmation dialog when pressing "Next" or a red text at the bottom next to the map description.

But again: 256x256 is pretty arbitrary. Calculation time probably scales linearly or roughly quadratically with the map size (mostly pathfinding I guess, and the number of objects) So even this size could be to much for some CPUs or some might handle 512x512 just fine... Hard to tell.

PR #1511, opened 2022-03-25
Update nix shell to current repository eco system.

Shawn8901:
The shell.nix file which was added to the project is a quite bit outdated and does not result in a runtime which is capable of compiling the game.
As i am currently running nixos (which is a distro build around the package manager nix, which at the end, the shell.nix can possibly be used in), i gave it a try to update it to the current state of the nix eco system.

The old version was mostly broken due to not having a SDL2 which includes a `libSDL2Main.a`, thus the cmake modules do not find the needed version information. 

edit:
sorry for force pushing again. but after cleaning old devShells i noticed that it also requires libsamplerate which was because of some reason on the shell before by something else.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,11 +1,11 @@
-# Copyright (C) 2005 - 2021 Settlers Freaks <sf-team at siedler25.org>
+# Copyright (C) 2005 - 2022 Settlers Freaks <sf-team at siedler25.org>
 #
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 { pkgs ? import <nixpkgs> {} }:
 
 with pkgs;
-stdenv.mkDerivation {
+mkShell {
   name = "s25client";
   buildInputs = [
     boost
@@ -17,5 +17,6 @@ stdenv.mkDerivation {
     miniupnpc
     SDL2
     SDL2_mixer
+    libsamplerate
   ];
 }
--- a/shell.nix
+++ b/shell.nix
@@ -18,7 +18,15 @@
 with pkgs;
 mkShell {
   name = "s25client";
+
+  # Set SOURCE_DATE_EPOCH to the newest file in the directory.
+  # RTTR_BUILD_DATE is calculated by that epoch value.
+  shellHook = ''
+    updateSourceDateEpoch .
+  '';
+
   buildInputs = [
+    stdenv
     boost
     bzip2
     cmake
--- a/shell.nix
+++ b/shell.nix
@@ -2,7 +2,18 @@
 #
 # SPDX-License-Identifier: GPL-2.0-or-later
 
-{ pkgs ? import <nixpkgs> {} }:
+{
+  pkgs ? import <nixpkgs> { overlays = [
+    # This is a workaround for https://github.com/NixOS/nixpkgs/issues/146759
+    # NixOS does currently ship an incomple SDL2 installation when withStatic=false.
+    # See https://nixpk.gs/pr-tracker.html?pr=149601 as state for a PR on NixOS side.
+    (final: prev: {
+      SDL2 = prev.SDL2.override {
+        withStatic = true;
+      };
+    })];
+  }
+}:
 
 with pkgs;
 mkShell {


shell.nix
```diff
@@ -1,11 +1,19 @@
...
+# Copyright (C) 2005 - 2022 Settlers Freaks <sf-team at siedler25.org>
 #
 # SPDX-License-Identifier: GPL-2.0-or-later
 
-{ pkgs ? import <nixpkgs> {} }:
+{
+  pkgs ? import <nixpkgs> { overlays = [
+    (final: prev: {
+      SDL2 = prev.SDL2.override {
+        withStatic = true;
```
Flamefire, line 9: What does this actually do? Note that in upstream SDL2 for either autotools as well as the CMake build a static SDL2main library is always build. And we don't need a static SDL2, i.e. a `libSDL2.a`

It appears to me to be rather a bug in the Nix SDL2 build/installation, but would need clarification here. I think an SDL2 installation without the libSDL2main.a library can be considered incomplete. If that isn't the case we could also make the dependency on that library optional in our CMake find module as IIRC we don't need that anyway,

Shawn8901: Maybe as additional info about what does it do, as i am not sure if you asked why i am doing it, or what is this technically.


```
<package> = prev.<package>.override {
        <some inputs>
};
```
is the nix packages pattern to override a package definition from the repository.
So there is a package SDL in that case, which has some inputs, but in  general everything from the package definition can be overriden. 
The SDL package itself has a input variable, which can be set, to define that static files should be kept in place after compiling, whilst the default behavior deletes them after compile was done.

By the shown definition, its defined, that the shell build by `shell.nix` should make a custom version of SDL on the PATH available, which still has the static files.

What can be set or overriden depends on the packages in the nixpkgs.
For SDL2 that build definion is located here:

https://github.com/NixOS/nixpkgs/blob/3838b7e07e8a79ece6cb249eb4bf0a2bfd6d388a/pkgs/development/libraries/SDL2/default.nix#L48


I hope that answers the technical side of what the heck is done there.
Perma link jumps to the parameter definition.

Flamefire: >  But libSDL2main.a is somehow special and needed by cmake to read the versions.

To be exact: The SDL2 installation consists of static or dynamic libraries AND this static library. CMake tries to find a (valid) SDL2 installation and as this static library is part of that the search fails if it is missing. So:

> Nixpkgs removes all .a files when not including withStatic,

Is a bug in Nix which must not remove the (always) static libSDLmain library, see https://github.com/NixOS/nixpkgs/issues/146759#issuecomment-974718200

I'm fine with including the workaround here if that does work but please add a comment above it to clearly mark it as a workaround for a Nix bug and possibly even link the issue there so it can be removed once the Nix package is fixed. Maybe this explanation is also useful for the Nix issue, although I think there should be a separate one about the missing `libSDL2main.a` not generically `*.a`. I'll leave that to you as you're more familiar with the Nix stuff.

Or correct me if I misunderstood anything.

Shawn8901: Absolut correct summarized. Ill add the mentioned improvement, as i am watcher on the PR which should fix  the libSDLmain.a topic ill get a notification if that gets merged downstream, in the meantime i am also trying to get some attention to that on their communication platforms, as the package maintainer on their side seems not very active (but thats just guessing on public github activity)

shell.nix
```diff
@@ -1,11 +1,27 @@
...
+{
+  pkgs ? import <nixpkgs> { overlays = [
+    # This is a workaround for https://github.com/NixOS/nixpkgs/issues/146759
+    # && https://github.com/NixOS/nixpkgs/pull/149601 .
+    # NixOS does currently ship an incomple SDL2 installation
+    # when withStatic is set to false, see the issue for more info.
+    # The overlay can be safely removed as soon as the PR hits the targeted
+    # OS channels, see https://nixpk.gs/pr-tracker.html?pr=149601 as state overview.
+    # nixos-unstable should be a reasonable target, when we could consider
+    # removing this workaround
```
Flamefire, line 14: The issue link is enough for the reference/explanation. Let's keep this short, e.g.:

```suggestion
    # This is a workaround for https://github.com/NixOS/nixpkgs/issues/146759 :
    # NixOS does currently ship an incomplete SDL2 installation when withStatic=false.
    # nixos-unstable should be a reasonable target, to remove this workaround.
    # See https://nixpk.gs/pr-tracker.html?pr=149601 for a state overview.
```

Flamefire (changes requested):
I would prefer to have this as 2 commits, so could you split this into 2? The first should contain the year update and add the missing libsamplerate dependency. The 2nd commit would be the SDL2 workaround. I.e. (example commit titles, feel free to reword them as you like):
- "Nix shell: Add libsamplerate dependency"
- "Nix shell: Workaround for incomplete SDL2 package"

First because this is cleaner: Those are 2 unrelated changes. And 2nd: It allows to simply (fully) revert the 2nd commit when the Nix package is fixed. Which is why "clean" is "good". I.e. one can later do `git revert <commit-sha>`

I also added a suggestion to make the comment a bit shorter and fix a typo. See if that's enough or if you'd add anything.

shell.nix
Shawn8901: I have shortened the text and split the commits, both makes absolute sense.

Flamefire (dismissed):

Flamefire:
LGTM. As I don't have any environment to test this: Have you tried this in the current state if it works?

Shawn8901:
Yes it results in an environment which can compile the game. 

Only open topic, but that was open since the `shell.nix` or the test has been added: `Test_rttrConfig` fails, as its relying on the current system date on build time. 
Nix build system is statically setting that to 19800101, which does fail the test as its checking for a year >2000.
And i don't see any possibility to change that behavior, at the moment.

Flamefire:
> Only open topic, but that was open since the `shell.nix` or the test has been added: `Test_rttrConfig` fails, as its relying on the current system date on build time. Nix build system is statically setting that to 19800101, which does fail the test as its checking for a year >2000. And i don't see any possibility to change that behavior, at the moment.

Why is that? Doesn't seem to really make sense to me.

FWIW: The test is this: https://github.com/Return-To-The-Roots/s25client/blob/397f2b2315e997504d4958bfbdea0af815ce559a/tests/rttrConfig/testRttrConfig.cpp#L68 and https://github.com/Return-To-The-Roots/s25client/blob/397f2b2315e997504d4958bfbdea0af815ce559a/tests/rttrConfig/testRttrConfig.cpp#diff-06081d187998cf02cc78e8743ff625930f76a9cd6323bee5d2820e862ddadc1aR78
So you may add a patch there

We set that at https://github.com/Return-To-The-Roots/s25client/blob/a24429ffa9426557304144d523729013da3799f4/libs/rttrConfig/CMakeLists.txt#L25 but I don't see any worth in changing this or the test as everything is working as intended and a "wrong" system time is outside the scope of that. So the Nix guys would need to figure out how to pass the build date/time to the build system (i.e. CMake)

Shawn8901:
> Why is that? Doesn't seem to really make sense to me.

When focusing reproducible builds a build should not to rely on things like systems epoch, as re-building the same code to a later time does produce different outputs, so it will not produce the same binary and so its not reproducible. 
RTTR does not focus that, as its using systems epoch at least on the version string for non-release builds (which is of course not bad, by itself).

A better behavior from the idea of having builds  reproducible, the code could use git commits date, so a build today has the exact same output like a fresh build tomorrow, if its on the same commit.
Maybe that does give a bit additional context: https://reproducible-builds.org/specs/source-date-epoch/

As cmake is aware of such things, and in nix the epoch is set to 315532800 (which is a deterministic non 0 value => 1980-01-01) _if no other source is given_.
As  I did some research where nix is setting that magic value l found that spec (before i did just know why) i also found the solution for the build date used for non release builds and also for the unit test.
When setting SOURCE_DATE_EPOCH to the commits date, which can be done in the `shell.nix` both issues are addressed: The programmer has a build date (ok for nix its the commit date in that case, but at least reproducible), and the unit test passes, as the commit date passes the sanity check.


> but I don't see any worth in changing this or the test as everything is working as intended and a "wrong" system time is outside the scope of that. So the Nix guys would need to figure out how to pass the build date/time to the build system (i.e. CMake)

I agree that its not the goal to patch that out of rttrConfig-Test nor of the cmake definition.
As its not per definition nix specific but reproducible build specific cmake luckily is already aware of such a thing if i had known that before...  :)

At the end i just want to repeat, that this has been an issue before, so its likely that the person who has added the `shell.nix` either did never run the tests, the test was added later, or the person did not care about the failing test, of course at the end its not important what went wrong, as its fixed now. :+1:

Flamefire:
> A better behavior from the idea of having builds reproducible, the code could use git commits date,

The problem with this is that this only works with git checkouts, not when the source code has been obtained/transferred in other means. E.g. the git tags on Github create source tarballs and we add an own one which includes the submodules, see e.g. https://github.com/Return-To-The-Roots/s25client/releases/tag/v0.9.5   
It also requires to have git installed, which is minor of course.

Anyway: Is it possible, to provide a fallback value? I.e. if the git command fails for any reason (git not found, not a git repo, ...) use some default with a year >= 2000, e.g. 2005 (same as the copyright notices)

Shawn8901:
i guess setting that to current date is the most desired solution without workaround all kind of things. that's at least the behavior you expect from cmake anyways.
Setting the build date to some other magical value, to just pass the sanity check, i really don't know...

As there wasn't any real user for the nix shell anyways (otherwise some one should have noticed that it was broken for long time), its maybe also more reasonable to drop the file, tbh.

Flamefire (approved):
Thanks!

PR #1518, opened 2022-04-02
Add CMake option to not install placeholder files

Shawn8901:
This option enables the possibility for a package maintainer to let RTTR_GAMEDIR to a place
which is not in control of the packaging system. Thats needed when RTTR_GAMEDIR should point
to $HOME, if the packaging system prevents writes to the RTTR_GAMEDIR on package build time.

The default keeps the old behavior.

To explain a bit more about the motivation. On discord there was a short discussion in the development channel between Flow and myself. I was asking how i can archive a build of RTTR which does not "depend" on having the game files on a predictable place in FHS. The reason for that is that i am trying to create a package for rttr for the packaging system "nix". 
This contains a read-only store located at `/nix/store`. As S2 games are in common cases stored either in the rttr folder structure, or would point somewhere else in FHS, it requires on nix to be present before the future rttr package got compiled. This is a bad user experience for a nix user on my opinion, as i will be build on the endusers machine in that case.

The nix people suggested to ask rttr project to change rttr itself to be able to define the gamedir on runtime (as dialogue or runtime argument), so l dropped a question about that on discord.

Flow suggested to let a nix rttr package to point to $HOME and patch out writing for the placeholder file (as nix will prevent any writes to $HOME on build time).

So this is the reason for this patch.
As i am not sure about the solution at all and do not yet have a integrated solution to the nix packages i start this PR as a Draft to get some discussion ongoing it thats okay or not, whilst doing the same for on the nix project.
As i am just planning on publishing stable versions (so upstream releases) a nix package has to ship this patch for 0.9.5 anyways.

I hope that the motivation was understandable, if there are questions feel free to ask here or ping me on discord.
If there is another suggestion how to get it running on nix, i would like to hear about them and test them.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -384,7 +384,11 @@ if(NOT WIN32)
 endif()
 
 # Placeholder for S2 installation
-install(FILES "${RTTR_S2_PLACEHOLDER_PATH}" DESTINATION "${RTTR_GAMEDIR}")
+option(RTTR_INSTALL_PLACEHOLDER "Install a placeholder file to the location the S2 game files should be copied to." ON)
+mark_as_advanced(RTTR_INSTALL_PLACEHOLDER)
+if(RTTR_INSTALL_PLACEHOLDER)
+    install(FILES "${RTTR_S2_PLACEHOLDER_PATH}" DESTINATION "${RTTR_GAMEDIR}")
+endif()
 
 ################################################################################
 # Postbuild


Shawn8901:
> I'm not sure on how Nix works. 
Nix builds a system store for its packages under `/nix/store`and does not have to known filesystem hierarchy (FHS) like other linux systems. When linking binaries it links to a version in the package store. The "names" in the store are predictable hashes over build instructions, to you can build binary A and let B depend on that and let binary C depend on a derivation of A without conflicts. Somehow a bit fancy and not really focus there. 

For this change the important info is that a package in nix-store is immutable, so a user can not place S2 files at `<RTTR_GAMEDIR>` (not sure what the default is), and also all write operations outside the build-directory are not allowed. 
As the store is read only `<RTTR_GAMEDIR>` has to be set on compile time to point somewhere else, so the user can drop files at that location.
But as write operations outside the packages build directory are now allowed that crashes the install phase.
And thats what i try to patch out.
From nix maintainer perspective the best would be to have `<RTTR_GAMEDIR>` set-able on runtime, so we could create a wrapper around the game binaries, so nix would handle everything for the user (thats at least what i got as feedback from their side).

In a discussion about that on discord Flow came with this idea, so that i can set `<RTTR_GAMEDIR>` to `~/.s25rttr/S2` (if the placeholder is not written).
 


> Am I right that it compiles RttR (in this case) from source on the end users machine? How are the CMake variables defined then? Especially: How is RTTR_GAMEDIR set? 

There is a open PR to have s25rttr as package in the nixpkgs repo, if that is accepted, it would be build by the nixos build system, and in that case the end-user will just download the compiled binary.
I guess it also explains the cmake questions.

> Setting it to `$HOME` is obviously non-portable/relocatable but likely enough if it is built on the users system.
 
`~/.s25rttr/S2` is not resolved on compile time, i know very nasty, but it works at least. Like written a argument like `-gamedir` on start would be the best, if such a PR would be accepted.


> 
> I especially struggle to understand the following sentence which sounds like it is missing some part:
> 
> > As S2 games are in common cases stored either in the rttr folder structure, or would point somewhere else in FHS, it requires on nix to be present before the future rttr package got compiled.
> 

So as the nix store is read only, i need everything which is required by RTTR to be one of the following 
* shipped by upstream, which is ofc not possible as S2 files are proprietary
* be present in nix store before we build rttr, which is also not possible as the nix built system will never have S2 files, so then compiling will be done on users machine (which works without issue, but sucks as solution) 
* To be able to have it no present on compile time, which is possible by setting `<RTTR_GAMEDIR>` to a predictable writeable folder on the system.
* Have it setable on runtime, which is currently not implemented in the game. 


> As for this PR: I'm fine with that approach but would suggest 2 changes:
> 
>     * Change description to describe a checkbox as that is where it is usually shown (i.e. what it does) and be complete, e.g. "Install a placeholder file to the location the S2 game files should be copied to"
> 
>     * Possible: Add a `mark_as_advanced` call on that variable so it isn't shown to regular users. Not sure if this variable would be used outside (very) special cases such as Nix, so hide it by default. Not sure here, just an idea

i will look into that.


I can explain more if you want on discord about the nix system and the issue, if you want. Maybe we also find a better solution, but its at least what Flow and i could figure out to get it working.

Shawn8901:
i have done the mentioned changes. 

beside that i am currently thinking about changing the S2 game dir to point to `/var/li/s25rttr` which is from FHS documentation for status data, and its disscusable if  the Game Data should go there, but that should result in a nicer package on nixos side, than the ~/. hack.
That just as side note and it will still require that PR at the end. 
(i am shipping an exact duplicate of that patch on downstream, so that's not a blocker on my side, but at the end its nicer to have less custom patching on downstream)

Flamefire:
Good to go from my side.

A more general note: Git guides recommend to use imperative & present tense in commit messages, i.e. describe what the commit **does** not what it **did**. See my change to the PR title. This reads nicer in the logs and often results in shorter messages and hence is easier to handle long-term.

Also a few typos in the commit message. Quote with suggestions for changes:
> This option enables the possibility for a package maintainer to ~~let~~ set RTTR_GAMEDIR to a place
which is not in control of the packaging system. ~~Thats~~That's needed when RTTR_GAMEDIR should point
to a directory which is not **in(?)** the package, ~~but~~**and(?)** the packaging system prevents writes on package build time **or the directory is otherwise read-only during build/install**.
>
>As the **use case** is special for some packaging systems and not ~~indended~~intended for daily use ~~its~~it is marked as **an** advanced option.

But that just as a note/suggestion, not really required.

Shawn8901:
Mostly thats just super bad habit on my side.
I did some changes on the commit message and also added some info about the use case (NixOS). So that a person has at least the chance to find that special use case.

Sadly i did not yet get any feedback on my PR on nixos side (they are currently busy with their next release), so it depends on you if the PR should be included even if the package is not available on the system which needs that.

Flamefire (approved):

Issue #1524, opened 2022-05-08
Flathub


Mailaender:
This is a notification that I submitted a package to Flathub which is an alternative to #1078.

zocker-160:
I just found this and your submission request for Flathub.

Couldn't you just add a simple start script, which checks if the games files are available and if not show a popup message, informing the user that the files are missing?

zocker-160:
@Mailaender I created a small script template, which should do everything you need

```bash
#! /usr/bin/env bash

TARGET_DIRECTORY=`pwd`
FILE="${TARGET_DIRECTORY}/lool"

if [ ! -s $FILE ]; then
    echo "$FILE does not exist"
    
    zenity --title="s25client data check" --width=300 --question --no-wrap --text="Missing required game files!" --cancel-label="Exit" --ok-label="Open folder"
    
    if [ $? == 0 ]; then
        xdg-open $TARGET_DIRECTORY
    else
        echo "exiting"
    fi
    
else
    # start s25client normally
    echo "starting s25client...."
fi
```

you could just put that as the start script into the flatpak, zenity is part of the runtime iirc

Mailaender:
I know how Zenity works. Checking for a single file is not sufficient here.

Flow86:
you can look at the MacOS Launcher application we have exactly for this on MacOS: https://github.com/Return-To-The-Roots/s25client/blob/master/extras/macosLauncher/starter.cpp

Mailaender:
https://flathub.org/apps/details/info.rttr.Return-To-The-Roots was already created months ago.

PR #1529, opened 2022-07-03
Added a popup when gamefiles are missing.

Mailaender:
required for #1524

### DIFF
--- a/extras/videoDrivers/WinAPI/WinAPI.cpp
+++ b/extras/videoDrivers/WinAPI/WinAPI.cpp
@@ -16,6 +16,7 @@
 #include <cstring>
 #include <iostream>
 #include <limits>
+#include <winuser.h>
 
 /**
  *  Zeiger auf die aktuelle Instanz.
@@ -441,6 +442,11 @@ bool VideoWinAPI::MessageLoop()
         DispatchMessage(&msg);
     }
     return true;
+}
+
+void VideoWinAPI::ShowErrorMessage(const char* title, const char* message)
+{
+    MessageBoxW(nullptr, boost::nowide::widen(message).c_str(), boost::nowide::widen(title).c_str(), MB_OK);
 }
 
 /**
--- a/extras/videoDrivers/WinAPI/WinAPI.h
+++ b/extras/videoDrivers/WinAPI/WinAPI.h
@@ -44,6 +44,9 @@ public:
     /// Die Nachrichtenschleife.
     bool MessageLoop() override;
 
+    /// Popup Window
+    void ShowErrorMessage(const char* title, const char* message) override;
+
     /// Funktion zum Auslesen des TickCounts.
     unsigned long GetTickCount() const override;
 
--- a/libs/driver/include/driver/VideoInterface.h
+++ b/libs/driver/include/driver/VideoInterface.h
@@ -72,6 +72,9 @@ public:
     virtual bool IsInitialized() const = 0;
     /// Shall we support OpenGL? (Disabled for tests)
     virtual bool IsOpenGL() const = 0;
+
+    // Display the problem to the gamer
+    virtual void ShowErrorMessage(const char* title, const char* message) = 0;
 };
 
 class VideoDriverLoaderInterface;
--- a/libs/s25main/desktops/dskSplash.cpp
+++ b/libs/s25main/desktops/dskSplash.cpp
@@ -12,12 +12,14 @@
 #include "WindowManager.h"
 #include "controls/ctrlTimer.h"
 #include "dskMainMenu.h"
+#include "helpers/format.hpp"
 #include "ingameWindows/iwMusicPlayer.h"
 #include "mygettext/mygettext.h"
 #include "ogl/glArchivItem_Bitmap.h"
 #include "gameData/ApplicationLoader.h"
 #include "s25util/Log.h"
 #include "s25util/error.h"
+#include <drivers/VideoDriverWrapper.h>
 
 using namespace std::chrono_literals;
 
@@ -76,9 +78,11 @@ void dskSplash::LoadFiles()
 
     } else
     {
-        s25util::error(_("Some files failed to load.\n"
+        s25util::error(_("Some essential game files failed to load.\n"
                          "Please ensure that the Settlers 2 Gold-Edition is installed \n"
                          "in the same directory as Return to the Roots."));
+        VIDEODRIVER.ShowErrorMessage("Please install Settlers II game files into",
+                                     RTTRCONFIG.ExpandPath("<RTTR_GAME>").string().c_str());
         GLOBALVARS.notdone = false;
     }
 }
--- a/libs/s25main/drivers/VideoDriverWrapper.h
+++ b/libs/s25main/drivers/VideoDriverWrapper.h
@@ -81,6 +81,8 @@ public:
     // Nachrichtenschleife
     bool Run();
 
+    void ShowErrorMessage(const char* title, const char* message);
+
     unsigned GetTickCount();
     /// Set framerate target (FPS)
     /// negative for unlimited, 0 for hardware VSync
--- a/tests/mockupDrivers/MockupVideoDriver.cpp
+++ b/tests/mockupDrivers/MockupVideoDriver.cpp
@@ -4,6 +4,7 @@
 
 #include "MockupVideoDriver.h"
 #include <boost/nowide/iostream.hpp>
+#include <iostream>
 
 MockupVideoDriver::MockupVideoDriver(VideoDriverLoaderInterface* CallBack) : VideoDriver(CallBack), tickCount_(1)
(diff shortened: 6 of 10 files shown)


libs/s25main/desktops/dskSplash.cpp
```diff
@@ -76,6 +77,9 @@ void dskSplash::LoadFiles()
 
     } else
     {
+        VIDEODRIVER.showMessageBox("Please install Settlers II game files into",
```
Flamefire, line 80: I'd rename this to `showErrorMessage` as that is what you implemented. 
Then better put the call below `s25util::error`

Flamefire (changes requested):
The WinAPI video driver needs an implementation. I guess this can be simply used with similar call as SDL: https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-messagebox

libs/s25main/desktops/dskSplash.cpp
Flow86: and please do logging before the message box. if a user aborts the program due to the message box, the log entry will never be written.
We also should be more explicit whats missing.

Flamefire: See
>  Then better put the call below `s25util::error`

tests/mockupDrivers/MockupVideoDriver.cpp
```diff
@@ -81,3 +81,9 @@ void* MockupVideoDriver::GetMapPointer() const
 {
     return nullptr;
 }
+
+void MockupVideoDriver::ShowErrorMessage(const char* title, const char* message)
+{
+    std::cout << title;
```
Flow86, line 87: `#include <iostream>` is missing in this file, also probably some newlines after the title/message should be good?

libs/s25main/drivers/VideoDriverWrapper.h
```diff
@@ -81,6 +81,8 @@ class VideoDriverWrapper : public Singleton<VideoDriverWrapper, SingletonPolicie
     // Nachrichtenschleife
     bool Run();
 
+    void ShowErrorMessage(const char* title, const char* message);
```
Flow86, line 84: `override` is missing here

libs/s25main/desktops/dskSplash.cpp
```diff
@@ -76,9 +77,11 @@ void dskSplash::LoadFiles()
 
     } else
     {
-        s25util::error(_("Some files failed to load.\n"
+        s25util::error(_("Some essential game files failed to load.\n"
                          "Please ensure that the Settlers 2 Gold-Edition is installed \n"
                          "in the same directory as Return to the Roots."));
+        VIDEODRIVER.ShowErrorMessage("Please install Settlers II game files into",
+                                   RTTRCONFIG.ExpandPath("<RTTR_GAME>").string().c_str());
```
Flamefire, line 84:
```suggestion
        VIDEODRIVER.ShowErrorMessage("Please install Settlers II game files into",
                                     RTTRCONFIG.ExpandPath("<RTTR_GAME>").string().c_str());
```

extras/videoDrivers/WinAPI/WinAPI.cpp
```diff
@@ -3,6 +3,7 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 
 #include "WinAPI.h"
+#include <winuser.h>
```
Flamefire, line 6: Move this to the bottom please (includes are sorted to find duplicates)

tests/mockupDrivers/MockupVideoDriver.cpp
```diff
@@ -81,3 +82,9 @@ void* MockupVideoDriver::GetMapPointer() const
 {
     return nullptr;
 }
+
+void MockupVideoDriver::ShowErrorMessage(const char* title, const char* message)
+{
+    std::cout << title << endl;
+    std::cout << message << endl;
```
Flamefire, line 89:
```suggestion
    std::cout << title << std::endl << message << std::endl;
```

Flamefire (changes requested):
Some minor issues as found by CI

extras/videoDrivers/WinAPI/WinAPI.cpp
```diff
@@ -443,6 +444,14 @@ bool VideoWinAPI::MessageLoop()
     return true;
 }
 
+/**
+ *  grafische Fehlermeldung
+ */
+void VideoWinAPI::ShowErrorMessage(const char* title, const char* message)
+{
+    MessageBox(NULL, (LPCWSTR)message, (LPCWSTR)title, MB_OK);
```
Flamefire, line 452: You need to actually create a wide string here, not reinterpret-cast the char string to a wide string. And maybe directly use the wide API, see also https://github.com/Return-To-The-Roots/s25client/blob/544f85a329932db1d5bf3ebd7caea46d062930aa/libs/s25client/s25client.cpp#L132:

```suggestion
    MessageBoxW(nullptr, boost::nowide::widen(message).c_str(), boost::nowide::widen(title).c_str(), MB_OK);
```

extras/videoDrivers/WinAPI/WinAPI.h
```diff
@@ -44,6 +44,9 @@ class VideoWinAPI final : public VideoDriver
     /// Die Nachrichtenschleife.
     bool MessageLoop() override;
 
+    /// Popup Fenster
```
Flamefire, line 47: Can you use English here, preferably the same wording as in `VideoInterface.h` or just remove this

libs/driver/include/driver/VideoInterface.h
```diff
@@ -72,6 +72,9 @@ class BOOST_SYMBOL_VISIBLE IVideoDriver
     virtual bool IsInitialized() const = 0;
     /// Shall we support OpenGL? (Disabled for tests)
     virtual bool IsOpenGL() const = 0;
+
+    // Fehlermeldung grafisch anzeigen
```
Flamefire, line 76: Please use English here and triple slash for documentation.
BTW: Any reason why you use C-strings here and not `const std::string&`? Especially as one of the strings you pass in here is already a `std::string`

Flamefire (changes requested):
Thanks for the update! Did a careful review and found a few issues. Most important are the fix for WinAPI and making the message(s) translatable.

Coverage stuff can be ignored here, not important for this change.

libs/driver/include/driver/VideoInterface.h
Mailaender: I don't really know what I am doing. This is my first time patching a C++ application.

Mailaender: > /usr/local/include/SDL2/SDL_messagebox.h:182:29: note: candidate function not viable: no known conversion from 'const std::string' (aka 'const basic_string<char, char_traits<char>, allocator<char> >') to 'const char *' for 2nd argument

This is the reason. Sorry, it has been too long since I submitted this.

libs/s25main/desktops/dskSplash.cpp
```diff
@@ -76,9 +77,11 @@ void dskSplash::LoadFiles()
 
     } else
     {
-        s25util::error(_("Some files failed to load.\n"
+        s25util::error(_("Some essential game files failed to load.\n"
                          "Please ensure that the Settlers 2 Gold-Edition is installed \n"
                          "in the same directory as Return to the Roots."));
+        VIDEODRIVER.ShowErrorMessage("Please install Settlers II game files into",
+                                     RTTRCONFIG.ExpandPath("<RTTR_GAME>").string().c_str());
```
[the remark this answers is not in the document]
Mailaender: > /Users/runner/work/s25client/s25client/libs/s25main/desktops/dskSplash.cpp:85:38: error: no viable conversion from 'std::string' (aka 'basic_string<char, char_traits<char>, allocator<char> >') to 'const char *'
                                     helpers::format(_("Please install Settlers II game files into %1%"), RTTRCONFIG.ExpandPath("<RTTR_GAME>")));

Flamefire:
Main question still open:
> Have you checked how this looks? I figure this might look a bit odd to have the message consists solely of the path and the actual actionable is in the title.

libs/s25main/desktops/dskSplash.cpp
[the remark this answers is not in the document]
Flamefire: You need to change the method signature to accept `const std::string&` or convert the `std::string` to a `const char*` here as you did before via `.c_str()`.

That depends on what you choose to use as the method signature, sorry for suggesting only the shortest variant here.
`helpers::format(_("Please install Settlers II game files into %1%"), RTTRCONFIG.ExpandPath("<RTTR_GAME>")).c_str()` would be the variant when keeping the current signature. (typed in the browser, so might still contain issues)

PR #1532, opened 2022-08-20
Plugin fixes

### LINKED ISSUE #1521: Nation specific catapult graphics broken
The addon "Race Specific Catapult Graphics" is currently broken. When fixing this, it should be renamed to "Nation Specific Catapult Graphics"

tehKaiN:
Fixes #1521 

obraz

Also brings back Military Aid to its full functionality (note the icon which indicates that enemy guardhouse can be attacked):

obraz

Submodule which needs to be merged prior to this one so that I can update submodule refs: https://github.com/Return-To-The-Roots/languages/pull/20

### DIFF
--- a/libs/s25main/world/GameWorldView.cpp
+++ b/libs/s25main/world/GameWorldView.cpp
@@ -353,9 +353,10 @@ void GameWorldView::DrawNameProductivityOverlay(const TerrainRenderer& terrainRe
             // Is object not belonging to local player?
             if(no->GetPlayer() != gwv.GetPlayerId())
             {
-                auto* attackAidImage = LOADER.GetImageN("map_new", 20000);
                 if(GetWorld().GetGGS().getSelection(AddonId::MILITARY_AID) == 2 && gwv.GetNumSoldiersForAttack(pt) > 0)
-                    attackAidImage->DrawFull(curPos - DrawPoint(0, attackAidImage->getHeight()));
+                {
+                    LOADER.GetImageN("map_new", 20000)->DrawFull(curPos - DrawPoint(0, attackAidImage->getHeight()));
+                }
                 continue;
             }
 
--- a/libs/s25main/world/GameWorldView.cpp
+++ b/libs/s25main/world/GameWorldView.cpp
@@ -347,12 +347,18 @@ void GameWorldView::DrawNameProductivityOverlay(const TerrainRenderer& terrainRe
             if(!no)
                 continue;
 
-            // Is object not belonging to local player?
-            if(no->GetPlayer() != gwv.GetPlayerId())
-                continue;
-
             Position curPos = GetWorld().GetNodePos(pt) - offset + curOffset;
             curPos.y -= 22;
+
+            // Is object not belonging to local player?
+            if(no->GetPlayer() != gwv.GetPlayerId())
+            {
+                auto *attackAidImage = LOADER.GetImageN("map_new", 20000);
+                if(GetWorld().GetGGS().isEnabled(AddonId::MILITARY_AID))
+                    if(gwv.GetNumSoldiersForAttack(pt) > 0) // soldiers available for attack?
+                        attackAidImage->DrawFull(curPos - DrawPoint(0, attackAidImage->getHeight()));
+                continue;
+            }
 
             // Draw object name
             if(show_names)
@@ -501,13 +507,6 @@ void GameWorldView::DrawObject(const MapPoint& pt, const DrawPoint& curPos)
         return;
 
     obj->Draw(curPos);
-
-    return;
-    // TODO: military aid - display icon overlay of attack possibility
-    RTTR_IGNORE_UNREACHABLE_CODE
-    if(gwv.GetNumSoldiersForAttack(pt) > 0) // soldiers available for attack?
-        LOADER.GetImageN("map_new", 20000)->DrawFull(curPos + DrawPoint(1, -5));
-    RTTR_POP_DIAGNOSTIC
 }
 
 void GameWorldView::DrawBoundaryStone(const MapPoint& pt, const DrawPoint pos, Visibility vis)
--- a/libs/s25main/world/GameWorldView.cpp
+++ b/libs/s25main/world/GameWorldView.cpp
@@ -355,7 +355,8 @@ void GameWorldView::DrawNameProductivityOverlay(const TerrainRenderer& terrainRe
             {
                 if(GetWorld().GetGGS().getSelection(AddonId::MILITARY_AID) == 2 && gwv.GetNumSoldiersForAttack(pt) > 0)
                 {
-                    LOADER.GetImageN("map_new", 20000)->DrawFull(curPos - DrawPoint(0, attackAidImage->getHeight()));
+                    auto* attackAidImage = LOADER.GetImageN("map_new", 20000);
+                    attackAidImage->DrawFull(curPos - DrawPoint(0, attackAidImage->getHeight()));
                 }
                 continue;
             }
--- a/libs/s25main/world/GameWorldView.cpp
+++ b/libs/s25main/world/GameWorldView.cpp
@@ -354,7 +354,7 @@ void GameWorldView::DrawNameProductivityOverlay(const TerrainRenderer& terrainRe
             if(no->GetPlayer() != gwv.GetPlayerId())
             {
                 auto *attackAidImage = LOADER.GetImageN("map_new", 20000);
-                if(GetWorld().GetGGS().isEnabled(AddonId::MILITARY_AID))
+                if(GetWorld().GetGGS().getSelection(AddonId::MILITARY_AID) == 2)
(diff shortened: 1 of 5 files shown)


tehKaiN:
I've made attack indicators optional, for ppl who may dislike them. Added relevant translations

obraz

obraz

libs/s25main/world/GameWorldView.cpp
```diff
@@ -347,12 +347,18 @@ void GameWorldView::DrawNameProductivityOverlay(const TerrainRenderer& terrainRe
...
+            Position curPos = GetWorld().GetNodePos(pt) - offset + curOffset;
+            curPos.y -= 22;
+
             // Is object not belonging to local player?
             if(no->GetPlayer() != gwv.GetPlayerId())
+            {
+                auto* attackAidImage = LOADER.GetImageN("map_new", 20000);
+                if(GetWorld().GetGGS().getSelection(AddonId::MILITARY_AID) == 2)
+                    if(gwv.GetNumSoldiersForAttack(pt) > 0) // soldiers available for attack?
+                        attackAidImage->DrawFull(curPos - DrawPoint(0, attackAidImage->getHeight()));
```
Flamefire, line 359: Braces for multi-line code in branches:
```suggestion
                if(GetWorld().GetGGS().getSelection(AddonId::MILITARY_AID) == 2)
                {
                    if(gwv.GetNumSoldiersForAttack(pt) > 0) // soldiers available for attack?
                        attackAidImage->DrawFull(curPos - DrawPoint(0, attackAidImage->getHeight()));
                }
```

libs/s25main/world/GameWorldView.cpp
```diff
@@ -484,7 +490,7 @@ void GameWorldView::DrawConstructionAid(const MapPoint& pt, const DrawPoint& cur
         // Draw building quality icon
         bm->DrawFull(curPos);
         // Show ability to construct military buildings
-        if(GetWorld().GetGGS().isEnabled(AddonId::MILITARY_AID))
+        if(GetWorld().GetGGS().getSelection(AddonId::MILITARY_AID) > 0)
```
Flamefire, line 493: Why this change? `isEnabled` already does exactly this.

Flamefire (changes requested):
Thanks a lot! Only 2 minor issues

And please use the imperative as commit messages. E.g. "Fix loading of plugin graphics". If you are comfortable with git amend and force push you can change the existing commits (message and include style fixes in the commit that added/changed the code) otherwise just keep it in mind for the future.

libs/s25main/world/GameWorldView.cpp
tehKaiN: Because I've changed setting to list instead of bool, so that there are now 3 states of it: "off", "only construction aid", and "construction and attack aid". I assumed that `isEnabled` is only part of bool-based plugins, perhaps I was wrong about it. I'll check that out tomorrow

tehKaiN:
Thanks for feedback! I'll rewrite my git history tomorrow so that commit messages are more uniform with your guidelines.

libs/s25main/world/GameWorldView.cpp
Flamefire: `isEnabled` is a shortcut for `isNotDefault`, i.e. the default for an addon should be the original S2 behavior, i.e. "disabled"
Hence the name.

tehKaiN: fixed!

libs/s25main/world/GameWorldView.cpp
```diff
@@ -347,12 +347,18 @@ void GameWorldView::DrawNameProductivityOverlay(const TerrainRenderer& terrainRe
...
 
+            Position curPos = GetWorld().GetNodePos(pt) - offset + curOffset;
+            curPos.y -= 22;
+
             // Is object not belonging to local player?
             if(no->GetPlayer() != gwv.GetPlayerId())
+            {
+                auto* attackAidImage = LOADER.GetImageN("map_new", 20000);
+                if(GetWorld().GetGGS().getSelection(AddonId::MILITARY_AID) == 2)
+                    if(gwv.GetNumSoldiersForAttack(pt) > 0) // soldiers available for attack?
```
Flow86, line 358: still missing the braces here.

tehKaiN: oops! I've just merged those into single `if` instead.

libs/s25main/world/GameWorldView.cpp
```diff
@@ -347,12 +347,17 @@ void GameWorldView::DrawNameProductivityOverlay(const TerrainRenderer& terrainRe
             if(!no)
                 continue;
 
+            Position curPos = GetWorld().GetNodePos(pt) - offset + curOffset;
+            curPos.y -= 22;
+
             // Is object not belonging to local player?
             if(no->GetPlayer() != gwv.GetPlayerId())
+            {
+                auto* attackAidImage = LOADER.GetImageN("map_new", 20000);
```
Flamefire, line 356: Try to avoid doing needless work. E.g. getting the image when you rarely need it (the condition below is mostly false)
So put this into the block guarded by the `if`

Just for context what happens in `LOADER.GetImageN("map_new", 20000);`

- Look up the Singleton object
- Construct a string-like
- Search a map
- Access a large vector (cache miss likely)
- Cast the result

Flamefire (changes requested):
One last remark concerning performance, besides that it looks good! 👍

libs/s25main/world/GameWorldView.cpp
tehKaiN: Yeah, I thought about it, but then I saw that current code for military aid for construction is doing basically the same:

```cpp
void GameWorldView::DrawConstructionAid(const MapPoint& pt, const DrawPoint& curPos)
{
    BuildingQuality bq = gwv.GetBQ(pt);
    if(bq != BuildingQuality::Nothing)
    {
        constexpr auto bqImgs = getBqImgs();
        auto* bm = LOADER.GetMapTexture(bqImgs[bq]);
        // Draw building quality icon
        bm->DrawFull(curPos);
        // Show ability to construct military buildings
        if(GetWorld().GetGGS().isEnabled(AddonId::MILITARY_AID))
        {
            if(!GetWorld().IsMilitaryBuildingNearNode(pt, gwv.GetPlayerId())
               && (bq == BuildingQuality::Hut || bq == BuildingQuality::House || bq == BuildingQuality::Castle
                   || bq == BuildingQuality::Harbor))
                LOADER.GetImageN("map_new", 20000)->DrawFull(curPos - DrawPoint(-1, bm->GetSize().y + 5));
        }
    }
}
```

so I thought that we're good here and by the power of constexpr, compiler optimization and other black magic we're good with this approach. Ideally, this icon should be loaded once for whole GameWorldView, provided that plugin is enabled, and be used straight from dedicated object field.

tehKaiN:
I've moved LOADER thingy inside `if` as you requested. Perhaps I should do additional work by moving that call into constructor and storing result in the field as I mentioned earlier?

libs/s25main/world/GameWorldView.cpp
Flamefire: Sorry? In that code every image that is loaded is also used.

libs/s25main/world/GameWorldView.cpp
```diff
@@ -347,12 +347,18 @@ void GameWorldView::DrawNameProductivityOverlay(const TerrainRenderer& terrainRe
...
 
+            Position curPos = GetWorld().GetNodePos(pt) - offset + curOffset;
+            curPos.y -= 22;
+
             // Is object not belonging to local player?
             if(no->GetPlayer() != gwv.GetPlayerId())
+            {
+                if(GetWorld().GetGGS().getSelection(AddonId::MILITARY_AID) == 2 && gwv.GetNumSoldiersForAttack(pt) > 0)
+                {
+                    LOADER.GetImageN("map_new", 20000)->DrawFull(curPos - DrawPoint(0, attackAidImage->getHeight()));
```
Flamefire, line 358:
```suggestion
                    auto* attackAidImage =  LOADER.GetImageN("map_new", 20000);
                    attackAidImage->DrawFull(curPos - DrawPoint(0, attackAidImage->getHeight()));
```

Flamefire:
> Perhaps I should do additional work by moving that call into constructor and storing result in the field as I mentioned earlier?

Yes that might make sense. But then we might want to do that for every texture used which can be an own PR as it will likely be large

libs/s25main/world/GameWorldView.cpp
tehKaiN: yeah but it is loaded very often nonetheless - for each construction icon which qualify as spot for military building as I understand it.

Flamefire:
Seems there was an extra space in the suggestion. Sorry for that, could you fix that? Rest looks good

tehKaiN:
fixed!

tehKaiN:
so, can we merge this, or is there something else which I have to fix? :)

Flamefire (approved):

PR #1534, opened 2022-09-16
Portrait support

### LINKED ISSUE #798: Player avatars

### LINKED ISSUE #1181: [LUA] function setPortrait
In order to recreate the original mission we also need to set the correct portrait images for players.

tehKaiN:
fixes #798 
fixes #1181 

I still have unit tests and other stuff to fix, but I think it's ready for reviews.

There's small problem with column names on 800x600 resolution, but who plays using that res anymore?

obraz

This will be split into lesser PRs:

- [x] https://github.com/Return-To-The-Roots/s25client/pull/1535
- [x] https://github.com/Return-To-The-Roots/s25client/pull/1536
- [x] https://github.com/Return-To-The-Roots/s25client/pull/1773


tehKaiN:
the thing about lua test is funny:

```
/home/runner/work/s25client/s25client/tests/s25Main/integration/testWorld.cpp(261): fatal error: in "MapTestSuite/LoadLua": Unexpected log: Lua script did not provide the function getRequiredLuaVersion()! It is probably outdated.
```
on my local machine it prints space between function name and `()` - very strange

libs/s25main/ogl/glSmartBitmap.cpp
```diff
@@ -255,35 +262,39 @@ void glSmartBitmap::drawPercent(DrawPoint drawPt, unsigned percent, unsigned col
...
     colors[0].a = GetAlpha(color);
     colors[3] = colors[2] = colors[1] = colors[0];
 
+    // vertical coords
     curTexCoords[0] = texCoords[0];
     curTexCoords[1] = texCoords[1];
     curTexCoords[2] = texCoords[2];
     curTexCoords[3] = texCoords[3];
-    curTexCoords[0].y = curTexCoords[3].y = curTexCoords[1].y - (curTexCoords[1].y - curTexCoords[0].y) * partDrawn;
+    curTexCoords[0].y = curTexCoords[3].y = helpers::lerp<float>(
```
Flamefire, line 289: You don't need the `<float>` AND the cast to float. Either is enough. With the suggestion to only implement it for `float` both can go making this simpler.

Flamefire (changes requested):
Great work!

> on my local machine it prints space between function name and () - very strange

This is a translation bug. Fixed by https://github.com/Return-To-The-Roots/languages/commit/d46539024d5df6b37eac889af9848e407e7260ff

Besides that the PR is really big although the changes are related. This may make the review and fixing iterations a bit longer, but should work out. So don't mind the many comments.

libs/common/include/helpers/mathFuncs.h
```diff
@@ -66,4 +66,20 @@ constexpr T interpolate(const T startVal, const T endVal, const U elapsedTime, c
     else // Special case for unsigned values
         return static_cast<T>(startVal - ((startVal - endVal) * elapsedTime) / duration);
 }
+
+// Linear interpolation, identical to C++20's std::lerp()
```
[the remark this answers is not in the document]
tehKaiN: I tend to write intermediate values before returns so that debugger stepping is a bit easier. I'll ofc revert this.

doc/lua/functions.md
```diff
@@ -21,32 +21,32 @@ There are the following objects defined:
 
 Reference: [libs/libGamedata/lua/LuaInterfaceBase.cpp](../../libs/libGamedata/lua/LuaInterfaceBase.cpp)
 
-**rttr:GetVersion()**
```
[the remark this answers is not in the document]
tehKaiN: I've read about it and see that:

> By default, two spaces in markdown is the same as a `<br>`.

I never knew about that, thanks!

libs/s25main/gameData/PortraitConsts.h
```diff
@@ -0,0 +1,32 @@
...
+  PortraitInfo{gettext_noop("Erik"), ResourceId("io"), 261},
+  PortraitInfo{gettext_noop("Knut"), ResourceId("io"), 256},
+  PortraitInfo{gettext_noop("Olof"), ResourceId("io"), 255},
+  PortraitInfo{gettext_noop("Yamauchi"), ResourceId("io"), 254},
+  PortraitInfo{gettext_noop("Tsunami"), ResourceId("io"), 253},
+  PortraitInfo{gettext_noop("Hakirawashi"), ResourceId("io"), 260},
+  PortraitInfo{gettext_noop("Shaka"), ResourceId("io"), 258},
+  PortraitInfo{gettext_noop("Todo"), ResourceId("io"), 262},
+  PortraitInfo{gettext_noop("Mnga Tscha "), ResourceId("io"), 257},
+  PortraitInfo{gettext_noop("Baby"), ResourceId("io_new"), 007},
```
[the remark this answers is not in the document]
tehKaiN: :D

libs/s25main/network/GameMessages.h
```diff
@@ -336,6 +336,38 @@ class GameMessage_Player_Name : public GameMessageWithPlayer
...
+class GameMessage_Player_Portrait : public GameMessageWithPlayer
+{
+public:
+    unsigned int playerPortraitIndex;
+
+    GameMessage_Player_Portrait() : GameMessageWithPlayer(NMS_PLAYER_PORTRAIT) {}
+    GameMessage_Player_Portrait(uint8_t player, unsigned int portraitIndex)
+        : GameMessageWithPlayer(NMS_PLAYER_PORTRAIT, player), playerPortraitIndex(portraitIndex)
+    {
+        LOG.writeToFile(">>> NMS_PLAYER_PORTRAIT(%u)\n") % portraitIndex;
```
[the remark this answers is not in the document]
tehKaiN: do you want me to remove it in all messages then?

libs/s25main/BasePlayerInfo.cpp
```diff
@@ -36,6 +38,7 @@ void BasePlayerInfo::Serialize(Serializer& ser, bool lightData) const
     if(!lightData || ps == PlayerState::AI)
         aiInfo.serialize(ser);
     ser.PushLongString(name);
+    ser.PushUnsignedInt(portraitIndex);
```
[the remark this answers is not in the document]
tehKaiN: that's true!

tehKaiN:
I'm a bit worried about breaking changes in serialization protocol - I'm planning to tackle some more issues blocking campaign implementations and perhaps there will be more serialization-intrusive things along the way. Perhaps we should merge it into separate branch (something like `campaign`) and then merge in all the changes at once into master, breaking the stuff only once?

libs/s25main/network/GameMessages.h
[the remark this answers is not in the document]
Flamefire: If you want yes. Best do that in a separate PR though to keep the diffs small.
Generally: If you can factor out changes into self-contained parts better do more small PRs. Same for commits: Fixups (e.g. formatting) can go into the commit that introduced the issue as long as the PR is not yet merged (or not even reviewed). This makes it possible to revert a commit later on. So small, self-contained commits are good.

Flamefire:
> I'm a bit worried about breaking changes in serialization

Then maybe factor out the other changes into separate PRs. E.g. the use of enum-IDs for the controls, the image-handling etc. So all that is left is really only the portrait.

If you don't know yet: `git add -p` allows to chose which changes to stage at the hunk level. So you can then commit only parts of the changes in a file. I found that very useful when splitting up commits.

libs/s25main/network/GameMessages.h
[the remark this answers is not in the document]
tehKaiN: ok, we'll do this separately then.

tehKaiN:
The thing is, image handling was required to make portraits in lobby in reasonable shape. Splitting settings and lobby portrait into separate PRs seems like an overkill to me. I can make lua thingy separate, but dunno if it's really worth it.

What I meant regarding breaking serialization is that with portrait change I introduce serialization version 10. Perhaps with future PRs I'll introduce another breaking changes, bumping it e.g. to 11, 12 and 13. Perhaps it's wise to cram up as many serialization-breaking changes to a single version bump (e.g. by preparing them on separate feature branch), so that nightly players won't have their save files unusable too often.

Flamefire:
> The thing is, image handling was required to make portraits in lobby in reasonable shape. Splitting settings and lobby portrait into separate PRs seems like an overkill to me. I can make lua thingy and IDs separate, but dunno if it's really worth it.

https://github.com/Return-To-The-Roots/s25client/pull/1534/commits/ca7a173d8971dc781fcf19b1038c61556dfca7ad and https://github.com/Return-To-The-Roots/s25client/pull/1534/commits/1367fccc4e3960859627dfe80b5da1dddc8d67a9 potentially with tests can go into a separate PR, especially as there is a remark for the latter. IMO that is a standalone change worth of its own PR and maybe we can come up with a few (automatic or manual) tests to check that it works as expected.

The ID change could be extracted. Yes it might not be really worth it but the idea is to bring as much into master as possible. I'm even thinking about making a change to the Serializer so the portrait change can be backwards compatible.

tehKaiN:
Ok, I'll split my changes into separate branches and we'll merge them in more granular manner - probably not today though.

libs/s25main/desktops/dskGameLobby.cpp
```diff
@@ -48,6 +50,15 @@
 namespace {
 enum CtrlIds
 {
+    ID_PLAYER_ROW_NAME = 1,
```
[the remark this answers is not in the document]
tehKaiN: I'm not so sure about this - aren't all enums in project in uppercase with words separated by underscore?

libs/s25main/controls/ctrlImageButton.cpp
```diff
@@ -11,11 +11,18 @@ ctrlImageButton::ctrlImageButton(Window* parent, unsigned id, const DrawPoint& p
 
 void ctrlImageButton::DrawContent() const
 {
-    DrawPoint pos = GetDrawPos() + DrawPoint(GetSize()) / 2;
```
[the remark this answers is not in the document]
tehKaiN: answered in split PR

tehKaiN:
Somehow I've missed some of your comments - all are fixed now. I'll continue to split the PR into smaller ones and I guess we'll leave portrait part unmerged until we decide if save compatibility should be broken or add support so that it'll be prevented

libs/s25main/desktops/dskGameLobby.cpp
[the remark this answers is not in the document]
Flamefire: True, "normal" enums are as they are global constants. We shouldn't introduce plain enums, but use enum classes instead which avoids this.
Anyway in this case those are constants for ids and local to the file so an unnamed, plain enum with only `ID` in uppercase is enough.

tehKaiN:
Oops, looks like this thing got buried under the rubble of my other activities, sorry! 👀 

Is that thing still desired for the project? If yes, then I could get back to this #1536 soon.

Flamefire:
> Is that thing still desired for the project? If yes, then I could get back to this #1536 soon.

Sure. IIRC the main issue in that other PR was a clear definition of the behavior (also in corner cases). Best to be done in code (tests). But has been a while and would need to read up on it again when you work on it

tehKaiN:
I've tweaked the field sizes a bit so that it looks better on 800x600:

obraz

The only problem is that AI (medium) doesn't fit the display. I could probably make some room by moving the map preview to the right, but that would also mean moving bottom controls so that they end at the same X pos.

The problem would be non-existant if we'd somehow get rid of "Ready?" column on top - perhaps adding a tooltip to the checkbox class? Nevermind that each row displays the checkbox - I've hacked it in for sake of preview.

Apart from that, setting the preferred portrait in options desktop works as intended, as well as (de)serialize in game config. I still need to check if those portraits survive the game save/load correctly.

EDIT: the column names are all over the place with some languages and sadly there's not much to do about them. But still, I guess that proper display of English/German is enough on such low resolution.

obraz

libs/s25main/SerializedGameData.cpp
```diff
@@ -101,8 +101,9 @@
 /// 8: noFlag::Wares converted to static_vector
 /// 9: Drop serialization of node BQ
 /// 10: troop_limits state introduced to military buildings
-/// 11:: wineaddon added, three new building types and two new goods
-static const unsigned currentGameDataVersion = 11;
+/// 11: wineaddon added, three new building types and two new goods
+/// 12: Add portrait index serialization in player state
+static const unsigned currentGameDataVersion = 12;
```
Flamefire: You don't use this anywhere, do you? The player data is outside the game data and we don't have any versioning for that, do we? So I guess we can reset this and remove the checks if we go forward with this change

Flamefire (commented):
I think this is a breaking change in many parts and I'm not sure how we do and can handle that:
- Savegames and replays are all incompatible due to a new field in the player info. 
- Sending player info packets contain an additional field making them potentially invalid on the receiver side (portrait treated as nation etc)
- There is a new network message but it uses the ID of an existing message and others get changed making half the messages invalid

I think the last is the most serious one. I can't remember if we have anything in place to avoid clients connecting to others with a differing version. But maybe we do. This needs to be checked.

For the others maybe we can use `VersionedSerializer`. In e.g. the replay files we have a version that we could increment to later check if the potrait index is there. Can you see if this is possible for all other parts too? I'd like to avoid being unable to use older replays and save games. But maybe we need to which means we can at least clean some stuff up.   
If possible do this in a separate branch with a PR against this branch to track it more easily unless you can do it in a single commit.

I don't fully understand the change to `dskOptions`. What does this do?

libs/s25main/SerializedGameData.cpp
tehKaiN: Nah, I don't use it, it seems it was a blind change on my end years ago and I've just mindlessly upped the number in recent merge to make it relevant again.

I see that `Savegame::GetVersion()` returns a magic number - perhaps this should be upped and some kind of migration code added so that old stuff can be read properly?

tehKaiN:
> Savegames and replays are all incompatible due to a new field in the player info.

This indeed requires some kind of migration code.

> Sending player info packets contain an additional field making them potentially invalid on the receiver side 

Well, I guess the firest packet in communication could contain some kind of version field so that incompatible clients are rejected early on or they refuse to connect to too new servers? That would mean another breaking change if there's no field like that, but that would be future-proof.

The other way to handle it is to assume that the player has the default portrait and the game should query for the proper one with completely new and independent packet. Not sure how older clients will react to it though.

> There is a new network message but it uses the ID of an existing message

Sounds like sloppiness ~~/auto-merge break~~ and shouldn't be the case!

> I don't fully understand the change to dskOptions. What does this do?

Similar to gif in the issue page (with correct label right now, of course):

189997894-4509549d-fcc3-45d1-9afc-99013b37d575

Flamefire:
> > Savegames and replays are all incompatible due to a new field in the player info.
> 
> This indeed requires some kind of migration code.

Would you be up to adding such? I guess you'd need to start at `GamePlayerInfo` and go up from there.

> Well, I guess the firest packet in communication could contain some kind of version field so that incompatible clients are rejected early on or they refuse to connect to too new servers? That would mean another breaking change if there's no field like that, but that would be future-proof.

I found we already require clients to have the same revision:
https://github.com/Return-To-The-Roots/s25client/blob/ec289ff848d06bf8c7687ff471bd760b93971ca0/libs/s25main/network/GameServer.cpp#L926-L927

So that should be enough.

> > There is a new network message but it uses the ID of an existing message
> 
> Sounds like sloppiness ~/auto-merge break~ and shouldn't be the case!

Given it is not possible that a client communicates with an incompatible client I think we can keep the messages ordered by name as you had before.

> > I don't fully understand the change to dskOptions. What does this do?
> 
> Similar to gif in the issue page (with correct label right now, of course):

I see. Where is this used? I though portraits are only for ingame use and hence the selection only makes sense in the lobby screen.

tehKaiN:
> Would you be up to adding such? I guess you'd need to start at GamePlayerInfo and go up from there.

Of course!

> Given it is not possible that a client communicates with an incompatible client I think we can keep the messages ordered by name as you had before.

Alright, I'll revert my change on my nearest commit spree.

> I see. Where is this used? I though portraits are only for ingame use and hence the selection only makes sense in the lobby screen.

The portrait selection in options screen sets up the default portrait, so that if you create a game and want to play as Knut, you don't have to cycle through portraits each time.

libs/s25main/ingameWindows/iwStatistics.cpp
```diff
@@ -62,34 +63,10 @@ iwStatistics::iwStatistics(const GameWorldViewer& gwv)
...
-                               LOADER.GetImageN("io", 256), curPlayer.name)
-                  ->SetBorder(false);
-                break;
-            case Nation::Babylonians:
-                AddImageButton(1 + i, DrawPoint(startX + pos * 34 - 17, 45 - 23), Extent(34, 47), TextureColor::Green1,
-                               LOADER.GetImageN("io_new", 7), curPlayer.name)
-                  ->SetBorder(false);
-                break;
-        }
+        const auto& portrait = Portraits[curPlayer.portraitIndex];
```
Flow86, line 67: this index *must* be checked if its in range. You can set an invalid index via LUA. We're not on C++26 where the behavior *may* be defined

tehKaiN: I agree on range check, but shouldn't it be placed in lua fn / config loading / save loading / etc? Basically everywhere where the value is being set.

libs/s25main/BasePlayerInfo.cpp
```diff
@@ -8,20 +8,22 @@
...
     if(lightData && !isUsed())
     {
+        portraitIndex = 0;
         nation = Nation::Romans;
         team = Team::None;
         color = PLAYER_COLORS[0];
     } else
     {
         name = ser.PopLongString();
+        portraitIndex = ser.PopUnsignedInt();
```
tehKaiN: It seems to me that this one particular line could be encompassed within an `if` which would check for the serialized data version. Is there any way to do such thing? If not, would it be reasonable to add `hasPortraits` param to the constructor and pass the correct value via checking with GetVersion()?

The only trouble that I see is that saves/replays have different version numbering.

libs/s25main/ingameWindows/iwStatistics.cpp
Flow86: yes. everywhere where it can be set is best practice for this kind of usage (game engine)

libs/s25main/BasePlayerInfo.cpp
Flamefire: You can use `VersionedSerializer`, see the current usage for GameData. You could make a new class for the playerinfo and depending on a version for save or replay pass the relevant version to it. I.e. it would be a "playerinfo" version (or higher up the classes if required) and the replay version just determines if this is e.g. 0 or 1, so you have an indirection there.

libs/s25main/ingameWindows/iwStatistics.cpp
Flamefire: Yes, the check shouldn't be here IMO but where the value is set. E.g. during loading/deserialization or on receiving the packet.
You could add an assert here though.

libs/s25main/BasePlayerInfo.cpp
tehKaiN: Thanks for the tips! I've got most of the code complete, but I'm not certain how to handle different file versions.

There's this logic around message `File has an old version and cannot be used` in `SavedFile::ReadFileHeader` which breaks replay/save file loading when the version changes. There's also no way of getting the read version from the SavedFile, so I can't set up the proper version for `VersionedSerializer`.

I guess it needs some more work:
1. Add `SaveFile::GetMinimumSupportedVersion()` and make `SavedFile::ReadFileHeader` use it in comparison for old version error
1. Rename `SaveFile::GetVersion()` to `SaveFile::GetCurrentVersion()` so that the name reflects that it returns the latest version and keep it in line with method expected by `VersionedSerializer()` for extra consistency
1. Add `SaveFile::GetVersion()` which would return the version read from the file,
1. Use the `SaveFile::GetVersion()` to set up the `VersionedDeserializer<BasePlayerInfo>` (I used BasePlayerInfo instead of PlayerInfo as it makes more sense, I think)

Thoughts?

Flamefire: What we need: Have playerdata version to `SavedFile::ReadPlayerData`

Best case would be to have a format version in the file header which we can use. Currently we have one: https://github.com/Return-To-The-Roots/s25client/blob/ec289ff848d06bf8c7687ff471bd760b93971ca0/libs/s25main/SavedFile.cpp#L30-L31

But that is only used to check for an exact match. What we need would be something like a major-minor version.

Even worse: We have that version intended for the derived classes, i.e. Savegame & Replay. So not sure if we can use that.

However in the extended header we could sneak in a version: The [playerNames size](https://github.com/Return-To-The-Roots/s25client/blob/master/libs/s25main/SavedFile.cpp#L95) doesn't need to be 4 bytes. 1 (or 2) is enough.

So we can use a similar trick as used in the [replay](https://github.com/Return-To-The-Roots/s25client/blob/ec289ff848d06bf8c7687ff471bd760b93971ca0/libs/s25main/Replay.cpp#L194-L197) to only use the first 2 bytes and store the version in the following 2 bytes.

So suggestion: Introduce a "minor version" member to `SavedFile`, possibly using an optional so we don't read garbage.   
Set that in `ReadExtHeader` by stealing 2 bytes from the playernames field.   
Use that to set the version for the deserializer in `ReadPlayerData`   
Add a TODO note to add a major & minor version for the derived class and one for the parent class in `ReadFileHeader` instead of that hack.

Or maybe slightly better: `GetVersion` currently only returns a small value which fits in a single byte.   
So we can use the 2nd byte for the version of `SavedFile`. For the current (non-portrait) that is `0` and with the portrait it is `1`. This way it will be incompatible for loading with old versions but we can detect if it was saved with portraits and later extend this to a more flexible versioning scheme when we do need to break it.   
Don't forget to put an assert that `GetVersion() <= 255`, just in case and for clarity

We can even use a major-minor version:
- Read 1 byte `read_version` as before which we cannot touch yet.
- Read 1 byte `header version` which for the current format is `0`, for the new format is `1`
- Treat the 4 bits each as major minor version, i.e. the new format is `0.1`. That leaves room for 14 more backwards-compatible changes. At some point we can use a full byte each or more.

As for the interface similar to what you have:
- `GetCurrentVersion()` pure virtual in place of the current `GetVersion`
- `GetVersion` to return the sub-class version read from file
- `GetHeaderVersion` to return the other part, i.e. the new one for use in e.g. `ReadPlayerData`. If you can come up with a better name, feel free ;-) In the future we can extent that as required. The only requirement is then that the major version from the file is equal to the current major and the minor is less-or-equal than the current minor.
- `GetCurrentHeaderVersion`

tehKaiN: Done! Tested as follows:

- created a singleplayer game with AI enemies with different portraits
- played a bit, saved the game
- quit the game, loaded game, observed that portraits are intact
- quit the game, loaded some custom single player save file, observed that all players are set to Octavianus

libs/s25main/SerializedGameData.cpp
Flamefire: This still needs to be reverted

libs/s25main/BasePlayerInfo.h
```diff
@@ -18,14 +18,15 @@ struct BasePlayerInfo
...
+    unsigned portraitIndex;
     Nation nation;
     /// Actual color (ARGB)
     unsigned color;
     Team team;
 
     BasePlayerInfo();
     /// Deserialize data. If lightData is true, unused data is not read (e.g. unused slot -> Skip rest)
-    BasePlayerInfo(Serializer& ser, bool lightData);
+    BasePlayerInfo(Serializer& ser, int version, bool lightData);
```
Flamefire: I was about to say accept `VersionedSerializer` directly here but seen you changed that for Clang.

I think you can still use `VersionedSerializer`:

PlayerInfo could already accept that, maybe make it a subclass:

`class BasePlayerInfo {
class Deserializer: VersionedSerializer<Deserializer> {static unsigned getCurrentVersion();} // Implement in BasePlayerInfo.cpp with comments
`

That's what I meant with "propagates upwards". E.g. `JoinPlayerInfo` then needs to accept a `PlayerInfo::Deserializer`. That would also make it explicit at the invocation site of its ctor: Only used for the GameMessage where the most recent version is accepted. Setting the version in the `JoinPlayerInfo` ctor looks unmotivated (although correct).

Could you investigate how much work that is? Or do you see any potential issues?

I guess an easy alternative would be to only accept rvalues here: `BasePlayerInfo(s25util::VersionedDeserializer<BasePlayerInfo>&& ser`.
My hunch is that `version` argument isn't descriptive enough

Flamefire (changes requested):

libs/s25main/BasePlayerInfo.h
tehKaiN: I don't like this part too and I'm not sure what to do with it.

As you pointed out, the problem with using `VersionedDeserializer` directly is that `PlayerInfo::PlayerInfo(Serializer& ser, int basePlayerInfoVersion)` needs to pass the `VersionedDeserializer &` in its initializer list for `BasePlayerInfo` ctor and I can't create that in-line there, be it `VersionedDeserializer` or `BasePlayerInfo::Deserializer` as you propose, because the lvalue is expected. I don't really see how declaring `BasePlayerInfo::Deserializer` helps here.

I also looked at creating `VersionedDeserializer` in `PlayerInfo`, but that moves the problem one class higher because there's the same problem with passing `VersionedDeserializer &` / `PlayerInfo::Deserializer &` in initializer list in `JoinPlayerInfo`.

I wasn't sure about passing `&&ser` but ultimately didn't liked it because as I understand it in non-rvalue scenarios it will pass by copy and that seems undesired for serializers? I'm not that good with rvalues and elaborate templates, so perhaps my worries are insubstantial.

libs/s25main/BasePlayerInfo.h
```diff
@@ -37,4 +38,8 @@ struct BasePlayerInfo
     /// Returns index of color in PLAYER_COLORS array or -1 if not found
     int GetColorIdx() const;
     static int GetColorIdx(unsigned color);
+
+    // 0: Initial
+    // 1: Added portraitIndex
+    static inline constexpr int getCurrentVersion() { return 1; }
```
Flamefire: Can you move this to the cpp file? No need to have this in the header causing recompiles on change

libs/s25main/BasePlayerInfo.h
Flamefire: The only way I see is using an overload to "convert" the rvalue to an lvalue reference: https://godbolt.org/z/Wes8aqYKz

I.e. `BasePlayerInfo(Deserializer&& ser, bool lightData): BasePlayerInfo(ser, lightData) {}`
Note that `X&&` does never make a copy unless `X` is a template parameter

`BasePlayerInfo::Deserializer` is just a "nicer way" for `libutil::VersionedSerializer<BasePlayerInfo>` and putting the version to the serializer instead of the BasePlayerInfo.

Not much better. With the renaming of `version` I'm fine with the current one

libs/s25main/BasePlayerInfo.cpp
```diff
@@ -4,24 +4,28 @@
...
 BasePlayerInfo::BasePlayerInfo()
-    : ps(PlayerState::Free), nation(Nation::Romans), color(PLAYER_COLORS[0]), team(Team::None)
+    : ps(PlayerState::Free), portraitIndex(0), nation(Nation::Romans), color(PLAYER_COLORS[0]), team(Team::None)
 {}
 
-BasePlayerInfo::BasePlayerInfo(Serializer& ser, bool lightData)
+BasePlayerInfo::BasePlayerInfo(Serializer& ser, int serializedVersion, bool lightData)
     : ps(helpers::popEnum<PlayerState>(ser)), aiInfo(!lightData || ps == PlayerState::AI ? ser : AI::Info())
 {
+    auto versionedDeserializer = s25util::VersionedDeserializer<BasePlayerInfo>(ser, serializedVersion);
```
Flamefire: Unused?

tehKaiN: Indeed! Too many refactorings ;)

libs/s25main/BasePlayerInfo.h
tehKaiN: Done!

libs/s25main/SavedFile.cpp
```diff
@@ -67,14 +78,15 @@ bool SavedFile::ReadFileHeader(BinaryFile& file)
         }
 
         // Version überprüfen
-        uint16_t read_version = file.ReadUnsignedShort();
-        if(read_version != GetVersion())
+        majorVersion_ = file.ReadUnsignedChar();
+        minorVersion_ = file.ReadUnsignedChar();
```
Flamefire, line 82: I think we need a check for minor version too: We can only read older and current ones, but not newer ones. This is important for future updates.

libs/s25main/SavedFile.h
```diff
@@ -22,8 +22,14 @@ class SavedFile
...
     virtual std::string GetSignature() const = 0;
-    /// Return the file format version
-    virtual uint16_t GetVersion() const = 0;
+    /// Return the maximum supported file format version
+    virtual uint8_t GetLatestMinorVersion() const = 0;
+    virtual uint8_t GetLatestMajorVersion() const = 0;
+
+    /// Return the file format version returned from file - minor part
+    uint8_t GetMinorVersion() const;
+    /// Return the file format version returned from file - major part
```
Flamefire:
```suggestion
    /// Return the file format version read from file - minor part
    uint8_t GetMinorVersion() const;
    /// Return the file format version read from file - major part
```

libs/s25main/SavedFile.h
```diff
@@ -74,4 +80,6 @@ class SavedFile
     /// Mapname
     std::string mapName_;
     std::vector<std::string> playerNames_;
+    uint8_t minorVersion_;
+    uint8_t majorVersion_;
```
Flamefire: Can we make this `std::optional<uint8_t>` so we detect when reading them before loading them from file? No need to check, just let the exception be thrown on invalid access.

```suggestion
    std::optional<uint8_t> minorVersion_;
    std::optional<uint8_t> majorVersion_;
```

libs/s25main/desktops/dskGameLobby.cpp
```diff
@@ -936,6 +961,14 @@ void dskGameLobby::ChangeNation(const unsigned player, const Nation nation)
     GetCtrl<ctrlGroup>(ID_grpPlayerStart + player)->GetCtrl<ctrlBaseText>(ID_btNation)->SetText(_(NationNames[nation]));
 }
 
+void dskGameLobby::ChangePortrait(const unsigned player, const unsigned portraitIndex)
+{
+    const auto& portrait = Portraits[portraitIndex];
```
Flamefire, line 967: Probably useful here:
```suggestion
    RTTR_Assert(portraitIndex < Portraits.size());
    const auto& portrait = Portraits[portraitIndex];
```

libs/s25main/gameData/PortraitConsts.h
```diff
@@ -0,0 +1,32 @@
+// Copyright (C) 2005 - 2022 Settlers Freaks (sf-team at siedler25.org)
```
Flamefire:
```suggestion
// Copyright (C) 2005 - 2025 Settlers Freaks (sf-team at siedler25.org)
```

libs/s25main/ingameWindows/iwStatistics.cpp
Flamefire:
```suggestion
        RTTR_Assert(curPlayer.portraitIndex < Portraits.size());
        const auto& portrait = Portraits[curPlayer.portraitIndex];
```
At least to show intent and avoid future questions

tehKaiN: The assert's already there 👀

libs/s25main/gameData/PortraitConsts.h
tehKaiN: Never liked those date comments - they outdate themselves regurarily

libs/s25main/ingameWindows/iwStatistics.cpp
Flamefire: I can't seem to find it. Can you link to it?

libs/s25main/BasePlayerInfo.cpp
```diff
@@ -4,24 +4,25 @@
...
     if(lightData && !isUsed())
     {
+        portraitIndex = 0;
         nation = Nation::Romans;
         team = Team::None;
         color = PLAYER_COLORS[0];
     } else
     {
         name = ser.PopLongString();
+        portraitIndex = (serializedVersion >= 1) ? ser.PopUnsignedInt() : 0;
```
Flamefire, line 26: Please add verification here (see discussion: Every point where it is set should be verified.). Could be just a `std::min`

libs/s25main/ingameWindows/iwStatistics.cpp
Flamefire: Wrong diff version, sorry

Flamefire (changes requested):
One final place in need for range check I missed before

libs/s25main/BasePlayerInfo.cpp
tehKaiN: Done! the BasePlayerInfo.cpp was complaining about unknown std::array, so I've refactored the array definition to make_array instead of just adding the header in PortraitConsts.

libs/s25main/gameData/PortraitConsts.h
```diff
@@ -15,18 +16,15 @@ struct PortraitInfo
     unsigned int resourceIndex;
 };
 
-constexpr std::array<PortraitInfo, 13> SUPPRESS_UNUSED Portraits = {
+constexpr auto SUPPRESS_UNUSED Portraits = helpers::make_array<PortraitInfo>(
```
Flamefire, line 19: Just FYI: Without the template param it will use the common type of the arguments which is already `PotraitInfo`. Doesn't hurt though due to use of `auto` on the lhs

Flamefire (approved):
Took a lot of work, so thanks!

PR #1535, opened 2022-09-18
Add lerp and inverseLerp helper functions

tehKaiN:
split from #1534


tehKaiN:
I guess I should add some unit tests for them - will do that asap

PR #1536, opened 2022-09-20
Add support for image cropping inside ctrlImageButton

tehKaiN:


libs/s25main/controls/ctrlBaseImage.h
```diff
@@ -24,8 +24,8 @@ class ctrlBaseImage
     /// Swap the images of those controls
     void SwapImage(ctrlBaseImage& other);
     Rect GetImageRect() const;
-    void DrawImage(const DrawPoint& pos) const;
-    void DrawImage(const DrawPoint& pos, unsigned color) const;
+    void DrawImage(const Rect& dstArea) const;
```
Flamefire, line 27: Can you add a docstring here what the method is supposed to do when given an area expecially when the area is different than the image size.

libs/s25main/controls/ctrlImageButton.cpp
```diff
@@ -11,11 +11,18 @@ ctrlImageButton::ctrlImageButton(Window* parent, unsigned id, const DrawPoint& p
 
 void ctrlImageButton::DrawContent() const
 {
-    DrawPoint pos = GetDrawPos() + DrawPoint(GetSize()) / 2;
+    DrawPoint pos = GetDrawPos() - GetImageRect().getOrigin() + DrawPoint::all(2);
```
Flamefire, line 14: Please add a comment here how/where the image is supposed to be drawn, i.e. why the offset of 2 and 4 at various places. Maybe introduce some named constants for those values.

libs/s25main/ogl/ITexture.h
```diff
@@ -15,4 +16,5 @@ class ITexture
     virtual Position GetOrigin() const = 0;
     virtual Extent GetSize() const = 0;
     virtual void DrawFull(const Position& dstPos, unsigned color = 0xFFFFFFFFu) = 0;
+    virtual void Draw(Rect dstArea, Rect srcArea, unsigned color = 0xFFFFFFFFu) = 0;
```
Flamefire, line 19: Please add a docstring here and what the method is supposed to do. `DrawFull(dstPos, color)` fully explains itself, but here it is unclear how the 2 areas are handled if they are not the same as the image size.

libs/s25main/ogl/glSmartBitmap.cpp
```diff
@@ -314,3 +325,18 @@ void glSmartBitmap::drawPercent(DrawPoint drawPt, unsigned percent, unsigned col
     glDrawArrays(GL_QUADS, 0, numQuads);
     glDisableClientState(GL_COLOR_ARRAY);
 }
+
+void glSmartBitmap::drawPercent(DrawPoint drawPt, unsigned percent, unsigned color, unsigned player_color)
```
Flamefire, line 329: Can you put the `drawPercent` method back above the `draw` method and amend the commit? I know this now matches the order in the header but it makes the diff hard to judge as it looks like you added this method.

Flamefire (changes requested):
Some minor annotations. Also can you please add tests at least for the controls?

You can easily create a Mock for `ITexture` (if it doesn't exist yet) and use that to check that the correct Draw-calls are done. See similar tests using the `MOCK_*` macros.   
Those tests should be simple enough (e.g. use "nice" sizes) that they serve also as documentation on how the feature is supposed to work.

Unfortunately I don't see an easy way to test the `gl*` classes, but the changes look correct.

tehKaiN:
all things fixed. I'll add unit tests across following days.

libs/s25main/ogl/glSmartBitmap.cpp
```diff
@@ -255,35 +277,39 @@ void glSmartBitmap::drawPercent(DrawPoint drawPt, unsigned percent, unsigned col
...
     colors[3] = colors[2] = colors[1] = colors[0];
 
+    // vertical coords
     curTexCoords[0] = texCoords[0];
     curTexCoords[1] = texCoords[1];
     curTexCoords[2] = texCoords[2];
     curTexCoords[3] = texCoords[3];
-    curTexCoords[0].y = curTexCoords[3].y = curTexCoords[1].y - (curTexCoords[1].y - curTexCoords[0].y) * partDrawn;
+    curTexCoords[0].y = curTexCoords[3].y = helpers::lerp(
+      texCoords[0].y, texCoords[1].y, helpers::inverseLerp(.0f, float(size_.y), float(srcArea.getOrigin().y)));
```
Flamefire, line 305: No need for the explicit float casts (anymore). The implicit conversion is safe and the explicit casts make it harder to read.

Flamefire:
Thanks, looks good! If you need help with the tests feel free to ask.

tehKaiN:
I've added some tests as well as corrected inverseLerp to not be templated. Let me know it this suffices, or if you want some more elaborate testing.

libs/s25main/ogl/glSmartBitmap.cpp
Flamefire: Still the float casts? Any reason?

libs/s25main/ogl/glSmartBitmap.cpp
```diff
@@ -240,12 +242,32 @@ void glSmartBitmap::generateTexture()
...
 {
     drawPercent(drawPt, 100, color, player_color);
 }
 
 void glSmartBitmap::drawPercent(DrawPoint drawPt, unsigned percent, unsigned color, unsigned player_color)
+{
+    // nothing to draw?
+    if(!percent)
+        return;
+    RTTR_Assert(percent <= 100);
```
Flamefire, line 260: I think this should be tested too. Especially coming up with a few corner cases. I.e. at least percent = 0 or 100 and some values where you notice the (correct) usage of floor/ceil below. I.e. choose appropriate image size and percent values.

tests/s25Main/UI/testImageButton.cpp
```diff
@@ -0,0 +1,64 @@
...
+// LCOV_EXCL_STOP
+
+BOOST_FIXTURE_TEST_SUITE(ImageButton, uiHelper::Fixture)
+
+BOOST_AUTO_TEST_CASE(DrawSmallerImageThanButton)
+{
+    TestTexture img(Extent(20, 10));
+    MOCK_EXPECT(img.Draw).once().with(Rect(20, 15, 20, 10), Rect(0, 0, 20, 10), 0xFFFFFFFFu);
+
+    Window wnd(nullptr, randomValue<unsigned>(), DrawPoint::all(0));
```
Flamefire, line 46: I'd suggest to move the Window constructor to the top, so the values for the button, img and mock-expectation are closer together. I'd also suggest to place the `MOCK_EXPECT` right before the `draw()` just to make it more clear.

tests/s25Main/UI/testImageButton.cpp
```diff
@@ -0,0 +1,64 @@
...
+    ctrlImageButton btn(&wnd, randomValue<unsigned>(), DrawPoint(10, 5), Extent(40, 30), TextureColor::Green1, &img,
+                        "");
+    btn.Draw();
+}
+
+BOOST_AUTO_TEST_CASE(DrawLargerImageThanButton)
+{
+    // Note the 2px border to prevent dirtying 3d button's ridge
```
Flamefire, line 54: That comment should be 1 line below so it is above to what it refers to not above another line
And maybe explain what is meant by "2px" border. I.e. that the image is shifted, drawn with 4px reduced size and the center part of the image is chosen, i.e. `(Extend(100,80) +- Extend(36,26))/ 2`  (the 2nd Rect)

See those behavior test case strategies used by other projects which here would go like `When(DrawLargerImageThanButton).Then(UseCenterPieceOfImage AND Offset by border from all 4 sides)`

tests/s25Main/UI/testImageButton.cpp
```diff
@@ -0,0 +1,64 @@
...
+#include <boost/test/unit_test.hpp>
+
+using namespace rttr::test;
+
+namespace {
+MOCK_BASE_CLASS(TestTexture, ITexture)
+{
+public:
+    TestTexture(Extent size) : size_(size) {}
+    Position GetOrigin() const override { return Position::all(0); }
```
Flamefire, line 19: As you use the origin and size in the code, the test should likely also set the origin and maybe have multiple images (or you reset the size and/or origin in the test and show how the draw call changes.

tests/s25Main/UI/testImageButton.cpp
```diff
@@ -0,0 +1,64 @@
...
+    ctrlImageButton btn(&wnd, randomValue<unsigned>(), DrawPoint(10, 5), Extent(40, 30), TextureColor::Green1, &img,
+                        "");
+    btn.Draw();
+}
+
+BOOST_AUTO_TEST_CASE(DrawLargerImageThanButton)
+{
+    // Note the 2px border to prevent dirtying 3d button's ridge
+    TestTexture img(Extent(100, 80));
+    MOCK_EXPECT(img.Draw).once().with(Rect(12, 22, 36, 26), Rect(32, 27, 36, 26), 0xFFFFFFFFu);
```
Flamefire, line 56: Same as above: Window to the top, MOCK_EXPECT above the Draw.

Flamefire (approved):
A bit on improvements of the tests while we are at it. Especially we could use a test for the drawPercent which has been missing for to long.

You could create mock functions for the OpenGL functions like `glVertexPointer` and either redirect them to some `MOCK_FUNCTION(...)` mocks or to some handwritten ones storing the last passed values. See https://github.com/Return-To-The-Roots/s25client/blob/40d94b40b33e3a491214627e27dc46a289145d7f/tests/s25Main/UI/testVideoDriver.cpp#L73 on how to temporarily replace the OpenGL functions.

tehKaiN:
oops, I've got some comments to the code in "pending" status and they were not there for you to see.

I'll try to resolve stuff you've mentioned by tomorrow.

tests/s25Main/UI/testImageButton.cpp
```diff
@@ -0,0 +1,68 @@
...
+    }
+    MOCK_NON_CONST_METHOD(DrawFull, 2)
+    MOCK_NON_CONST_METHOD(Draw, 3)
+
+    Extent size_;
+};
+} // namespace
+
+namespace mock {
+static mock::stream& operator<<(mock::stream& s, const Rect& r)
```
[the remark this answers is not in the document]
tehKaiN: oops, I haven't found it because I was looking specifically for mock::stream!

tests/s25Main/UI/testImageButton.cpp
tehKaiN: I'm not sure if I get what you mean, especially in 2nd part, but I've reworded the comment to be more descriptive

tests/s25Main/UI/testImageButton.cpp
```diff
@@ -0,0 +1,68 @@
...
+using namespace rttr::test;
+
+namespace {
+MOCK_BASE_CLASS(TestTexture, ITexture)
+{
+public:
+    TestTexture(Extent size) : size_(size) {}
+    Position GetOrigin() const override
+    {
+        return Position::all(0);
```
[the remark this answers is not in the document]
tehKaiN: will do. I've started doing it and found some problems. I don't quite get the idea of image origin - I assume this should generally work so that if I set origin to (5,8) and draw image somewhere in the game at (30,40), then at this position the pixel from the (5,8) is drawn and the image's top left corner starts at (25,32).

This is problematic when drawing image buttons - my assumption is that they should always be centered in the buttons, regardless of the origin values. Should this be the case or am I getting something wrong?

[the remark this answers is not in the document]
Flamefire: Yes IIRC the origin is mostly used for the textures of buildings etc where the origin may vary depending on the nation etc. I don't think it is useful for button textures, which I mentioned in https://github.com/Return-To-The-Roots/s25client/pull/1534/files#r973562992
So yes I do share your opinion that the image should always be centred and the origin ignored for buttons.

Maybe @Spikeone has something to add.

libs/s25main/ogl/glSmartBitmap.cpp
Flamefire: Alright in that case we can use templated lerp/inverselerp and static_assert that the type is a float value (as using an integral will make it fail due to truncation in the division)

tehKaiN: though the templated version doesn't change anything regarding casts - explicit `float(int)` still needs to be used or else warning will occur.

Flamefire: Yes but if we need the casts anyway we can make it templated now to not have to change it in the future.

tests/s25Main/UI/testImageButton.cpp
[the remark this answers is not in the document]
tehKaiN: I've had a slight brain meltdown because unit tests returned for me different position values passed to `img.Draw()` when specifying different origins, but then I remembered that the origin thingy is resolved inside that function ( https://github.com/Return-To-The-Roots/s25client/blob/master/libs/s25main/ogl/glArchivItem_Bitmap.cpp#L45 ) so I think it has to be that way. :)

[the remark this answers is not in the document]
Flamefire: Ok, that means that the button control MUST not use the origin when calling `Draw` as that is handled by `Draw`.
Or actually if `Draw` handles `origin` then in order to have the image always centered the button control must **add** the origin to account for it being removed in the draw call. So the current change at https://github.com/Return-To-The-Roots/s25client/pull/1536/files#diff-adfa4d0f1a6924716e377c3f74e76a930efc15607563954cee41631febf87c4fR14 would make the image being offset by 2 times the origin.

I guess current button textures don't have an origin set so this wasn't an issue before.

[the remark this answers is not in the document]
tehKaiN: I you look closely then GetImageRect() creates rect with `position = -origin` so the calculations are correct there, because my subtraction effectively adds ;) I guess that's confusing though, so I'll look into getting origin directly and add it there instead.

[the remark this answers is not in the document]
Flamefire: That would make a great end-to-end test: Specify a non-zero origin for an image used in a button and then check that the actual draw call (i.e. the OpenGL coords) will draw the image at the exact coordinates. I.e. that you modification and the origin handling in the draw call actually do cancel each other out as you describe here (I checked: you are right)
And maybe add a short comment to the line removing the origin that this "compensates" (or any better word) the offset so the image is drawn exactly where specified.

libs/s25main/ogl/glSmartBitmap.cpp
tehKaiN: I've added some unit test here, but I'm not quite sure on how I should approach the float/ceil thingy - I can't really subclass `glSmartBitmap` and mock `drawRect` unless I make it virtual, which I think shouldn't be done just for unit tests. The texture coords don't follow image size in pixels, for example I tried to create 50x100 bitmap and coord span for the texture was something like 0 .. 0.7825. Checking for magic float values seems not to be the best approach either.

tests/s25Main/UI/testImageButton.cpp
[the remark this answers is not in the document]
tehKaiN: I've updated the comment, doing the end-to-end test is yet to follow.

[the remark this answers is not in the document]
Spikeone: > Yes IIRC the origin is mostly used for the textures of buildings etc where the origin may vary depending on the nation etc. I don't think it is useful for button textures, which I mentioned in https://github.com/Return-To-The-Roots/s25client/pull/1534/files#r973562992 So yes I do share your opinion that the image should always be centred and the origin ignored for buttons.
> 
> Maybe @Spikeone has something to add.

Only if we plan to add buttons with text and image - but currently I don't think we'd need that in any of the current interfaces.

tehKaiN:
This was way too overdue.

I've updated the `drawPercent()` which was a bit tricky, because it failed for some bitmap heights - their height was less than 100 so checking if a given percentage was drawn didn't do the trick.

Also, I had problems with `BOOST_TEST(a, b, tolerance)` as well as `BOOST_CHECK_CLOSE()` as it appeared that they do strong comparison (`<`) instead of accepting equity - and that caused problems due to float granularity.

libs/s25main/ogl/glSmartBitmap.cpp
```diff
@@ -253,35 +275,39 @@ void glSmartBitmap::drawPercent(DrawPoint drawPt, unsigned percent, unsigned col
...
-    curTexCoords[0].y = curTexCoords[3].y = curTexCoords[1].y - (curTexCoords[1].y - curTexCoords[0].y) * partDrawn;
+    curTexCoords[0].y = curTexCoords[3].y = helpers::lerp(
+      texCoords[0].y, texCoords[1].y, helpers::inverseLerp(.0f, float(size_.y), float(srcArea.getOrigin().y)));
+    curTexCoords[1].y = curTexCoords[2].y = helpers::lerp(
+      texCoords[0].y, texCoords[1].y, helpers::inverseLerp(.0f, float(size_.y), float(srcArea.getEndPt().y)));
+    // horizontal coords
+    curTexCoords[0].x = curTexCoords[1].x = helpers::lerp(
+      texCoords[0].x, texCoords[3].x, helpers::inverseLerp(.0f, float(size_.x), float(srcArea.getOrigin().x)));
+    curTexCoords[2].x = curTexCoords[3].x = helpers::lerp(
+      texCoords[0].x, texCoords[3].x, helpers::inverseLerp(.0f, float(size_.x), float(srcArea.getEndPt().x)));
```
Flamefire, line 310: I have trouble understanding this lerp of inverseLerp. Can you introduce variables for the inverseLerp please? Giving those values names that might be enough to understand what they are. If not a comment what the inverseLerp caluclates will do (in addition to the variables)
Especially as the difference in the inverseLerps is easy to miss.

tests/s25Main/UI/testSmartBitmap.cpp
```diff
@@ -26,6 +30,77 @@ static std::ostream& boost_test_print_type(std::ostream& os, const ColorBGRA col
...
+std::vector<Point<float>> vertexCoords;
+int callCount;
+
+void APIENTRY glVertexPointer(GLint size, GLenum type, GLsizei stride, const void* pointer)
+{
+    ++callCount;
+    RTTR_Assert(type == GL_FLOAT);
+    RTTR_Assert(stride == 0);
+    RTTR_Assert(size == 2);
+    vertexPointer = static_cast<const Point<float>*>(pointer);
```
Flamefire, line 48: This is a pointer to a stack variable, isn't it? So you'd need to copy the values.

tests/s25Main/UI/testSmartBitmap.cpp
```diff
@@ -26,6 +30,77 @@ static std::ostream& boost_test_print_type(std::ostream& os, const ColorBGRA col
...
+    RTTR_Assert(type == GL_FLOAT);
+    RTTR_Assert(stride == 0);
+    RTTR_Assert(size == 2);
+    texturePointer = static_cast<const Point<float>*>(pointer);
+}
+
+void APIENTRY glColorPointer(GLint size, GLenum type, GLsizei stride, const void* pointer)
+{
+    ++callCount;
+    RTTR_UNUSED(size);
```
Flamefire, line 82: Just FYI: You can just remove the parameter names if you only care about the callcount to avoid the unused macros. Unless you want them to be shown, but if you don't use any I don't see a reason for the names

libs/s25main/ogl/glSmartBitmap.cpp
tehKaiN: I don't quite remember RTTR's internals but it appears that texCoords and srcArea were using coordinate systems - the lerp(inverseLerp()) combo was for remapping between their ranges.

I've simplified the code a bit so perhaps it's more readable what's going on.

libs/s25main/ogl/glSmartBitmap.cpp
```diff
@@ -238,12 +240,32 @@ void glSmartBitmap::generateTexture()
...
+    // nothing to draw?
+    if(!percent)
+        return;
+    RTTR_Assert(percent <= 100);
+
+    const float partDrawn = percent / 100.f;
+    auto startY = int(std::round(size_.y * (1 - partDrawn)));
+    auto endY = size_.y - startY;
+    Rect dstArea(drawPt.x, drawPt.y + startY, size_.x, endY);
+    Rect srcArea(0, startY, size_.x, endY);
```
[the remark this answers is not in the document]
tehKaiN: Good catch!

tests/s25Main/UI/testSmartBitmap.cpp
```diff
@@ -278,4 +353,51 @@ BOOST_AUTO_TEST_CASE(MultiPlayerBitmap)
...
+        // All bottom texture coords have same Y component
+        BOOST_TEST(rttrOglMock2::textureCoords[1].y == rttrOglMock2::textureCoords[2].y);
+        // Bottom starts at bottom
+        BOOST_TEST(rttrOglMock2::textureCoords[1].y == smartBmp.texCoords[1].y);
+        // Top starts at given percentage
+        BOOST_TEST(std::abs((rttrOglMock2::textureCoords[0].y / smartBmp.texCoords[1].y) - .5f) <= 1.f / size.y);
+
+        smartBmp.drawPercent(DrawPoint::all(0), 100);
+        BOOST_TEST(rttrOglMock2::textureCoords.size() == size_t(4));
+        // All top texture coords have same Y component
```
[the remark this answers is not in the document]
tehKaiN: Yup, that simplifies code by a lot.

tehKaiN (commented):
>  Checking the code it looks pretty complete. Just a few straight forward minor improvements. And I noticed the button tests has (sizes) img < button and img > button tests. Does a img == button test make sense or would you say that is covered well enough already or not a relevant use case?

I... don't remember :D I'll look at it soon.

I've made some amendments, also elaborated on the problematic BOOST_CHECK() in one of my comments.

tests/s25Main/UI/testSmartBitmap.cpp
```diff
@@ -278,4 +351,56 @@ BOOST_AUTO_TEST_CASE(MultiPlayerBitmap)
     }
 }
 
+BOOST_AUTO_TEST_CASE(DrawPercent)
+{
+    // Make sure that test passes across different image heights
+    std::vector<Extent> testCases = {
+      {rttr::test::randomValue<unsigned int>(10, 50), 50},
+      {rttr::test::randomValue<unsigned int>(10, 50), 35},
```
tehKaiN, line 359: Is my tooling acting up or does clangformat format such initializers with different indent?

Probably also should've used std::to_array instead...

tests/s25Main/UI/testSmartBitmap.cpp
Flamefire: > size is number of floats in point

--> Points is number of float pairs, i.e. size/2

> OpenGL API that it expects the pointer to be valid the moment the glDrawArrays() is called.

Yes but here you have:
```
void foo() {
Points[...] pts
glVertexPointer(pts)
glDraw()
}
void test() {
foo();
//acccess pts;
}
```

I.e. OpenGL gets the valid pointer but you check the pointer after the function has returned and the pointer is no longer valid.

tests/s25Main/UI/testSmartBitmap.cpp
```diff
@@ -278,4 +353,51 @@ BOOST_AUTO_TEST_CASE(MultiPlayerBitmap)
...
+        smartBmp.drawPercent(DrawPoint::all(0), 50);
+        BOOST_TEST(rttrOglMock2::textureCoords.size() == size_t(4));
+        // All top texture coords have same Y component
+        BOOST_TEST(rttrOglMock2::textureCoords[0].y == rttrOglMock2::textureCoords[3].y);
+        // All bottom texture coords have same Y component
+        BOOST_TEST(rttrOglMock2::textureCoords[1].y == rttrOglMock2::textureCoords[2].y);
+        // Bottom starts at bottom
+        BOOST_TEST(rttrOglMock2::textureCoords[1].y == smartBmp.texCoords[1].y);
+        // Top starts at given percentage
+        BOOST_TEST(std::abs((rttrOglMock2::textureCoords[0].y / smartBmp.texCoords[1].y) - .5f) <= 1.f / size.y);
```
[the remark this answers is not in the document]
Flamefire: But `<= 1.f / size.y` doesn't convey what is tested here, at least I can't tell what this checks just by looking at this.
`1/size` is the full size in texture coords, isn't it? So why should the difference from the calculated 50% to actual 50% be anyhow related to the full size?

That error looks correct to me: Result is `0.514285769 != 0.5` which does look off. So something int the coords **is** off: Either the first value is wrong and should be ~`0.2734375` (half of the 2nd) or you are missing something in the test.   
Because as it stands you don't get 50% of the image but rather 51% if the calculation is correct which I doubt. My guess is `smartBmp.texCoords[0].y != 0`:

Assume (integer coords): BMP: `Top=11`, `Bottom=111`,` size=100 `-> `50% top=66` And with the current check: `66 / 111 != 0.5`

So I guess your assumption was that all texture coords start at zero but they don't. Especially for `smartBitmaps` this isn't true at all: They are parts of a unified texture. I.e. during loading as many images as possible are combined to a single texture and each `glSmartBitmap` gets texture coordinates corresponding to the position in that texture. I.e. it could be the rect from`0.254 0.314` to `0.270 0.412`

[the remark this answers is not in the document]
tehKaiN: Consider a bitmap with height of 35px. Suppose you want to draw 50% of it - that percentage can't be cleanly calculated from that height, so it would be either 17 (48.57%) or 18px (51.43%), depending on roundings. Note that percentage between those values increases by 2.85 percent points. So the check needs to be done with precision of the granularity of the percentage for given height, which is calculated in percents like `100/35 = 2.85%`, or `1 / height` if we don't want the percentage unit.

I'm more and more convinced that this test should be done with height divisible by 100, so that we can skip all that complex math in it.

tests/s25Main/UI/testSmartBitmap.cpp
```diff
@@ -278,4 +353,51 @@ BOOST_AUTO_TEST_CASE(MultiPlayerBitmap)
     }
 }
 
+BOOST_AUTO_TEST_CASE(DrawPercent)
+{
+    // Make sure that test passes across different image heights
+    for(int i = 0; i < 100; ++i)
+    {
+        auto bmpSrc = createRandBmp(0);
+        const Extent size(bmpSrc->getWidth(), bmpSrc->getHeight());
```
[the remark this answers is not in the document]
tehKaiN: That convinced me, but when I tried to declare it, it became apparent that libsiedler doesn't depend on s25Common which I would need to add `#include <Point.h>`. Do we want to add this dependency?

tests/s25Main/UI/testSmartBitmap.cpp
tehKaiN: Yes, but...

My code assumes that both `glVertexPointer` and `glDrawArrays` are done back to back and in the latter call the points from `texturePointer` and `vertexPointer` are copied into the `textureCoords` and `vertexCoords` for later access. So it's more like:

```cpp
void foo() {
Points[...] pts;
glVertexPointer(pts); // store pts ptr for glDraw()
glDraw(); // copy pts contents to textureCoords/vertexCoords
}
void test() {
foo();
//access textureCoords/vertexCoords
}
```

What am I missing here?

tests/s25Main/UI/testSmartBitmap.cpp
Flamefire: There are "full indents" and "continuation indents". I.e. when a new block starts or when the previous line/statement is continued after being broken up due to max. line length

Flamefire (changes requested):
Ok, getting close.

BTW: Please don't force-push in the middle of a review. It's hard to see what has exactly changed after the last time I looked at it. Especially the rebase before that where the "compare changes" then shows your changes and all changes from develop together.

tests/s25Main/UI/testSmartBitmap.cpp
```diff
@@ -278,4 +352,55 @@ BOOST_AUTO_TEST_CASE(MultiPlayerBitmap)
...
+            // All top texture coords have same Y component
+            BOOST_TEST(rttrOglMock2::textureCoords[0].y == rttrOglMock2::textureCoords[3].y);
+            // All bottom texture coords have same Y component
+            BOOST_TEST(rttrOglMock2::textureCoords[1].y == rttrOglMock2::textureCoords[2].y);
+            // Bottom starts at bottom
+            BOOST_TEST(rttrOglMock2::textureCoords[1].y == smartBmp.texCoords[1].y);
+            // Top starts at given percentage
+            BOOST_TEST(std::abs((rttrOglMock2::textureCoords[0].y / smartBmp.texCoords[1].y) - .5f) <= 1.f / size.y);
+            // BOOST_TEST(rttrOglMock2::textureCoords[0].y / smartBmp.texCoords[1].y == .5f, tt::tolerance(1.f /
+            // size.y));
```
Flamefire, line 395: > Not really. Take a look at the drawPercent() function itself - the startY / height are calculated using the percentage from the pixel height, and rounded to pixel precision. This causes the function to not exactly draw the percentage that it was asked to.
> 
> From what I see this was my deliberate change and I think I remember it was necessary to achieve pixel-perfect look on the portrait buttons.

I see the percentage is now the on discrete image pixels, no longer on the continuous texture. I'm not sure this is a good idea. We use this mainly for drawing the building construction progress which would now result in less smooth transitions, doesn't it? 

Given your explanation this needs another testcase: Drawing a rect should draw a given number of pixels. I.e. test that your actual use case works as expected.

The change seems to be: Previously we used `drawPercent(100)` for regular drawing the whole bitmap. Now you go the other way round and use `drawRect` from inside `drawPercent` so now that gets distorted in favor of `drawRect`.   
If we want both I'd say we need support for floating point rects so that the `drawPercent` can  pass that instead of a `Rect` and support both cases. On the other hand: We have a granularity of 1% only anyway so that 1px +- might not be noticable there, e.g. usage here: https://github.com/Return-To-The-Roots/s25client/blob/0d2a837baa3b72c1154940a4ff910f701ca02fa0/libs/s25main/buildings/noBuildingSite.cpp#L221. Did you check already how it looks before and after for a big building construction possibly with zoom to get the worst case?

If we stay with int-rects I still think the test is not fully right (generically): image `textureCoords` can be anywhere. I.e. `[0].y=0.5` & `[1].y=0.7`. So I'd say we at least need a comment and potentially a `BOOST_TEST_REQUIRE(smartBmp.texCoords[0] == PointF(0,0))` to ensure this expectation that for a single image directly added the texture coords of `glSmartBitmap` start at zero. Or simpler: Use `smartBmp.texCoords[1].y - smartBmp.texCoords[0].y` instead of only the former which is the same as the last one should be zero.

And would a better test be that the drawn part of the image is within 1px of the percentage of the height? `TEST(abs(drawHeight - expectedHeight) <= 1)` This is what you have here but I had trouble seeing  it until writing this sentence due to the ` / height` basically applied on both sides on the current check.

Something like:
```suggestion
            // Top starts at given percentage (normalized texture coords)
            const auto expectedStart = smartBmp.texCoords[0].y + percentage * (smartBmp.texCoords[1].y - smartBmp.texCoords[1].y); 
            BOOST_TEST(std::abs(rttrOglMock2::textureCoords[0].y - expectedStart) * size.y <= 1); // 1px tolerance due to rounding of source coordinates in pixels
```

Not sure if this is clearer, or is it? Or move the size:
`BOOST_TEST(std::abs((rttrOglMock2::textureCoords[0].y / smartBmp.texCoords[1].y) - percentage) * size.y <= 1);` // 1px tolerance due to rounding of source coordinates in pixels

Flamefire (changes requested):
> My clang-format version is 20.1.3 - perhaps the wrong one?

Yes we and CI use clang-format 10. The CMake target fixes this to this version because as you noticed different versions are not fully consistent. We might need to increase this to easier support newer systems. However I'd like to keep the short functions on a single line. Getters like `Position GetOrigin() const override { return origin_; }` don't "deserve" 4 lines ;-)

libs/common/include/helpers/mathFuncs.h
```diff
@@ -70,7 +70,8 @@ constexpr T interpolate(const T startVal, const T endVal, const U elapsedTime, c
 }
 
 /// Linear interpolation, similar to C++20's std::lerp()
-constexpr float lerp(const float startVal, const float endVal, const float ratio) noexcept
+template<typename T>
+constexpr T lerp(const T startVal, const T endVal, const T ratio) noexcept
```
Flamefire, line 74: Shall we use a 2nd template param for the ratio and maybe an enable-if to be a floating point type. That would allow `lerp(int, int, float)` and even `lerp(PointF, PointF, float)`. Might need making the return type `auto` as e.g.`int*float=float` and `Point*float = PointF`   
Might be confusing though, what do you think? At least allowing different float types for T and ratio looks reasonable and the PointF-float could be used here already.

tests/s25Main/UI/testSmartBitmap.cpp
```diff
@@ -278,4 +352,61 @@ BOOST_AUTO_TEST_CASE(MultiPlayerBitmap)
...
+                // All top texture coords have same Y component
+                BOOST_TEST(rttrOglMock2::textureCoords[0].y == rttrOglMock2::textureCoords[3].y);
+                // All bottom texture coords have same Y component
+                BOOST_TEST(rttrOglMock2::textureCoords[1].y == rttrOglMock2::textureCoords[2].y);
+                // Bottom starts at bottom
+                BOOST_TEST(rttrOglMock2::textureCoords[1].y == smartBmp.texCoords[1].y);
+                // Top starts at given percentage, draw accuracy is +- 1px
+                auto expectedPixelHeight = ratio * size.y;
+                auto actualRatio =
+                  ((rttrOglMock2::textureCoords[1].y - rttrOglMock2::textureCoords[0].y) / smartBmp.texCoords[1].y);
```
Flamefire, line 398: Nit: Take the actual start coordinate into account:
```suggestion
                const auto actualRatio =
                  (rttrOglMock2::textureCoords[1].y - rttrOglMock2::textureCoords[0].y) / (smartBmp.texCoords[1].y -  smartBmp.texCoords[0].y);
```

Flamefire (changes requested):
Just a nit left, thanks! :-)

> I'm reverting this to the old way without rounding. 

What exactly did (or will you?) revert? Couldn't find it.

> On a personal note though, I'd assume that a pixel-graphics game should have a pixel precision of drawing, so probably at a time it felt for me like a natural improvement, but that's highly subjective.

Sure, we shall always strive for the best, hence my very picky reviews ;-)   
So best if we can achieve both use cases well. But "well enough" is enough

> Since other glSmartBitmap methods use int-based Position/DrawPoint, I guess that having it all int-based is somewhat desired. So it looks like drawPercent()'s float-based internals were somewhat inconsistent with other methods.

"Inconsistent" because special ;-) All other use cases do indeed use pixel-perfect drawing intentionally. But `drawPercent` is used for an animation where we could easily support the sub-pixel accuracy.

> Introduce RectBase<float> to drawRect() to preserve that inconsistent, but well-known behavior? Seems to be the safest bet, but a bit messy on the API side. Tough nut to crack. ;)

That was my idea too: Have an internal implementation method taking a float-Rect which trivially supports the public, pixel perfect version and `drawPercent` can keep the sub-pixel values.

>>    Did you check already how it looks before and after for a big building construction possibly with zoom to get the worst case?
>
> I've played the game a bit with romans, built small buildings as well as castles and guard towers, harbors and ships, on max zoom. I don't see any downside, looks okay to my taste.

"Well enough" then. No need to make this more complicated than it needs to be :-)

tehKaiN:
> What exactly did (or will you?) revert? Couldn't find it.

Yeah, sorry if my post wasn't clear. I was about to revert to old behavior, but then I've edited this msg and elaborated why it's int-rounded and my hand faltered and I didn't make those changes. I'll need your definite decision on what to do with it - since "well enough" is enough, and there are no visual artifacts from current PR code, perhaps we're overthinking it and there's no wrong in keeping it as it is rn. Still, it's not really my project, so I don't want to do too big decisions here.

Flamefire:
>  since "well enough" is enough, and there are no visual artifacts from current PR code, perhaps we're overthinking it and there's no wrong in keeping it as it is rn. 

Yep that was my thought that I wanted to convey with my (2nd) last sentence. So yes keep it and we can revisit if needed later.   
I just wanted to me sure I didn't miss anything

libs/common/include/helpers/mathFuncs.h
tehKaiN: Not sure how to go about it in `lerp(int, int, float)` scenario - what would user expect in the return type? `float` as the value is somewhere between ints, or `int` as the input from/to were passed with that type? I can see scenarios where both ways would be preferred.

I'd leave it as simple as it is, and if given use case emerges then perhaps expand template thingy.

Flamefire (approved):
Looks good to me. The CI failure is unrelated and should be fixed by https://github.com/Return-To-The-Roots/s25client/pull/1774

PR #1537, opened 2022-09-20
Add lua function GetStatisticsValue()

### LINKED ISSUE #1175: [LUA] getStatisticsValue(INDEX)
In order to recreate the original mission as close to the original as possible we need a function that returns the land size to recreate the MET_LAND_SIZE event. This could be achieved by being able to read values from the statistics, not only the land size but for example slain soldiers as well.

tehKaiN:
fixes #1175


Flamefire (approved):
LGTM! Thanks!

PR #1547, opened 2023-01-07
Remove use of sprintf

Flamefire:
Use better alternatives like `format` or `snprintf` to avoid warnings about using this deprecated, unsafe function.

Mainly affects the submodules so updated them all.

Closes #1546


Flamefire:
@Flow86 Jenkins gives a 500-error for me. No idea why it failed.

FixedSpeed (approved):

Issue #1549, opened 2023-01-28
hide the info bar at the top


Lopez73:
Is there a way to hide the info bar at the top, somehow it bothers me a bit when playing. Thanks!! :)
https://www.imagebam.com/view/MEIDQE8

PR #1550, opened 2023-01-29
Add option to disable GF debug data

Flamefire:
The information about current GF, GF length, NWF length and ping are usually not relevant.
So add an option to enable them when required.

Also introduce an enum for the IDs in dskOptions to be able to add/change stuff.
Closes https://github.com/Return-To-The-Roots/s25client/issues/1549

### DIFF
--- a/libs/s25main/Settings.cpp
+++ b/libs/s25main/Settings.cpp
@@ -72,9 +72,10 @@ void Settings::LoadDefaults()
     // {
     // 0 = ask user at start,1 = enabled, 2 = disabled
     global.submit_debug_data = 0;
-    global.use_upnp = 2;
+    global.use_upnp = false;
     global.smartCursor = true;
     global.debugMode = false;
+    global.showGFInfo = false;
     // }
 
     // video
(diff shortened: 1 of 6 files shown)


Spikeone:
Just one quick thing I always thought about: You can not enable this in normal play anymore (thus you can not write down a GF for checking later in a replay) - therefore you'd need to create a savegame to find out which GF you should jump to.

libs/s25main/Settings.cpp
```diff
@@ -72,9 +72,10 @@ void Settings::LoadDefaults()
     // {
     // 0 = ask user at start,1 = enabled, 2 = disabled
     global.submit_debug_data = 0;
-    global.use_upnp = 2;
+    global.use_upnp = false;
```
Flow86, line 75: this breaks existing, did you increase the settings version?

libs/s25main/desktops/dskOptions.cpp
```diff
@@ -154,54 +219,68 @@ dskOptions::dskOptions() : Desktop(LOADER.GetImageN("setup013", 0))
...
+    upnp->AddTextButton(0, DrawPoint(280, 385), Extent(190, 22), TextureColor::Grey, _("Off"), NormalFont);
+    upnp->AddTextButton(1, DrawPoint(480, 385), Extent(190, 22), TextureColor::Grey, _("On"), NormalFont);
+    upnp->SetSelection(SETTINGS.global.use_upnp);
+
+    groupAllgemein->AddText(ID_txtSmartCursor, DrawPoint(80, 420), _("Smart Cursor"), COLOR_YELLOW, FontStyle{},
+                            NormalFont);
+    ctrlOptionGroup* smartCursor = groupAllgemein->AddOptionGroup(ID_grpSmartCursor, GroupSelectType::Check);
     smartCursor->AddTextButton(
-      10103, DrawPoint(280, 415), Extent(190, 22), TextureColor::Grey, _("Off"), NormalFont,
+      0, DrawPoint(280, 415), Extent(190, 22), TextureColor::Grey, _("Off"), NormalFont,
```
Flow86, line 241: why not using enums here then?

Flow86 (changes requested):

libs/s25main/Settings.cpp
Flamefire: Well we assign this to a bool (even prior to this PR) so `2` is "true" which is likely not intended. Nothing is broken, the currently stored "2" will be converted to "true" on load and to "1" on save so this is still compatible, isn't it?

libs/s25main/desktops/dskOptions.cpp
Flamefire: I could but chose not to: We basically have boolean text buttons: On and Off so I used IDs 0 and 1 for those, no need for an enum, although we could add constants manually `ID_btOn/Off` if you really want.

Flow86: yea add those, that makes the "0" and "1" much more readable

libs/s25main/Settings.cpp
Flow86: ah true though, thats alright then.

libs/s25main/desktops/dskOptions.cpp
Flamefire: Done. The debug-data thing needed some special handling and looks actually odd. We may need to look at that later if we want to implement a "never" option. I just changed/fixed the button label here as that is unrelated to this PR.

Flow86 (approved):

Flamefire:
Rebased to fix the submodule update. No change as I ordered the submodule commits such that it works.

Issue #1102, opened 2019-08-19
Smartmoves got stuck


Spikeone:
I encountered a strange bug which was observable by the following behavior:
- Scrolling by rightclicking felt like it had some kind of force feedback, means by move the mouse a bit the screen moved quite a lot the further you scrolled. Also the scrollspeed was increased, making it easy to scroll the map several times by move the mouse less than a node
- Smartmove for mouse positioning didn't work anymore. The cursor was always placed in the upper left corner (close) which means you had to wait before you could do the 2nd click, move the mouse to the button and then click or you'd just close the window
- The cursor was sometimes positioned off, that way the window for a new building often was missplaced as the cursor first jumped to the possible upper left corner of the new window, then the window opened and the cursor jumped again

Sadly I could no reproduce this play playing the replay. I only know after which GF this problem occured (248000+) so not sure if it could possibly help.

Flamefire:
I don't get it exactly. Do you say this happens only sometimes? Because this stuff is independent from any game specific stuff...

- Could not reproduce with SDL2
- Don't really understand. Just tried with place-flag window and cursor was placed at the button.
- Same here. 

Which videoDriver are you using? Try SDL2 and see if it works there. IMO we should fully switch to SDL2 and remove support for changing it and dynamic loading. Just makes development harder and bug reports less precise.

Spikeone:
Can't reproduce it either. 
- SDL2
- Windowed 1920x1080

Rewatched my video but you can't really see it there either. Still can only say:
- moving the mouse 5 nodes in one direction results in nearly scroling the whole map
- Smartmoves did not work correctly, mouse was placed but not on buttons

Spikeone:
Happened again today, mouse moves to the upper left corner when opening any window

Flamefire:
Is this fully reproducible? I.e. happens everytime?
What exactly do you mean by "context window"?

Spikeone:
Sadly it doesn't happen everytime but this time I got a video, gotta check that out later.

Issue #1500, opened 2022-02-18
Panning with mouse erratic


Homo-carbonis:
I've just started playing the game after building from git and I'm getting an enormous input lag when panning with the mouse. The camera just whizzes around the map uncontrollably. Both the mouse cursor and moving the camera with the arrow keys work normally.

phoenix147:
Same here, panning with mouse is almost unusable, although mine is stuck mostly and just moves very little
I'm using latest Manjaro Linux

Same problem was described here (in german): https://www.siedler25.org/index.php?com=forum&mod=forum&action=thread&id=1704&start=-1#last

Disabling smart cursor is a workaround

Is there anything we can do to track this down?

Spikeone:
Hm, I have no idea, but hope @Flamefire has - Could this be related to high DPI mouses when using linux?

chewi:
Same here with 0.9.5 on Gentoo. 0.9.1 worked fine last I checked. I don't have a high DPI screen or mouse.

chewi:
Actually I've retested 0.9.1 again and it happens there too. Perhaps it's a change in SDL or Xorg. Disabling the Smart Cursor option does indeed help although the panning is much faster than I remember.

Spikeone:
```
vsync=0
```
seems to work fine
```
vsync=1
```
seems to cause 1 FPS

Spikeone:
this should be fixed and the vsync issue is not related to this one

PR #1557, opened 2023-02-27
Fix mouse panning issues in SDL2 video driver

### LINKED ISSUE #1526: Cant Scroll with right Mouse
Hi,

I installed a fresh Garuda DE with KDE on my System.
Now I installed rttr, but, I cant scroll in the Game the Map with the right Mouse, it just scroll for for a part of a second and then it stops scrolling. If I click again it also scrolls for a small part of a second, again, and stops.

It makes no diferend if I play in windowed or fullscreen mode.
In Windows 10 I dont get sutch problems in rtttr.

Other games work like a charm(some Steam games like HL2, UT3, Skyrim, C&C Remaster, ect...) or Games like Widelands(also use the right Mouse for scrolling).

siredmar:
This should fix #1526 and #1500.
The SDL2 video driver got stuck on the logic regarding `mouseMoved`. I didn't see any reason why to catch the first mouse move only. The problem is, that the Msg_MouseMove is only sent on the first time -> no mouse pan working. So this PR removes the logic around `mouseMoved` and basically sends every mouse motion event.

### DIFF
--- a/extras/videoDrivers/SDL2/VideoSDL2.cpp
+++ b/extras/videoDrivers/SDL2/VideoSDL2.cpp
@@ -251,8 +251,6 @@ bool VideoSDL2::SwapBuffers()
 
 bool VideoSDL2::MessageLoop()
 {
-    static bool mouseMoved = false;
-
     SDL_Event ev;
     while(SDL_PollEvent(&ev))
     {
@@ -387,19 +385,12 @@ bool VideoSDL2::MessageLoop()
             }
             break;
             case SDL_MOUSEMOTION:
-                // Handle only 1st mouse move
-                if(!mouseMoved)
-                {
-                    mouse_xy.pos = Position(ev.motion.x, ev.motion.y);
-
-                    CallBack->Msg_MouseMove(mouse_xy);
-                    mouseMoved = true;
-                }
+                mouse_xy.pos = Position(ev.motion.x, ev.motion.y);
+                CallBack->Msg_MouseMove(mouse_xy);
                 break;
         }
     }
 
-    mouseMoved = false;
     return true;
 }
 


Spikeone:
Guess Flamefire will have a proper review, only thing I could imagine is: this was for smartmoves and maybe the mouse event fired twice in some cases. 

maybe this even fixes https://github.com/Return-To-The-Roots/s25client/issues/1102

Flamefire:
I guess this was for the case where a mouse-warp caused a motion event breaking things, but not sure anymore.   
I can't see why this would be the cause: It just means that only the first queued motion event will be handled. Then stuff will be redrawn etc and in the next event handling part the next (new) motion event will be handled

> The problem is, that the Msg_MouseMove is only sent on the first time -> no mouse pan working.

But it does work doesn't it? It may ignore some parts of the movements if they happen "to fast" or drawing etc takes to long.

Anyway: @siredmar Have you reproduced the issue from #1526 and this patches fixes it for you? I tried in virtual box and there it is still broken (but I think this is a bug in SDL)

siredmar:
@Flamefire yes, I can reproduce this issue with my vanilla ubuntu 22.04 installation using Intel graphic drivers.
This PR fixes the issue. Since the patch i had no issues anymore.

siredmar:
@Flamefire during my debugging i saw that SDL fired MOUSEMOTION event on right click enabled permanently. So once clicked and moved one pixel there were a lot of events. Can you reproduce it with your virtual box environment?

Flamefire:
> @Flamefire during my debugging i saw that SDL fired MOUSEMOTION events on right click enabled permanently. So once clicked and moved one pixel there were a lot of events. Can you reproduce it with your virtual box environment?

In VirtualBox the events were fired correctly but the warp-back doesn't work as it has no effect on future motion events. Reported at: https://github.com/libsdl-org/SDL/issues/5741

>  This PR fixes the issue. Since the patch i had no issues anymore.

As I wasn't able to produce any issues with this patch (maybe the bug I worked around with that code does no longer exist in SDL2) I'm fine with merging this and see if that also helps others.

siredmar:
@Flamefire the codecov/patch workflow fails because of missing unit tests. I'm not quite sure how to unit test this. Do you have any idea or simply skip this time?

Flamefire (approved):

Flamefire:
> @Flamefire the codecov/patch workflow fails because of missing unit tests. I'm not quite sure how to unit test this. Do you have any idea or simply skip this time?

It was just missing an approving review. Coverage is only informational.

Merging, thanks for this!

PR #1558, opened 2023-03-06
Change runner definition due to deprecation

Flow86:


Flow86:
Don't know why those are still failing, we have to investigate

Flow86:
yea it seems its a bug in lcov or so:

grafik

it catches the line, but due to formatting not the full one

.github/workflows/unit-tests.yml
```diff
@@ -13,70 +13,72 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  BOOST_VERSION: 1.69.0
+  CMAKE_VERSION: 3.26.0
+  BOOST_VERSION: 1.81.0
```
Flamefire, line 17: The Boost version of 1.69 corresponds to the minimum boost version we support, see https://github.com/Return-To-The-Roots/s25client/blob/master/CMakeLists.txt#L309

Why the change?

.github/workflows/unit-tests.yml
```diff
@@ -13,70 +13,72 @@ concurrency:
...
   Linux:
     strategy:
       matrix:
         include:
-          - { compiler: gcc-6,    os: ubuntu-18.04, buildType: Release }
-          - { compiler: gcc-6,    os: ubuntu-18.04, buildType: Debug, coverage: true }
-          - { compiler: clang-7,  os: ubuntu-18.04, buildType: Release }
-          - { compiler: clang-7,  os: ubuntu-18.04, buildType: Debug }
-          - { compiler: clang  ,  os: macos-10.15,  buildType: Debug, boostVersion: 1.74.0 } # Multiple bugs with recent OSX until 1.74
-          # GCC 9 is known to show a few warnings that GCC 10 has "fixed", make sure this doesn't happen for us
```
Flamefire, line 29: Is GCC-9 really not available on any of the runner images? This entry was especially to avoid a regression on GCC 9 (hence the comment)   
Also the current change means that our supported minimum  GCC version changes from 6 to 10 which is a lot IMO.

It looks like GCC-7 is available in the [Ubuntu 20 repos](https://packages.ubuntu.com/focal/allpackages?format=txt.gz)

So I'd use GCC-7 as the minimum and keep the GCC-9 entry

.github/workflows/unit-tests.yml
```diff
@@ -13,70 +13,72 @@ concurrency:
...
+          # Linux GCC
+          - { compiler: gcc-10,   os: ubuntu-20.04, buildType: Debug, cxxStandard: 14, coverage: true }
+          - { compiler: gcc-10,   os: ubuntu-20.04, buildType: Debug, cxxStandard: 17 }
+          - { compiler: gcc-10,   os: ubuntu-20.04, buildType: Release, cxxStandard: 14 }
+          # Linux clang
+          - { compiler: clang-11, os: ubuntu-20.04, buildType: Debug, cxxStandard: 14, externalSanitizer: true }
+          - { compiler: clang-11, os: ubuntu-20.04, buildType: Debug, cxxStandard: 17 }
+          - { compiler: clang-11, os: ubuntu-20.04, buildType: Release, cxxStandard: 14 }
+          # Linux Latest GCC
+          #- { compiler: gcc-12,   os: ubuntu-20.04, buildType: Debug, cxxStandard: 14, coverage: true }
```
Flamefire, line 36: This should work:
```suggestion
          - { compiler: gcc-12,   os: ubuntu-22.04, buildType: Debug, coverage: true }
```

Flamefire (changes requested):
I'd change the compiler matrix a bit. What we had before:

- Oldest GCC supported (GCC 6)
- Oldest Clang supported (Clang 7)
- Both with their default C++ standard and in debug and release mode
- Regression test for GCC 9
- Latest GCC and Clang in the oldest possible C++ standard (14, they default to something newer) and their default C++ standard
- Test for latest compiler with latest boost (default Boost is the oldest supported by us)

Given the possible runner images I see GCC-7 and Clang-7 available in Ubuntu 20 as the oldest possible and GCC-12 in Ubuntu 22 and Clang-16 via the apt-repo as the newest compilers and think we should use those similar to what we had before.

.github/workflows/unit-tests.yml
Flow86: 1.69.0 does not work anymore on those runners

.github/workflows/unit-tests.yml
Flow86: gcc-9 is also so outdated, lets skip those old compilers without c++14 support (I'll update the whole toolchain)

.github/workflows/unit-tests.yml
Flamefire: I'll check that. I'm a huge fan of "what isn't tested is broken" hence I'll try to get an older Boost to work on the runners.

.github/workflows/unit-tests.yml
Flamefire: What doesn't it support? And I wouldn't say it is outdated: It is the default for Ubuntu 20

PR #1562, opened 2023-03-11
glad: Fix GLES and OGL-Compat detection

kloptops:
Changed `CMAKE_MATCH_0` to `CMAKE_MATCH_1` otherwise it will never match `GLES`. Changed `set(RTTR_OGL_COMPAT 0)` on line 30, not sure if this was intended or a bug.


Flamefire (approved):
Thanks!

PR #1565, opened 2023-03-18
Quickstart AI Battle

wichern:
This change allows to quick-start into an AI-only game.

Example:
`./bin/s25client -m share/s25rttr/RTTR/MAPS/NEW/Bagel5.SWD --ai aijh --ai aijh --ai dummy`

Will result in 
Screenshot from 2023-03-18 22-15-57

Implementation:
This change adds a new program option (`--ai`) and extends `QuickStartGame()`.
In order to have the requested AIs ready in `dskGameLobby`, the AI configuration is stored in `GameClient`.
In `GameClient::GameLoaded()` we call `ToggleHumanAIPlayer()` to turn the human player into an AI as well.

Notes:
There are no further limitations. The user can still add buildings and roads.


libs/s25main/network/GameClient.cpp
```diff
@@ -1535,6 +1539,17 @@ bool GameClient::StartReplay(const boost::filesystem::path& path)
     return true;
 }
 
+void GameClient::SetAIBattlePlayers(std::vector<JoinPlayerInfo>&& playerInfos)
+{
+    aiBattlePlayers_ = playerInfos;
```
Flamefire, line 1544: The move needs to be explicit (which I admit is not intuitive, but see https://godbolt.org/z/j3zb3c5a3)

And then you can take it by value, as by R-value ref is unusual:
```suggestion
void GameClient::SetAIBattlePlayers(std::vector<JoinPlayerInfo> playerInfos)
{
    aiBattlePlayers_ = std:move(playerInfos);
```

libs/s25main/network/GameClient.cpp
```diff
@@ -1593,7 +1608,7 @@ void GameClient::SkipGF(unsigned gf, GameWorldView& gwv)
 
     unsigned start_ticks = VIDEODRIVER.GetTickCount();
 
-    if(!replayMode)
+    if(!replayMode && !aiBattleMode_)
```
Flamefire, line 1611: I think this won't work: Wouldn't the AI do nothing in this case (i.e. not in a replay)? The AI IIRC is handled on the server so simply running GFs will not let the server (and hence the AI) be involved.   
Can you verify this?

libs/s25main/network/GameClient.cpp
```diff
@@ -1695,7 +1710,7 @@ void GameClient::SetPause(bool pause)
             return;
         framesinfo.isPaused = true;
         framesinfo.frameTime = FramesInfo::milliseconds32_t::zero();
-    } else if(replayMode)
+    } else if(replayMode || aiBattleMode_)
```
Flamefire, line 1713: Similar to above: The server needs to be involved.

libs/s25main/network/GameClient.h
```diff
@@ -300,6 +310,10 @@ class GameClient final :
 
     std::unique_ptr<ReplayInfo> replayinfo;
     bool replayMode;
+
+    // Configured players for an AI battle.
```
Flamefire, line 314: Docstrings need to start with 3 slashes

```suggestion
    /// Configured players for an AI battle.
```

libs/s25main/network/GameClient.h
```diff
@@ -164,7 +172,9 @@ class GameClient final :
     void SystemChat(const std::string& text) override;
     void SystemChat(const std::string& text, unsigned char fromPlayerIdx);
 
-    void ToggleHumanAIPlayer();
+    /// @brief Toggle current player to be an AI player
+    /// @param aiInfo           AI to replace with
```
Flamefire, line 176: Maybe shorten it to:
```suggestion
    /// Toggle current player to be an AI player of the given type
```

libs/s25main/network/GameClient.h
```diff
@@ -129,6 +130,13 @@ class GameClient final :
 
     /// Lädt ein Replay und startet dementsprechend das Spiel
     bool StartReplay(const boost::filesystem::path& path);
+
+    /// @brief Set the player infos for the AI battle.
+    /// @param playerInfos  List of AI players
```
Flamefire, line 135: Maybe rather: 
```suggestion
    /// When a non-empty vector is given then an AI battle with the given AIs is started
```

as that is what it does mean. The current description is pretty much the same as the name (so the name is already very good as it doesn't need much description)

Flamefire (changes requested):
Good idea, well done overall, just a few bits for polishing :+1:

libs/s25client/s25client.cpp
```diff
@@ -473,8 +473,19 @@ int RunProgram(po::variables_map& options)
...
+        if(options.count("map"))
+        {
+            if(options.count("ai"))
+            {
+                if(!QuickStartGame(options["map"].as<std::string>(), options["ai"].as<std::vector<std::string>>(),
+                                   true))
+                    return 1;
+            } else
+            {
+                if(!QuickStartGame(options["map"].as<std::string>()))
```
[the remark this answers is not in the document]
wichern: Yes, you are right. That is much better.

libs/s25client/s25client.cpp
```diff
@@ -522,6 +533,7 @@ int main(int argc, char** argv)
     desc.add_options()
         ("help,h", "Show help")
         ("map,m", po::value<std::string>(),"Map to load")
+        ("ai", po::value<std::vector<std::string>>(),"AI player to add")
```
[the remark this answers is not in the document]
wichern: agreed

libs/s25main/QuickStartGame.cpp
```diff
@@ -56,8 +58,42 @@ bool QuickStartGame(const boost::filesystem::path& mapOrReplayPath, bool singleP
 
     WINDOWMANAGER.Switch(std::make_unique<dskSelectMap>(csi));
 
-    if((extension == ".sav" && GAMECLIENT.HostGame(csi, mapOrReplayPath, MapType::Savegame))
-       || ((extension == ".swd" || extension == ".wld") && GAMECLIENT.HostGame(csi, mapOrReplayPath, MapType::OldMap)))
+    static_cast<void>(ais);
+    if(!ais.empty() && (extension == ".swd" || extension == ".wld")
+       && GAMECLIENT.HostGame(csi, mapOrReplayPath, MapType::OldMap))
+    {
+        std::vector<JoinPlayerInfo> playerInfos;
```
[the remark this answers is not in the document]
wichern: agreed

libs/s25main/QuickStartGame.cpp
```diff
@@ -56,8 +58,42 @@ bool QuickStartGame(const boost::filesystem::path& mapOrReplayPath, bool singleP
...
+
+            playerInfos.push_back(pi);
+        }
+        GAMECLIENT.SetAIBattlePlayers(std::move(playerInfos));
+
+        WINDOWMANAGER.ShowAfterSwitch(std::make_unique<iwConnecting>(csi.type, nullptr));
+        return true;
+    } else if((extension == ".sav" && GAMECLIENT.HostGame(csi, mapOrReplayPath, MapType::Savegame))
+              || ((extension == ".swd" || extension == ".wld")
+                  && GAMECLIENT.HostGame(csi, mapOrReplayPath, MapType::OldMap)))
```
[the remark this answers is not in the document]
wichern: You are right.
I had multiple iterations with lots of changes. The required changes finally collapsed to this and I did not see the similarity to the existing code. I still wonder though, what the use-case of a savegame in an ai-only battle would look like. I will try it.

libs/s25main/QuickStartGame.cpp
```diff
@@ -56,8 +58,42 @@ bool QuickStartGame(const boost::filesystem::path& mapOrReplayPath, bool singleP
...
     WINDOWMANAGER.Switch(std::make_unique<dskSelectMap>(csi));
 
-    if((extension == ".sav" && GAMECLIENT.HostGame(csi, mapOrReplayPath, MapType::Savegame))
-       || ((extension == ".swd" || extension == ".wld") && GAMECLIENT.HostGame(csi, mapOrReplayPath, MapType::OldMap)))
+    static_cast<void>(ais);
+    if(!ais.empty() && (extension == ".swd" || extension == ".wld")
+       && GAMECLIENT.HostGame(csi, mapOrReplayPath, MapType::OldMap))
+    {
+        std::vector<JoinPlayerInfo> playerInfos;
+        for(unsigned playerId = 0; playerId < ais.size(); ++playerId)
```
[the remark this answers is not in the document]
wichern: agreed

libs/s25main/network/GameClient.cpp
wichern: Yes, you are right. Skipping won't work like this.

libs/s25main/network/GameClient.cpp
wichern: Pause seems to work either way. I removed this change.

libs/s25main/network/GameClient.h
wichern: agreed

libs/s25main/network/GameClient.h
wichern: agreed

libs/s25main/network/GameClient.h
wichern: agreed

wichern:
@Flamefire 
Thank you very much for your review!
I tried to fix all your findings.

Tomorrow, I'll check if it is possible to save and load an ai battle.

Flamefire (changes requested):
Looking good so far, just a few minor improvements, especially making stuff const correct/const-where-possible,   
Just a remark: Short-ish single-line statements after `if`/`for`/... don't need braces. Indentation (enforced via clang-format) already take care of potential bugs. No need to remove them as they don't hurt, just might save you some work.

There is still a bug though now that I'm thinking about it: `aiBattlePlayers_` is never reset, this means after the first such quickstart **every** game will be considered an AI battle. I guess this reset (clear) can be best done in `GameClient::Stop`

libs/s25main/desktops/dskGameLobby.cpp
```diff
@@ -254,10 +254,36 @@ dskGameLobby::dskGameLobby(ServerType serverType, std::shared_ptr<GameLobby> gam
...
+                lobbyHostController->SetPlayerState(i, PlayerState::AI, GAMECLIENT.GetAIBattlePlayers()[i]);
+            } else
+            {
+                // Close remaining slots
+                lobbyHostController->CloseSlot(i);
+            }
+        }
+
+        // Set name of host to the corresponding AI for his Id.
+        auto jpi = gameLobby_->getPlayer(localPlayerId_);
```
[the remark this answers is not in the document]
wichern: agree

libs/s25main/network/GameClient.h
```diff
@@ -129,6 +130,12 @@ class GameClient final :
 
     /// Lädt ein Replay und startet dementsprechend das Spiel
     bool StartReplay(const boost::filesystem::path& path);
+
+    /// When a non-empty vector is given then an AI battle with the given AIs is started
+    void SetAIBattlePlayers(std::vector<AI::Info>&& aiInfos);
```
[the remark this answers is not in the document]
wichern: You are right, this approach will still create a copy. I am not sure how I could avoid this with pass-by-instance.
Since I do not need the value on caller side anymore, I switched to `swap` here. Should I rename the method to `SwapAIBattlePlayers`?

wichern:
I did not see an existing UT for `QuickStartGame()`. How should I handle missing code coverage? Would you like me to create a UT for it?

libs/s25main/QuickStartGame.cpp
```diff
@@ -59,6 +63,27 @@ bool QuickStartGame(const boost::filesystem::path& mapOrReplayPath, bool singleP
...
+            if(ai_lower == "aijh")
+            {
+                type = AI::Type::Default;
+            } else if(ai_lower == "dummy")
+            {
+                type = AI::Type::Dummy;
+            } else
+            {
+                LOG.write(_("Invalid AI player name: %1%\n")) % ai;
+                return false;
```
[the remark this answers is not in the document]
Flamefire: Yes I think in this file is a good idea. The test for that (if you want) can be in testSimple or so, as it doesn't require any initialization. For the interface I'd rather have it return a `std::vector<AI::Info>`. Out-parameters are confusing to read. If you want a `bool` you could use an `optional` (from Boost). You need something to produce the error anyway (e.g. "Invalid AI player name"), so maybe an exception can be used instead which can be logged by the caller. But only logging it and returning an empty optional also works and is maybe easier.

libs/s25main/desktops/dskGameLobby.cpp
```diff
@@ -254,10 +254,36 @@ dskGameLobby::dskGameLobby(ServerType serverType, std::shared_ptr<GameLobby> gam
...
+    {
+        // Initialize AI battle players
+        unsigned aiPlayerCount =
+          std::min(gameLobby_->getNumPlayers(), static_cast<unsigned>(GAMECLIENT.GetAIBattlePlayers().size()));
+
+        RTTR_Assert(localPlayerId_ < aiPlayerCount);
+
+        for(unsigned i = 0; i < gameLobby_->getNumPlayers(); i++)
+        {
+            if(i < aiPlayerCount)
```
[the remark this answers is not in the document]
Flamefire: Ok so:
```suggestion
            if(i < aiBattlePlayers.size())
```
and `aiPlayerCount` can be removed

libs/s25main/desktops/dskGameLobby.cpp
```diff
@@ -594,6 +614,10 @@ void dskGameLobby::Msg_Group_ComboSelectItem(const unsigned group_id, const unsi
 
 void dskGameLobby::GoBack()
 {
+    // We have to clean up the configured ai players, so that subsequent games can
+    // start with the default setup again.
+    GAMECLIENT.ClearAIBattlePlayers();
```
Flamefire, line 619: I still think that `GameClient::Stop` is the best place to do this as here it can be easily forgotten at some point. I don't even know if `GameClient::ExitGame()` is the only place how a game is exited past this point, but `GameClient::Stop` is called in `GameClient::Start` (which would also be a good place: In `Start` it makes sense to clear AI players, but as it calls `Stop` already that might be even better. Maybe even an assert in `Start` that no AI battle is active could be added in case `Stop` is removed in the future)

Yes it's a bit of a mess ;-)

libs/s25main/network/GameClient.h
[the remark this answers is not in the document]
Flamefire: > you are right, this approach will still create a copy. I am not sure how I could avoid this with pass-by-instance.

On the call site: `SetAIBattlePlayers(std::move(aiInfos))` and in the function move the argument. To my understanding this is the "modern C++" for sink arguments (clang-tidy even has a check for that, at least for constructors and `std::string`)

Note that the move is in this case a "cheap swap", I'm pretty sure the move-ctor/assignment of `std::vector` simply swaps 3 pointers at most. Hence the swap inside the function is not required as that's what the move will already do (at worst) so the change to an instance parameter will at worst (and likely)  just involve an addition move-construction which as mentioned is cheap enough.

> Since I do not need the value on caller side anymore

This is what `std::move(aiInfos)` at the call site expresses very well.

Having this an instance parameter allows to make a copy IF the value on the call site would be needed, hence that provides more flexibility at the cost of a (cheap) additional move.

Flamefire:
> I did not see an existing UT for `QuickStartGame()`. How should I handle missing code coverage? Would you like me to create a UT for it?

As that method is not critical I'm OK with missing coverage for that, as we have the review at least. If you can come up with a test then feel free. However I imagine that this would be very complicated due to the client, server, windowmanager and desktop being involved (so it's rather an integration test). Hence the idea to factor out the AI name parsing to have at least that tested, and skip the rest which "should work" after inspection ;-)

Flamefire (approved):

libs/s25main/desktops/dskGameLobby.cpp
wichern: Could it be you mean `GameClient::Connect` instead of `GameClient::Start`? `GameClient::Stop` is called by `GameClient::Connect`. I added an assert there.

Flamefire (approved):

libs/s25main/desktops/dskGameLobby.cpp
Flamefire: Yes indeed. Perfect!

Issue #1176, opened 2020-01-14
[LUA] onAttack(attacker, defender)


Spikeone:
This function should be called whenever one player attacks another player. This function is needed to change the agression level of the AI if the feature described in #743 is implemented.

Flamefire:
Implemented in #1570

PR #1570, opened 2023-04-03
#1176 onAttack for lua

ottml:


Flow86 (approved):

libs/s25main/world/GameWorld.cpp
```diff
@@ -779,6 +779,11 @@ void GameWorld::Attack(const unsigned char player_attacker, const MapPoint pt, c
     // Send the soldiers to attack
     unsigned curNumSoldiers = 0;
 
+    if(!potentialAttackers.empty() && HasLua())
+    {
+        GetLua().EventAttack(player_attacker, attacked_building->GetPlayer());
+    }
```
Spikeone, line 785: maybe It'd be cool to have the number of attackers as well, since we have the count later anyway.

ottml: Good improvement. Added additional parameters for the number of attackers.

Flamefire:
@Flow86 What's up with Jenkins?

Flow86:
> @Flow86 What's up with Jenkins?

some of the packages seem to be broken, I'll try to renovate the pipeline (to get a newer cross compiler etc) but this will take a while

JarnoNL79:
In the gameworld.cpp file the lua onAttack event is added to the _GameWorld::Attack_, if I understand the code correctly this is only for land attacks and there's a seperate method / function to do sea attacks, the _GameWorld::AttackViaSea_ the lua event seems to be missing from there. 

I could misinterpret the code, but I think adding the code below after line 836 should do the trick (could be something I missed for the harbour instead of building) 

```
    if(counter > 0 && HasLua())
    {
        GetLua().EventAttack(player_attacker, attacked_building->GetPlayer(), counter );
    }
```

ottml:
> In the gameworld.cpp file the lua onAttack event is added to the _GameWorld::Attack_, if I understand the code correctly this is only for land attacks and there's a seperate method / function to do sea attacks, the _GameWorld::AttackViaSea_ the lua event seems to be missing from there.
> 
> I could misinterpret the code, but I think adding the code below after line 836 should do the trick (could be something I missed for the harbour instead of building)
> 
> ```
>     if(counter > 0 && HasLua())
>     {
>         GetLua().EventAttack(player_attacker, attacked_building->GetPlayer(), counter );
>     }
> ```

Added the missing call to seaAttack too.

Flamefire (approved):

Flow86 (changes requested):
Please rebase on latest master (and perhaps squash some commits to get the history a bit more clean)

Flamefire (approved):

ottml:
@Flow86 Can you review and approve and merge if everything is okay. Thanks

Flow86 (approved):

PR #1572, opened 2023-04-07
Granular military control

ii8:
This changes the military control addon to allow the player to send and order their little guys however they like.

The original button for sending max rank soldiers home is still there but it is now simply a shortcut to change the garrison of a building to just one private.


Spikeone:
I'd be really happy if we could also keep the 'original' control addon. With that implementation you only got a bit control and that's it - now you can only decide to either have nothing or the big package.

Options could be:
Original
Enhanced
Full

ii8:
As requested the old control addon is back now. This is ready I think if what I did with the replay version is right.

libs/s25main/FindWhConditions.h
```diff
@@ -35,10 +35,10 @@ struct HasWareAndFigure : protected HasMinWares, protected HasFigure
     bool operator()(const nobBaseWarehouse& wh) const;
 };
 
-struct HasMinSoldiers
+struct HasSoldiers
 {
-    const unsigned count;
-    HasMinSoldiers(unsigned count) : count(count) {}
+    const std::array<bool, NUM_SOLDIER_RANKS> ranks;
+    HasSoldiers(std::array<bool, NUM_SOLDIER_RANKS> ranks) : ranks(ranks) {}
```
Flamefire, line 41: Pass non-trivial/large-ish values by reference:
```suggestion
    HasSoldiers(const std::array<bool, NUM_SOLDIER_RANKS>& ranks) : ranks(ranks) {}
```

libs/s25main/FindWhConditions.h
```diff
@@ -35,10 +35,10 @@ struct HasWareAndFigure : protected HasMinWares, protected HasFigure
     bool operator()(const nobBaseWarehouse& wh) const;
 };
 
-struct HasMinSoldiers
+struct HasSoldiers
```
Flamefire, line 38: Maybe a better name: `HasAnyMatchingSoldier` as passing an array of ranks had me thinking this would return true if ALL are available, not ANY.

libs/s25main/GamePlayer.cpp
```diff
@@ -1172,20 +1173,25 @@ bool GamePlayer::IsAttackable(const unsigned char playerId) const
         return GetPactState(PactType::NonAgressionPact, playerId) != PactState::Accepted;
 }
 
-void GamePlayer::OrderTroops(nobMilitary* goal, unsigned count, bool ignoresettingsendweakfirst) const
+void GamePlayer::OrderTroops(nobMilitary* goal, std::array<unsigned, NUM_SOLDIER_RANKS> counts, unsigned max) const
```
Flamefire, line 1176:
```suggestion
void GamePlayer::OrderTroops(nobMilitary* goal, const std::array<unsigned, NUM_SOLDIER_RANKS>& counts, unsigned max) const
```

libs/s25main/GamePlayer.cpp
```diff
@@ -1172,20 +1173,25 @@ bool GamePlayer::IsAttackable(const unsigned char playerId) const
...
+
+        wh = FindWarehouse(*goal, FW::HasSoldiers(desiredRanks), false, false);
         if(wh)
         {
-            unsigned order_count = std::min(wh->GetNumSoldiers(), count);
-            count -= order_count;
-            wh->OrderTroops(goal, order_count, ignoresettingsendweakfirst);
+            wh->OrderTroops(goal, counts, max);
+            sum = 0;
+            std::accumulate(counts.begin(), counts.end(), sum);
```
Flamefire, line 1192:
```suggestion
            sum = std::accumulate(counts.begin(), counts.end(), 0u);
```

libs/s25main/GamePlayer.cpp
```diff
@@ -1172,20 +1173,25 @@ bool GamePlayer::IsAttackable(const unsigned char playerId) const
         return GetPactState(PactType::NonAgressionPact, playerId) != PactState::Accepted;
 }
 
-void GamePlayer::OrderTroops(nobMilitary* goal, unsigned count, bool ignoresettingsendweakfirst) const
+void GamePlayer::OrderTroops(nobMilitary* goal, std::array<unsigned, NUM_SOLDIER_RANKS> counts, unsigned max) const
 {
     // Solange Lagerhäuser nach Soldaten absuchen, bis entweder keins mehr übrig ist oder alle Soldaten bestellt sind
     nobBaseWarehouse* wh;
+    int sum = 0;
```
Flamefire, line 1180: That's never negative as its a sum of unsigneds:
```suggestion
    unsigned sum = 0;
```

libs/s25main/GamePlayer.h
```diff
@@ -198,7 +198,7 @@ class GamePlayer : public GamePlayerInfo
     /// Are these players allied? (-> Teamview, attack support, ...)
     bool IsAlly(unsigned char playerId) const;
     /// Truppen bestellen
-    void OrderTroops(nobMilitary* goal, unsigned count, bool ignoresettingsendweakfirst = false) const;
+    void OrderTroops(nobMilitary* goal, std::array<unsigned, NUM_SOLDIER_RANKS> counts, unsigned max) const;
```
Flamefire, line 201: What is the purpose of `max`? Can you enhance the docstring clarifying what this is supposed to do exactly, e.g. order the given amount of (additional) soldiers and what that `max` is, please

libs/s25main/ai/aijh/AIPlayerJH.cpp
```diff
@@ -1369,9 +1369,17 @@ void AIPlayerJH::HandleBorderChanged(const MapPoint pt)
...
         {
-            aii.SetCoinsAllowed(pt, true);
+            if(mil->IsGoldDisabled())
+                aii.SetCoinsAllowed(pt, true);
+
+            // Fill up with soldiers
+            for(unsigned rank = 0; rank < NUM_SOLDIER_RANKS; ++rank)
+            {
+                if(mil->GetDesiredTroops(rank) != mil->GetMaxTroopsCt())
+                    aii.SetDesiredTroops(mil->GetPos(), rank, mil->GetMaxTroopsCt());
```
Flamefire, line 1381: What is this supposed to do? Doesn't this set the desired troops such that each rank should be there fully filling the building up? I.e. wouldn't this request for a building with capacity for 2 soldiers, 2 soldiers of each rank, i.e. 2*NUM_SOLDIER_RANKS soldiers overfilling the building?

Or is "desired_troops" rather a limit on how many soldiers of that rank are desired? Maybe a "max" in the name would help then.

libs/s25main/ai/aijh/AIPlayerJH.cpp
```diff
@@ -1442,11 +1457,14 @@ void AIPlayerJH::MilUpgradeOptim()
...
                 aii.SetCoinsAllowed(milBld->GetPos(), true);
             }
-            if(milBld->HasMaxRankSoldier()) // has max rank soldier? send it/them out!
-                aii.SendSoldiersHome(milBld->GetPos());
-            if(SoldierAvailable(0)
-               && milBld->GetNumTroops() < milBld->GetMaxTroopsCt()) // building not full and privates in a warehouse?
-                aii.OrderNewSoldiers(milBld->GetPos());              // order new!
+            // Keep 0 generals, 1 of each other rank and fill the rest with privates
+            unsigned rank = 0;
+            for(unsigned count : {milBld->GetMaxTroopsCt(), 1u, 1u, 1u, 0u})
```
Flamefire, line 1462: I think you should use an array of size `NUM_SOLDIER_RANKS` here and take also the maximum rank into account (can be limited by an addon)

libs/s25main/buildings/nobBaseWarehouse.h
```diff
@@ -247,7 +247,7 @@ class nobBaseWarehouse : public nobBaseMilitary, public DataChangedObservable
                + GetNumRealFigures(Job::Sergeant) + GetNumRealFigures(Job::Officer) + GetNumRealFigures(Job::General);
     }
     /// Bestellt Soldaten
-    void OrderTroops(nobMilitary* goal, unsigned count, bool ignoresettingsendweakfirst = false);
+    void OrderTroops(nobMilitary* goal, std::array<unsigned, NUM_SOLDIER_RANKS> &counts, unsigned &max);
```
Flamefire, line 250: Please provide a docstring what this is supposed to do especially in regard to the (IMO unusual) reference parameters

libs/s25main/buildings/nobMilitary.cpp
```diff
@@ -81,6 +82,22 @@ size_t nobMilitary::GetTotalSoldiers() const
...
+        ++counts[troop->GetRank()];
+    for(auto& troop : ordered_troops)
+        ++counts[troop->GetRank()];
+    for(auto& troop : troops_on_mission)
+        ++counts[troop->GetRank()];
+    if(defender_ && (defender_->IsWaitingAtFlag() || defender_->IsFightingAtFlag()))
+        ++counts[defender_->GetRank()];
+    for(auto& troop : far_away_capturers)
+        ++counts[troop->GetRank()];
+    return counts;
```
Flamefire, line 98: Not fully sure here. Are you sure that the additional conditions for `defender_` are required and it is not in `troops_on_mission` or `far_away_capturers`?

Just double checking.

libs/s25main/buildings/nobMilitary.cpp
```diff
@@ -136,6 +153,8 @@ void nobMilitary::Serialize(SerializedGameData& sgd) const
     sgd.PushObjectContainer(ordered_coins, true);
     sgd.PushObjectContainer(troops, true);
     sgd.PushObjectContainer(far_away_capturers, true);
+    for(unsigned n : desired_troops)
+        sgd.PushUnsignedInt(n);
```
Flamefire, line 157:
```suggestion
    sgd.PushContainer(desired_troops);
```

libs/s25main/buildings/nobMilitary.cpp
```diff
@@ -149,6 +168,15 @@ nobMilitary::nobMilitary(SerializedGameData& sgd, const unsigned obj_id)
     sgd.PopObjectContainer(troops, GO_Type::NofPassivesoldier);
     sgd.PopObjectContainer(far_away_capturers, GO_Type::NofAttacker);
 
+    if(sgd.GetGameDataVersion() < 10)
+    {
+        desired_troops.fill(GetMaxTroopsCt());
+    } else
+    {
+        for(unsigned i = 0; i < NUM_SOLDIER_RANKS; ++i)
+            desired_troops[i] = sgd.PopUnsignedInt();
```
Flamefire, line 177:
```suggestion
        sgd.PopContainer(desired_troops);
```

libs/s25main/buildings/nobMilitary.h
```diff
@@ -71,6 +71,7 @@ class nobMilitary : public nobBaseMilitary
     bool is_regulating_troops;
     /// Soldatenbesatzung
     OwnedSortedTroops troops;
+    std::array<unsigned, NUM_SOLDIER_RANKS> desired_troops;
```
Flamefire, line 74: Fixed size required for `sgd.PushContainer(desired_troops);`:

E.g.:
```suggestion
    std::array<uint8_t, NUM_SOLDIER_RANKS> desired_troops;
```

libs/s25main/buildings/nobMilitary.h
```diff
@@ -138,6 +140,10 @@ class nobMilitary : public nobBaseMilitary
     auto GetTroops() const { return helpers::nonNullPtrSpan(troops); }
     bool IsInTroops(const nofPassiveSoldier& soldier) const;
 
+    const unsigned* GetDesiredTroopsPointer(unsigned rank) const { return &desired_troops[rank]; }
+    unsigned GetDesiredTroops(const unsigned rank) const { return desired_troops[rank]; }
+    void SetDesiredTroops(const unsigned rank, const unsigned count);
```
Flamefire, line 145: `const not required on declaration:
```suggestion
    void SetDesiredTroops(unsigned rank, unsigned count);
```

Also please add a docstring, especially what "desired" means here exactly

libs/s25main/buildings/nobMilitary.h
```diff
@@ -138,6 +140,10 @@ class nobMilitary : public nobBaseMilitary
     auto GetTroops() const { return helpers::nonNullPtrSpan(troops); }
     bool IsInTroops(const nofPassiveSoldier& soldier) const;
 
+    const unsigned* GetDesiredTroopsPointer(unsigned rank) const { return &desired_troops[rank]; }
```
Flamefire, line 143: Not really a big fan of handing out a pointer. Maybe use `GetDesiredTroops` to update the UI texts periodically/before draw instead?

libs/s25main/ingameWindows/iwMilitaryBuilding.cpp
```diff
@@ -32,6 +32,13 @@ iwMilitaryBuilding::iwMilitaryBuilding(GameWorldView& gwv, GameCommandFactory& g
                    _(BUILDING_NAMES[building->GetBuildingType()]), LOADER.GetImageN("resource", 41)),
       gwv(gwv), gcFactory(gcFactory), building(building)
 {
+    unsigned offset = 0;
```
Flamefire, line 35: Maybe a better name: 
```suggestion
    unsigned btOffset = 0;
```

so that it is clear it refers to the buttons, especially as it is declared a bit away from its use

libs/s25main/ingameWindows/iwMilitaryBuilding.cpp
```diff
@@ -204,9 +238,41 @@ void iwMilitaryBuilding::Msg_ButtonClick(const unsigned ctrl_id)
...
+            else
+            {
+                // Send all soldiers home except one
+                gcFactory.SetDesiredTroops(building->GetPos(), 0, 1);
+                for(unsigned rank = 1; rank < NUM_SOLDIER_RANKS; ++rank)
+                    gcFactory.SetDesiredTroops(building->GetPos(), rank, 0);
+            }
+        }
+        break;
+        case 11: case 16: case 21: case 26: case 31:
```
Flamefire, line 262: Magic numbers like that might not be a good idea.

I can imagine:
- `if(ctrl_id>10)` --> SetDesiredTroops
- `ctrl_id % 5 == 0` --> plus, else minus (with an assert `ctrl_id % 5 == 1`)

libs/s25main/ingameWindows/iwMilitaryBuilding.cpp
```diff
@@ -204,9 +238,41 @@ void iwMilitaryBuilding::Msg_ButtonClick(const unsigned ctrl_id)
...
+        {
+            unsigned rank = (ctrl_id - 11) / 5;
```
Flamefire, line 264: Style: If you don't intend to change the value mark the variable `const`, like:
```suggestion
            const unsigned rank = (ctrl_id - 11) / 5;
```

Flamefire (changes requested):
Well done, thx!

I added a few remarks especially to clarify what "desired soldiers" exactly means. Neither from the naming nor from the docstrings or code it is clear to me what happens e.g. when you have more desired soldiers than can be stationed or when you have less than what the military control switch states.

libs/s25main/buildings/nobMilitary.cpp
ii8: Fairly sure, `defender_` always comes out of the building(`troops`) from which he is removed in `ProvideDefender` and not added anywhere else.

libs/s25main/buildings/nobMilitary.h
ii8: This is for use with `Window::AddVarText`, I think this is how it is intended to be used currently. If `AddVarText` should be changed could that be a separate PR?

libs/s25main/buildings/nobMilitary.cpp
Flamefire: Ok, but IIRC as soon as he is removed from `troops` it is set as `defender_`, so I don't understand the ` && (defender_->IsWaitingAtFlag() || defender_->IsFightingAtFlag())` condition part. Wouldn't that leave a (short time) where the defender is wrongly not counted?

ii8: That's to match the number returned by `GetTotalSoldiers` which calculates it like this. I'm struggling to figure out why it does that though, I don't understand it either.

libs/s25main/GamePlayer.cpp
ii8: I can't make this const like this it would get modified. It's just a few bytes, it's fine to copy it I think.

libs/s25main/buildings/nobMilitary.cpp
ii8: I can't find this function, I used the one from `helpers` instead.

libs/s25main/buildings/nobMilitary.cpp
ii8: This function doesn't work for fixed size containers, used the one from `helpers` again.

libs/s25main/buildings/nobMilitary.cpp
ii8: See here: https://github.com/Return-To-The-Roots/s25client/commit/a8fbe91a23edd08151943f43f3e6135fbf5bae9b

libs/s25main/ingameWindows/iwMilitaryBuilding.cpp
```diff
@@ -204,11 +243,48 @@ void iwMilitaryBuilding::Msg_ButtonClick(const unsigned ctrl_id)
...
+                for(unsigned rank = 1; rank < NUM_SOLDIER_RANKS; ++rank)
+                    gcFactory.SetTroopLimit(building->GetPos(), rank, 0);
+            }
         }
         break;
+        default:
+        {
+            if(ctrl_id > 10)
+            {
+                if(ctrl_id % NUM_SOLDIER_RANKS == 4)
```
Flamefire, line 271: This isn't correct. The IDs for increasing the rank are `11 + (5 * i)`, i.e. not related to `NUM_SOLDIER_RANKS` although that is `5` (currently) by coincidence.

So what we have:
- MINUS button: `11 + (5 * rank) = 10 + 1 + (5 * rank)`
- PLUS button `14 + (5 * rank) = 10 + 4 + (5 * rank)`

So the `ctrl_id % x` is the same `x` as in `rank = (ctrl_id - y) / x`, i.e. 5. The modulo should just strip off the `(5 * rank)` and coincidently the 10 leaving only the 1 and 4.
Hence:
```suggestion
                if(ctrl_id % 5 == 4)
```

Flamefire (changes requested):
Minor change remaining.
Also you'd need to regenerate test replays for the tests `Play200kReplay` & `PlaySeaReplay`.

Instructions are there, so basically:
- start a game with the mentioned map and players
- jump forward the amount of GFs
- End the game
- Copy the created replay to the expected location with the name replacing the current one(s)

tests/testData/200kGFs.rpl:
```
    // Map: Big Slaughter v2
    // 7 x Hard KI
    // 2 KIs each in Teams 1-3, 1 in Team 4
    // 200k GFs
```

tests/testData/SeaMap300kGfs.rpl:
```
    // Map: Island by Island
    // 2 x Hard KI + Player KI
    // No teams, Sea attacks enabled, ships fast
    // 300k GFs run (+ a bit)
```

ii8:
How do I jump forward the GFs? And does it matter if I accidentally go a few too many because I don't click quit fast enough?
Also, somehow `make test` is not failing currently, I expect it should, does it not run these tests?

Flamefire:
> ~Somehow `make test` is not failing currently, I expect it should, does it not run these tests?~ Only runs them in release builds.

Yes those 2 tests are so slow that we run them only in release IIRC.

Flamefire (dismissed):
Looks good to me, thanks!

ii8:
Do I have to make the auto-nitpicker happy or something? Can I automatically apply the changes it suggests somehow?

Flamefire:
For the formatting: If you have clang-format-10 installed the CMake process will create a "clangFormat" target which you can "build". That will run the formatting over all files changed since the last format run so this is kinda automatic. Easiest is to get s astic clang-format from e.g. https://github.com/muttleyxd/clang-tools-static-binaries/releases. Download the corresponding file, e.g. `clang-format-10_windows-amd64.exe` (assuming Windows), rename it to `clang-format-10.exe` and put it somewhere that is in `PATH` then CMake will find it. Or put it anywhere and set `ClangFormat_BINARY` to the full path of it in CMake.

For clang-tidy we don't have anything like it but shouldn't be much to do there.

I know that static analysis stuff looks like nitpicking and often doesn't help much but it proved to be valuable and reduces maintenance effort, so we need it.

As for the GCC-12 job failure: Just ignore it, should be fixed once @Flow86 merges #1580

Flamefire (approved):

ii8:
@Flamefire I don't think that test failure is related to my changes. Locally the tests pass most of the time, only sometimes does the network test fail like this.

ii8:
@Flamefire Do I need to do something?

Flamefire:
Not from my side. Failure on Jenkins seems to be a timeout. Restarted that and will merge as soon as that finished.

Issue #1566, opened 2023-03-19
vsync = 1 causes 1 FPS


Spikeone:
In the latest nightly people only get 1 FPS if vsync is enabled (which seems to be the default case often). Setting the value in the config to 0 seems to fix the problem, although it sometimes enables itself.

This maybe is related to the fix: https://github.com/Return-To-The-Roots/s25client/pull/1557

Xellzul:
Forcing refresh rate in-game (from dynamic to fixed number e.g. 60) seems to fix it also, although it doesn't seem persistent.

PR #1577, opened 2023-04-10
FIX issue-1566 - framerate stuck to 1

Xellzul:
Fixes https://github.com/Return-To-The-Roots/s25client/issues/1566
Issue was when vsync is set to number above 0, it is read as bool -> 1 and that is used as FPS limit

### DIFF
--- a/libs/s25main/Settings.cpp
+++ b/libs/s25main/Settings.cpp
@@ -224,7 +224,7 @@ void Settings::Load()
         video.fullscreenSize.width = iniVideo->getIntValue("fullscreen_width");
         video.fullscreenSize.height = iniVideo->getIntValue("fullscreen_height");
         video.fullscreen = iniVideo->getBoolValue("fullscreen");
-        video.framerate = iniVideo->getIntValue("framerate");
+        video.framerate = iniVideo->getValue("framerate", 0);
         video.vbo = iniVideo->getBoolValue("vbo");
         video.shared_textures = iniVideo->getBoolValue("shared_textures");
         // };
--- a/libs/s25main/Settings.cpp
+++ b/libs/s25main/Settings.cpp
@@ -90,7 +90,7 @@ void Settings::LoadDefaults()
         video.windowedSize = video.fullscreenSize = VideoMode(800, 600);
         video.fullscreen = false;
     }
-    video.vsync = 0;
+    video.framerate = 0; // Special value for HW vsync
     video.vbo = true;
     video.shared_textures = true;
     // }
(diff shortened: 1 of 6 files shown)


libs/s25main/Settings.cpp
```diff
@@ -90,7 +90,7 @@ void Settings::LoadDefaults()
         video.windowedSize = video.fullscreenSize = VideoMode(800, 600);
         video.fullscreen = false;
     }
-    video.vsync = 0;
+    video.framerate = 0; // default vsync
```
Flamefire, line 93: Better:
```suggestion
    video.framerate = 0; // Special value for HW vsync
```

libs/s25main/desktops/Desktop.cpp
```diff
@@ -26,11 +26,7 @@ Desktop::Desktop(glArchivItem_Bitmap* background)
 {
     SetScale(true);
     SetFpsDisplay(true);
-    // By default limit the maximum frame rate to 60 FPS
-    if(SETTINGS.video.vsync < 0)
-        VIDEODRIVER.setTargetFramerate(60);
-    else
-        VIDEODRIVER.setTargetFramerate(SETTINGS.video.vsync);
+    VIDEODRIVER.setTargetFramerate(SETTINGS.video.framerate);
```
Flamefire, line 29: This is intentional: For all non-ingame purposes "unlimitted" should mean "60" as otherwise you'll get insane framerates for no benefit.
So please revert this part of the change

libs/s25main/desktops/dskOptions.cpp
```diff
@@ -275,9 +275,10 @@ dskOptions::dskOptions() : Desktop(LOADER.GetImageN("setup013", 0))
     optiongroup->AddTextButton(ID_btOff, DrawPoint(280, 125), Extent(190, 22), TextureColor::Grey, _("Windowed"),
                                NormalFont);
 
-    // "VSync"
-    groupGrafik->AddText(ID_txtVSync, DrawPoint(80, 180), _("Limit Framerate:"), COLOR_YELLOW, FontStyle{}, NormalFont);
-    groupGrafik->AddComboBox(ID_cbVSync, DrawPoint(280, 175), Extent(390, 22), TextureColor::Grey, NormalFont, 150);
+    // "framerate"
```
Flamefire, line 278: Can be removed:
```suggestion
```

Flamefire (changes requested):
Good catch! This was actually introduced by me with #1550

Hence the important part is the `iniVideo->getIntValue("framerate");` change but I agree with the renaming. Just some smaller annotations.

libs/s25main/desktops/Desktop.cpp
Xellzul: Oh, you are right, I misunderstood the point of this code

Xellzul: reverted

libs/s25main/Settings.cpp
```diff
@@ -224,7 +224,7 @@ void Settings::Load()
         video.fullscreenSize.width = iniVideo->getIntValue("fullscreen_width");
         video.fullscreenSize.height = iniVideo->getIntValue("fullscreen_height");
         video.fullscreen = iniVideo->getBoolValue("fullscreen");
-        video.vsync = iniVideo->getBoolValue("vsync");
+        video.framerate = iniVideo->getIntValue("framerate");
```
Flamefire, line 227: This fails for existing configs using the old name, better provide a default:

```suggestion
        video.framerate = iniVideo->getValue("framerate", 0);
```

Flamefire (changes requested):

PR #1579, opened 2023-04-10
Scaling font and text fields

Xellzul:
So I'd like to resolve the lack of HDPI support - ISSUE-533

Only thing that is scaled for now is ctrlText (bottom on the screenshots), the rest of the text based controls should be straight forward to update. 

The approach with adding `ScaleHeigthIf` was easiest, however it is not very elegant, if you have better idea I am all for it

Another option would be for user-selectable scaling - 100% 125% ...
 
Example before at 5120x2160
image
After at 5120x2160
image

Before at 800x600 
image
After at 800x600
image

### DIFF
--- a/libs/s25main/RescaleWindowProp.h
+++ b/libs/s25main/RescaleWindowProp.h
@@ -4,13 +4,16 @@
 
 #pragma once
 
+#define DefaultWidth 800
+#define DefaultHeigth 600
+
 /// Functor or static method to scale a window propertie from the real coordinates
 /// (relative to 800x600 resolution) to new coordinates given a size
 struct ScaleWindowPropUp
 {
     Extent size;
     ScaleWindowPropUp(const Extent& size) : size(size) {}
-
+    
     template<typename T_Pt>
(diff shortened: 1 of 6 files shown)


falbrechtskirchinger:
@Xellzul I've implemented a different solution in PR #1594 but don't have access to a HighDPI display. Would you mind taking a look and maybe giving it a try?

Xellzul:
@falbrechtskirchinger 
I would love to test it, however it might take a week or so.
You can test it even with low dpi display using
AMD: https://www.amd.com/en/support/kb/faq/dh-010
Or 
NVIDIA: https://www.nvidia.com/en-us/geforce/technologies/dsr/technology/

falbrechtskirchinger:
> I would love to test it, however it might take a week or so.

@Xellzul No problem at all. I plan on making a few more changes anyway.

> You can test it even with low dpi display using
> NVIDIA: https://www.nvidia.com/en-us/geforce/technologies/dsr/technology/

Unfortunately, that isn't available on Linux AFAICT. Thanks for the tip, though. There might be another solution.

PR #1580, opened 2023-04-12
GHA: Test latest compilers & Boost 1.69

Flamefire:
Followup to https://github.com/Return-To-The-Roots/s25client/pull/1558, see discussion at https://github.com/Return-To-The-Roots/s25client/pull/1558/files#r1161563495

Combination of #1575 & #1576, hence closes #1575 & closes #1576

Add GCC 12/11 to test and build with Clang 16 and its default C++ standard. So both latest GCC & Clang use the minimum required C++ standard and their default.

For GCC 12 we need Boost 1.82 which will be released soon.

@Flow86 This needs manual merge and adjustment of the required checks afterwards.


Flow86:
superseded by #1581

Issue #1581, opened 2023-04-17
Clean Up Toolchains and GHA


Flow86:
To ease up matrix builds, I'll checked all favorited platforms and their compilers / boost versions:

| OS                   | Github runner? | clang/gcc versions         | boost version |
|----------------------|----------------|----------------------------|---------------|
| MacOS                | No             | clang 14.0.0 (darwin 21.6) | None (*)      |
| macos-12             | Yes            | clang 14.0.0 (darwin 21.6) | 1.81          |
| Debian Bullseye      | No             | clang 9, 11, 13            | 1.74          |
|                      |                | gcc 9, 10                  |               |
| Ubuntu Focal (20.04) | No             | clang 11, 13               | 1.69, 1.74    |
|                      |                | gcc 9, 10                  |               |
| ubuntu-20.04         | Yes            | clang 10, 11, 12, 16*         | 1.69, 1.74    |
|                      |                | gcc 9, 10                  |               |
| Ubuntu Focal (22.04) | No             | clang 11, 12, 13, 14, 15   | 1.74          |
|                      |                | gcc 9, 10, 11, 12          |               |
| ubuntu-22.04         | Yes            | clang 12, 13, 14      | 1.74          |
|                      |                | gcc 9, 10, 11, 12          |               |
| Windows              | No             | mingw gcc 10                  | None (*)      |

*) has to be manually installed/packaged

Clang / G++ Default C++ versions

| Compiler | C++ Version | 
|----------|-------------|
| gcc 9    | 14          |
| gcc 10   | 14          |
| gcc 11   | 17          |
| gcc 12   | 17          |
| clang 9  | 14          |
| clang 11 | 14          |
| clang 12 | 14          |
| clang 13 | 14          |
| clang 14 | 14          |
| clang 15 | 14          |
| clang 16 | 17          |

So the resulting runner/GHA matrix is:

| Runner / OS     | compiler   | boost version | c++ version | remarks     |
|-----------------|------------|---------------|-------------|-------------|
| macos-12        | clang 14   | 1.81.0        | default(14) |             |
| ubuntu-20.04    | gcc 9 *oldest*      | 1.69.0        | default(14) | coverage, warnings |
|                 | gcc 10     | 1.69.0        | default(14) | also release |
|                 | clang 10 *oldest*   | 1.69.0        | default(14) | also release |
|                 | ~clang 11~   | 1.69.0        | default(14) | nothing of value |
[...]

PR #1582, opened 2023-04-17
sort matrix according to issue description

Flow86:
see #1581

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -13,6 +13,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
+  # Default versions from Jenkins Toolchain
   CMAKE_VERSION: 3.26.0
   BOOST_VERSION: 1.81.0
   ADDITIONAL_CMAKE_FLAGS: -DRTTR_ENABLE_BENCHMARKS=ON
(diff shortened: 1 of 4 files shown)


.github/workflows/unit-tests.yml
```diff
@@ -121,7 +121,6 @@ jobs:
       - name: Install dependencies
         if: "!startsWith(runner.os, 'macos')"
         run: |
-          # Minimum requirements: CMake 3.9, Boost 1.69
```
Flamefire, line 124: IMO this is kind of important information. We currently don't have a test for CMake 3.9 at all anymore. Hence I'd rather make that the default in the global env. Same for Boost: 1.69 is the minimum so why use another version as the default?

.github/workflows/unit-tests.yml
```diff
@@ -23,21 +23,21 @@ jobs:
       matrix:
         include:
           # MacOSX
-          - { compiler: clang,    os: macos-12,     buildType: Debug, cxxStandard: 14 }
+          - { compiler: clang,    os: macos-12,     buildType: Debug }
```
Flamefire, line 26: IMO it makes sense to fix/set the Boost version here as it is not arbitrary but depends on the choice of macos version. See also the debug output of https://github.com/Return-To-The-Roots/s25client/pull/1580/files#diff-cd0effce06f425599e96952f4d5f684641a683f46d9909e7c74a9f8b2f03ccd3R113-R114 which may help when something changes. So I'd include that (debug output and explicit version here)

.github/workflows/unit-tests.yml
```diff
@@ -23,21 +23,21 @@ jobs:
...
           # Linux GCC
           # GCC 9 is known to show a few warnings that GCC 10 has "fixed", make sure this doesn't happen for us
-          - { compiler: gcc-9,    os: ubuntu-20.04, buildType: Debug }
-          - { compiler: gcc-10,   os: ubuntu-20.04, buildType: Debug, cxxStandard: 14, coverage: true }
-          - { compiler: gcc-10,   os: ubuntu-20.04, buildType: Debug, cxxStandard: 17 }
-          - { compiler: gcc-10,   os: ubuntu-20.04, buildType: Release, cxxStandard: 14 }
+          - { compiler: gcc-9,    os: ubuntu-20.04, buildType: Debug, boostVersion: 1.69.0 }
+          - { compiler: gcc-10,   os: ubuntu-20.04, buildType: Debug, boostVersion: 1.69.0, coverage: true }
+          - { compiler: gcc-10,   os: ubuntu-20.04, buildType: Release }
+          - { compiler: gcc-11,   os: ubuntu-22.04, buildType: Debug, boostVersion: 1.74.0 }
```
Flamefire, line 32: I think it makes sense noting that GCC < 11 is C++14 and GCC >= 11 is C++17 at least as a comment so it is clear we test both. Similar for clang. You already gathered that info in the issue but that will be lost/hard to find when this is merged. Maybe even make it explicit by using `cxxStandard: 14` for the gcc-10 Release build and `cxxStandard: 17` for the gcc-12 release build.

.github/workflows/unit-tests.yml
```diff
@@ -23,21 +23,21 @@ jobs:
...
           # Linux clang
-          - { compiler: clang-11, os: ubuntu-20.04, buildType: Debug, cxxStandard: 14, externalSanitizer: true }
-          - { compiler: clang-11, os: ubuntu-20.04, buildType: Debug, cxxStandard: 17 }
-          - { compiler: clang-11, os: ubuntu-20.04, buildType: Release, cxxStandard: 14 }
+          - { compiler: clang-10, os: ubuntu-20.04, buildType: Debug, boostVersion: 1.69.0, externalSanitizer: true }
+          - { compiler: clang-10, os: ubuntu-20.04, buildType: Release }
+          - { compiler: clang-14, os: ubuntu-22.04, buildType: Debug, boostVersion: 1.74.0 }
```
Flamefire, line 36: I'd add a comment here why Boost 1.74 is used

.github/workflows/unit-tests.yml
```diff
@@ -23,21 +23,21 @@ jobs:
...
           # Linux Latest GCC
-          - { compiler: gcc-12,   os: ubuntu-22.04, buildType: Debug, cxxStandard: 14, coverage: true }
+          - { compiler: gcc-12,   os: ubuntu-22.04, buildType: Debug, boostVersion: 1.82.0 }
```
Flamefire, line 38: Same here, e.g.:
```suggestion
          - { compiler: gcc-12,   os: ubuntu-22.04, buildType: Debug, boostVersion: 1.82.0 } # GCC 12 detects an issue fixed in Boost 1.82
```

.github/workflows/unit-tests.yml
```diff
@@ -23,21 +23,21 @@ jobs:
...
           # Linux Latest clang
-          - { compiler: clang-16, os: ubuntu-20.04, buildType: Debug, cxxStandard: 14, externalSanitizer: true }
+          - { compiler: clang-16, os: ubuntu-20.04, buildType: Debug }
```
Flamefire, line 40: I'd keep the sanitizer enabled: So we have it for the oldest and newest supported clang

Flamefire (changes requested):
Looks good to me, just missing a bit of information in the file WHY a specific version was chosen, especially the "global" Boost/CMake version look very random.

See the inline comments.

.github/workflows/unit-tests.yml
Flow86: thats exactly one point, we should shift the minimum up to get rid of all the workarounds. 

Stock cmake from debian bullseye is 3.18.4 - we should probably make this the new minimum

.github/workflows/unit-tests.yml
Flow86: oh yes, that was by accident

.github/workflows/unit-tests.yml
Flow86: I wrote it down now in the comments

.github/workflows/unit-tests.yml
Flow86: I wrote it down now in the comments

.github/workflows/unit-tests.yml
Flow86: I wrote it down now in the comments

.github/workflows/unit-tests.yml
Flow86: I wrote it down now in the comments

.github/workflows/unit-tests.yml
```diff
@@ -22,39 +23,76 @@ jobs:
...
+          - { compiler: clang-10, os: ubuntu-20.04, type: Debug, cmake: 3.16.3, boost: 1.69.0, externalSanitizer: true }
+          #
+          #   Default compiler for Ubuntu 20.04
+          #   Use (compiler) default C++ standard (14)
+          #   Use default cmake
+          #   Use default boost
+          - { compiler: gcc-10,   os: ubuntu-20.04, type: Debug, coverage: true }
+          #
+          #   Default compilers for Ubuntu 22.04
+          #   Use C++ standard (17, default for gcc)
```
Flamefire, line 49: C&P made this hard to understand, maybe: 
```suggestion
          #   Use C++17 standard (default for gcc-11)
```

.github/workflows/unit-tests.yml
```diff
@@ -22,39 +23,76 @@ jobs:
...
 
     steps:
       - run: echo "DEPS_DIR=${{runner.temp}}/.cache" >> $GITHUB_ENV
-      - run: echo "CMAKE_VERSION=${{matrix.cmakeVersion}}" >> $GITHUB_ENV
-        if: matrix.cmakeVersion
-      - run: echo "BOOST_VERSION=${{matrix.boostVersion}}" >> $GITHUB_ENV
-        if: matrix.boostVersion
-      - run: echo "ADDITIONAL_CMAKE_FLAGS=${ADDITIONAL_CMAKE_FLAGS} -DRTTR_EXTERNAL_BUILD_TESTING=ON -DRTTR_ENABLE_SANITIZERS=ON" >> $GITHUB_ENV
+
+      - name: "Set BUILD_TYPE"
```
Flamefire, line 68: IMO giving this names is overly verbose but fine for me. At least get rid of the quotes (here and below) as in `name: Set C++ standard`:
```suggestion
      - name: Set build type (Release/Debug/...)
```

.github/workflows/unit-tests.yml
```diff
@@ -22,39 +23,76 @@ jobs:
...
+
+      - name: Enable coverage collection
+        if: matrix.coverage
+        run: echo "ADDITIONAL_CMAKE_FLAGS=${ADDITIONAL_CMAKE_FLAGS} -DRTTR_ENABLE_COVERAGE=ON" >> $GITHUB_ENV
+
+      - name: "Enable external sanitizer"
         if: matrix.externalSanitizer
+        run: echo "ADDITIONAL_CMAKE_FLAGS=${ADDITIONAL_CMAKE_FLAGS} -DRTTR_EXTERNAL_BUILD_TESTING=ON -DRTTR_ENABLE_SANITIZERS=ON" >> $GITHUB_ENV
 
+      # Coverage collection requires access to history to find merge-base
```
Flamefire, line 91: Why move this above the part that does **not** apply to coverage? Then I'd swap the 2 so this looks more logical and maybe add a comment like  `fetch-depth: 0  # Faster` or so while at it

.github/workflows/unit-tests.yml
```diff
@@ -109,52 +149,48 @@ jobs:
...
-
-      - name: Set C++ standard
-        if: matrix.cxxStandard
-        run: echo "ADDITIONAL_CMAKE_FLAGS=${ADDITIONAL_CMAKE_FLAGS} -DCMAKE_CXX_STANDARD=${{matrix.cxxStandard}}" >> $GITHUB_ENV
-
       - name: Setup environment
-        run: |
-          echo "BUILD_TYPE=${{matrix.buildType}}" >> $GITHUB_ENV
-          echo "TRAVIS_BUILD_DIR=$GITHUB_WORKSPACE" >> $GITHUB_ENV
+        run: echo "TRAVIS_BUILD_DIR=$GITHUB_WORKSPACE" >> $GITHUB_ENV
```
Flamefire, line 180: As we'll never use travis again, maybe rather change that in the script(s) and get rid of this step which is now only for `$TRAVIS_BUILD_DIR` and rename `tools/ci/travisBuild.sh` to `githubBuild.sh` or just `build.sh`?

This is an ugly clutch anyway the build script should have been able to handle either variable.

libs/s25main/Debug.cpp
```diff
@@ -62,16 +62,16 @@ bool captureBacktrace(std::vector<void*>& stacktrace, LPCONTEXT ctx = nullptr) n
...
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wcast-function-type"
-    // error: cast between incompatible function types from
-    // 'FARPROC' {aka 'int (__attribute__((stdcall)) *)()'}
-    // to
-    // 'RtlCaptureContextType' {aka 'void (__attribute__((stdcall)) *)(CONTEXT*)'}
-    // [-Werror=cast-function-type]
-    // and so on
-#endif
+#            if __GNUC__ >= 9
```
Flamefire, line 65: I do agree this looks odd/ugly here but serves the point to highlight that it is in multiple conditions. And e.g. https://github.com/Return-To-The-Roots/s25client/pull/1582/files#diff-e93d2e7c03a97a61287ef951276877e6d0cacac56ce2e0daecc3c935bae907eaR24-R35 looks much better due to this.

I'm wondering if we couldn't use Boost for this (backtrace collection) anyway or at least add a wrapper around `GetProcAddress` to do the casting with the warning disabled or use Boost.DLL which IIRC can do that too.

But that's another PR...

.github/workflows/unit-tests.yml
Flamefire: Ubuntu 20 having CMake 3.16 is a reasonable argument to having that as the minimum as IIRC Ubuntu 18 is EOL. Maybe mention that somewhere but we shouldn't increase the required version if there isn't a benefit to not offend/inconvenience potential contributors.

Flamefire (commented):
This has gotten a bit verbose increasing vertical spacing a lot. But whatever. A few remarks inline

.github/workflows/unit-tests.yml
Flow86: so that the comment is above the block

.github/workflows/unit-tests.yml
Flow86: oh there were some with quotes and some without. Ive removed it

.github/workflows/unit-tests.yml
Flow86: good point

.github/workflows/unit-tests.yml
Flow86: I don't know if the script is used at another place, so I'll leave it like it is for now

PR #1586, opened 2023-04-26
#1509 Do not cache texture for animated objects (like open gate)

### LINKED ISSUE #1509: Opened gate not animated
A gate added by ```rttr:GetWorld():AddStaticObject(70, 72, 561, 0xFFFF, 2)``` is not animated anymore. Guess that happened when fixing the crash related to non player colors or something alike.

ottml:
Fixes #1509

### DIFF
--- a/libs/s25main/nodeObjs/noStaticObject.cpp
+++ b/libs/s25main/nodeObjs/noStaticObject.cpp
@@ -75,7 +75,7 @@ BlockingManner noStaticObject::GetBM() const
  */
 void noStaticObject::Draw(DrawPoint drawPt)
 {
-    if(!textures.bmp)
+    if(IsAnimated() || !textures.bmp)
     {
         textures = getTextures(file, id);
         RTTR_Assert(textures.bmp);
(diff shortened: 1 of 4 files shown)


tests/s25Main/lua/testLua.cpp
```diff
@@ -588,6 +588,7 @@ BOOST_AUTO_TEST_CASE(World)
     BOOST_TEST(obj2->GetItemID() == 501u);
     BOOST_TEST(obj2->GetItemFile() == 0xFFFFu);
     BOOST_TEST(obj2->GetSize() == 1u);
+    BOOST_TEST(obj2->IsAnimated() == false);
```
Flamefire, line 591: Please don't use `== false` or `== true` but rather e.g.:
```suggestion
    BOOST_TEST(!obj2->IsAnimated());
```

libs/s25main/nodeObjs/noStaticObject.cpp
```diff
@@ -111,3 +111,8 @@ noStaticObject::Textures noStaticObject::getTextures(unsigned short file, unsign
 
     return textures;
 }
+
+bool noStaticObject::IsAnimated() const
+{
+    return file == 0xFFFF && id == 561;
```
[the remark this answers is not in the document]
ottml: I have added a new function isOpenGateway and and changed the calling code to use it in the texture and animated function.

libs/s25main/nodeObjs/noStaticObject.h
```diff
@@ -37,8 +37,10 @@ class noStaticObject : public noCoordBase
     void Draw(DrawPoint drawPt) override;
 
     static Textures getTextures(unsigned short file, unsigned short id);
+    bool IsAnimated() const;
```
Flow86, line 40: One small inconsistency: `IsAnimated` vs `isOpenGateway` - I think the `I` should be lower case. but this is very picky, I know :-)

That the only thing left I think. Thank you :-)

ottml: Changed to upper case because all other functions are named this way. (IsOpenGateway)

PR #1594, opened 2023-07-05
Implement GUI scaling for HighDPI displays and accessibility (SDL2 only)

### LINKED ISSUE #533: [wish] some kind of hidpi support
Hey guys,
on hidpi screens the text in the menus (in the main menu as well as ingame) is barely readable.

I don't expect you guys to introduce complete hidpi support for the game. But at least the font size and menu sizes could be made dpi-dependent so that you can read the text properly.

Just a thought.
Thanks.

falbrechtskirchinger:
This PR implements a global scaling of the window content in the SDL2 driver (a stub has been added to the the WinAPI driver).

The window zoom can be adjusted in the options screen by selecting the desired zoom level from a combo box or by holding down Control and scrolling.

The zoom behavior in the menu screens is a little janky due to how the controls are sized. ~~I've not investigated this any further.~~ After looking at #1579 again, I suspect this is related to `RescaleWindowProp`/`ScaleWindowPropUp`. Maybe there's something to be done?

I don't own a HighDPI display myself and have only tested it on setups where `windowSize == renderSize`. Some optional improvements are under consideration for this PR and are listed in the to-do list below.

Fixes #533
Closes #1579

To-do:
- [x] Actual play testing.
- [x] Dynamically generate the list of window zooms and filter to avoid UI elements becoming inaccessible on small(er) screens.
- [x] Auto-select sensible `windowZoom` when a HighDPI setup is detected instead of always defaulting to 100%.
- [X] Add tests.
- [x] ~~Maybe add a `windowZoom` parameter to `CreateScreen()` to avoid unnecessary calls at the start of the game.~~ (An unnecessary micro-optimization.)
- [x] Split out `PointF` using alias into a separate PR. There're a few other places that could be updated but are functionally unrelated to this PR. (Done: #1597)
- [x] Split out rounding functions. (Done: #1599)
- [x] Refactor based on split-out PRs.
- [x] ~~Investigate `RescaleWindowProp`/`ScaleWindowPropUp`.~~ (Current behavior seems fine.)
- [x] Fix soft cursor drifting on zoom change.
- [x] Should the setting be renamed to `(G)UI scale` and `Zoom Game World` be removed and always disabled?
- [x] Remove CI commit. (~~Submit separate PR.~~ #1596)
- [x] Requires #1618, which requires:
    - [x] https://github.com/Return-To-The-Roots/libutil/pull/30
    - [x] https://github.com/Return-To-The-Roots/libsiedler2/pull/24

### DIFF
--- a/extras/videoDrivers/SDL2/VideoSDL2.cpp
+++ b/extras/videoDrivers/SDL2/VideoSDL2.cpp
@@ -352,7 +352,7 @@ bool VideoSDL2::MessageLoop()
                 break;
             }
             case SDL_MOUSEBUTTONDOWN:
-                mouse_xy.pos = Position(ev.button.x, ev.button.y);
+                mouse_xy.pos = getGuiScale().screenToView(Position(ev.button.x, ev.button.y));
 
                 if(/*!mouse_xy.ldown && */ ev.button.button == SDL_BUTTON_LEFT)
                 {
@@ -366,7 +366,7 @@ bool VideoSDL2::MessageLoop()
                 }
                 break;
             case SDL_MOUSEBUTTONUP:
-                mouse_xy.pos = Position(ev.button.x, ev.button.y);
+                mouse_xy.pos = getGuiScale().screenToView(Position(ev.button.x, ev.button.y));
 
                 if(/*mouse_xy.ldown &&*/ ev.button.button == SDL_BUTTON_LEFT)
                 {
@@ -393,7 +393,7 @@ bool VideoSDL2::MessageLoop()
             }
             break;
             case SDL_MOUSEMOTION:
-                mouse_xy.pos = Position(ev.motion.x, ev.motion.y);
+                mouse_xy.pos = getGuiScale().screenToView(Position(ev.motion.x, ev.motion.y));
                 CallBack->Msg_MouseMove(mouse_xy);
                 break;
         }
@@ -433,8 +433,18 @@ OpenGL_Loader_Proc VideoSDL2::GetLoaderFunction() const
 
 void VideoSDL2::SetMousePos(Position pos)
 {
+    const auto screenPos = getGuiScale().viewToScreen(pos);
     mouse_xy.pos = pos;
-    SDL_WarpMouseInWindow(window, pos.x, pos.y);
+    SDL_WarpMouseInWindow(window, screenPos.x, screenPos.y);
+}
+
+void VideoSDL2::setGuiScalePercent(unsigned percent)
+{
+    if(setGuiScaleInternal(percent))
+    {
+        UpdateCurrentSizes();
+        CallBack->WindowResized();
+    }
 }
 
 KeyEvent VideoSDL2::GetModKeyState() const
--- a/extras/videoDrivers/SDL2/VideoSDL2.cpp
+++ b/extras/videoDrivers/SDL2/VideoSDL2.cpp
@@ -443,11 +443,6 @@ void VideoSDL2::SetMousePos(Position pos)
     SDL_WarpMouseInWindow(window, screenPos.x, screenPos.y);
 }
 
-void VideoSDL2::onGuiScaleChanged()
-{
-    UpdateCurrentSizes();
-}
-
 KeyEvent VideoSDL2::GetModKeyState() const
 {
     const SDL_Keymod modifiers = SDL_GetModState();
--- a/extras/videoDrivers/SDL2/VideoSDL2.cpp
+++ b/extras/videoDrivers/SDL2/VideoSDL2.cpp
@@ -128,15 +128,20 @@ bool VideoSDL2::CreateScreen(const std::string& title, const VideoMode& size, bo
     int wndPos = SDL_WINDOWPOS_CENTERED;
 
     const auto requestedSize = fullscreen ? FindClosestVideoMode(size) : size;
+    unsigned commonFlags = SDL_WINDOW_OPENGL;
+
+#if SDL_VERSION_ATLEAST(2, 0, 1)
+    commonFlags |= SDL_WINDOW_ALLOW_HIGHDPI;
+#endif
 
     window = SDL_CreateWindow(title.c_str(), wndPos, wndPos, requestedSize.width, requestedSize.height,
-                              SDL_WINDOW_OPENGL | (fullscreen ? SDL_WINDOW_FULLSCREEN : SDL_WINDOW_RESIZABLE));
+                              commonFlags | (fullscreen ? SDL_WINDOW_FULLSCREEN : SDL_WINDOW_RESIZABLE));
 
     // Fallback to non-fullscreen
     if(!window && fullscreen)
     {
         window = SDL_CreateWindow(title.c_str(), wndPos, wndPos, requestedSize.width, requestedSize.height,
-                                  SDL_WINDOW_OPENGL | SDL_WINDOW_RESIZABLE);
+                                  commonFlags | SDL_WINDOW_RESIZABLE);
     }
 
     if(!window)
--- a/extras/videoDrivers/SDL2/VideoSDL2.cpp
+++ b/extras/videoDrivers/SDL2/VideoSDL2.cpp
@@ -445,11 +445,12 @@ void VideoSDL2::SetMousePos(Position pos)
 
 void VideoSDL2::setGuiScalePercent(unsigned percent)
 {
-    if(setGuiScaleInternal(percent))
-    {
-        UpdateCurrentSizes();
-        CallBack->WindowResized();
-    }
+    setGuiScaleInternal(percent);
+}
+
+void VideoSDL2::onGuiScaleChanged()
+{
+    UpdateCurrentSizes();
 }
 
 KeyEvent VideoSDL2::GetModKeyState() const
--- a/extras/videoDrivers/SDL2/VideoSDL2.cpp
+++ b/extras/videoDrivers/SDL2/VideoSDL2.cpp
@@ -443,11 +443,6 @@ void VideoSDL2::SetMousePos(Position pos)
     SDL_WarpMouseInWindow(window, screenPos.x, screenPos.y);
 }
 
-void VideoSDL2::setGuiScalePercent(unsigned percent)
-{
-    setGuiScaleInternal(percent);
-}
-
 void VideoSDL2::onGuiScaleChanged()
 {
     UpdateCurrentSizes();
--- a/libs/driver/include/driver/GuiScale.h
+++ b/libs/driver/include/driver/GuiScale.h
@@ -8,6 +8,14 @@
 #include "helpers/mathFuncs.h"
 #include <type_traits>
 
+/// Holds the range of supported GUI scale percentages
+struct GuiScaleRange
+{
+    unsigned minPercent;         ///< Minimum supported GUI scale in percent
+    unsigned maxPercent;         ///< Maximum supported GUI scale in percent
+    unsigned recommendedPercent; ///< Recommended GUI scale in percent
+};
+
 /// Represents the scale factors – precomputed from a percentage – to translate between screen and view dimensions
 class GuiScale
 {
--- a/libs/driver/include/driver/GuiScale.h
+++ b/libs/driver/include/driver/GuiScale.h
@@ -1,0 +1,61 @@
+// Copyright (C) 2005 - 2023 Settlers Freaks (sf-team at siedler25.org)
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#pragma once
+
+#include "Point.h"
+#include "helpers/mathFuncs.h"
+#include <type_traits>
+
+/// Represents the scale factors – precomputed from a percentage – to translate between screen and view dimensions
+class GuiScale
+{
+public:
+    constexpr GuiScale(unsigned percent) noexcept
+        : percent_(percent), scaleViewToScreen_(percent_ / 100.f), scaleScreenToView_(1.f / scaleViewToScreen_)
+    {}
+
+    constexpr unsigned percent() const noexcept { return percent_; }
+    constexpr float scaleFactor() const noexcept { return scaleViewToScreen_; }
+    constexpr float invScaleFactor() const noexcept { return scaleScreenToView_; }
+
+    /// Translate a scalar from screen to view space
+    template<typename R = float>
+    constexpr R screenToView(float val) const noexcept
+    {
+        if constexpr(std::is_integral_v<R>)
+            return helpers::iround<R>(val * scaleScreenToView_);
+        else
+            return static_cast<R>(val * scaleScreenToView_);
+    }
+
+    /// Translate a point from screen to view space
+    template<typename R = Position, typename T>
+    constexpr R screenToView(Point<T> pt) const noexcept
+    {
+        return R(PointF(pt) * scaleScreenToView_);
+    }
+
+    /// Translate a scalar from view to screen space
+    template<typename R = float>
+    constexpr R viewToScreen(float val) const noexcept
+    {
+        if constexpr(std::is_integral_v<R>)
+            return helpers::iround<R>(val * scaleViewToScreen_);
+        else
+            return static_cast<R>(val * scaleViewToScreen_);
+    }
+
+    /// Translate a point from view to screen space
+    template<typename R = Position, typename T>
+    constexpr R viewToScreen(Point<T> pt) const noexcept
+    {
+        return R(PointF(pt) * scaleViewToScreen_);
+    }
+
+private:
+    unsigned percent_;        ///< GUI scaling factor in percent
+    float scaleViewToScreen_; ///< Decimal GUI scaling factor
+    float scaleScreenToView_; ///< Inverse of the decimal GUI scaling factor
+};
--- a/libs/driver/include/driver/VideoDriver.h
+++ b/libs/driver/include/driver/VideoDriver.h
@@ -33,6 +33,8 @@ public:
 
     const GuiScale& getGuiScale() const override final { return guiScale_; }
 
+    GuiScaleRange getGuiScaleRange() const override;
+
     /// prüft auf Initialisierung.
     bool IsInitialized() const override final { return initialized; }
     bool IsOpenGL() const override { return true; }
--- a/libs/driver/include/driver/VideoDriver.h
+++ b/libs/driver/include/driver/VideoDriver.h
@@ -31,6 +31,8 @@ public:
     Extent GetRenderSize() const override final { return renderSize_; }
     bool IsFullscreen() const override final { return isFullscreen_; }
 
+    const GuiScale& getGuiScale() const override final { return guiScale_; }
+
     /// prüft auf Initialisierung.
     bool IsInitialized() const override final { return initialized; }
     bool IsOpenGL() const override { return true; }
@@ -38,6 +40,9 @@ public:
 protected:
     VideoMode FindClosestVideoMode(const VideoMode& mode) const;
     void SetNewSize(VideoMode windowSize, Extent renderSize);
+
+    /// Set the GUI scale in percent and return true if changed
+    bool setGuiScaleInternal(unsigned percent);
 
     VideoDriverLoaderInterface* CallBack; /// Das DriverCallback für Rückmeldungen.
     bool initialized;                     /// Initialisierungsstatus.
@@ -48,4 +53,6 @@ private:
     // cached as possibly used often
     VideoMode windowSize_;
     Extent renderSize_;
+
+    GuiScale guiScale_; ///< Scale factor applied to the user interface
 };
--- a/libs/driver/include/driver/VideoDriver.h
+++ b/libs/driver/include/driver/VideoDriver.h
@@ -34,6 +34,7 @@ public:
     float getDpiScale() const override final { return dpiScale_; }
 
     const GuiScale& getGuiScale() const override final { return guiScale_; }
+    void setGuiScalePercent(unsigned percent) override;
 
     GuiScaleRange getGuiScaleRange() const override;
 
@@ -44,9 +45,6 @@ public:
 protected:
     VideoMode FindClosestVideoMode(const VideoMode& mode) const;
     void SetNewSize(VideoMode windowSize, Extent renderSize);
-
-    /// Set the GUI scale in percent
-    void setGuiScaleInternal(unsigned percent);
 
     /// Callback to update state in response to GUI scale changes before events are dispatched
     virtual void onGuiScaleChanged() {}
--- a/libs/driver/include/driver/VideoDriver.h
+++ b/libs/driver/include/driver/VideoDriver.h
@@ -46,9 +46,6 @@ protected:
     VideoMode FindClosestVideoMode(const VideoMode& mode) const;
     void SetNewSize(VideoMode windowSize, Extent renderSize);
 
-    /// Callback to update state in response to GUI scale changes before events are dispatched
-    virtual void onGuiScaleChanged() {}
-
     VideoDriverLoaderInterface* CallBack; /// Das DriverCallback für Rückmeldungen.
     bool initialized;                     /// Initialisierungsstatus.
     MouseCoords mouse_xy;                 /// Status der Maus.
--- a/libs/driver/include/driver/VideoDriver.h
+++ b/libs/driver/include/driver/VideoDriver.h
@@ -28,7 +28,7 @@ public:
     bool GetMouseStateR() const override;
 
     VideoMode GetWindowSize() const override final { return windowSize_; }
-    Extent GetRenderSize() const override final { return renderSize_; }
+    Extent GetRenderSize() const override final { return scaledRenderSize_; }
     bool IsFullscreen() const override final { return isFullscreen_; }
 
     const GuiScale& getGuiScale() const override final { return guiScale_; }
@@ -51,8 +51,9 @@ protected:
     bool isFullscreen_;                   /// Vollbild an/aus?
 private:
     // cached as possibly used often
-    VideoMode windowSize_;
-    Extent renderSize_;
+    VideoMode windowSize_;    ///< Size of the window or fullscreen resolution
+    Extent renderSize_;       ///< Size of the renderable surface
+    Extent scaledRenderSize_; ///< Size of the projection clipping area
 
     GuiScale guiScale_; ///< Scale factor applied to the user interface
 };
--- a/libs/driver/include/driver/VideoDriver.h
+++ b/libs/driver/include/driver/VideoDriver.h
@@ -45,8 +45,11 @@ protected:
     VideoMode FindClosestVideoMode(const VideoMode& mode) const;
     void SetNewSize(VideoMode windowSize, Extent renderSize);
 
-    /// Set the GUI scale in percent and return true if changed
-    bool setGuiScaleInternal(unsigned percent);
+    /// Set the GUI scale in percent
+    void setGuiScaleInternal(unsigned percent);
+
+    /// Callback to update state in response to GUI scale changes before events are dispatched
+    virtual void onGuiScaleChanged() {}
 
     VideoDriverLoaderInterface* CallBack; /// Das DriverCallback für Rückmeldungen.
     bool initialized;                     /// Initialisierungsstatus.
--- a/libs/driver/include/driver/VideoDriver.h
+++ b/libs/driver/include/driver/VideoDriver.h
@@ -31,6 +31,8 @@ public:
     Extent GetRenderSize() const override final { return scaledRenderSize_; }
     bool IsFullscreen() const override final { return isFullscreen_; }
 
+    float getDpiScale() const override final { return dpiScale_; }
+
     const GuiScale& getGuiScale() const override final { return guiScale_; }
 
     GuiScaleRange getGuiScaleRange() const override;
@@ -57,5 +59,7 @@ private:
     Extent renderSize_;       ///< Size of the renderable surface
     Extent scaledRenderSize_; ///< Size of the projection clipping area
 
+    float dpiScale_;    ///< Scale factor required to convert "normal" DPI to the display DPI
     GuiScale guiScale_; ///< Scale factor applied to the user interface
+    bool autoGuiScale_; ///< Whether the recommended GUI scale is used
 };
--- a/libs/driver/src/VideoDriver.cpp
+++ b/libs/driver/src/VideoDriver.cpp
@@ -4,6 +4,7 @@
 
 #include "driver/VideoDriver.h"
 #include "commonDefines.h"
+#include "driver/VideoDriverLoaderInterface.h"
 #include "helpers/mathFuncs.h"
 #include <algorithm>
 #include <stdexcept>
@@ -86,7 +87,15 @@ bool VideoDriver::setGuiScaleInternal(unsigned percent)
     if(guiScale_.percent() == percent)
         return false;
 
+    // translate current mouse position to screen space
+    const auto screenPos = guiScale_.viewToScreen(mouse_xy.pos);
+
     guiScale_ = GuiScale(percent);
+
+    // move cursor to new position in view space
+    mouse_xy.pos = guiScale_.screenToView(screenPos);
+    CallBack->Msg_MouseMove(mouse_xy);
+
     return true;
 }
 
--- a/libs/driver/src/VideoDriver.cpp
+++ b/libs/driver/src/VideoDriver.cpp
@@ -4,6 +4,7 @@
 
 #include "driver/VideoDriver.h"
 #include "commonDefines.h"
+#include "helpers/mathFuncs.h"
 #include <algorithm>
 #include <stdexcept>
 
@@ -84,4 +85,18 @@ bool VideoDriver::setGuiScaleInternal(unsigned percent)
 
     guiScale_ = GuiScale(percent);
     return true;
+}
+
+GuiScaleRange VideoDriver::getGuiScaleRange() const
+{
+    const auto ratioXY = PointF(renderSize_) / PointF(windowSize_.width, windowSize_.height);
+    const auto ratioAvg = (ratioXY.x + ratioXY.y) / 2.f;
+    const auto min = helpers::iround<unsigned>(ratioAvg * 50.f);
+    const auto recommended = std::max(min, helpers::iround<unsigned>(ratioAvg * 100.f));
+    const auto maxScaleXY = renderSize_ / PointF(800.f, 600.f);
+    const auto maxScale = std::min(maxScaleXY.x, maxScaleXY.y);
+    // if the window shrinks below its minimum size of 800x600, max can be smaller than recommended
+    const auto max = std::max(helpers::iround<unsigned>(maxScale * 100.f), recommended);
+
+    return GuiScaleRange{min, max, recommended};
 }
--- a/libs/driver/src/VideoDriver.cpp
+++ b/libs/driver/src/VideoDriver.cpp
@@ -20,7 +20,7 @@ IVideoDriver::~IVideoDriver() = default;
  */
 VideoDriver::VideoDriver(VideoDriverLoaderInterface* CallBack)
     : CallBack(CallBack), initialized(false), isFullscreen_(false), renderSize_(0, 0), scaledRenderSize_(0, 0),
-      dpiScale_(1.f), guiScale_(100)
+      dpiScale_(1.f), guiScale_(100), autoGuiScale_(false)
 {
     std::fill(keyboard.begin(), keyboard.end(), false);
 }
@@ -76,15 +76,20 @@ void VideoDriver::SetNewSize(VideoMode windowSize, Extent renderSize)
 {
     windowSize_ = windowSize;
     renderSize_ = renderSize;
-    scaledRenderSize_ = guiScale_.screenToView<Extent>(renderSize);
 
     const auto ratioXY = PointF(renderSize_) / PointF(windowSize_.width, windowSize_.height);
     dpiScale_ = (ratioXY.x + ratioXY.y) / 2.f; // use the average ratio of both axes
+
+    if(autoGuiScale_)
+        guiScale_ = GuiScale(getGuiScaleRange().recommendedPercent);
+
+    scaledRenderSize_ = guiScale_.screenToView<Extent>(renderSize);
 }
 
 void VideoDriver::setGuiScalePercent(unsigned percent)
 {
-    if(percent == 0)
+    autoGuiScale_ = (percent == 0);
+    if(autoGuiScale_)
         percent = getGuiScaleRange().recommendedPercent;
 
     if(guiScale_.percent() == percent)
@@ -94,9 +99,7 @@ void VideoDriver::setGuiScalePercent(unsigned percent)
     const auto screenPos = guiScale_.viewToScreen(mouse_xy.pos);
 
     guiScale_ = GuiScale(percent);
-
-    onGuiScaleChanged(); // give backends an opportunity to update their state
-
+    scaledRenderSize_ = guiScale_.screenToView<Extent>(renderSize_);
     CallBack->WindowResized();
 
     // move cursor to new position in view space
--- a/libs/driver/src/VideoDriver.cpp
+++ b/libs/driver/src/VideoDriver.cpp
@@ -82,7 +82,7 @@ void VideoDriver::SetNewSize(VideoMode windowSize, Extent renderSize)
     dpiScale_ = (ratioXY.x + ratioXY.y) / 2.f; // use the average ratio of both axes
 }
 
-void VideoDriver::setGuiScaleInternal(unsigned percent)
+void VideoDriver::setGuiScalePercent(unsigned percent)
 {
     if(percent == 0)
         percent = getGuiScaleRange().recommendedPercent;
--- a/libs/driver/src/VideoDriver.cpp
+++ b/libs/driver/src/VideoDriver.cpp
@@ -80,6 +80,9 @@ void VideoDriver::SetNewSize(VideoMode windowSize, Extent renderSize)
 
 bool VideoDriver::setGuiScaleInternal(unsigned percent)
 {
+    if(percent == 0)
+        percent = getGuiScaleRange().recommendedPercent;
+
     if(guiScale_.percent() == percent)
         return false;
 
--- a/libs/driver/src/VideoDriver.cpp
+++ b/libs/driver/src/VideoDriver.cpp
@@ -20,7 +20,7 @@ IVideoDriver::~IVideoDriver() = default;
  */
 VideoDriver::VideoDriver(VideoDriverLoaderInterface* CallBack)
     : CallBack(CallBack), initialized(false), isFullscreen_(false), renderSize_(0, 0), scaledRenderSize_(0, 0),
-      guiScale_(100)
+      dpiScale_(1.f), guiScale_(100)
 {
     std::fill(keyboard.begin(), keyboard.end(), false);
 }
@@ -77,6 +77,9 @@ void VideoDriver::SetNewSize(VideoMode windowSize, Extent renderSize)
     windowSize_ = windowSize;
     renderSize_ = renderSize;
     scaledRenderSize_ = guiScale_.screenToView<Extent>(renderSize);
+
+    const auto ratioXY = PointF(renderSize_) / PointF(windowSize_.width, windowSize_.height);
+    dpiScale_ = (ratioXY.x + ratioXY.y) / 2.f; // use the average ratio of both axes
 }
 
 bool VideoDriver::setGuiScaleInternal(unsigned percent)
@@ -101,10 +104,8 @@ bool VideoDriver::setGuiScaleInternal(unsigned percent)
 
 GuiScaleRange VideoDriver::getGuiScaleRange() const
 {
-    const auto ratioXY = PointF(renderSize_) / PointF(windowSize_.width, windowSize_.height);
-    const auto ratioAvg = (ratioXY.x + ratioXY.y) / 2.f;
-    const auto min = helpers::iround<unsigned>(ratioAvg * 50.f);
-    const auto recommended = std::max(min, helpers::iround<unsigned>(ratioAvg * 100.f));
+    const auto min = helpers::iround<unsigned>(dpiScale_ * 50.f);
+    const auto recommended = std::max(min, helpers::iround<unsigned>(dpiScale_ * 100.f));
     const auto maxScaleXY = renderSize_ / PointF(800.f, 600.f);
     const auto maxScale = std::min(maxScaleXY.x, maxScaleXY.y);
     // if the window shrinks below its minimum size of 800x600, max can be smaller than recommended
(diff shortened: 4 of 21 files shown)


falbrechtskirchinger:
I've incorporated @Spikeone's feedback and added an option to decouple scaling the game world from scaling the GUI. On HighDPI setups, some scaling is applied regardless of the setting.

I want to note that I very specifically chose the term "window zoom" (inspired by VSCode) over "GUI scale" because the game world was scaled as well (I wasn't eager to figure out how not to when I started and was far less familiar with the code base than I am now). Unlike in 3D games where scaling is limited to the GUI/HUD and scaling the game world makes no sense due to the perspective projection.

Arguably, this setting could be removed, "Window Zoom" be renamed to "GUI Scale," and the game world scaled depending on DPI, as is the case with the setting turned off.

Xellzul:
@falbrechtskirchinger 
Hey, had quick look at this PR (e64943ef6c81c0d21013403049c62f4b9ec55a80) on Windows 11 at 6880x2880 (21:9) resolution:

Everything works as expected except:
Taking in-game screenshots using F12, this was the result with scaling (note the size 1720 x 720 of the screenshot was exact 1/4 of used resolution of 6880x2880 using 400% scaling):


In-game graph was thin
Screenshot 2023-07-11 200349

Few screenshots:
https://imgur.com/a/VX9xn1u

falbrechtskirchinger:
@Xellzul Thank you! I'm away for a few days but should be able to fix those issues over the weekend.

falbrechtskirchinger:
Both issues should be fixed. As expected, they required minimal changes, though the line drawing could be optimized by calculating the line width only once at a higher level – sacrificing a degree of abstraction – or by implementing `BeginDrawLine()`/`EndDrawLine()`  functions to calculate and set the width once per multiple draw calls.

`iwStatistics` could also benefit from a significant rewrite, as many draw positions are specified manually. (In a separate PR and at a later date, of course.)

falbrechtskirchinger:
> * Should the setting be renamed to `(G)UI scale` and `Zoom Game World` be removed and always disabled?

I'm making an executive decision and going ahead with the change. "Window Zoom" becomes "GUI Scale" and the DPI scale always scales the game world view. I'm also moving from using `unsigned` to a new type that encapsulates all the precomputed values and the functions to transform points, obviating some of the additions to `IVideoDriver`.

falbrechtskirchinger:
> * Should the setting be renamed to `(G)UI scale` and `Zoom Game World` be removed and always disabled?

I'm making an executive decision and going ahead with the change. "Window Zoom" becomes "GUI Scale" and the DPI scale always scales the game world view. I'm also moving from using `unsigned` to a new type that encapsulates all the precomputed values and the functions to transform points, obviating some of the additions to `IVideoDriver, as well as further additions or redundant calculations.

extras/videoDrivers/SDL2/VideoSDL2.cpp
```diff
@@ -346,7 +351,7 @@ bool VideoSDL2::MessageLoop()
                 break;
             }
             case SDL_MOUSEBUTTONDOWN:
-                mouse_xy.pos = Position(ev.button.x, ev.button.y);
+                mouse_xy.pos = Position(GetGuiScale().screenToViewCoords(Position(ev.button.x, ev.button.y)));
```
falbrechtskirchinger, line 354: Originally, this was written as `mouse_xy.pos = GetGuiScale().screenToViewCoords(Position(ev.button.x, ev.button.y));` when `screenToViewCoords()` had a template parameter for the return type that was defaulted to `Position`.
In other places, one could write `GetGuiScale().screenToViewCoords<Extent>(...)`, for example. I removed the template parameter because it mirrored the style of the rest of the code.
The upside of the template parameter here is fewer parentheses, which improves readability.
It would be easy to revert the commit that introduced this change.

libs/driver/include/driver/GuiScale.h
```diff
@@ -0,0 +1,51 @@
...
+    constexpr GuiScale(unsigned percent) noexcept
+        : percent_(percent), scaleViewToScreen_(percent_ * .01f), scaleScreenToView_(1.f / scaleViewToScreen_)
+    {}
+
+    constexpr unsigned percent() const noexcept { return percent_; }
+    constexpr float scaleFactor() const noexcept { return scaleViewToScreen_; }
+    constexpr float invScaleFactor() const noexcept { return scaleScreenToView_; }
+
+    template<typename T, detail::EnableIfPoint_t<T> = 0>
+    constexpr PointF screenToViewCoords(T pos) const noexcept
```
falbrechtskirchinger, line 36: Why so complicated?
```cpp
template<typename T>
    constexpr PointF screenToViewCoords(Point<T> pos) const noexcept
```

libs/driver/src/VideoDriver.cpp
```diff
@@ -72,5 +75,57 @@ VideoMode VideoDriver::FindClosestVideoMode(const VideoMode& mode) const
...
+    if(guiScale_.percent() == percent)
+        return;
+
+    // translate current mouse position to screen space
+    const auto screenPos = Position(guiScale_.viewToScreenCoords(mouse_xy.pos));
+
+    guiScale_ = GuiScale(percent);
+
+    // move cursor to new position in view space
+    mouse_xy.pos = Position(guiScale_.screenToViewCoords(screenPos));
```
falbrechtskirchinger, line 99: Remove here. Duplicated below.

libs/s25main/desktops/dskOptions.cpp
```diff
@@ -570,6 +587,17 @@ void dskOptions::Msg_ButtonClick(const unsigned ctrl_id)
 
             SETTINGS.Save();
 
+            // Does the selected backend support GUI scaling?
+            unsigned recommendedGuiScale;
+            VIDEODRIVER.GetGuiScaleRange(nullptr, nullptr, &recommendedGuiScale);
+            if(VIDEODRIVER.GetGuiScale().percent() == 100
+               && (SETTINGS.video.guiScale != 100 || (SETTINGS.video.guiScale == 0 && recommendedGuiScale != 100)))
```
falbrechtskirchinger, line 594: Could be clearer. Really, this tries to figure out if the backend *needs* to support GUI scaling to fulfill the user request and warns if it cannot.

libs/s25main/desktops/dskOptions.cpp
```diff
@@ -655,3 +681,91 @@ void dskOptions::loadVideoModes()
...
+    unsigned index = 0;
+    auto* group = GetCtrl<ctrlGroup>(ID_grpGraphics);
+    auto* combo = group->GetCtrl<ctrlComboBox>(ID_cbGuiScale);
+    const auto& selection = combo->GetSelection();
+    if(!selection || *selection == 0)
+        index = recommendedGuiScaleIndex + 1;
+    else if(!up && *selection > 1)
+        index = *selection - 1;
+    else if(up && *selection < guiScales.size())
+        index = *selection + 1;
```
falbrechtskirchinger, line 762: Add some comments.

libs/s25main/drivers/VideoDriverWrapper.cpp
```diff
@@ -300,10 +300,11 @@ void VideoDriverWrapper::RenewViewport()
         return;
 
     const Extent renderSize = videodriver->GetRenderSize();
+    const VideoMode windowSize = videodriver->GetWindowSize();
 
     // Viewport mit widthxheight setzen
-    glViewport(0, 0, renderSize.x, renderSize.y);
-    glScissor(0, 0, renderSize.x, renderSize.y);
+    glViewport(0, 0, windowSize.width, windowSize.height);
+    glScissor(0, 0, windowSize.width, windowSize.height);
```
falbrechtskirchinger, line 307: Update comment.

libs/s25main/ogl/OpenGLRenderer.cpp
```diff
@@ -111,7 +112,7 @@ void OpenGLRenderer::DrawLine(DrawPoint pt1, DrawPoint pt2, unsigned width, unsi
     glDisable(GL_TEXTURE_2D);
     glColor4ub(GetRed(color), GetGreen(color), GetBlue(color), GetAlpha(color));
 
-    glLineWidth(static_cast<GLfloat>(width));
+    glLineWidth(static_cast<GLfloat>(width) * VIDEODRIVER.GetGuiScale().scaleFactor());
```
falbrechtskirchinger, line 115: Add a scalar overload to `screenToView()` and `viewToScreen()` and use it here. Should be clearer.

libs/s25main/world/GameWorldView.cpp
```diff
@@ -95,7 +97,11 @@ void GameWorldView::SetZoomFactor(float zoomFactor, bool smoothTransition /* = t
     else
         targetZoomFactor_ = zoomFactor;
     if(!smoothTransition)
+    {
         zoomFactor_ = targetZoomFactor_;
+        UpdateEffectiveZoomFactor();
+        CalcFxLx();
```
falbrechtskirchinger, line 103: I asked on Discord whether this call to `CalcFxLx()` was missing. Added here because I think it is.

libs/s25main/world/GameWorldView.h
```diff
@@ -62,6 +62,8 @@ class GameWorldView
 
     /// How much the view is scaled (1=normal, >1=bigger, >0 && <1=smaller)
     float zoomFactor_;
+    /// Cache the possibly DPI scale corrected zoom factor (based on setting)
+    float effectiveZoomFactor_;
```
falbrechtskirchinger, line 66: This is no longer based on a setting.

falbrechtskirchinger (commented):
Discovered a few minor things, but otherwise am quite happy with the state of this PR.

extras/videoDrivers/SDL2/VideoSDL2.cpp
falbrechtskirchinger: I've removed c07bd33 (f0700f4c1) after seeing other places in the code use the same template-based cast pattern. (And `iround()` uses it too, of course.)

falbrechtskirchinger:
I've started using C++17 features. CI is expected to fail accordingly.

Flamefire:
Please use force-push more carefully. After starting to review a PR a force-push means loosing any information about the state the last conversation was about and what the new changes are. It may make sense to do a final rebase and/or cleanup once all is resolved right before merging but until then especially for those large changes I find it confusing.

Idea: Prefix a commit meant to fix an older commit with "fixup: <old title>" such that it sticks out and it is easy to combine those commits later on in a single final rebase.

falbrechtskirchinger:
> Please use force-push more carefully.

The last force-push was an unfortunate accident. I moved my work to a new branch and pushed the way outdated old `window-zoom` branch. (Renamed it now and set the tracking branch properly to avoid this in the future.)

> Idea: Prefix a commit meant to fix an older commit with "fixup: " such that it sticks out and it is easy to combine those commits later on in a single final rebase.

I'm using `git commit --fixup` with `git rebase -i` anyway. I can just push my "fixup! " commits without the rebase.

I can merge master instead of rebasing for C++17 if you prefer.

libs/s25main/desktops/dskOptions.cpp
```diff
@@ -663,3 +688,91 @@ void dskOptions::loadVideoModes()
...
+
+    for(unsigned percent = roundGuiScale(minPercent); percent <= maxPercent; percent += stepSize)
+    {
+        if(percent == recommendedPercentRounded)
+            recommendedGuiScaleIndex_ = guiScales_.size();
+        guiScales_.push_back(percent);
+
+        combo->AddString(helpers::toString(percent) + "%");
+        if(percent == SETTINGS.video.guiScale)
+            combo->SetSelection(combo->GetNumItems() - 1);
```
Flamefire, line 743: Might be easier to read if you swap the lines a bit:
```suggestion
        if(percent == SETTINGS.video.guiScale)
            combo->SetSelection(combo->GetNumItems());

        combo->AddString(helpers::toString(percent) + "%");
        guiScales_.push_back(percent);
```

libs/s25main/desktops/dskOptions.cpp
```diff
@@ -663,3 +688,91 @@ void dskOptions::loadVideoModes()
...
+{
+    if(!VIDEODRIVER.GetModKeyState().ctrl)
+        return false;
+
+    unsigned index = 0;
+    auto* group = GetCtrl<ctrlGroup>(ID_grpGraphics);
+    auto* combo = group->GetCtrl<ctrlComboBox>(ID_cbGuiScale);
+    const auto& selection = combo->GetSelection();
+    if(!selection || *selection == 0) // No selection or "Auto" item selected
+        index = recommendedGuiScaleIndex_ + 1;
```
Flamefire, line 765: This ignores the "up" parameter AND the range of `guiScales_`. Idea:   
Set `index` to `selection` or `recommendedGuiScaleIndex_`, then check `up` and range using `index` instead of `selection`.

libs/s25main/desktops/dskOptions.cpp
```diff
@@ -663,3 +688,91 @@ void dskOptions::loadVideoModes()
...
+    else if(!up && *selection > 1) // Scrolling down and not at the first item
+        index = *selection - 1;
+    else if(up && *selection < guiScales_.size()) // Scrolling up and not at the last item
+        index = *selection + 1;                   // (no "- 1" needed because of "Auto" item)
+    else
+        return true;
```
Flamefire, line 771: It might be clearer if instead of returning you wrap the change below into something like `if(index != selection)`, especially with the above suggested fix.

libs/driver/include/driver/GuiScale.h
```diff
@@ -0,0 +1,61 @@
...
+#include "Point.h"
+#include "helpers/mathFuncs.h"
+#include <type_traits>
+
+/// Represents the scale factors – precomputed from a percentage – to translate between screen and view dimensions
+class GuiScale
+{
+public:
+    constexpr GuiScale(unsigned percent) noexcept
+        : percent_(percent), scaleViewToScreen_(percent_ * .01f), scaleScreenToView_(1.f / scaleViewToScreen_)
```
Flamefire, line 16: No need for manual optimizations, be explicit:
```suggestion
        : percent_(percent), scaleViewToScreen_(percent_ / 100f), scaleScreenToView_(1.f / scaleViewToScreen_)
```

libs/driver/src/VideoDriver.cpp
```diff
@@ -72,5 +75,54 @@ VideoMode VideoDriver::FindClosestVideoMode(const VideoMode& mode) const
 void VideoDriver::SetNewSize(VideoMode windowSize, Extent renderSize)
 {
     windowSize_ = windowSize;
-    renderSize_ = renderSize;
+    screenRenderSize_ = renderSize;
+    viewRenderSize_ = guiScale_.screenToView<Extent>(renderSize);
+
+    const auto ratioXY = PointF(screenRenderSize_) / PointF(windowSize_.width, windowSize_.height);
+    dpiScale_ = (ratioXY.x + ratioXY.y) * .5f; // use the average ratio of both axes
```
Flamefire, line 82: Similar: Don't transform division by hand:
```suggestion
    dpiScale_ = (ratioXY.x + ratioXY.y) / 2f; // use the average ratio of both axes
```

libs/s25main/drivers/VideoDriverWrapper.cpp
```diff
@@ -493,3 +494,24 @@ bool VideoDriverWrapper::IsFullscreen() const
 {
     return videodriver->IsFullscreen();
 }
+
+float VideoDriverWrapper::getDpiScale() const
+{
+    return videodriver->getDpiScale();
+}
+
+const GuiScale& VideoDriverWrapper::getGuiScale() const
```
Flamefire, line 503: Maybe it makes more sense to call the struct and hence the getter `GuiScaler`? Because you set an unsigned but get a struct although the name is seemingly the same.   
Just an idea.

libs/s25main/world/GameWorldView.cpp
Flamefire: Yeah, I guess we never used the non-smooth transition often and hence didn't notice.

Flamefire (changes requested):
> I'm using git commit --fixup with git rebase -i anyway. I can just push my "fixup! " commits without the rebase.

Perfect! Yes please

> I can merge master instead of rebasing for C++17 if you prefer.

Added review comments, so yes that would be better.

libs/driver/include/driver/GuiScale.h
```diff
@@ -0,0 +1,61 @@
...
+    {
+        if constexpr(std::is_integral_v<R>)
+            return helpers::iround<R>(val * scaleViewToScreen_);
+        else
+            return static_cast<R>(val * scaleViewToScreen_);
+    }
+
+    /// Translate a point from screen to view space
+    template<typename R = Position, typename T>
+    constexpr R screenToView(Point<T> pt) const noexcept
```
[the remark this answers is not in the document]
falbrechtskirchinger: `Point<T>` returns `Point<int>` by default and `float` returns `float` by default. Both functions round when going from floating-point to integer. These choices match the common usage of both overloads.

libs/driver/include/driver/VideoDriver.h
```diff
@@ -28,9 +28,16 @@ class VideoDriver : public IVideoDriver
     bool GetMouseStateR() const override;
 
     VideoMode GetWindowSize() const override final { return windowSize_; }
-    Extent GetRenderSize() const override final { return renderSize_; }
+    Extent GetRenderSize() const override final { return viewRenderSize_; }
```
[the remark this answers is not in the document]
falbrechtskirchinger: No. This is all by careful design. The render size in view space coordinates is the one users are interested in. It is scaled by GUI scale, which is fundamental to implementing the scaling effect. `glViewport` and `glScissor` are set to the window size, and `glOrtho` uses render size (as returned here). The original render size is only needed for internal calculations.

`viewRenderSize_` has to be updated by the backend in the `onGuiScaleChanged` callback. These steps are ordered this way to avoid spurious hover events. Arguably, I'm complicating things unnecessarily to avoid a minor visual glitch, but is carefully considered. :man_shrugging:

libs/driver/src/VideoDriver.cpp
```diff
@@ -72,5 +75,54 @@ VideoMode VideoDriver::FindClosestVideoMode(const VideoMode& mode) const
...
+    // move cursor to new position in view space
+    // must happen after window resize event to avoid drawing spurious hover events
+    mouse_xy.pos = guiScale_.screenToView(screenPos);
+    CallBack->Msg_MouseMove(mouse_xy);
+}
+
+void VideoDriver::getGuiScaleRange(unsigned* minPercent, unsigned* maxPercent, unsigned* recommendedPercent) const
+{
+    if(minPercent)
+        *minPercent = 100u;
```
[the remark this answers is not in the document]
falbrechtskirchinger: This is partly a historic artifact and partly to have one authoritative location to set the minimum. I originally had 50% DPI-adjusted here and then decided after feedback from testing, that 50% is definitely too low. Someday we may still decide to support values lower than 100% (maybe 80%?) and this would be the place to change it.

libs/s25main/desktops/dskOptions.cpp
```diff
@@ -663,3 +688,91 @@ void dskOptions::loadVideoModes()
...
+    if(!combo)
+    {
+        auto* group = GetCtrl<ctrlGroup>(ID_grpGraphics);
+        combo = group->GetCtrl<ctrlComboBox>(ID_cbGuiScale);
+    }
+
+    guiScales_.clear();
+    combo->DeleteAllItems();
+
+    combo->AddString((boost::format(_("Auto (%u%%)")) % recommendedPercent).str());
```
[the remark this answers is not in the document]
falbrechtskirchinger: Great. I miss `fmtlib`, so this helps.

libs/s25main/desktops/dskOptions.cpp
falbrechtskirchinger: I'll check. I fiddled around with these lines quite a bit trying to make them easier to read while still retaining the desired behavior.

libs/s25main/desktops/dskOptions.cpp
falbrechtskirchinger: (See above.)

libs/driver/include/driver/GuiScale.h
falbrechtskirchinger: I know `-ffast-math` does it for me, but it's a hard habit to break.

libs/s25main/drivers/VideoDriverWrapper.cpp
falbrechtskirchinger: In my mind, `GuiScale` encapsulates the percentage as well as the derived scaling factors. The fact that it also bundles the operations is incidental and simply for convenience. I wouldn't want to emphasize the operations by calling it `GuiScaler`.

libs/driver/include/driver/GuiScale.h
Flamefire: Even if it doesn't the performance penalty for our occasional use of floats is negligible and readability is more important. But yes, I know that force of habbit ;-)

libs/s25main/desktops/dskOptions.cpp
falbrechtskirchinger: > This ignores the "up" parameter AND the range of `guiScales_`.

I didn't quite understand what you meant by this earlier. No, this is deliberate. Regardless of direction, we start at the recommended index. The `+ 1` accounts for "Auto" in the combo box. Note that "Auto" and the recommended percentage in the combo box aren't required to be the same due to rounding. Hence, storing the index instead of using the value obtained via `GetGuiScaleRange()`.

Nevertheless, I'll see if I can improve clarity.

libs/driver/include/driver/VideoDriver.h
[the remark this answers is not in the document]
Flamefire: > The render size in view space coordinates is the one users are interested in. It is scaled by GUI scale, which is fundamental to implementing the scaling effect.

So if the guiScale chantes the `viewRenderSize_` needs to change too, doesn't it?

> `viewRenderSize_` has to be updated by the backend in the `onGuiScaleChanged` callback.

I don't see this happening in this PR, am I missing it? Also `viewRenderSize_` is a property of this class so isn't this class responsible for keeping the values in sync?

libs/driver/src/VideoDriver.cpp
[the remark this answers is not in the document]
Flamefire: Ok, agreed to keep it for future use

libs/s25main/desktops/dskOptions.cpp
Flamefire: I guess using an extra variable to hold the "old selection" (taking the unset & auto case into account) and only checking `up` together with `std::min/max` might work. Just a suggestion.   
Important to me is that this case is handled correctly, i.e. not choosing a higher scale when scrolling down

libs/driver/include/driver/GuiScale.h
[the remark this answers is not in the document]
falbrechtskirchinger: > I don't understand how this relates to this function. `Point<T>` here is whatever is passed in, no defaults here AFAIK.

The return type is defaulted: `typename R = Position` (which is `Point<int>`, right?)

> Using the other overload for the components makes this very clear and even might be faster for float-points as you save a potential copy (you unconditionally construct/cast the input point to `PointF`)

I removed a bunch of template code to do the conversion conditionally after convincing myself that copy elision makes it unnecessary in practice.

> But this is only about being more readable.

More readable where? Here in `GuiScale.h` or at the call site? Because I care more about where the function is used and wanted that to be more compact. I've also only very recently added the `float` overload for clarity.

I don't see how duplicating the `Point` rounding logic here helps anyone. The rounding in `float` is just done for parity. Not that it's actually used anywhere at the moment AFAICR.

Or am I still misunderstanding you?

libs/s25main/drivers/VideoDriverWrapper.cpp
Flamefire: It looks like (almost) the only uses of `GuiScale` are those functions. The encapsulated values are only queried by the options screen. So why don't "emphasize the operations" if that's how it is (mainly) used?

And this looks a bit odd to me:
```
const GuiScale& getGuiScale();
void setGuiScale(float);
```

libs/s25main/desktops/dskOptions.cpp
falbrechtskirchinger: "Auto" and none always move to the recommended index first. That means that given a recommended scale of 152%, moving in either direction would select 150% first. If we want "Auto" to move to 140% and 160% respectively, which would make sense, we can make the logic more complicated. ;-)

I'll give it a shot later and report back.

(FYI, `std::clamp` was added in C++17. Yay!)

libs/driver/include/driver/GuiScale.h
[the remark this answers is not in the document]
Flamefire: > The return type is defaulted: `typename R = Position` (which is `Point<int>`, right?)

I see, so you meant "`Point<T>` returns `Point<int>`" as in "passing x to **this** function returns y".

> I removed a bunch of template code to do the conversion conditionally after convincing myself that copy elision makes it unnecessary in practice.

Agreed.

> More readable where? Here in `GuiScale.h` or at the call site? 

readable here. You have 2 overloads: one working on a single value and one on a pair of values. From the code here I cannot tell that `foo(pt)` and `Position(foo(pt.x), foo(pt.y))` do the same thing. Which is why I was asking in the first place as the other operations on points are basically implemented that way: Apply the function on each component.   
So it begs the question whether the operation on Points is supposed to be different especially as it duplicates logic: In this case only the choice of `scaleScreenToView_` vs `scaleViewToScreen_`

> I don't see how duplicating the `Point` rounding logic here helps anyone. 

You already duplicated the rounding logic in the single-value function (for IMO good reasons) so why not make use of that avoiding the next person wondering whether there is a difference in the overloads?

At least put the overloads next to each other for discoverability.

libs/s25main/desktops/dskOptions.cpp
Flamefire: > "Auto" and none always move to the recommended index first. That means that given a recommended scale of 152%, moving in either direction would select 150% first. If we want "Auto" to move to 140% and 160% respectively, which would make sense, we can make the logic more complicated. ;-)

The first thing makes sense to me: Moving in either directly selects the recommended index first, i.e. effectively a no-op. So `        index = recommendedGuiScaleIndex_ + 1;` is a bit misleading because `index` is not the "index" but the new selection

So your approach makes sense to me, just the naming `index = index-we-want + 1` threw me off.

> (FYI, std::clamp was added in C++17. Yay!)

Oh right! IIRC we even had a replacement in `helpers`-namespace.

falbrechtskirchinger: > Oh right! IIRC we even had a replacement in `helpers`-namespace.

I think so too.

libs/s25main/drivers/VideoDriverWrapper.cpp
Flamefire: > The places that set the GUI scale don't care about the operations and only know the scaling percentage, while the opposite holds for the callers of `getGuiScale()`. IMO, the asymmetric requirements justify the asymmetry in the API.

True, so the first variant is better I guess which leaves open adding `getGuiScale` wrapping `getGuiScaler().getPercent()` if we need that often enough.

falbrechtskirchinger:
```sh
$ grep -r 'percent()' extras/ libs/
libs/driver/include/driver/GuiScale.h:    constexpr unsigned percent() const noexcept { return percent_; }
libs/driver/src/VideoDriver.cpp:    if(guiScale_.percent() == percent)
libs/s25main/desktops/dskOptions.cpp:            if(VIDEODRIVER.getGuiScale().percent() == 100
```
Not really.

libs/driver/include/driver/GuiScale.h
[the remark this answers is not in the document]
falbrechtskirchinger: > At least put the overloads next to each other for discoverability.

No objections here.

> You have 2 overloads: one working on a single value and one on a pair of values. From the code here I cannot tell that `foo(pt)` and `Position(foo(pt.x), foo(pt.y))` do the same thing.

Why though? Multiplying a point (aka. vector) and a scalar is a perfectly well-defined and commonly understood operation.

With casts, this becomes less readable to me:
```cpp
return R(screenToView(static_cast<float>(pt.x)), screenToView(static_cast<float>(pt.y)));
````

And rewriting the scalar overloads is dangerous, as conceptually, a point is more likely to be in a value range that is clearly representable by `float`.
Scalars can represent anything and so I like keeping the cast and the responsibility for it at the call site.

Without the casts:
```cpp
return R(screenToView(pt.x), screenToView(pt.y));
````
This looks fine to me, but requires the dangerous change below:

```cpp
/// Translate a scalar from view to screen space
template<typename R = float, typename T, std::enable_if_t<std::is_arithmetic_v<T>, int> = 0>
constexpr R viewToScreen(T val) const noexcept
{
    if constexpr(std::is_integral_v<R>)
        return helpers::iround<R>(static_cast<float>(val) * scaleViewToScreen_);
    else
        return static_cast<R>(val * scaleViewToScreen_);
}
```

So ultimately, I'd still only group the overloads and keep the rest as-is. :man_shrugging:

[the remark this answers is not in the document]
Flamefire: > Why though? Multiplying a point (aka. vector) and a scalar is a perfectly well-defined and commonly understood operation.

`R(PointF(pt) * scaleScreenToView_) ==  R(screenToView(pt.x), screenToView(pt.y))` isn't immediately obvious due to the rounding when written more accurately: `R(screenToView<R::element_type>(pt.x)...`) 

Anyway your cast argument makes sense although `PointF(pt)` has the exact same issue although yes, it is "less likely". So I guess it is not worth the inconvenience on use.

> So ultimately, I'd still only group the overloads and keep the rest as-is. 🤷‍♂️

Agreed.   
I'd then like to have a test clearly stating `viewToScreen(pt) == Position(viewToScreen<int>(pt.x), viewToScreen<int>(pt.y))` with some random `pt` (via `test::random::randomPoint` or so)

libs/driver/include/driver/VideoDriver.h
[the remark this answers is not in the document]
falbrechtskirchinger: Dang it. My original response didn't make it.

Long story short. It is happening. But the fact that this causes confusion makes me want to refactor it.
The code has organically evolved by nudging existing functions to do new things, and it's time to rethink this part of the code from the ground up instead of shoe-horning it.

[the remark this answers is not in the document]
falbrechtskirchinger: Also, thinking about terminology, `viewRenderSize_` and `screenRenderSize_` don't make sense. All render sizes are always in screen coordinates. I should use `renderSize_` and `scaledRenderSize_` or similar.

Or, more accurately, the render size isn't even measured in pixels but points. But no need to make things more confusing.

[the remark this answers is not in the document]
falbrechtskirchinger: The easiest solution is to stop trying to be clever and duplicate the calculations in `SetNewSize()` and `setGuiScale()`. Both functions need things done slightly differently, so implementing one via the other is more complicated.

libs/s25main/drivers/VideoDriverWrapper.cpp
falbrechtskirchinger: As I'm about to rename it, `GuiScaler` still sounds wrong to me in context. It implies that this class is doing the actual scaling of the UI when that is accomplished by setting the viewport and projection accordingly. The logic that defining operations on `GuiScale` should turn it into `GuiScaler` doesn't quite track when, for example, adding addition operations to `Vector2D` wouldn't turn it into `Vector2DAdder` – it's still just a vector. A class that accepts two or more vectors and adds them would be different.

When I read `mouse_xy.pos = getGuiScaler().screenToView(Position(ev.button.x, ev.button.y));` `GuiScaler` sounds like a much more complicated entity than a simple wrapper of pre-computed values with some operations that are bundled for convenience. `GuiScale` just hides the implementation detail that division by the scaling factor is done by multiplying with the reciprocal. And with the functions, users don't have to think about which direction requires multiplying or dividing by the scaling factor.

libs/s25main/desktops/dskOptions.cpp
falbrechtskirchinger: Yeah, I assume `SetSelection()` won't allow me to go past the number of items.

libs/s25main/drivers/VideoDriverWrapper.cpp
falbrechtskirchinger: If it's just the asymmetry in the getter and setter that's bothering you, might I suggest:
```cpp
void setGuiScalePercent(unsigned);
```

Flamefire: The whole thing was just a suggestion how the asymmetry might be (less) confusing. So I'm fine with keeping it.

However I like your idea of `setGuiScalePercent`. That is even better as it includes the unit in the name.

libs/driver/include/driver/VideoDriver.h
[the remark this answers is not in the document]
Flamefire: Not sure I can follow. Where is `viewRenderSize_` updated? I only see `viewRenderSize_ = guiScale_.screenToView<Extent>(renderSize);` in `SetNewSize`

Would it make sense to simply not store/cache `viewRenderSize_` and calculate it on-the-fly in `GetRenderSize`?

Or even the other way round: `screenRenderSize_` is only used internally isn't it? Maybe rename that to `unscaledRenderSize_` or `logicalRenderSize_`(?) or don't store it at all and calculate from `viewRenderSize_` when required.   
On the other hand: When changing the scale you'd need to rescale `viewRenderSize_` anyway, so might as well store the "requested" (logical?) render size.

Damn, that stuff gets confusing ;-)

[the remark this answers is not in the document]
falbrechtskirchinger: It used to be updated this way: `onGuiScaleChanged()` -> `SDL2Video::UpdateCurrentSizes()` -> `VideoDriver::SetNewSize()`

Anyway, I've simplified it already and am handling a previously unhandled case (GUI scale re-calculates if DPI changes). Let's see what you think.

Caching it does make sense IMO, as it's quite frequently used.

falbrechtskirchinger:
Disclaimer: I can't do a functional test of the changes right now (graphics driver was updated and I can't reboot at the moment).

libs/s25main/Settings.h
```diff
@@ -63,6 +63,7 @@ class Settings : public Singleton<Settings, SingletonPolicies::WithLongevity>
         bool fullscreen;
         bool vbo;
         bool shared_textures;
+        unsigned guiScale;
```
Flamefire, line 66:
```suggestion
        unsigned guiScale; /// in %, 0="automatic"
```

libs/s25main/desktops/dskOptions.cpp
```diff
@@ -29,6 +30,7 @@
 #include "ogl/FontStyle.h"
 #include "s25util/StringConversion.h"
 #include "s25util/colors.h"
+#include <boost/format.hpp>
```
Flamefire, line 33: Can be removed now?

libs/s25main/desktops/dskOptions.cpp
```diff
@@ -499,6 +508,13 @@ void dskOptions::Msg_Group_ComboSelectItem(const unsigned group_id, const unsign
             VIDEODRIVER.setTargetFramerate(SETTINGS.video.framerate);
             break;
         case ID_cbVideoDriver: SETTINGS.driver.video = combo->GetText(selection); break;
+        case ID_cbGuiScale:
+            if(selection == 0)
+                SETTINGS.video.guiScale = 0;
+            else
+                SETTINGS.video.guiScale = guiScales_[selection - 1];
```
Flamefire, line 515: To avoid the special case and confusing difference between "selection" and "index", I'd suggest to add a zero entry at the beginning of `guiScales` and then:

```suggestion
            SETTINGS.video.guiScale = guiScales_[selection];
```

libs/s25main/desktops/dskOptions.cpp
```diff
@@ -663,3 +689,89 @@ void dskOptions::loadVideoModes()
...
+void dskOptions::scrollGuiScale(bool up)
+{
+    auto* group = GetCtrl<ctrlGroup>(ID_grpGraphics);
+    auto* combo = group->GetCtrl<ctrlComboBox>(ID_cbGuiScale);
+    const auto& selection = combo->GetSelection();
+    unsigned newSelection = 0;
+    if(!selection || *selection == 0) // No selection or "Auto" item selected
+        newSelection = recommendedGuiScaleIndex_ + 1;
+    else
+        newSelection = std::clamp(*selection + (up ? 1 : -1), 0u, static_cast<unsigned>(guiScales_.size()));
```
Flamefire, line 769: Previously you didn't allow scrolling to the zero index, i.e. "auto". Was that change intentional? IMO it might be confusing as it will be  a large jump.

```suggestion
        newSelection = std::clamp(*selection + (up ? 1 : -1), 1u, static_cast<unsigned>(guiScales_.size()));
```

libs/s25main/desktops/dskOptions.cpp
```diff
@@ -663,3 +689,89 @@ void dskOptions::loadVideoModes()
...
+    constexpr auto stepSize = 10u;
+    const auto roundGuiScale = [=](unsigned percent) {
+        return helpers::iround<unsigned>(static_cast<float>(percent) / stepSize) * stepSize;
+    };
+
+    unsigned minPercent, maxPercent, recommendedPercent;
+    VIDEODRIVER.getGuiScaleRange(&minPercent, &maxPercent, &recommendedPercent);
+    const auto recommendedPercentRounded = roundGuiScale(recommendedPercent);
+    auto* group = GetCtrl<ctrlGroup>(ID_grpGraphics);
+    auto* combo = group->GetCtrl<ctrlComboBox>(ID_cbGuiScale);
```
Flamefire, line 731: Any reason not to use the following as `group` is only used in this single line:

```suggestion
    auto* combo = GetCtrl<ctrlGroup>(ID_grpGraphics)->GetCtrl<ctrlComboBox>(ID_cbGuiScale);
```

Flamefire (changes requested):

libs/s25main/desktops/dskOptions.cpp
falbrechtskirchinger: That was unintentional.

libs/s25main/Settings.h
falbrechtskirchinger: FYI, `///` attributes the documentation comment to the following item. Use `///<` for trailing comments (see https://www.doxygen.nl/manual/docblocks.html#memberdoc), which reminds me that I wanted to update the affected comments in this PR.

Flow86: yep, `///< xxx` is the right way

libs/s25main/desktops/dskOptions.cpp
```diff
@@ -663,3 +685,89 @@ void dskOptions::loadVideoModes()
...
+}
+
+void dskOptions::scrollGuiScale(bool up)
+{
+    auto* group = GetCtrl<ctrlGroup>(ID_grpGraphics);
+    auto* combo = group->GetCtrl<ctrlComboBox>(ID_cbGuiScale);
+    const auto& selection = combo->GetSelection();
+    unsigned newSelection = 0;
+    if(!selection || *selection == 0) // No selection or "Auto" item selected
+        newSelection = recommendedGuiScaleIndex_ + 1;
```
Flamefire, line 763:
```suggestion
        newSelection = recommendedGuiScaleIndex_;
```

This is now a 1:1 match, isn't it?

libs/s25main/desktops/dskOptions.cpp
```diff
@@ -663,3 +685,89 @@ void dskOptions::loadVideoModes()
...
+    else
+        newSelection = std::clamp(*selection + (up ? 1 : -1), 1u, static_cast<unsigned>(combo->GetNumItems() - 1));
+
+    if(newSelection != selection)
+    {
+        combo->SetSelection(newSelection);
+        SETTINGS.video.guiScale = guiScales_[newSelection - 1];
```
Flamefire, line 770: Same here:
```suggestion
        SETTINGS.video.guiScale = guiScales_[newSelection];
```

Flamefire (changes requested):

libs/s25main/desktops/dskOptions.cpp
falbrechtskirchinger: Indeed. I'll notice all these issues once I can actually launch and test it.

falbrechtskirchinger:
OK, the code seems to work as expected. I didn't find anything else beyond the selection/index issue.

If this iteration gets the seal of approval, remember to let me do a final rebase to clean up history before merging.

libs/driver/src/VideoDriver.cpp
```diff
@@ -73,4 +76,60 @@ void VideoDriver::SetNewSize(VideoMode windowSize, Extent renderSize)
...
     renderSize_ = renderSize;
+
+    const auto ratioXY = PointF(renderSize_) / PointF(windowSize_.width, windowSize_.height);
+    dpiScale_ = (ratioXY.x + ratioXY.y) / 2.f; // use the average ratio of both axes
+
+    if(autoGuiScale_)
+    {
+        unsigned percent;
+        getGuiScaleRange(nullptr, nullptr, &percent);
+        guiScale_ = GuiScale(percent);
```
Flamefire, line 87: Triggered by the codecov annotation IMO it makes sense to add a short test case explaining & showing what effect `autoGuiScale_` is supposed to have. Is that easily possible?

libs/driver/src/VideoDriver.cpp
```diff
@@ -73,4 +76,60 @@ void VideoDriver::SetNewSize(VideoMode windowSize, Extent renderSize)
...
+    scaledRenderSize_ = guiScale_.screenToView<Extent>(renderSize_);
+    CallBack->WindowResized();
+
+    // move cursor to new position in view space
+    // must happen after window resize event to avoid drawing spurious hover events
+    mouse_xy.pos = guiScale_.screenToView(screenPos);
+    CallBack->Msg_MouseMove(mouse_xy);
+}
+
+void VideoDriver::getGuiScaleRange(unsigned* minPercent, unsigned* maxPercent, unsigned* recommendedPercent) const
```
Flamefire, line 115: I was checking this noting that I didn't fully understood it initially. I'd rather have this return a simple struct containing min/max/recommended percent than out-parameters passed as pointers where you don't know if the original values matter until you read the code twice.

Maybe even split out the recommended part into an extra function which would even simplify https://github.com/Return-To-The-Roots/s25client/pull/1594/files#diff-334108d4fb1ef8a8a66c17c2706fc8f30d515828a6d739e53d2e819d842835ddR595-R598

tests/s25Main/UI/testVideoDriver.cpp
```diff
@@ -100,3 +102,121 @@ BOOST_FIXTURE_TEST_CASE(CreateAndDestroyTextures, uiHelper::Fixture)
...
+    // Regular DPI configuration
+    driver->getGuiScaleRange(&minPercent, &maxPercent, &recommendedPercent);
+    BOOST_TEST(minPercent == 100u);
+    BOOST_TEST(maxPercent == 300u);
+    BOOST_TEST(recommendedPercent == 100u);
+
+    // Maximum is constrained by width
+    driver->ResizeScreen(VideoMode(800 * 2, 600 * 3), false);
+    driver->getGuiScaleRange(nullptr, &maxPercent, nullptr);
+    BOOST_TEST(maxPercent == 200u);
```
Flamefire, line 208: This could look like:
```suggestion
    BOOST_TEST(driver->getGuiScaleRange().max == 200u);
```

I think that's easier to read/use, isn't it?

Flamefire (changes requested):
Looking almost finished. A test explaining and asserting the behavior the auto-scale should have would be nice, if this isn't to much work.

Bonus points for adding a small test for the dskOptions that would have caught the index off-by-one issue you just fixed. But I know tests for those classes are often a nightmare so not required.

If those are to much work feel free to rebase and I'll accept it without those tests.    
So just the range-get function return the values as a struct and possibly split it (IMO it makes sense to have `getRange` and `getRecommended` and not mix those although the latter may call the former. But feel free to keep it as a single function)

Thanks a lot! :)

libs/driver/src/VideoDriver.cpp
falbrechtskirchinger: This OK?

```cpp
// in GuiScale.h
struct GuiScaleRange {
    unsigned minPercent; // wanted to include the unit somewhere and picked the struct members
    unsigned maxPercent;
    unsigned recommendedPercent;
};

// range function is part of the interface
virtual GuiScaleRange VideoInterface::getGuiScaleRange() const;

// additional convenience function only exists in the wrapper
unsigned VideoDriverWrapper::getRecommendedGuiScalePercent() const {
    return driver->getGuiScaleRange().recommendedPercent;
}
```

libs/driver/src/VideoDriver.cpp
falbrechtskirchinger: Can do.

falbrechtskirchinger:
> Bonus points for adding a small test for the dskOptions that would have caught the index off-by-one issue you just fixed. But I know tests for those classes are often a nightmare so not required.

I'm already planning a small follow-up PR anyway (multi-monitor DPI awareness). The tests aren't related but I'd *maybe* include them then. I can think of more things that could be tested in `dskOptions`.

libs/driver/src/VideoDriver.cpp
Flamefire: Yeah looks good!   
This way I'd even omit the `getRecommendedGuiScalePercent` as it doesn't provide  much benefit if `getGuiScaleRange` has it has a member.

falbrechtskirchinger:
@Flamefire The rebase is done. Once you approve the changes a force-push *should* happen automatically within about a minute.
I've added a few more doc comments and applied one of your suggestions to another line.

Flamefire (approved):
LGTM!

Flamefire (approved):
Thanks a lot for your effort!

Blackpearl1:
I just used the newest nightly(e37a322)

I change the GUI zoom to 200% and.... nothing happen.
Restartet the game and... nothing happen.

I Use 3840x2160 (16:9)

Whats wrong?

falbrechtskirchinger:
> I just used the newest nightly([e37a322](https://github.com/Return-To-The-Roots/s25client/commit/e37a322209cffb00fda4afd50c87b8f7e8261083))
> 
> I change the GUI zoom to 200% and.... nothing happen. Restartet the game and... nothing happen.
> 
> I Use 3840x2160 (16:9)
> 
> Whats wrong?

Changing the GUI scale in the options screen should have an immediate effect.
I need more information:
* What's your OS?
* Are you using the SDL2 video backend/driver? (WinAPI is **not** supported.)
* Did you get any warning messages when leaving the options screen?
* Could you post screenshots of the options screen with two different GUI scales selected?

The most likely scenario is that you're using the WinAPI backend. If you didn't get a warning message when leaving the options screen, I made some mistake in determining whether the selected GUI scale is supported.

Blackpearl1:
@falbrechtskirchinger 

> Are you using the SDL2 video backend/driver? (WinAPI is not supported.)
That was my fault.

I used the win...whatever

I changed it to SDL2 and now it works like a charm. Thank you :D nice job.

The Message that it wont work with the win...whatever comes after changing to SDL2...

PR #1595, opened 2023-07-06
CI: Bump Boost version to 1.82.0

falbrechtskirchinger:
CI (MacOS) fails due to Boost 1.82.0 being installed and 1.81.0 being expected.


Flamefire:
Thanks for the PR!   
We still need to test what we can with Boost 1.81. So I added another PR changing only that single job and also fixing the similar failure on Appveyor

falbrechtskirchinger:
Great. I was about to look at the AppVeyor failure myself.

PR #1600, opened 2023-07-09
Fix cursor overlapping tooltips

falbrechtskirchinger:
Tooltips are placed at a consistent distance and the cursor doesn't overlap.

Before:
(Excuse the screen tear. Didn't catch that when I took the screenshots.)
fix_tooltip_overlap_before

After:
fix_tooltip_overlap_after


Flamefire (approved):
Makes sense, thanks!

PR #1601, opened 2023-07-09
Improve window positioning when cycling sizes

falbrechtskirchinger:
Remember the last window position set by a move and try to restore it after a resize. This makes windows stick to the screen corners when resized.

The images below depict the window position between clicking the resize button.

Before:
improved_positioning_old

After:
improved_positioning_new


libs/s25main/ingameWindows/iwObservate.cpp
```diff
@@ -143,11 +143,6 @@ void iwObservate::Msg_ButtonClick(const unsigned ctrl_id)
             for(unsigned i = 1; i <= 4; ++i)
                 GetCtrl<ctrlImageButton>(i)->SetPos(
                   DrawPoint(GetCtrl<ctrlImageButton>(i)->GetPos().x - diff, GetSize().y - 50));
-
-            DrawPoint maxPos(VIDEODRIVER.GetRenderSize() - GetSize() - Extent::all(1));
```
Flamefire, line 147: Why was this removed? This is required if you have the window at right or bottom border and increase its size

libs/s25main/Settings.cpp
```diff
@@ -347,8 +347,10 @@ void Settings::LoadIngame()
             const auto* iniWindow = static_cast<const libsiedler2::ArchivItem_Ini*>(settingsIngame.find(window.second));
             if(!iniWindow)
                 continue;
-            windows.persistentSettings[window.first].lastPos.x = iniWindow->getIntValue("pos_x");
-            windows.persistentSettings[window.first].lastPos.y = iniWindow->getIntValue("pos_y");
+            const auto x = windows.persistentSettings[window.first].lastPos.x = iniWindow->getIntValue("pos_x");
+            const auto y = windows.persistentSettings[window.first].lastPos.y = iniWindow->getIntValue("pos_y");
+            windows.persistentSettings[window.first].restorePos.x = iniWindow->getValue("restore_pos_x", x);
+            windows.persistentSettings[window.first].restorePos.y = iniWindow->getValue("restore_pos_y", y);
```
Flamefire, line 353: It might make sense to reduce the duplication a bit here:
```suggestion
            const auto lastPos = windows.persistentSettings[window.first].lastPos = DrawPoint(iniWindow->getIntValue("pos_x"), iniWindow->getIntValue("pos_y"));
            windows.persistentSettings[window.first].restorePos = DrawPoint(iniWindow->getValue("restore_pos_x", lastPos.x), iniWindow->getValue("restore_pos_y", lastPos.y));
```

The whole idea of the `Point` class is to not have separated x,y-pairs anymore as those may easily lead to mistakes

Flamefire (changes requested):

libs/s25main/ingameWindows/iwObservate.cpp
falbrechtskirchinger: No? This is already done by `IngameWindow` and simply duplicated here, which breaks the newly introduced behavior.

libs/s25main/ingameWindows/IngameWindow.h
```diff
@@ -71,7 +71,7 @@ class IngameWindow : public Window
     DrawPoint GetRightBottomBoundary();
 
     /// Set the position for the window after adjusting newPos so the window is in the visible area
-    void SetPos(DrawPoint newPos);
+    void SetPos(DrawPoint newPos, bool saveRestorePos = true);
```
falbrechtskirchinger, line 74: How do you feel about this publicly exposed parameter? I was considering hiding it in a private function. I don't think it's useful outside of `IngameWindow`.

libs/s25main/ingameWindows/iwObservate.cpp
Flamefire: Ah you are right!

libs/s25main/ingameWindows/IngameWindow.h
Flamefire: Not strongly either way as I'm not sure if it is worth adding an extra private function for this. Might even lead to possible issues.

falbrechtskirchinger: I'm worried about potential issues too. But because an internal feature is now accidentally usable from the outside. :man_shrugging:

falbrechtskirchinger: Guess we'll find out.

libs/s25main/Settings.cpp
falbrechtskirchinger: Makes sense.

Flamefire:
Looks good but upon a final look it isn't immediately obvious why we have `lastPos` and `restorePos`, i.e. their difference.

1. Isn't the map window the only one which can be resized, i.e. for every other window the 2 positions are always the same? 
2. Can't we fold `restorePos` into `lastPos`? I mean the main change here is the edge-handling when storing the position, isn't it?

So wouldn't we get the same effect by storing the value stored into `restorePos` into `lastPos` instead and removing `restorePos`?

Spikeone:
>     1. Isn't the map window the only one which can be resized, i.e. for every other window the 2 positions are always the same?
You can resize the observation window as well

falbrechtskirchinger:
> So wouldn't we get the same effect by storing the value stored into `restorePos` into `lastPos` instead and removing `restorePos`?

The code evolved quite a bit, so let me think about that a bit. I know there were edge cases that didn't work initially.

falbrechtskirchinger:
I removed the `restorePos` member forgetting that the observation window doesn't persist. The fix is to bring that member back, then we can remove the setting and update `lastPos` accordingly instead.

Hopefully, there isn't another edge case I'm forgetting. This seemingly trivial feature isn't quite so easy to get right.

falbrechtskirchinger:
> 1. Isn't the map window the only one which can be resized, i.e. for every other window the 2 positions are always the same?

In addition to the map and observation window, every non-modal window can be shaded/minimized, which is also a resize.

Flamefire:
> In addition to the map and observation window, every non-modal window can be shaded/minimized, which is also a resize.

So currently moving a window to a border and minimizing it (I prefer that word as it seems more conventional for windows) would move the minimized window even closer to the border?   
Does this really make sense?

In any case: The actual behavior change intended here is clipping the window to borders if they were clipped before, correct? I.e. the change in [`SetIwSize`](https://github.com/Return-To-The-Roots/s25client/pull/1601/files#diff-e8829f77b7d4215f69ca07f04cc1f57f8b90097254156db6437152c9591ad7f5R95)

Can't we have it simpler: we get an adjusted position there before the resize which will clip the window correctly again after the resize and not store anything in the config? And/or to persist the clipping over restarts do the modification to `newRestorePos` to `lastPos` instead? We could even use values of `0` and `INT_MAX` for the left/top and right/bottom positions to ensure that the restored window after resolution change is still clipped.

Or am I misunderstanding anything?

falbrechtskirchinger:
> So currently moving a window to a border and 

> minimizing it (I prefer that word as it seems more conventional for windows)

Windows® (capital W, registered trademark), sure. Mac and Linux users – depending on the desktop environment – will disagree with you. I have known the function of rolling the window into its title bar as "shade" since the late 90s when I was using MacOS (pre-MacOS X, so, may not be a thing today). It's still available on KDE, though.
And being a pedantic a\*\*hole, I take great offense at calling it "minimize" when there's a more accurate term. :wink:
The BlueByte devs – going by the icon name – seem to have called the feature "iconify". They didn't know either…

> would move the minimized window even closer to the border? Does this really make sense?

I think so. But I concede that this is a less straightforward and more opinionated feature than I assumed.

> In any case: The actual behavior change intended here is clipping the window to borders if they were clipped before, correct? I.e. the change in [`SetIwSize`](https://github.com/Return-To-The-Roots/s25client/pull/1601/files#diff-e8829f77b7d4215f69ca07f04cc1f57f8b90097254156db6437152c9591ad7f5R95)

Not quite. Critically, it's about restoring the position the user last dragged the window to. Which is subtly different for windows with more than two resize states (map and observation window).

If a window is placed close to an edge, but not yet in clipping distance, following a resize that puts it in clipping distance `lastPos` and `restorePos` diverge. On the next resize information is lost when only keeping one value.

> Can't we have it simpler: we get an adjusted position there before the resize which will clip the window correctly again after the resize and not store anything in the config? And/or to persist the clipping over restarts do the modification to `newRestorePos` to `lastPos` instead? We could even use values of `0` and `INT_MAX` for the left/top and right/bottom positions to ensure that the restored window after resolution change is still clipped.

`INT_MAX` sounds like a good idea and I have assigned `newRestorePos` to `lastPos`, as you suggested, which results in (IMO) incorrect behavior. 

> Or am I misunderstanding anything?

In summary, yes. Even though it's only relevant for one window (the observation window isn't persistent and storing `restorePos` as a member works), to get the desired behavior, we'd need to save `restorePos` to the settings.

Let's hear some more opinions. Do we want windows to snap to edges – including when "minimized" (ugh) – and do we care about the very subtle difference I've explained?

Flamefire:
> I have known the function of rolling the window into its title bar as "shade" since the late 90s when I was using MacOS (pre-MacOS X, so, may not be a thing today). It's still available on KDE, though.

Didn't know that, only went by the only name I've read for this and what was used throughout this project before. So thanks for the explanation where this name comes from. Makes sense as "minimized" may mean "smaller" which is more general than the folding into the title bar.

> > Or am I misunderstanding anything?
> 
> In summary, yes. Even though it's only relevant for one window (the observation window isn't persistent and storing `restorePos` as a member works), to get the desired behavior, we'd need to save `restorePos` to the settings.

I see. Quite a lot of edge cases here, so thanks for tackling that!

falbrechtskirchinger:
Alright. I've implemented the "correct" behavior as envisioned and greatly expanded the unit test to cover the edge cases.

Bonus: `Point` gained two members `Min/MaxElementValue` to simplify 9 uses of `std::numeric_limits<Point::ElementType>`.

 I'll do one more round of manual testing and mark this PR as ready for review if all goes well. *\*knock on wood\**

tests/s25Main/UI/testIngameWindow.cpp
```diff
@@ -228,4 +232,169 @@ BOOST_AUTO_TEST_CASE(SaveAndRestoreMinimized)
...
+              BOOST_TEST(wnd.GetPos() == DrawPoint(renderSize - wndSizeL));
+              BOOST_TEST(wnd.GetSize() == wndSizeL);
+          },
+          [&]() {
+              BOOST_TEST(wnd.GetPos() == renderSize - DrawPoint(wndSizeL.x, minHeight));
+              BOOST_TEST(wnd.GetSize() == Extent(wndSizeL.x, minHeight));
+          });
+
+        WindowPositioning_testOne(
+          wnd, "Resize down (S), window no longer connects with screen edges and should move to restorePos",
```
Flamefire, line 387:
```suggestion
          wnd, "Decrease size, window no longer connects with screen edges and should move to restorePos",
```

falbrechtskirchinger:
I had to add braces for formatting (IfMacros is a clang-format 13 addition). Also, `BOOST_TEST_CONTEXT` expands to an `if`. So your initial suggestion didn't quite work, but thanks for pointing me in the right direction. I'm a Catch2/doctest user myself.

Improved output with `BOOST_TEST(false)`:
```
tests/s25Main/UI/testIngameWindow.cpp(398): info: check wnd.GetSize() == Extent(wndSizeS.x, minHeight) has passed
Assertion occurred in a following context:
    Non-persisted window, posAtMouse
    Decrease size, window no longer connects with screen edges and should move to restorePos
    Minimized
tests/s25Main/UI/testIngameWindow.cpp(399): error: in "IngameWindows/WindowPositioning": check false has failed
Failure occurred in a following context:
    Non-persisted window, posAtMouse
    Decrease size, window no longer connects with screen edges and should move to restorePos
    Minimized
```

tests/s25Main/UI/testIngameWindow.cpp
```diff
@@ -228,4 +232,171 @@ BOOST_AUTO_TEST_CASE(SaveAndRestoreMinimized)
     }
 }
 
+namespace {
+void WindowPositioning_testOne(IngameWindow& wnd, const char* msg, const std::function<void()>& main,
+                               const std::function<void()>& minimize)
```
Flamefire, line 237: Maybe the param names can be cleaner:
```suggestion
void WindowPositioning_testOne(IngameWindow& wnd, const char* context, const std::function<void()>& checkNormal,
                               const std::function<void()>& checkMinimized)
```

And with the current usage it might be even better to move the context out of this function, i.e. have `BOOST_TEST_CONTEXT` at the call site instead of passing it as a param. But up to you.

Flamefire (commented):
> I had to add braces for formatting (IfMacros is a clang-format 13 addition). Also, `BOOST_TEST_CONTEXT` expands to an `if`. So your initial suggestion didn't quite work, but thanks for pointing me in the right direction. I'm a Catch2/doctest user myself.

A true. I meant to omit the semicolon too: `BOOST_TEST_CONTEXT("Before minimize/un-minimize") main();` would also work (best with a newline but that might be confusing), but let's not get too detailed.

> Edit: And clang-format-17 formats differently than v10. :-(

Note that there is a build target "clangFormat" which formats modified files (e.g. `make clangFormat -j`) which is created if the correct clang-format version is found. Maybe that helps.

What is the `BOOST_INFO` in front of the `SetMinimized` for? There is no assertion inside the function and the next check is guarded by `BOOST_TEST_CONTEXT`. So I'd rather remove that

tests/s25Main/UI/testIngameWindow.cpp
falbrechtskirchinger: I prefer keeping it in the function. Waiting for `apt.llvm.org/focal` to come back before I push.

falbrechtskirchinger:
> > Edit: And clang-format-17 formats differently than v10. :-(
>  
> Note that there is a build target "clangFormat" which formats modified files (e.g. `make clangFormat -j`) which is created if the correct clang-format version is found. Maybe that helps.

I'm using that already, thanks! My default is just to format from the editor using clang-format in PATH and it usually produces compatible results.

> What is the `BOOST_INFO` in front of the `SetMinimized` for? There is no assertion inside the function and the next check is guarded by `BOOST_TEST_CONTEXT`. So I'd rather remove that

I think I meant to use `BOOST_TEST_MESSAGE` there. I just wanted something in the log about these key steps.

Flamefire (approved):

Flamefire:
Woops, I wanted to enable auto-merge as Appveyor is so slow. Well, I guess/hope it is fine ;-)

Thanks for your efforts!

Issue #889, opened 2018-11-01
Can't set windowed resolution


Spikeone:
Guess the image says it all: If you set the video mode to windowed, you can not change the screen resolution anymore as the option only applies for the fullscreen resolution. When switching to windowed mode one should be able to set the windowed resolution in the options menu.
![image](https://user-images.githubusercontent.com/1276342/47845630-d6646680-ddc5-11e8-9640-dcff96d7bcb0.png)

Setting the size manually works but once you open the settings the size is reverted to 800x600.

Flamefire:
Just tested: When you close the game normally the setting is kept. Only when you ALT+F4 it (or press red X) it is not stored.

Spikeone:
Well, wenn streaming I'd really like to keep the aspect ratio and screensize I set. Just add another dropdown for the windowed resolution? At least for me it is quite odd that one can not set a fixed windowed size.

PR #1602, opened 2023-07-10
Add support for additional display modes

### LINKED ISSUE #1512: Lock window size
I like to play windowed, sadly you can resize the window when misclicking. As you can only change the window size in the config, you have to open that file and restore the size. It'd be nice to lock the size so resizing is disabled.

### LINKED ISSUE #1615: Status of WinAPI backend and querying capabilities
What is the official status of the WinAPI video backend?

I've mostly ignored it in my PRs and the feature set continues to diverge. When new features are exposed in the options screen, users should be informed if selected options are ineffective or not able to select them in the first place.
Is it worth adding a `GetCapabilities()` function to `IVideoDriver`?

It could look something like this:
```cpp
// Bitset details to be figured out separately
enum class DisplayMode : unsigned {
	None,
	Window = …,
	NonResizableWindow = …,
	Fullscreen = ...,
	WindowedFullscreen = …
};

struct VideoDriverCapabilities {
	DisplayMode displayMode{};
	bool guiScale = false;
	// someday maybe?
	// bool copy;
	// bool paste;
};

class IVideoDriver {
	virtual VideoDriverCapabilities GetCapabilities() const = 0;
};

class VideoSDL2 {
	VideoDriverCapabilities GetCapabilities() const override {
		VideoDriverCapabilities cap{};
		cap.displayMode = DisplayMode::Window | DisplayMode::NonResizableWindow | DisplayMode::Fullscreen | DisplayMode::WindowedFullscreen;
		cap.guiScale = true;
		return cap;
	}
};

class VideoWinAPI {
	VideoDriverCapabilities GetCapabilities() const override {
		VideoDriverCapabilities cap{};
		cap.displayMode = DisplayMode::Window | DisplayMode::Fullscreen;
		return cap;
	}
};
```

Further, to reduce duplication between `dskOptions` and `iwSettings`:
```cpp
struct DisplayModeItem {
	std::string text;
	DisplayMode value;
};

std::vector<DisplayModeItem> displayModeItemsFromCapabilities(const VideoDriverCapabilities& cap);
```
The `DisplayModeItem`s would then be used to populate a ComboBox and map the selection index to a `DisplayMode` value.

falbrechtskirchinger:
Implement additional display modes in SDL2 and WinAPI backends:
* Windowed fullscreen
* Windowed (non-resizable)

Fixes #1512
Closes #1615

To-do:
- [ ] Rebase after #1612.
- [ ] Rework options screen and in-game settings window.

<details>
<summary>Notes and references:</summary>

* https://learn.microsoft.com/en-us/windows/win32/winmsg/window-styles
* https://src.chromium.org/viewvc/chrome/trunk/src/ui/views/win/fullscreen_handler.cc
</details>

### DIFF
--- a/extras/videoDrivers/WinAPI/WinAPI.cpp
+++ b/extras/videoDrivers/WinAPI/WinAPI.cpp
@@ -129,26 +129,25 @@ void VideoWinAPI::CleanUp()
 /**
  *  Erstellt das Fenster mit entsprechenden Werten.
  *
- *  @param[in] width      Breite des Fensters
- *  @param[in] height     Höhe des Fensters
- *  @param[in] fullscreen Vollbildmodus ja oder nein
+ *  @param[in] width       Breite des Fensters
+ *  @param[in] height      Höhe des Fensters
+ *  @param[in] displayMode Fullscreen on/off, window resizable?
  *
  *  @return @p true bei Erfolg, @p false bei Fehler
  *
  *  @bug Hardwarecursor ist bei Fenstermodus sichtbar,
  *       Cursor deaktivieren ist fehlerhaft
  */
-bool VideoWinAPI::CreateScreen(const std::string& title, const VideoMode& newSize, bool fullscreen)
+bool VideoWinAPI::CreateScreen(const std::string& title, const VideoMode& newSize, DisplayMode displayMode)
 {
     if(!initialized)
         return false;
 
-    if(!RegisterAndCreateWindow(title, newSize, fullscreen))
+    if(!RegisterAndCreateWindow(title, newSize, displayMode))
         return false;
 
-    if(fullscreen && !MakeFullscreen(GetWindowSize()))
+    if(bitset::isSet(displayMode, DisplayMode::Fullscreen) && !MakeFullscreen(GetWindowSize()))
         return false;
-    isFullscreen_ = fullscreen;
 
     if(!InitOGL())
         return false;
@@ -174,30 +173,31 @@ bool VideoWinAPI::CreateScreen(const std::string& title, const VideoMode& newSiz
  *
  *  @todo Vollbildmodus ggf. wechseln
  */
-bool VideoWinAPI::ResizeScreen(const VideoMode& newSize, bool fullscreen)
+bool VideoWinAPI::ResizeScreen(const VideoMode& newSize, DisplayMode displayMode)
 {
     if(!initialized || !isWindowResizable)
         return false;
 
-    if(isFullscreen_ == fullscreen && newSize == GetWindowSize())
+    const auto fullscreen = bitset::isSet(displayMode, DisplayMode::Fullscreen);
+    if(IsFullscreen() == fullscreen && newSize == GetWindowSize())
         return true;
 
     ShowWindow(screen, SW_HIDE);
 
     VideoMode windowSize = fullscreen ? FindClosestVideoMode(newSize) : newSize;
     // Try to switch full screen first
-    if(isFullscreen_ && !fullscreen)
+    if(IsFullscreen() && !fullscreen)
     {
         if(ChangeDisplaySettings(nullptr, 0) != DISP_CHANGE_SUCCESSFUL)
             return false;
-    } else if(isFullscreen_ || fullscreen)
+    } else if(IsFullscreen() || fullscreen)
     {
         if(!MakeFullscreen(windowSize))
             return false;
     }
 
     // Fensterstyle ggf. ändern
-    std::pair<DWORD, DWORD> style = GetStyleFlags(isFullscreen_);
+    std::pair<DWORD, DWORD> style = GetStyleFlags(IsFullscreen());
     SetWindowLongPtr(screen, GWL_STYLE, style.first);
     SetWindowLongPtr(screen, GWL_EXSTYLE, style.second);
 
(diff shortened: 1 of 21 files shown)


falbrechtskirchinger:
@Spikeone There you go!

extras/videoDrivers/WinAPI/WinAPI.cpp
```diff
@@ -253,7 +254,7 @@ RECT VideoWinAPI::CalculateWindowRect(bool fullscreen, VideoMode& size) const
     return wRect;
 }
 
-bool VideoWinAPI::RegisterAndCreateWindow(const std::string& title, const VideoMode& wndSize, bool fullscreen)
+bool VideoWinAPI::RegisterAndCreateWindow(const std::string& title, const VideoMode& wndSize, WindowMode windowMode)
```
Flow86, line 257: I think you accidently removed `bool fullscreen` here as last parameter

falbrechtskirchinger: Yes, thanks! I did see the AppVeyor failure. I had initially folded the `fullscreen` parameter into the `enum` and realized later that this would lose information. I'm still considering renaming the `enum` to `WindowFlags` and turning it into a flag `enum` (I remember seeing a convenience macro elsewhere).
That way, we could extend it later to add – for example – borderless fullscreen if anyone ever wanted to do that.

falbrechtskirchinger: A quick image search shows this setting is typically called "Display Mode". A `DisplayMode` flag `enum` would allow folding the `fullscreen` parameter back into the `enum`. Settings would still store one `bool` per bit for compatibility.

Flamefire: I think that is a good idea, as I don't like having (only) `WindowMode::Resizable` vs `WindowMode::NonResizable`, i.e. a boolean enum with a non-obvious type name. Hence I already wanted to suggest `WindowResizable::Yes/No` instead. But a flags-type is better I guess.

SDL calls it `WindowFlags`. I like having the "Flags" in the name so it is clear that this isn't meant to be used directly. Not sure how other projects usually call it, but my tendency would be rather to have "Window" in the name than "Display" as to me it wouldn't be obvious what "Display" refers to in a window creation function. That notwithstanding: If there is some commonly used name we should use that too.

falbrechtskirchinger: I looked at `enumUtils.h` and found it lacking. For example, it implements only `operator&()` and `operator|()`. I've extended it a bit and put all the code at the start of `VideoInterface.h` **for illustration purposes only**. Obviously, that would go into a libutil PR and the `enumUtils.h` header.

I'd like a thumbs up or thumbs down on the `enum` stuff before I adapt the rest of the code and work on that libutil PR.

I've at least one other issue once that is settled and then we can talk about the name.

libs/driver/src/VideoDriver.cpp
```diff
@@ -17,7 +17,7 @@ IVideoDriver::~IVideoDriver() = default;
  *  @param[in] CallBack DriverCallback für Rückmeldungen.
  */
 VideoDriver::VideoDriver(VideoDriverLoaderInterface* CallBack)
-    : CallBack(CallBack), initialized(false), isFullscreen_(false), renderSize_(0, 0)
+    : CallBack(CallBack), initialized(false), renderSize_(0, 0)
```
falbrechtskirchinger, line 20: Add initializer.

extras/videoDrivers/WinAPI/WinAPI.cpp
falbrechtskirchinger: Oh, the unary `~` is of course a bad idea – circumvents type safety. Would be moved into utility functions.

Flamefire:
This (and the other) PR is old but I'd like to clean the PRs up. I had the conflicts resolved (in a local branch) but am wondering what the intention was. @Spikeone You seem to have been involved too?
I can only imagine 3 modes:
1. Fullscreen
1. Windows w/o resize in full desktop size (aka windowed-fullscreen)
1. Resizeable window.

I'm not sure if we really need option 2 as it looks like it can be done with 3 already. Or am I missing anything?

Spikeone:
@Flamefire maybe I'm weird but I do play RttR in windowed mode (1920x1080 on a 2560x1440 monitor) but not fullscreen.
Sadly I sometimes fail to click the menu buttons on the bottom, resulting in a resize of the window.
This is annyoing for two reasones:
- You can not change the windowed resolution from our settings, thus you have to change the config.ini everytime
- When you are streaming, this adds either some missing areas on the bottom or a black line (depending on if it became larger or not)

So the case described is exactly what I'd love to have. Funny enough I was looking forward to this and was thinking about taking a look at it since it happened quite often to me recently as I went back into streaming RttR 😅

Flamefire:
Ok so you basically want a non-resizable window, not a fullscreen window?

Do you remember any other game that has similar options? Especially for changing the windowed size. I'd like to see how their options UI looks.

Spikeone:
Most other games simply don't support resizing the window, at least older games did not (when they even supported windowed mode at all). Sadly I can not remember any game that I played windowed that prevented resizing at all. So normally you don't have the choice - either its always enabled (so like we do it) or it doesn't work at all (as I'd like it in my case). Only other game I currently know about is my powershellgame where I added that option as well 😅

We do have an option to change the resolution, but it's always fullscreen resolution, even if windowed is enabled.

Flamefire:
> We do have an option to change the resolution, but it's always fullscreen resolution, even if windowed is enabled.

That might be worth changing :) Just switch the label to "Window size" when "windowed" is selected.
And make "Fullscreen" to "Display mode: FullScreen 'Windowed' 'Borderless Window'"

Spikeone:
> That might be worth changing :) Just switch the label to "Window size" when "windowed" is selected.
> And make "Fullscreen" to "Display mode: FullScreen 'Windowed' 'Borderless Window'"

would be great - although we discussed this in #889 where, at least I had the feeling, you didn't like the idea. Would be happy to have all the features in any case :)

PR #1603, opened 2023-07-15
Update text in shown tooltip

falbrechtskirchinger:
When a tooltip is currently being shown and its text is changed, update the displayed text.

### DIFF
--- a/libs/s25main/controls/ctrlBaseTooltip.h
+++ b/libs/s25main/controls/ctrlBaseTooltip.h
@@ -14,7 +14,7 @@ public:
     ctrlBaseTooltip(std::string tooltip = "") : tooltip_(std::move(tooltip)) {}
     virtual ~ctrlBaseTooltip();
 
-    void SetTooltip(const std::string& tooltip) { tooltip_ = tooltip; }
+    void SetTooltip(const std::string& tooltip);
     const std::string& GetTooltip() const { return tooltip_; }
     /// Swap the tooltips of those controls
     void SwapTooltip(ctrlBaseTooltip& other);
--- a/libs/s25main/WindowManager.cpp
+++ b/libs/s25main/WindowManager.cpp
@@ -798,12 +798,21 @@ class WindowManager::Tooltip
     const glFont* font;
     std::vector<std::string> lines;
     unsigned width = 0, height = 0;
+    unsigned short maxWidth;
 
 public:
     Tooltip(const ctrlBaseTooltip* showingCtrl, const std::string& text, unsigned short maxWidth)
-        : showingCtrl(showingCtrl), font(NormalFont),
-          lines(font->GetWrapInfo(text, maxWidth, maxWidth).CreateSingleStrings(text))
+        : showingCtrl(showingCtrl), font(NormalFont), maxWidth(maxWidth)
     {
+        setText(text);
+    }
+
+    auto getShowingCtrl() const { return showingCtrl; }
+    auto getWidth() const { return width; }
+
+    void setText(const std::string& text)
+    {
+        lines = font->GetWrapInfo(text, maxWidth, maxWidth).CreateSingleStrings(text);
         if(lines.empty())
             return;
         width = 0;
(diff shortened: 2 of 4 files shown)


libs/s25main/controls/ctrlBaseTooltip.h
```diff
@@ -26,4 +26,5 @@ class ctrlBaseTooltip
 
 protected:
     std::string tooltip_;
+    mutable bool showing_ = false;
```
Flow86, line 29: please only init inside the constructor (so we have *one* place to initialize things)

I always find `mutable` ugly, but I also don't see a better way to do it in this case.

So basically it lgtm, besides the one small thing.

Thanks

falbrechtskirchinger: > please only init inside the constructor (so we have _one_ place to initialize things)

Will do. Maybe worth considering changing that policy to always initialize in the class body at some point. Tends to avoid mistakes, in my experience, and is generally considered best practice since C++11.

> I always find `mutable` ugly, but I also don't see a better way to do it in this case.

Dito.

Flamefire: I don't mind having in-class initializers and ctors setting stuff. That allows to have defaults in-class and ctors only set base on their arguments. Especially good for classes with multiple different ctors having an attribute like this "show" think which is always off by default.

Flamefire (changes requested):
Triggered by the "mutable" I feel this is the wrong approach:  `WindowManager` already knows the currently shown tooltip (used for hiding it)   
So how about adding a "UpdateTooltip" method to `WindowManager` that works similarly to `WindowManager::SetToolTip` in checking if the currently shown tooltip is "THIS"? Or even better: Add an extra param to `WindowManager::SetToolTip`: `bool updateCurrent = false` to combine that into a single function. Maybe even use `updateCurrent = updateCurrent || tooltip.empty()` and do the check at https://github.com/Return-To-The-Roots/s25client/blob/bdf4110d761ddf73f235a8bdbd56fca8cf7cf941/libs/s25main/WindowManager.cpp#L839 only once (when `updateCurrent` (now) is set.)

falbrechtskirchinger:
> Triggered by the "mutable" I feel this is the wrong approach: `WindowManager` already knows the currently shown tooltip (used for hiding it)

Great. I was unaware of this.

> So how about adding a "UpdateTooltip" method to `WindowManager` that works similarly to `WindowManager::SetToolTip` in checking if the currently shown tooltip is "THIS"? Or even better: Add an extra param to `WindowManager::SetToolTip`: `bool updateCurrent = false` to combine that into a single function. Maybe even use `updateCurrent = updateCurrent || tooltip.empty()` and do the check at

I think `tooltip.empty()` should override `updateCurrent`. Unless I'm missing something, the conditions don't mix cleanly the way you're describing. You'll see in a second.
 
> https://github.com/Return-To-The-Roots/s25client/blob/bdf4110d761ddf73f235a8bdbd56fca8cf7cf941/libs/s25main/WindowManager.cpp#L839
> only once (when `updateCurrent` (now) is set.)

Flamefire:
> I think `tooltip.empty()` should override `updateCurrent`. Unless I'm missing something, the conditions don't mix cleanly the way you're describing. You'll see in a second.

Can you elaborate? `updateCurrent = updateCurrent || tooltip.empty()` accomplishes the override, doesn't it?   
I'd have kept the ToolTip instance (in WindowManager) immutable and recreated it on change.

Anyway your solution works too 👍

falbrechtskirchinger:
> > I think `tooltip.empty()` should override `updateCurrent`. Unless I'm missing something, the conditions don't mix cleanly the way you're describing. You'll see in a second.
> 
> Can you elaborate? `updateCurrent = updateCurrent || tooltip.empty()` accomplishes the override, doesn't it? I'd have kept the ToolTip instance (in WindowManager) immutable and recreated it on change.

Oh, right, because you would have just taken the branch re-creating the tooltip.
I'm not a fan of short-lived, frequent heap allocations, not because of the cost (which is often negligible), but because of heap fragmentation and resulting excessive memory consumption. glibc's `malloc()` really sucks in that regard and applications can end up reserving obscene amounts of completely unused memory because it's full of unused holes.

Of course, the `lines` vector is still bugging me for the same reason. :wink:

falbrechtskirchinger:
FYI, I have just confirmed that this does in fact work by testing these commits against #1604.

Flamefire (approved):

Flow86 (approved):

PR #1606, opened 2023-07-16
Allow excluding windows from being closed via Escape key

falbrechtskirchinger:
Add a "pin" function to in-game windows. The function replaces the shade (aka. minimize) button and the window can instead be shaded by double-clicking the title bar.
Pinned windows are excluded from being closed by the window manager when hitting the Escape key.

The setting `interface.enable_window_pinning` has to be set to `1` for the new behavior to take effect.

An option can be added to the UI once the options screen has been reorganized.

To-do:
- [x] `Alt-W` should still close all windows.
- [x] Right-clicking also closes pinned windows. Should it? I'm leaning toward "No" at the moment.
- [x] Save and restore the pinned state as part of the window settings.
- [x] ~~The feature has poor discoverability. Should it be added as a separate button instead?~~
- [x] Rebase after #1608 and #1609.
- [x] Add unit tests?


Flow86:
> The feature has poor discoverability. Should it be added as a separate button instead?

adding a small "pin" button left of the minimize button of windows would be great.

Spikeone:
Although one should be able to disable the pin button (for people who like the original interface). Will think about a pin icon.

falbrechtskirchinger:
FYI, this is what the button looks like at the moment:

pin_window_button

(A reddened hover state of the corner button.)

falbrechtskirchinger:
I've updated the PR description with the new mechanic, developed after some back-and-forth with @Spikeone on Discord.

To summarize the rationale: An additional button would've been difficult to fit on some windows (e.g., HQ) and required new assets for the top part of the window chrome. Once we add a setting to the options screen – with a tooltip explaining the behavior – this should be discoverable enough.

The pin/unpin icons are a bit difficult to tell apart (at least to my eyes and without #1594 :wink:) and @Spikeone created another set of icons using a lock, which is visually easier to distinguish. We could rename the feature and change the icons as well.

@Flow86 Thoughts?

Otherwise, work is paused until after #1608 and #1609 have been (ideally) merged.

falbrechtskirchinger:
For easier viewing, these are all the new assets at 400%:
resource_new


<details>
  <summary>Commands used for future reference:</summary>

```sh
scale=4
pad=4
s=$((16*scale + pad*2))
for f in *.bmp; do magick "$f" -filter point -resize $((scale*100))% -background white -gravity center \
                               -extent ${s}x${s} "${f%.bmp}.png"; done
magick montage -background white -tile 2x3 000*.png 003*.png 001*.png 004*.png 002*.png 005*.png
               -mode Concatenate resource_new.png
```
</details>

libs/s25main/Loader.cpp
```diff
@@ -215,7 +215,7 @@ bool Loader::LoadFilesAtStart()
     if(!LoadSounds())
         return false;
 
-    return LoadResources({"io_new", "client", "languages", "logo", "menu", "rttr"});
+    return LoadResources({"io_new", "client", "languages", "logo", "menu", "rttr", "resource_new"});
```
Flow86, line 218: please don't call it `new` ("its new until its old") - perhaps use `resource_rttr` instead?

Flamefire: We have `io_new` already... Maybe simply put the new icons into that folder?

falbrechtskirchinger: I don't really care either way. Just tell me where to put it … \*pause for inadvertent joke\*

Flamefire: Ok, then I make the call and say `io_new` as that contains other icons as well.

falbrechtskirchinger: Alright. FYI, with regards to naming, I followed the original in case that seems odd.

E.g., `closicdn`/`closicsel`/`closicup` = Close Icon Down/Selected/Up

Flamefire: Don't mind that but doesn't matter much as we currently refer to images by index rather than by name. Hence we need names such as `014.nx14.ny0.bmp` where it encodes the index and the (draw-)origin

falbrechtskirchinger: Yeah, I could have made use of the origin, but exporting full-size images was easier.

falbrechtskirchinger:
I want to add a new category called `[tweakables]` to the settings. Intended for all the little knobs like `enable_window_pinning` etc. This would be useful for #1604 as well.
Thoughts anyone?

Flamefire:
> I want to add a new category called `[tweakables]` to the settings. Intended for all the little knobs like `enable_window_pinning` etc. This would be useful for #1604 as well.

What would you put there from the other PR? IMO the interface category is fitting, i.e. similar to revert-mouse

falbrechtskirchinger:
> > I want to add a new category called `[tweakables]` to the settings. 
>
> What would you put there from the other PR? IMO the interface category is fitting, i.e. similar to revert-mouse

One or two settings for the "Follow object" changes. @Spikeon was arguing for some control to retain the vanilla experience.

I'm fine with those going into `[interface]` but was thinking ahead a little. Might be nice to have particularly granular stuff in one place. Just a thought.

spikeon:
I believe you meant to tag @Spikeone i am unfamiliar with this repo

falbrechtskirchinger:
> I believe you meant to tag @Spikeone i am unfamiliar with this repo

My apologies! You're correct.

Spikeone:
> What would you put there from the other PR? IMO the interface category is fitting, i.e. similar to revert-mouse

I think the idea to put multiple settings into that screen, maybe some that don't change the interface but other behaviors (I have no idea at the moment what settings could be client side only and change behavior, and are not interface options). Maybe at some point we'd split tweakables and interface again.

falbrechtskirchinger:
Oh, the original purpose of the `[tweakables]` category came back to me. I wanted to suggest putting some of the things that are now hardcoded constants there. For example, the (not yet implemented) pick radius in pixels for #1604.
All these values would also be perfectly at home in the `[interface]` category, I just thought it might be useful to separate out the smaller things, before `[interface]` gets crowded. :man_shrugging:

libs/s25main/ingameWindows/IngameWindow.cpp
```diff
@@ -294,11 +320,27 @@ void IngameWindow::Draw_()
     using ButtonStateResIds = helpers::EnumArray<unsigned, ButtonState>;
     constexpr ButtonStateResIds closeResIds = {47, 55, 51};
     constexpr ButtonStateResIds minimizeResIds = {48, 56, 52};
+    constexpr ButtonStateResIds pinBaseResIds = {47, 47, 51};
+    constexpr auto overlayIdsStart = 15u;
+    constexpr ButtonStateResIds pinOverlayResIdsRel = {0, 1, 2};
+    constexpr ButtonStateResIds unpinOverlayResIdsRel = {3, 4, 5};
```
falbrechtskirchinger, line 326: This seems a bit silly. Should just pre-calculate the arrays.

libs/s25main/Settings.h
```diff
@@ -104,6 +105,7 @@ class Settings : public Singleton<Settings, SingletonPolicies::WithLongevity>
     {
         unsigned autosave_interval;
         bool revert_mouse;
+        bool enableWindowPinning;
```
falbrechtskirchinger, line 108: I want to rename this to just `windowPinning`. Shorter and leaves the door open if `true`/`false` aren't enough. (For example, to add a third value for closing on right-click.)

falbrechtskirchinger (commented):
Besides two minor things and a final decision on `[tweakables]`, this seems ready to me.

libs/s25main/Settings.h
falbrechtskirchinger: On second thought, doesn't make much of a difference. The current setting names are a mix between imperative mood, past participles, and noun phrases … it's not going to be more consistent one way or the other.

Flamefire (approved):

libs/s25main/Settings.h
Flamefire: Yes I know it's a bit "grown historically"...

Flamefire (approved):

falbrechtskirchinger:
The required test runs for the auto-merge seem to have been canceled so I rebased onto master and force-pushed over your merge commit to trigger another run. There's one cosmetic change I forgot to push previously, otherwise, it's unchanged.

Flamefire (approved):

Flamefire:
The appveyor build failed due to a random issue unrelated to this. I added a commit to fix this to your branch so you don't need to do anything.

PR #1607, opened 2023-07-17
Jenkins CI: Increase timeout for Apple build

Flamefire:
2hrs doesn't seem to be enough on OSX.


Flow86:
the 120 should be enough - the problem is that since an update sometimes the jenkins agent containers do not spin up correctly and time out

Flamefire:
What do you mean by "do not spin up correctly"? I checked one such timed-out job and have seen that it started building most of the stuff on OSX but at some point stopped (killed by the timeout)

Flow86:
ah okay then it was one of those where the docker container got stopped in the background and did not notice that he lost its container and timed out

Flamefire:
Not sure I understood that. So the timeout doesn't need to be raised? If so please close this PR.

And the required macos GHA check still needs updating

PR #1609, opened 2023-07-18
Refactor PersistentWindowSettings and persist window minimized state

falbrechtskirchinger:
* Remove unneeded `PersistentWindowSettings` constructors and move default initialization in class.
* Look up  `PersistentWindowSettings` only once in `IngameWindow` (assumes pointer stability). This will make more of a difference when more settings are saved in these PRs:
  * #1601
  * #1606
* Save and restore the minimized state of in-game windows. This required some more changes to avoid calling unqualified virtual `Resize()` from the constructor.

### DIFF
--- a/libs/s25main/Settings.cpp
+++ b/libs/s25main/Settings.cpp
@@ -350,6 +350,7 @@ void Settings::LoadIngame()
             windows.persistentSettings[window.first].lastPos.x = iniWindow->getIntValue("pos_x");
             windows.persistentSettings[window.first].lastPos.y = iniWindow->getIntValue("pos_y");
             windows.persistentSettings[window.first].isOpen = iniWindow->getIntValue("is_open");
+            windows.persistentSettings[window.first].isMinimized = iniWindow->getValue("is_minimized", false);
         }
     } catch(std::runtime_error& e)
     {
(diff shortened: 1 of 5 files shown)


Flamefire:
Any chance you can come up with a test especially for the minimized state restore? Should be enough to use a dummy window with a CGI ID matching an "enabled" window

Just want to be sure we'd catch e.g. the virtual call bug in the ctor you found.

falbrechtskirchinger:
> Any chance you can come up with a test especially for the minimized state restore? Should be enough to use a dummy window with a CGI ID matching an "enabled" window

Done. Let me know if this is what you had in mind.

> Just want to be sure we'd catch e.g. the virtual call bug in the ctor you found.

That's rather difficult to test. I couldn't call `SetMinimized()` from the constructor, because that ultimately calls `Resize()`, and `clang-tidy` would rightfully warn, that this doesn't call the most derived `Resize()` as one might expect/want.
Instead, I'm calling `Window::Resize()` myself, which is already called from the constructor, so I wouldn't expect any unanticipated side effects.

Flamefire (dismissed):

falbrechtskirchinger:
@Flamefire Rebased and resolved the one-line merge conflict in `IngameWindow.h`. You can re-enable auto-merge.

Flow86 (approved):

PR #1611, opened 2023-07-20
Implement platform-agnostic paste in SDL2 backend

falbrechtskirchinger:
Add paste functionality on non-Windows platforms using `SDL_GetClipboardText()` et al. instead of WinAPI functions.

### DIFF
--- a/extras/videoDrivers/SDL2/VideoSDL2.cpp
+++ b/extras/videoDrivers/SDL2/VideoSDL2.cpp
@@ -15,6 +15,7 @@
 #include <boost/nowide/iostream.hpp>
 #include <SDL.h>
 #include <algorithm>
+#include <memory>
 
 #ifdef _WIN32
 #    include <boost/nowide/convert.hpp>
@@ -27,6 +28,17 @@
         if((call) == -1)                \
             PrintError(SDL_GetError()); \
     } while(false)
+
+namespace {
+template<typename T>
+struct SDLMemoryDeleter
+{
+    void operator()(T* p) const { SDL_free(p); }
+};
+
+template<typename T>
+using SDL_memory = std::unique_ptr<T, SDLMemoryDeleter<T>>;
+} // namespace
 
 IVideoDriver* CreateVideoInstance(VideoDriverLoaderInterface* CallBack)
 {
@@ -213,29 +225,19 @@ void VideoSDL2::PrintError(const std::string& msg) const
 
 void VideoSDL2::HandlePaste()
 {
-#ifdef _WIN32
-    if(!IsClipboardFormatAvailable(CF_UNICODETEXT))
+    if(!SDL_HasClipboardText())
         return;
 
-    OpenClipboard(nullptr);
+    SDL_memory<char> text(SDL_GetClipboardText());
+    if(!text || *text == '\0') // empty string indicates error
+        PrintError(text ? SDL_GetError() : "Paste failed.");
 
-    HANDLE hData = GetClipboardData(CF_UNICODETEXT);
-    const wchar_t* pData = (const wchar_t*)GlobalLock(hData);
-
-    KeyEvent ke = {KeyType::Invalid, 0, false, false, false};
-    while(pData && *pData)
+    KeyEvent ke = {KeyType::Char, 0, false, false, false};
+    for(const char32_t c : s25util::utf8to32(text.get()))
     {
-        ke.c = *(pData++);
-        if(ke.c == L' ')
-            ke.kt = KeyType::Space;
-        else
-            ke.kt = KeyType::Char;
+        ke.c = static_cast<unsigned>(c);
         CallBack->Msg_KeyDown(ke);
     }
-
-    GlobalUnlock(hData);
-    CloseClipboard();
-#endif
 }
 
 void VideoSDL2::DestroyScreen()


extras/videoDrivers/SDL2/VideoSDL2.cpp
```diff
@@ -213,29 +213,27 @@ void VideoSDL2::PrintError(const std::string& msg) const
...
-#ifdef _WIN32
-    if(!IsClipboardFormatAvailable(CF_UNICODETEXT))
+    if(!SDL_HasClipboardText())
         return;
 
-    OpenClipboard(nullptr);
-
-    HANDLE hData = GetClipboardData(CF_UNICODETEXT);
-    const wchar_t* pData = (const wchar_t*)GlobalLock(hData);
+    auto* text = SDL_GetClipboardText();
```
Flamefire, line 219: Can you please put that into a `std::unique_ptr` with a deleter calling `SDL_free`? Might even make sense to use a type alias like `SDL_memory` or so.  
We should only use full RAII in new/changed code.

extras/videoDrivers/SDL2/VideoSDL2.cpp
```diff
@@ -213,29 +213,27 @@ void VideoSDL2::PrintError(const std::string& msg) const
...
+    if(text[0] == '\0')
+        PrintError(SDL_GetError());
 
     KeyEvent ke = {KeyType::Invalid, 0, false, false, false};
-    while(pData && *pData)
+    auto* p = text;
+    while(*p)
     {
-        ke.c = *(pData++);
+        ke.c = static_cast<unsigned char>(*(p++));
```
Flamefire, line 229: The result of `SDL_GetClipboardText` seems to be UTF-8 encoded. So you need to decode each codepoint here.

You can use `s25util::utf8to32`, as in the `SDL_TEXTINPUT` handling especially as I don't think we should have handled space differently when pasting.

Flamefire (changes requested):
Good point, thanks!

extras/videoDrivers/SDL2/VideoSDL2.cpp
falbrechtskirchinger: Oh, I didn't know that was handled at this level. Now the `int` type makes sense.

extras/videoDrivers/SDL2/VideoSDL2.cpp
falbrechtskirchinger: Sure.

falbrechtskirchinger: I put it in the source file since it isn't needed in the header and didn't put the `SDLMemoryDeleter` in a `detail` namespace.

extras/videoDrivers/SDL2/VideoSDL2.cpp
falbrechtskirchinger: I just double-checked, that pasting a space still works. :+1:

Flamefire: > Oh, I didn't know that was handled at this level. Now the int type makes sense.

by now we should use `char32_t` for that member (`c`), but not sure if anything would break by that.

extras/videoDrivers/SDL2/VideoSDL2.cpp
```diff
@@ -28,6 +29,21 @@
             PrintError(SDL_GetError()); \
     } while(false)
 
+namespace {
+template<typename T>
+struct SDLMemoryDeleter
+{
+    void operator()(T* p) const
+    {
+        if(p)
```
Flow86, line 38: check is not necessary, `SDL_free(nullptr)` is also valid like `free(nullptr)`

Flamefire: Furthermore `unique_ptr` doesn't even call the deleter for a nullptr

extras/videoDrivers/SDL2/VideoSDL2.cpp
```diff
@@ -213,29 +229,22 @@ void VideoSDL2::PrintError(const std::string& msg) const
...
         return;
 
-    OpenClipboard(nullptr);
-
-    HANDLE hData = GetClipboardData(CF_UNICODETEXT);
-    const wchar_t* pData = (const wchar_t*)GlobalLock(hData);
+    SDL_memory<char> rawText(SDL_GetClipboardText());
+    if(!rawText)
+        return;
+    if(*rawText == '\0') // empty string indicates error
```
Flamefire, line 238: I don't see a null-pointer getting returned here per the documentation. So I'd say combine them:
```suggestion
    if(!rawText || *rawText == '\0')
```

extras/videoDrivers/SDL2/VideoSDL2.cpp
```diff
@@ -213,29 +229,22 @@ void VideoSDL2::PrintError(const std::string& msg) const
...
+        PrintError(SDL_GetError());
 
-    KeyEvent ke = {KeyType::Invalid, 0, false, false, false};
-    while(pData && *pData)
+    const std::u32string text = s25util::utf8to32(rawText.get());
+    KeyEvent ke = {KeyType::Char, 0, false, false, false};
+    for(char32_t c : text)
```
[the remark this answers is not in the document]
falbrechtskirchinger: Sure.

extras/videoDrivers/SDL2/VideoSDL2.cpp
falbrechtskirchinger: The former isn't documented, though. ´SDL_free()` may not use libc `free()`.
Didn't know about `unique_ptr` checking.

extras/videoDrivers/SDL2/VideoSDL2.cpp
falbrechtskirchinger: And constructing a string from a `nullptr` isn't valid anymore. Let's not make the C++ upgrade path more difficult.

Flamefire: Being cautious is good. I just wanted to not spend an extra 2 lines and a comment on something that doesn't/shouldn't happen (the SDL doc is pretty clear on that)   
Hence combine the NULL and the `\0` check into a single condition treating both as an error as-if it was documented like that.

falbrechtskirchinger: Well, the SDL doc is clearly lying. It's not magic. They can't return a free-able non-null pointer in an OOM condition. Unless it just aborts in that case, which isn't documented.

```c
char *SDL_GetClipboardText(void)
{
    SDL_VideoDevice *_this = SDL_GetVideoDevice();

    if (_this == NULL) {
        SDL_SetError("Video subsystem must be initialized to get clipboard text");
        return SDL_strdup("");
    }

    if (_this->GetClipboardText) {
        return _this->GetClipboardText(_this);
    } else {
        const char *text = _this->clipboard_text;
        if (text == NULL) {
            text = "";
        }
        return SDL_strdup(text);
    }
}

char *SDL_strdup(const char *string)
{   
    size_t len = SDL_strlen(string) + 1;
    char *newstr = (char *)SDL_malloc(len);
    if (newstr) {
        SDL_memcpy(newstr, string, len);
    }       
    return newstr;
}
```

I rest my case.

Flamefire: > Well, the SDL doc is clearly lying.

I thought so, too. Good job finding the relevant code :+1:

My point is that treating both `!text` and `*text == '\0' the same as "an error occurred" makes more sense than using 2 separated conditionals requiring 2 comments explaining each of them, i.e.:

```
    if(some condition or multiple)
        PrintError(SDL_GetError());
```

doesn't require any further elaboration that this is a check for an error and avoids tinkering about why one part would just return while the other prints etc. In one word: Simplicity :)

falbrechtskirchinger: But calling `SDL_GetError()` simply isn't valid if `!text`. That is my concern.
```cpp
 if(!text || *text == '\0')
        PrintError(text ? SDL_GetError() : "Paste failed.");
```

Is that an acceptable compromise?

Flamefire (approved):
Awesome, thanks!

PR #1613, opened 2023-07-25
Synchronize game world view HUD visibility settings

falbrechtskirchinger:
Synchronize visibility of game world view HUD elements from parent to child views (like the observation window).

@Spikeone identified this bug in #1604 [here](https://github.com/Return-To-The-Roots/s25client/pull/1604#issuecomment-1638412706).


Flow86 (approved):
lgtm

PR #1614, opened 2023-07-25
Fix button placement in minimized in-game windows

falbrechtskirchinger:
As mentioned, I've discovered a bug introduced in 938f355891d #1609. This is a very quick fix. Let me know if you'd prefer to store the full size in `IngameWindow` and add a `GetFullSize()` member function, or maybe a different solution entirely.

inventory_buttons

Commit message follows:
<hr>

Some windows position their buttons based on GetSize().y, which results in incorrect placement if the window is persisted and minimized. Replace uses of GetSize().y with the full window height.

Affected windows:
* iwBuildings
* iwDistribution
* iwWares

### DIFF
--- a/libs/s25main/ingameWindows/iwBuildings.cpp
+++ b/libs/s25main/ingameWindows/iwBuildings.cpp
@@ -67,10 +67,10 @@ iwBuildings::iwBuildings(GameWorldView& gwv, GameCommandFactory& gcFactory)
         }
     }
 
-    // Hilfe-Button
+    // "Help" button
     Extent btSize = Extent(30, 32);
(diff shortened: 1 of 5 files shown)


libs/s25main/ingameWindows/iwBuildings.cpp
```diff
@@ -67,10 +67,10 @@ iwBuildings::iwBuildings(GameWorldView& gwv, GameCommandFactory& gcFactory)
         }
     }
 
-    // Hilfe-Button
+    // "Help" button
```
Flamefire, line 70: Just a note for the future: Instead of the work of translating such comments you can simply remove those that don't provide additional information: `AddImageButton(..., "Help")` doesn't need such a comment, especially when the ID is replaced by the enum-IDs (e.g. `ID_btHelp`)

PR #1616, opened 2023-07-28
Implement window snapping

### LINKED ISSUE #574: Automatic window alignment
As I just played the 5th campaign mission, I missed a feature that is implemented in OTTD: automatic window alignment.

First of all the window(Visible screen) borders are some base align points you can align a wondow with (eg Screens right side aligns with the left side of a window, only vertical alignment).
But you can also align windows with other windows. When their edges are close, they also align horizontal/vertical.

As this makes it quite easy to align windows in a good looking way I think we also should implement this feature. Alignment strength is quite weak so custom, overlapping positioning should be still possible.

As I think the explanation is quite bad, I could either make a short video or one just downloads ottd and checks it for him/herself.

falbrechtskirchinger:
Make windows align along edges within a certain snap distance.

Fixes #574

To-do:
- [x] Add setting.
- [x] Choose a sensible default distance. (8 seems reasonable. That's about what KWin uses.)
- [x] Add unit tests.
- [x] Rename to `snapWindow()` during final rebase.

### DIFF
--- a/libs/s25main/SnapOffset.h
+++ b/libs/s25main/SnapOffset.h
@@ -1,0 +1,12 @@
+// Copyright (C) 2005 - 2023 Settlers Freaks (sf-team at siedler25.org)
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#pragma once
+
+#include "Point.h"
+
+/// Type used for specifying how much a window has to be moved to snap to another window's edge. Signed base type!
+using SnapOffset = Position;
+//-V:SnapOffset:801
+//-V:SnapOffset:813
--- a/libs/s25main/WindowManager.cpp
+++ b/libs/s25main/WindowManager.cpp
@@ -877,3 +877,53 @@ void WindowManager::DrawToolTip()
         curTooltip->draw(ttPos);
     }
 }
+
+SnapOffset WindowManager::snapWindow(Window* wnd, const Rect& wndRect) const
+{
+    constexpr int snapDistance = 15;
+
+    /// Progressive minimum distance to another window edge; initial value limits to at most snapDistance
+    auto minDist = Extent::all(snapDistance + 1);
+    /// (Smallest) offset (signed distance) the window should be moved; initially zero, so no snapping
+    SnapOffset minOffset = SnapOffset::all(0);
+
+    const auto setOffsetIfLowerDist = [](int a, int b, unsigned& minDist, int& minOffset) {
+        const int offset = b - a;
+        const unsigned dist = std::abs(offset);
+        if(dist < minDist)
+        {
+            minDist = dist;
+            minOffset = offset;
+        }
+    };
+    const auto setOffsetIfLowerDistX = [&](int x1, int x2) { setOffsetIfLowerDist(x1, x2, minDist.x, minOffset.x); };
+    const auto setOffsetIfLowerDistY = [&](int y1, int y2) { setOffsetIfLowerDist(y1, y2, minDist.y, minOffset.y); };
+
+    for(const auto& curWnd : windows)
+    {
+        if(curWnd.get() == wnd)
+            continue;
+
+        const Rect curWndRect = curWnd->GetBoundaryRect();
+
+        // Calculate smallest offset for each parallel pair of window edges, iff the windows overlap along the
+        // orthogonal axis (± snap distance)
+        if(wndRect.top <= (curWndRect.bottom + snapDistance) && wndRect.bottom >= (curWndRect.top - snapDistance))
+        {
+            setOffsetIfLowerDistX(wndRect.left, curWndRect.left);
+            setOffsetIfLowerDistX(wndRect.left, curWndRect.right);
+            setOffsetIfLowerDistX(wndRect.right, curWndRect.left);
+            setOffsetIfLowerDistX(wndRect.right, curWndRect.right);
+        }
+
+        if(wndRect.left <= (curWndRect.right + snapDistance) && wndRect.right >= (curWndRect.left - snapDistance))
+        {
+            setOffsetIfLowerDistY(wndRect.top, curWndRect.top);
+            setOffsetIfLowerDistY(wndRect.top, curWndRect.bottom);
+            setOffsetIfLowerDistY(wndRect.bottom, curWndRect.top);
+            setOffsetIfLowerDistY(wndRect.bottom, curWndRect.bottom);
+        }
+    }
+
+    return minOffset;
+}
--- a/libs/s25main/WindowManager.cpp
+++ b/libs/s25main/WindowManager.cpp
@@ -880,7 +880,9 @@ void WindowManager::DrawToolTip()
 
 SnapOffset WindowManager::snapWindow(Window* wnd, const Rect& wndRect) const
 {
-    constexpr int snapDistance = 15;
+    const auto snapDistance = static_cast<int>(SETTINGS.interface.windowSnapDistance);
+    if(snapDistance == 0)
+        return SnapOffset::all(0); // No snapping
 
     /// Progressive minimum distance to another window edge; initial value limits to at most snapDistance
     auto minDist = Extent::all(snapDistance + 1);
--- a/libs/s25main/ingameWindows/IngameWindow.cpp
+++ b/libs/s25main/ingameWindows/IngameWindow.cpp
@@ -5,7 +5,9 @@
 #include "IngameWindow.h"
 #include "CollisionDetection.h"
 #include "Loader.h"
+#include "RTTR_Assert.h"
 #include "Settings.h"
+#include "WindowManager.h"
 #include "driver/MouseCoords.h"
 #include "drivers/VideoDriverWrapper.h"
 #include "helpers/EnumRange.h"
@@ -201,6 +203,7 @@ void IngameWindow::MouseLeftDown(const MouseCoords& mc)
     {
         // start moving
         isMoving = true;
+        snapOffset_ = SnapOffset::all(0);
         lastMousePos = mc.GetPos();
     } else
     {
@@ -256,18 +259,30 @@ void IngameWindow::MouseLeftUp(const MouseCoords& mc)
 
 void IngameWindow::MouseMove(const MouseCoords& mc)
 {
-    // Fenster bewegen, wenn die Bewegung aktiviert wurde
     if(isMoving)
     {
-        DrawPoint newPos = GetPos() + mc.GetPos() - lastMousePos;
-        // Make sure we don't move outside window on either side
-        DrawPoint newPosBounded =
-          elMin(elMax(newPos, DrawPoint::all(0)), DrawPoint(VIDEODRIVER.GetRenderSize() - GetSize()));
-        // Fix mouse position if moved too far
-        if(newPosBounded != newPos)
-            VIDEODRIVER.SetMousePos(newPosBounded - GetPos() + lastMousePos);
+        // Calculate new window boundary rectangle without snapping
+        DrawPoint delta = mc.GetPos() - lastMousePos;
+        Rect wndRect = GetBoundaryRect();
+        RTTR_Assert(wndRect.getOrigin() == GetPos()); // The rest of the code assumes this to be true
+        wndRect.move(delta - snapOffset_);
 
-        SetPos(newPosBounded);
+        // Try to snap this window to any other
+        snapOffset_ = WINDOWMANAGER.snapWindow(this, wndRect);
+
+        // The cursor position is always relative to the position of the "unsnapped" window; bound the "unsnapped"
+        // window position to the screen…
+        DrawPoint newPos = wndRect.getOrigin();
+        DrawPoint newPosBounded =
+          elMin(elMax(newPos, DrawPoint::all(0)), DrawPoint(VIDEODRIVER.GetRenderSize() - wndRect.getSize()));
+        // …and use it to fix the mouse position if moved too far
+        if(newPosBounded != newPos)
+            VIDEODRIVER.SetMousePos(newPosBounded - wndRect.getOrigin() + mc.GetPos());
+
+        // Set new position and re-calculate snap offset (window position may have been out of bounds)
+        SetPos(newPos + snapOffset_);
+        snapOffset_ = GetPos() - newPosBounded;
+
         lastMousePos = mc.GetPos();
     } else
     {
--- a/libs/s25main/ingameWindows/IngameWindow.cpp
+++ b/libs/s25main/ingameWindows/IngameWindow.cpp
@@ -192,14 +192,8 @@ void IngameWindow::SetPinned(bool pinned)
 
 void IngameWindow::MouseLeftDown(const MouseCoords& mc)
 {
-    // Maus muss sich auf der Titelleiste befinden
-    Rect title_rect(LOADER.GetImageN("resource", 36)->getWidth(), 0,
-                    static_cast<unsigned short>(GetSize().x - LOADER.GetImageN("resource", 36)->getWidth()
-                                                - LOADER.GetImageN("resource", 37)->getWidth()),
-                    LOADER.GetImageN("resource", 43)->getHeight());
-    title_rect.move(GetDrawPos());
-
-    if(IsPointInRect(mc.GetPos(), title_rect))
+    // Check if the mouse is on the title bar
+    if(IsPointInRect(mc.GetPos(), GetButtonBounds(IwButton::Title)))
     {
         // start moving
         isMoving = true;
--- a/libs/s25main/ingameWindows/IngameWindow.cpp
+++ b/libs/s25main/ingameWindows/IngameWindow.cpp
@@ -454,7 +454,7 @@ void IngameWindow::MoveNextToMouse()
 
 bool IngameWindow::IsMessageRelayAllowed() const
 {
-    return !isMinimized_;
+    return !isMinimized_ && !isMoving;
 }
 
 void IngameWindow::SaveOpenStatus(bool isOpen) const
--- a/libs/s25main/Settings.cpp
+++ b/libs/s25main/Settings.cpp
@@ -141,6 +141,7 @@ void Settings::LoadDefaults()
     interface.autosave_interval = 0;
     interface.revert_mouse = false;
     interface.enableWindowPinning = false;
+    interface.windowSnapDistance = 8;
     // }
 
     // addons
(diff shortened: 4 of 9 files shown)


libs/s25main/WindowManager.cpp
```diff
@@ -871,3 +871,51 @@ void WindowManager::DrawToolTip()
         curTooltip->draw(ttPos);
     }
 }
+
+DrawPoint WindowManager::SnapWindow(unsigned wndId, const Rect& wndRect) const
+{
+    constexpr auto snapDistance = 15;
+    DrawPoint minDist(snapDistance + 1, snapDistance + 1);
```
falbrechtskirchinger, line 878: Use `DrawPoint::all(snapDistance + 1)`.

libs/s25main/WindowManager.cpp
```diff
@@ -871,3 +871,51 @@ void WindowManager::DrawToolTip()
...
+    for(const auto& wnd : windows)
+    {
+        if(wnd->GetID() == wndId)
+            continue;
+
+        DrawPoint dist(snapDistance + 1, snapDistance + 1);
+        DrawPoint snapOffset = DrawPoint::all(0);
+        const auto calcSnapOffsetX = [&](int a, int b) { calcSnapOffset(a, b, dist.x, snapOffset.x); };
+        const auto calcSnapOffsetY = [&](int a, int b) { calcSnapOffset(a, b, dist.y, snapOffset.y); };
+        const Rect rect = wnd->GetDrawRect();
```
falbrechtskirchinger, line 899: Should I use `GetBoundaryRect()` instead? What's the practical difference between the two?

libs/s25main/ingameWindows/IngameWindow.cpp
```diff
@@ -230,16 +233,24 @@ void IngameWindow::MouseLeftUp(const MouseCoords& mc)
...
 void IngameWindow::MouseMove(const MouseCoords& mc)
 {
-    // Fenster bewegen, wenn die Bewegung aktiviert wurde
     if(isMoving)
     {
-        DrawPoint newPos = GetPos() + mc.GetPos() - lastMousePos;
+        // Calculate window rectangle without snapping
+        DrawPoint delta = mc.GetPos() - lastMousePos;
+        DrawPoint oldPos = GetPos() - snapOffset_;
+        Rect wndRect(oldPos, GetSize());
```
falbrechtskirchinger, line 241: Similar question: is this the draw rect or the boundary rect? When are they different?

libs/s25main/WindowManager.cpp
Flamefire: For windows it doesn't matter: They are the same. Only controls may have a different size, e.g. size of text-controls depend on the current text.

For correctness here I'd say `GetBoundaryRect` looks more sensible: We want to actual, current extents of the window (in case we implement dynamic windows later)

libs/s25main/ingameWindows/IngameWindow.cpp
Flamefire: I'd use the boundary rect.

libs/s25main/WindowManager.cpp
```diff
@@ -876,3 +876,54 @@ void WindowManager::DrawToolTip()
...
+        {
+            minDist = dist;
+            snapOffset = diff;
+        }
+    };
+
+    for(const auto& wnd : windows)
+    {
+        if(wnd->GetID() == wndId)
+            continue;
```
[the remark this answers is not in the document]
Flamefire: Might be good to have a test for this case: I imagine this to be common for e.g. help windows (`CGI_HELP` IIRC)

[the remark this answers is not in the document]
falbrechtskirchinger: Yes, I was thinking of the observation windows. I'll be adding tests next and if I use the same ID (e.g., 0) this will automatically be tested.

falbrechtskirchinger:
* Adding thorough tests is quite the task. I count at least 64 window positions that could be tested. For that reason, the current tests are somewhat limited.
* I looked into the performance impact with a large-ish number of windows (about 70) by measuring the time spent in `IngameWindow::MouseMove()` for `isMoving == true`: Of about 2500 samples, ~70% fell below 10us, while ~90% fell below 15us. About 1% were above 100us with 5 outliers between 3-5ms. Negligible, or at least not significant enough to warrant a spatial data structure for the windows.
* My initial plan was to disable snapping by default, but users are most likely used to this function already. I chose 8 for the default snap distance after measuring the snap distance of my desktop window manager. Since the function is not yet configurable from the options screen, it might still be better to disable it by default. Opinions?

libs/s25main/WindowManager.cpp
```diff
@@ -876,3 +876,54 @@ void WindowManager::DrawToolTip()
...
+        calcSnapOffsetX(wndRect.left, rect.right);
+        calcSnapOffsetX(wndRect.right, rect.left);
+        calcSnapOffsetX(wndRect.right, rect.right);
+        calcSnapOffsetY(wndRect.top, rect.top);
+        calcSnapOffsetY(wndRect.top, rect.bottom);
+        calcSnapOffsetY(wndRect.bottom, rect.top);
+        calcSnapOffsetY(wndRect.bottom, rect.bottom);
+
+        // Only snap along an axis if the windows overlap (+/- snap distance) along the orthogonal axis
+        if(dist.x < minDist.x && (rect.bottom + snapDistance) >= wndRect.top
```
Flamefire, line 920: I think those conditions are redundant. Given the start value `dist == minDist == snapDistance` shouldn't a change imply that the windows overlap?

If not I think it makes more sense to only do the `calcSnapOffsetX` calls when the overlap condition is met.

Maybe that even simplifies your code: You basically do the same here as in `calcSnapOffset`: If the distance is less then set the offset. So if you do the `calcSnapOffsetX` only when you'd use it you can keep a global "minDist" and "snapOffset", don't you?

libs/s25main/WindowManager.cpp
```diff
@@ -876,3 +876,54 @@ void WindowManager::DrawToolTip()
...
+           && wndRect.bottom >= (rect.top - snapDistance))
+            minSnapOffset.x = snapOffset.x;
+        if(dist.y < minDist.y && (rect.right + snapDistance) >= wndRect.left
+           && wndRect.right >= (rect.left - snapDistance))
```
Flamefire, line 924: I'd rewrite the overlap condition to be clearer by putting the same values on the same side:
```
        if((rect.right + snapDistance) >= wndRect.left
           && (rect.left - snapDistance) <= wndRect.right)
```

or at least for symmetry: 

```
        if(wndRect.left <= (rect.right + snapDistance)
           && rect.left <= (wndRect.right + snapDistance))
```

Flamefire (changes requested):
Looks good so far, it's just a bit hard to understand/follow the code. Likely due to how this evolved. So I left a few remarks to clean that up/make it easier to read. As usual: I'm open to alternatives

libs/s25main/WindowManager.cpp
```diff
@@ -876,3 +876,54 @@ void WindowManager::DrawToolTip()
         curTooltip->draw(ttPos);
     }
 }
+
+DrawPoint WindowManager::SnapWindow(Window* wndPtr, const Rect& wndRect) const
+{
+    const auto snapDistance = static_cast<int>(SETTINGS.interface.windowSnapDistance);
```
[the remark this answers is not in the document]
falbrechtskirchinger: I didn't even want to think about any potential overflow issues, so I chose to perform all calculations using signed numbers.

libs/s25main/WindowManager.cpp
```diff
@@ -876,3 +876,54 @@ void WindowManager::DrawToolTip()
...
+    DrawPoint minSnapOffset = DrawPoint::all(0);
```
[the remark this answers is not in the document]
falbrechtskirchinger: "Snap offset" refers to the distances a window has to be moved along each axis to snap to a suitable position (i.e., roughly speaking, window edges). Zero is used to indicate that the window isn't moved at all. We`re trying to find distances in the interval `(0, snapDistance]` and minimize within that interval. I'll try to make that clearer.

libs/s25main/WindowManager.cpp
```diff
@@ -876,3 +876,54 @@ void WindowManager::DrawToolTip()
...
+    if(snapDistance == 0)
+        return minSnapOffset;
```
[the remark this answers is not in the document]
falbrechtskirchinger: (See `minSnapOffset`.)

libs/s25main/WindowManager.cpp
```diff
@@ -876,3 +876,54 @@ void WindowManager::DrawToolTip()
...
+
+    DrawPoint minDist(snapDistance + 1, snapDistance + 1);
```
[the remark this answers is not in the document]
falbrechtskirchinger: I could have used `INT_MAX`, but any number larger than `snapDistance` is sufficent. No argument about the numerous poor naming choices throughout this function.

libs/s25main/WindowManager.cpp
```diff
@@ -876,3 +876,54 @@ void WindowManager::DrawToolTip()
...
+            snapOffset = diff;
+        }
+    };
+
+    for(const auto& wnd : windows)
+    {
+        if(wnd.get() == wndPtr)
+            continue;
+
+        DrawPoint dist = DrawPoint::all(snapDistance + 1);
```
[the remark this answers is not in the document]
falbrechtskirchinger: `currentMinDist` would work. `minDist` is already taken, hence why I went with the shorter name.

libs/s25main/WindowManager.cpp
```diff
@@ -876,3 +876,54 @@ void WindowManager::DrawToolTip()
...
+        DrawPoint snapOffset = DrawPoint::all(0);
+        const auto calcSnapOffsetX = [&](int a, int b) { calcSnapOffset(a, b, dist.x, snapOffset.x); };
+        const auto calcSnapOffsetY = [&](int a, int b) { calcSnapOffset(a, b, dist.y, snapOffset.y); };
+        const Rect rect = wnd->GetBoundaryRect();
```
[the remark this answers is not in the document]
falbrechtskirchinger: I'm entirely unhappy with the naming of the function parameters and the corresponding loop variables.

libs/s25main/WindowManager.cpp
falbrechtskirchinger: I don't think so. First, we need to calculate the offset to all edges, given that a window could be positioned such that it is in snap range of multiple edges and any attempt to check beforehand seems comparably computationally expensive. Then, as the comment tries to explain, we can snap to an edge, even if the windows do not overlap along the same axis:

```
       |
+----+ |
|A   | |
|    | |
+----+ |
       |
       | +----+
       | |B   |
       | |    |
       | +----+
       |
```

This is the extreme case for which these conditions were added.

libs/s25main/WindowManager.cpp
falbrechtskirchinger: I'm simply moving left-to-right and top-to-bottom within the conditions. This logical ordering made the most sense to me, but I'll take another look.

falbrechtskirchinger:
> Looks good so far, it's just a bit hard to understand/follow the code. Likely due to how this evolved.

Indeed, I started naming the variables before I knew the specifics of the calculations ahead.

libs/s25main/WindowManager.cpp
falbrechtskirchinger: Oh, now that I've looked at the full code, you're right about that second part. The overlap conditions are independent of the calculation results and the lambdas could be called conditionally. :+1:

libs/s25main/WindowManager.cpp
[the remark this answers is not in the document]
falbrechtskirchinger: Overflows aren't a concern, but signed/unsigned comparisons are. Would you prefer to add the casts to the `if` conditions instead (e.g., `wndRect.top <= (curWndRect.bottom + static_cast<int>(snapDistance))`)?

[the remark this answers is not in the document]
Flamefire: I see. Ok then we can leave it

libs/s25main/WindowManager.cpp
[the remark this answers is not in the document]
Flamefire: That doesn't match how it is used: `if(curMinDist.x < minDist.x)` hence my question.

Flamefire (changes requested):
I still don't understand why we have `minDist`/`minOffset` and `curMinDist`/`curMinOffset`

From the current code I'd expect that the latter can be replaced by the former (i.e. no temporay dist/offset) allowing to remove 
```
if(curMinDist.x < minDist.x)
                minOffset.x = curMinOffset.x;
```
(and the same for y) without any change to the functionality, or am I misunderstanding it? This is what I wanted to suggest in https://github.com/Return-To-The-Roots/s25client/pull/1616#discussion_r1292817722

Or wait: A difference is: Currently if a later window is found which is in snap distance but further away than an earlier window in snap distance it would still snap to the later. I.e. the snapping isn't to the closest window because the global `minDist` is never updated. With the suggested change it would snap to the closest window as we have (only) a global `minDist` to minimize on.   
What was your intention?

If the current behavior is your intention then `minDist` can be a constant and a value, i.e. not a Point/Extent. The `const` would communicate the intend better (see https://github.com/Return-To-The-Roots/s25client/pull/1616#discussion_r1292813825) and not making it 2D matches that we only have a single snap distance, so no difference in X/Y snapping.

libs/s25main/WindowManager.cpp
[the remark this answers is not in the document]
falbrechtskirchinger: The global `minDist` should of course be updated. I've added a unit test for this case.

falbrechtskirchinger:
> I still don't understand why we have `minDist`/`minOffset` and `curMinDist`/`curMinOffset`

I see. Since the lambdas are now called conditionally, there's no need to track distance and offset in a temporary. Additionally, with `minDist` not being updated the behavior was not as intended.

I've added a unit test to ensure we snap to the *closest* window.

libs/s25main/WindowManager.cpp
```diff
@@ -876,3 +876,59 @@ void WindowManager::DrawToolTip()
...
+    {
+        if(curWnd.get() == wnd)
+            continue;
+
+        const auto setOffsetIfLowerDistX = [&](int x1, int x2) {
+            setOffsetIfLowerDist(x1, x2, minDist.x, minOffset.x);
+        };
+        const auto setOffsetIfLowerDistY = [&](int y1, int y2) {
+            setOffsetIfLowerDist(y1, y2, minDist.y, minOffset.y);
+        };
```
Flamefire, line 911: Those lambdas can now be moved outside the loop can't they?
```suggestion
    const auto setOffsetIfLowerDistX = [&](int x1, int x2) {
        setOffsetIfLowerDist(x1, x2, minDist.x, minOffset.x);
    };
    const auto setOffsetIfLowerDistY = [&](int y1, int y2) {
        setOffsetIfLowerDist(y1, y2, minDist.y, minOffset.y);
    };

    for(const auto& curWnd : windows)
    {
        if(curWnd.get() == wnd)
            continue;

```

libs/s25main/WindowManager.cpp
```diff
@@ -876,3 +876,59 @@ void WindowManager::DrawToolTip()
...
     }
 }
+
+DrawPoint WindowManager::SnapWindow(Window* wnd, const Rect& wndRect) const
+{
+    const auto snapDistance = static_cast<int>(SETTINGS.interface.windowSnapDistance);
+    if(snapDistance == 0)
+        return DrawPoint::all(0); // No snapping
+
+    /// The smallest distance to another window edge; values greater than snapDistance indicate no snapping
```
Flamefire, line 886: Maybe:
```suggestion
    /// Current min. distance to another window edge; start value limits to at most snapDistance
```

libs/s25main/WindowManager.cpp
```diff
@@ -876,3 +876,59 @@ void WindowManager::DrawToolTip()
...
+    auto minDist = Extent::all(snapDistance + 1);
+    /// The smallest offset (i.e., signed distance) to another window edge; zero indicates no snapping
```
Flamefire, line 888:
```suggestion
    /// The (smallest) offset (signed distance) the window should be moved, so zero indicates no snapping
```

Flamefire (changes requested):
Ok thanks! 2 suggestions for improved comments and maybe (as icing on the cake) introduce a new type alias for the return value, e.g. `using SnapOffset = Position<int>` as while `DrawPoint` works it isn't exactly what it is, i.e. it is neither something for drawing nor a point.

falbrechtskirchinger:
> introduce a new type alias for the return value

Since I need this in `WindowManager.h` and `IngameWindow.h`, where would you prefer I add this alias?
* `SnapOffset.h` (just like `DrawPoint.h`)
* `Window.h` (currently not included in `WindowManager.h`)
* Duplicate the alias in both files.
* Other?

falbrechtskirchinger:
I've modified your comment suggestions a bit further. See if that's OK.

libs/s25main/SnapOffset.h
```diff
@@ -0,0 +1,12 @@
+// Copyright (C) 2005 - 2023 Settlers Freaks (sf-team at siedler25.org)
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#pragma once
+
+#include "Point.h"
+
+/// Type used for specifying how much a window has to be moved to snap to another window's edge. Signed base type!
+using SnapOffset = Position;
```
Flamefire, line 10: I would have been explicit: `Point<int>` but I guess that is equally correct/confusing as it's neither a point nor a position :shrug:

Flamefire (dismissed):
LGTM, thanks!

libs/s25main/SnapOffset.h
falbrechtskirchinger: I've simply copied it from `DrawPoint.h`. I think this conveys/preserves compatibility with `DrawPoint`, which is the other `Point` type that `SnapOffset` interacts with frequently.

falbrechtskirchinger:
@Flamefire When I force-push the rebased version, there's no way for you to get a diff without the changes that have already been merged to `master`, right? I.e., both clicking the "Compare" link as well as " Show changes since your last review" in the "Files changed" tab show a diff cluttered with stuff that's already been reviewed.

Would it be easier if you used "Update branch" to merge `master` yourself and I force-push my changes over that? Then "Compare" should only show my changes made during the rebase (that'd just be the rename `SnapWindow()` -> `snapWindow()`).

Flamefire:
> Would it be easier if you used "Update branch" to merge master yourself and I force-push my changes over that? Then "Compare" should only show my changes made during the rebase (that'd just be the rename SnapWindow() -> snapWindow()).

Good idea!

falbrechtskirchinger:
> > Would it be easier if you used "Update branch" to merge master yourself and I force-push my changes over that? Then "Compare" should only show my changes made during the rebase (that'd just be the rename SnapWindow() -> snapWindow()).
> 
> Good idea!

Done.

Flamefire (approved):

PR #1617, opened 2023-07-31
Show Popup on missing game files

Flow86:
adapt PR from Mailaender

closes #1529

### DIFF
--- a/extras/videoDrivers/SDL2/VideoSDL2.cpp
+++ b/extras/videoDrivers/SDL2/VideoSDL2.cpp
@@ -223,6 +223,11 @@ void VideoSDL2::PrintError(const std::string& msg) const
     boost::nowide::cerr << msg << std::endl;
 }
 
+void VideoSDL2::ShowErrorMessage(const char* title, const char* message)
+{
+    SDL_ShowSimpleMessageBox(SDL_MESSAGEBOX_ERROR, title, message, window);
+}
+
 void VideoSDL2::HandlePaste()
 {
     if(!SDL_HasClipboardText())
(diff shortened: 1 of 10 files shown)


Flamefire (approved):

extras/videoDrivers/SDL2/VideoSDL2.cpp
```diff
@@ -223,6 +223,12 @@ void VideoSDL2::PrintError(const std::string& msg) const
     boost::nowide::cerr << msg << std::endl;
 }
 
+void VideoSDL2::ShowErrorMessage(const std::string& title, const std::string& message)
+{
+    if(window)
+        SDL_ShowSimpleMessageBox(SDL_MESSAGEBOX_ERROR, title.c_str(), message.c_str(), window);
```
Flamefire, line 229:
```suggestion
    SDL_ShowSimpleMessageBox(SDL_MESSAGEBOX_ERROR, title.c_str(), message.c_str(), window);
```

"the parent window, or NULL for no parent" -> NULL is fine, better than ignoring this.

Flamefire (changes requested):

extras/videoDrivers/SDL2/VideoSDL2.cpp
Flow86: done

extras/videoDrivers/SDL2/VideoSDL2.cpp
```diff
@@ -223,6 +223,12 @@ void VideoSDL2::PrintError(const std::string& msg) const
     boost::nowide::cerr << msg << std::endl;
 }
 
+void VideoSDL2::ShowErrorMessage(const std::string& title, const std::string& message)
+{
+    // window==nullptr is okay too ("no parent")
```
Flamefire, line 228: I would have left out this comment as it only repeats the documentation of the called function and if we did that more often the code would become harder to read.   
But whatever...

Flamefire (approved):

extras/videoDrivers/SDL2/VideoSDL2.cpp
Flow86: added a comment

PR #1618, opened 2023-08-05
Bump C++ standard version to C++17

falbrechtskirchinger:
There're about 4 commits that aren't strictly about C++17 but `clang-tidy` fixes I noticed along the way, like removing `// NOLINT` entirely or replacing it with a more specific suppression. Also, some changes apply to checks in newer `clang-tidy` versions only, like the De Morgan's theorem one in `RTTR_Assert`.

The majority of changes stem from the [Re-format after namespace concatenation](https://github.com/Return-To-The-Roots/s25client/commit/613869f7c452322dfcd1cc87cc20be1d8b2e679a) commit. It might be easier to remove that commit via an interactive rebase, recreate it by running `clang-format`, and compare it against my branch.

To-do:
- [x] Update submodule `libutil`: https://github.com/Return-To-The-Roots/libutil/pull/30
- [x] Update submodule `libsiedler2`: https://github.com/Return-To-The-Roots/libsiedler2/pull/24


tests/common/testMultiArray.cpp
```diff
@@ -7,10 +7,10 @@
...
 BOOST_AUTO_TEST_SUITE(MultiArrayTests)
 
-static_assert(std::is_same<int[1], helpers::detail::AddExtents_t<int, 1>>::value, "!");
-static_assert(std::is_same<int[2], helpers::detail::AddExtents_t<int, 2>>::value, "!");
-static_assert(std::is_same<int[3][5], helpers::detail::AddExtents_t<int, 3, 5>>::value, "!");
-static_assert(std::is_same<int[3][5][7], helpers::detail::AddExtents_t<int, 3, 5, 7>>::value, "!");
+static_assert(std::is_same_v<int[1], helpers::detail::AddExtents_t<int, 1>>, "!");
+static_assert(std::is_same_v<int[2], helpers::detail::AddExtents_t<int, 2>>, "!");
+static_assert(std::is_same_v<int[3][5], helpers::detail::AddExtents_t<int, 3, 5>>, "!");
+static_assert(std::is_same_v<int[3][5][7], helpers::detail::AddExtents_t<int, 3, 5, 7>>, "!");
```
falbrechtskirchinger, line 13: These (and similar cases) could be unary `static_assert`s as well.

libs/rttrConfig/src/RTTR_Version.cpp
```diff
@@ -5,20 +5,41 @@
 #include "RTTR_Version.h"
 #include "build_version_defines.h"
 
-namespace rttr { namespace version {
-
-    std::string GetTitle() { return "Return To The Roots"; }
```
Flamefire, line 10: Shame it starts splitting those functions although I'm not sure why it didn't [do that before](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#allowshortfunctionsonasingleline):

> SFS_Inline (in configuration: Inline) Only merge functions defined inside a class. Implies “empty”.

Flamefire (commented):
I like it, especially the compact namespaces and the unary static-assert.

> The majority of changes stem from the [Re-format after namespace concatenation](https://github.com/Return-To-The-Roots/s25client/commit/613869f7c452322dfcd1cc87cc20be1d8b2e679a) commit. It might be easier to remove that commit via an interactive rebase, recreate it by running clang-format, and compare it against my branch.

You can view the diff without whitespace changes which looks clean enough.

We do have the C++ standard set to 14 in [.clang-format](https://github.com/Return-To-The-Roots/s25client/blob/f9363fb9cb7b7adb080f15ad5a0915723f338f31/.clang-format#L165C1-L165C23) though. Do we need to update that too? Does it change anything?

I would love to keep functions containing only a return statement on a single line, but I guess that isn't (easily) possible (see the note). Have you checked anything related to that already?

falbrechtskirchinger:
> I like it, especially the compact namespaces and the unary static-assert.
> 
> > The majority of changes stem from the [Re-format after namespace concatenation](https://github.com/Return-To-The-Roots/s25client/commit/613869f7c452322dfcd1cc87cc20be1d8b2e679a) commit. It might be easier to remove that commit via an interactive rebase, recreate it by running clang-format, and compare it against my branch.
> 
> You can view the diff without whitespace changes which looks clean enough.

Huh, either never noticed that button on GH before, or forgot about it. Thanks!

> We do have the C++ standard set to 14 in [.clang-format](https://github.com/Return-To-The-Roots/s25client/blob/f9363fb9cb7b7adb080f15ad5a0915723f338f31/.clang-format#L165C1-L165C23) though. Do we need to update that too? Does it change anything?

Makes sense to bump that as well.

> I would love to keep functions containing only a return statement on a single line, but I guess that isn't (easily) possible (see the note). Have you checked anything related to that already?


Noticed that too but didn't look for a solution. I'll see what I can do.

libs/rttrConfig/src/RTTR_Version.cpp
falbrechtskirchinger: Either set that to "All" (which changes a lot) or use `// clang-format off`. Those seem to be the only options.

falbrechtskirchinger:
Bumping the standard in `.clang-format` had no impact on formatting.

Flow86 (approved):

libs/rttrConfig/src/RTTR_Version.cpp
Flow86: should be fine

PR #1619, opened 2023-08-16
Add player info (Number for human player, COMP for ai and change to defeated if player is defeated)

ottml:
Add info to statistic window to see if player is defeated. Add info for computer player or human player.  Like in the original settler 2. So it is obvious to see which players are already defeated.
noPlayerDefeated
onePlayerDefeated
allPlayerDefeated

### DIFF
--- a/libs/s25main/ingameWindows/iwStatistics.cpp
+++ b/libs/s25main/ingameWindows/iwStatistics.cpp
@@ -15,9 +15,24 @@
 #include "iwHelp.h"
 #include "network/GameClient.h"
 #include "ogl/FontStyle.h"
+#include "ogl/glFont.h"
 #include "world/GameWorldBase.h"
 #include "world/GameWorldViewer.h"
 #include "gameData/const_gui_ids.h"
+
+namespace {
+
+std::string getPlayerStatus(const GamePlayer& player)
+{
+    if(player.IsDefeated())
+        return "---";
+    else if(player.isHuman())
+        return "#" + std::to_string(player.GetPlayerId() + 1);
+    else
+        return _("COMP");
+}
+
+} // namespace
 
 iwStatistics::iwStatistics(const GameWorldViewer& gwv)
     : IngameWindow(CGI_STATISTICS, IngameWindow::posLastOrCenter, Extent(252, 336), _("Statistics"),
@@ -254,6 +269,17 @@ void iwStatistics::Draw_()
         return;
 
     // Die farbigen Boxen unter den Spielerportraits malen
+    DrawPlayerBoxes();
+
+    // Koordinatenachsen malen
+    DrawAxis();
+
+    // Statistiklinien malen
+    DrawStatistic(currentView);
+}
+
+void iwStatistics::DrawPlayerBoxes()
+{
     unsigned short startX = 126 - numPlayingPlayers * 17;
     DrawPoint drawPt = GetDrawPos() + DrawPoint(startX, 68);
     for(unsigned i = 0; i < gwv.GetWorld().GetNumPlayers(); ++i)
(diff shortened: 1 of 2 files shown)


libs/s25main/ingameWindows/iwStatistics.cpp
```diff
@@ -263,15 +275,16 @@ void iwStatistics::Draw_()
             continue;
 
         if(activePlayers[i])
+        {
             DrawRectangle(Rect(drawPt, Extent(34, 12)), player.color);
+
+            std::string playerId = std::to_string(player.GetPlayerId() + 1);
```
Flamefire, line 281: It makes sense to do that only when actually required. So either:

1. replace `playerId` by this value directly inline
2. make this a variable containing the whole logic to move it out the `Draw` call
3. introduce a static function (local to this file) or a lambda (local to this function) returning the text to show given a `player` which might be easier to read as you can use if-elseif-else

Also make the variable `const` if you can

Flamefire (changes requested):
Good point, I like it! Just minor polishing

Spikeone:
Weren't defeated players greyed out in the original or am I recalling that wrongly? (going to check that later)

libs/s25main/ingameWindows/iwStatistics.cpp
```diff
@@ -15,10 +15,26 @@
...
 
+namespace {
+
+std::string getPlayerStatus(const GamePlayer& player)
+{
+    const std::string playerId = std::to_string(player.GetPlayerId() + 1);
+    if(player.IsDefeated())
+        return "---";
+    else if(player.isHuman())
+        return "#" + playerId;
```
Flamefire, line 31: Just to avoid work when not required:
```suggestion
    if(player.IsDefeated())
        return "---";
    else if(player.isHuman())
        return "#" + std::to_string(player.GetPlayerId() + 1);
```

libs/s25main/ingameWindows/iwStatistics.cpp
```diff
@@ -263,15 +290,13 @@ void iwStatistics::Draw_()
             continue;
 
         if(activePlayers[i])
+        {
             DrawRectangle(Rect(drawPt, Extent(34, 12)), player.color);
+            SmallFont->Draw(drawPt + Extent(34, 12 + 2) / 2, getPlayerStatus(player),
```
Flamefire, line 295: Can you maybe introduce a variable for `Extend(34, 12)` used in both calls as they need to match, don't they?

I guess something like:
```suggestion
            SmallFont->Draw(rect.getOrigin() + rect.getSize() / 2 + DrawPoint(0, 1), getPlayerStatus(player),
```

This should also make it obvious that a (0,1)-offset is intended.

Flamefire (changes requested):

libs/s25main/ingameWindows/iwStatistics.cpp
ottml: You are right its rendering code it should be fast

Flow86:
can we show the name of the player instead of "# Nr"? (if necessary shortened to 4 chars or so). I think we now the names of all players, arent we?

libs/s25main/ingameWindows/iwStatistics.cpp
```diff
@@ -15,10 +15,25 @@
...
+namespace {
+
+std::string getPlayerStatus(const GamePlayer& player)
+{
+    if(player.IsDefeated())
+        return "---";
+    else if(player.isHuman())
+        return "#" + std::to_string(player.GetPlayerId() + 1);
+    else
+        return _("COMP");
```
Spikeone, line 32: Is this in all languages COMP? Checked english and it is, but I'm not that sure about it

Spikeone:
> can we show the name of the player instead of "# Nr"? (if necessary shortened to 4 chars or so). I think we now the names of all players, arent we?

Well, using this style is still honor the original, which I'd kinda prefer. We have a tooltip showing the full name which is fine I think.

Flow86:
> Well, using this style is still honor the original, which I'd kinda prefer. We have a tooltip showing the full name which is fine I think.

tooltip is also fine for me

libs/s25main/ingameWindows/iwStatistics.cpp
ottml: It must be translated because in france it's ordinateur.

ottml:
I have no idea how to write a test case for this feature. Is this necessary for a successful merge?

Flamefire (approved):

Flamefire:
> I have no idea how to write a test case for this feature. Is this necessary for a successful merge?

I'd say it's fine. Have you tested it that it looks like you intended with the latest changes?

ottml:
> > I have no idea how to write a test case for this feature. Is this necessary for a successful merge?
> 
> I'd say it's fine. Have you tested it that it looks like you intended with the latest changes?

Yes i tested the new implementation manual in single player and multiplayer everything is working as expected.

PR #1620, opened 2023-08-16
Add new function to enable/disable preview of map to lua

ottml:
We need lua functionality to have possibility to disable the map preview for missions. So the user can not see the map in advance.

previewdisabled
previewenabled

### DIFF
--- a/libs/s25main/desktops/dskGameLobby.cpp
+++ b/libs/s25main/desktops/dskGameLobby.cpp
@@ -231,7 +231,8 @@ dskGameLobby::dskGameLobby(ServerType serverType, std::shared_ptr<GameLobby> gam
     combo->AddString(_("Very fast")); // Sehr Schnell
 
     // Karte laden, um Kartenvorschau anzuzeigen
-    if(!gameLobby_->isSavegame())
+    bool isMapPreviewEnabled = !lua || lua->IsMapPreviewEnabled();
+    if(!gameLobby_->isSavegame() && isMapPreviewEnabled)
     {
         // Map laden
         libsiedler2::Archiv mapArchiv;
--- a/libs/s25main/desktops/dskGameLobby.cpp
+++ b/libs/s25main/desktops/dskGameLobby.cpp
@@ -231,27 +231,36 @@ dskGameLobby::dskGameLobby(ServerType serverType, std::shared_ptr<GameLobby> gam
     combo->AddString(_("Very fast")); // Sehr Schnell
 
     // Karte laden, um Kartenvorschau anzuzeigen
-    bool isMapPreviewEnabled = !lua || lua->IsMapPreviewEnabled();
-    if(!gameLobby_->isSavegame() && isMapPreviewEnabled)
+    if(!gameLobby_->isSavegame())
     {
-        // Map laden
-        libsiedler2::Archiv mapArchiv;
-        // Karteninformationen laden
-        if(int ec = libsiedler2::loader::LoadMAP(GAMECLIENT.GetMapPath(), mapArchiv))
+        bool isMapPreviewEnabled = !lua || lua->IsMapPreviewEnabled();
+        if(!isMapPreviewEnabled)
         {
-            WINDOWMANAGER.ShowAfterSwitch(
-              std::make_unique<iwMsgbox>(_("Error"), _("Could not load map:\n") + libsiedler2::getErrorString(ec), this,
-                                         MsgboxButton::Ok, MsgboxIcon::ExclamationRed, 0));
+            AddTextDeepening(70, DrawPoint(560, 40), Extent(220, 220), TextureColor::Grey, _("No preview"), LargeFont,
+                             COLOR_YELLOW);
+            AddText(71, DrawPoint(670, 40 + 220 + 10), _("Map: ") + GAMECLIENT.GetMapTitle(), COLOR_YELLOW,
+                    FontStyle::CENTER, NormalFont);
         } else
         {
-            auto* map = static_cast<libsiedler2::ArchivItem_Map*>(mapArchiv.get(0));
-            ctrlPreviewMinimap* preview = AddPreviewMinimap(70, DrawPoint(560, 40), Extent(220, 220), map);
+            // Map laden
+            libsiedler2::Archiv mapArchiv;
+            // Karteninformationen laden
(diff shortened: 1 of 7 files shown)


libs/s25main/lua/LuaInterfaceSettings.cpp
```diff
@@ -237,3 +237,15 @@ std::vector<AddonId> LuaInterfaceSettings::GetAllowedAddons()
...
+
+bool LuaInterfaceSettings::IsMapPreviewEnabled()
+{
+    kaguya::LuaRef getMapPreviewEnabled = lua["isMapPreviewEnabled"];
+    if(getMapPreviewEnabled.type() == LUA_TFUNCTION)
+    {
+        kaguya::LuaRef previewEnabled = getMapPreviewEnabled();
+        if(previewEnabled.typeTest<bool>())
+            return previewEnabled;
+    }
```
Flamefire, line 249: Simpler version similar to https://github.com/Return-To-The-Roots/s25client/blob/605bd80f5cbb7c123dc5ba581124baae70ec783c/libs/s25main/lua/LuaInterfaceSettings.cpp#L180

```suggestion
        return getMapPreviewEnabled.call<bool>();
```

And maybe rename to `func` as there is already a naming-mix between "get" and "is"

tests/s25Main/lua/testLuaSettings.cpp
```diff
@@ -149,6 +149,10 @@ BOOST_AUTO_TEST_CASE(Events)
     BOOST_TEST_REQUIRE(allowedAddons[0] == AddonId::LIMIT_CATAPULTS);
     BOOST_TEST_REQUIRE(allowedAddons[1] == AddonId::CHARBURNER);
     BOOST_TEST_REQUIRE(allowedAddons[2] == AddonId::TRADE);
+
+    executeLua("function isMapPreviewEnabled()\n  return { false }\nend");
+    BOOST_TEST_REQUIRE(lua.IsMapPreviewEnabled());
```
Flamefire, line 154: This is the opposite, isn't it?

tests/testData/maps/LuaFunctions.lua
```diff
@@ -20,6 +20,10 @@ function getAllowedChanges()
 	return {general=true, addonsAll=false, addonsSome=true, swapping=false, playerState = not isSinglePlayer, ownNation = true, ownColor=true, ownTeam=false, aiNation = false, aiColor=false, aiTeam=false}
 end
 
+function isMapPreviewEnabled()
```
Flamefire, line 23: Any reason to add this here?

ottml: I added it because other settings are there too.

Flow86:
can we perhaps instead of not showing any map if preview is disabled, showing some "Hidden" map view, don't know if we can generate a preview of something like that?

Flamefire (dismissed):

Flamefire:
> can we perhaps instead of not showing any map if preview is disabled, showing some "Hidden" map view, don't know if we can generate a preview of something like that?

Maybe some placeholder graphics. Not sure if it is worth the effort. But if e.g. @Spikeone wants to provide an image @ottml likes I'm fine with it. Otherwise I'd just keep the current state of the PR

Spikeone:
@Flamefire I'd to this without a graphic. Can't we just add a gray area (like for player colors) and either use a scaled up '?' or simply write "no preview" or something alike in that area?

Flow86:
> Can't we just add a gray area (like for player colors) and either use a scaled up '?' or simply write "no preview" or something alike in that area? 

thats a good idea

ottml:
disabledView
disabledView2

libs/s25main/desktops/dskGameLobby.cpp
```diff
@@ -231,27 +231,36 @@ dskGameLobby::dskGameLobby(ServerType serverType, std::shared_ptr<GameLobby> gam
...
-        libsiedler2::Archiv mapArchiv;
-        // Karteninformationen laden
-        if(int ec = libsiedler2::loader::LoadMAP(GAMECLIENT.GetMapPath(), mapArchiv))
+        bool isMapPreviewEnabled = !lua || lua->IsMapPreviewEnabled();
+        if(!isMapPreviewEnabled)
         {
-            WINDOWMANAGER.ShowAfterSwitch(
-              std::make_unique<iwMsgbox>(_("Error"), _("Could not load map:\n") + libsiedler2::getErrorString(ec), this,
-                                         MsgboxButton::Ok, MsgboxIcon::ExclamationRed, 0));
+            AddColorDeepening(70, DrawPoint(560, 40), Extent(220, 220), TextureColor::Grey, 0);
```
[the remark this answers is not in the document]
ottml: Yes. Good idea

tests/s25Main/lua/testLuaSettings.cpp
```diff
@@ -149,6 +149,10 @@ BOOST_AUTO_TEST_CASE(Events)
     BOOST_TEST_REQUIRE(allowedAddons[0] == AddonId::LIMIT_CATAPULTS);
     BOOST_TEST_REQUIRE(allowedAddons[1] == AddonId::CHARBURNER);
     BOOST_TEST_REQUIRE(allowedAddons[2] == AddonId::TRADE);
+
+    executeLua("function isMapPreviewEnabled()\n  return false\nend");
+    BOOST_TEST_REQUIRE(!lua.IsMapPreviewEnabled());
```
Flamefire, line 154: Also check the default:

```suggestion
    BOOST_TEST(lua.IsMapPreviewEnabled()); // Not defined in lua -> Default to enabled
    executeLua("function isMapPreviewEnabled()\n  return false\nend");
    BOOST_TEST(!lua.IsMapPreviewEnabled());
    executeLua("function isMapPreviewEnabled()\n  return true\nend");
    BOOST_TEST(lua.IsMapPreviewEnabled());
```

(Yes we erroneously used `BOOST_TEST_REQUIRE` for no good reason around this)

Flamefire (changes requested):
Looks good, so just a minor improvement on the test.

Flamefire (approved):

Spikeone:
hey @ottml just asking since I've seen something alike in the campagin.lua - wouldn't it be great to set some kind of preview graphic - so one could for example set the SETUP810.LBM in order to hint, that this is winter map.

Issue #568, opened 2016-08-12
Possible "Singleplayer -> Campaign" - Screen


lvanderwall:
I've been thinking about the "Singleplayer -> Campaign" screen that's not yet implemented, and how to improve the original one from S2. I know it's something for the distant future, so it's just a brainstorming to collect a few ideas:

![campaign_screen](https://cloud.githubusercontent.com/assets/16540971/17620971/8c24c2f4-6090-11e6-8e4c-d4ab8319568d.png)
- [ ] campaign screen lists all missions for that campaign. Missions you can't play are disabled / grayed out
- [ ] improvement to S2: A "Campaign-Dropdown" that lists all campaigns you have installed in the campaign folder (see below)
- [ ] don't show map overview on start, at least in single player mode

Store new campaigns (and possibly the campaign progress) in a separate subfolder. People shouldn't need to copy new campaign files into the `DATA` folder like in the original:
- [ ]  all `swd`\- and `lua`-files for a campaign "MyPaign" are stored in a subfolder `%RTTR_INSTALL%/RTTR/MAPS/CAMPAIGNS/MYPAIGN`
- [ ] save campaign progress, possibly by a `mission.dat` that is stored in the same folder
- [ ] enable new missions by a lua command `ActivateMission(i)`
- [ ] for the "Roman Campaign" there should be a `readme` to inform people which map files to copy into the folder, as those maps cannot be included included directly for copyright reasons

As I said before, it's just a brainstorming to collect some ideas, so feel free to add to them or crush them completely!

Spikeone:
Damn - gonna edit this post when I'm home

Flamefire:
We can store the campaign progress (which mission is next) in the general settings file. Avoids problems if the maps directory is not writable. 
The masterfile must be designed carefully. What do we really need? 

1. Name
1. Description 
2. Author
3. Map file names or shall we use a clever naming? Maybe both: map names must be same as masterfile but with _m1-9 attached and the master must state the number of missions so we can detect missing ones. 

Anything else? 
Translations are hard. Maybe make ini files with "[Name]" as the section and entries for each language? Map names are even harder because I'd take the name from the map file to keep it consistent with the preview/join screen. But no translations there yet. 
I would allow subdirs but don't require them. 

I like the old mission screen. Why not keep it for mission selection and add a list (similar to map list) before that to select a campaign which is skipped if there only is one?

[...]

PR #1591, opened 2023-05-10
Campaing screen

ottml:
Fixes #568
This is only a frist draft for further discussion
- [x] Support for lua file path different from map file path #1628
- [x] Campaign gui #1628
- [x] Lua file for campaign settings, including translation #1628
- [x] Campaign mission selection gui #1628
- [x] Add roman and continent campaign #1628
- [x] Disable map preview #1620 
- [ ] Enable next mission lua support
- [ ] Storing of campaing progress in settings
- [ ] Update campaigns lua files with new functions

PR #1628, opened 2023-10-11
Campaign screen final

ottml:
Fixes https://github.com/Return-To-The-Roots/s25client/issues/568
Relates too #1591


ottml:
romancampaign
kontinetcampaign

Spikeone:
Looking pretty good - although it's sad we don't have the original worldmap screen, I don't think we need that (for now).

Spikeone:
Oh, and we still need some kind of "patch map" function where one is able to override parts of the original map file (to fix BBs maps) - since this is only required for the original roman campaign, it could be just giving an index and new hex value which changes the mapfile after initially loading it.

ottml:
@Flamefire @Flow86 can you have a look at the pull request and make a review. Would be nice. Thanks.

Flamefire (changes requested):
Awesome work!

A bit of cleanup would be great and I did my best to find potential issues and suggest solutions & improvements.

data/RTTR/CAMPAIGNS/world/AFRICA.LUA
```diff
@@ -0,0 +1,68 @@
...
+-------------------------------- TODO -----------------------------------------
+-- EnableNextMissions()
+-- Set Portraits
+-- Set AI Agression Level
+-------------------------------------------------------------------------------
+
+
+-------------------------------- Lua Version used -----------------------------
+function getRequiredLuaVersion()
+    return 1.3
```
Spikeone, line 17: I think it should be 1 instead of 1.3, guess 3 was the feature level which has been removed (in all scripts, also the roman version)

data/RTTR/CAMPAIGNS/roman/MISS200.LUA
```diff
@@ -0,0 +1,514 @@
...
+-- Set AI Agression Level
+-------------------------------------------------------------------------------
+
+
+-------------------------------- Lua Version used -----------------------------
+function getRequiredLuaVersion()
+    return 1.3
+end
+
+
```
Spikeone, line 20: Guess all roman scripts require the preview disabled, haven't done that yet

Spikeone:
@Flamefire shouldn't we put campaigns into the s25maps repo as well?

libs/rttrConfig/src/files.h
```diff
@@ -35,6 +35,7 @@ namespace folders {
     constexpr auto mapsPlayed = "<RTTR_USERDATA>/MAPS"; // downloaded maps
     constexpr auto mapsRttr = "<RTTR_RTTR>/MAPS/NEW";   // maps added by RTTR
     constexpr auto mapsSea = "<RTTR_RTTR>/MAPS/SEA";    // seafaring maps
+    constexpr auto campaigns = "<RTTR_RTTR>/CAMPAIGNS"; // contains one folder for each campaign
```
Flamefire, line 38: Should we maybe have a user folder for campaigns, e.g. `<RTTR_USERDATA>/campaigns`? The `RTTR_RTTR` folder is in the install dir and might not be user-writable.

Flamefire:
> @Flamefire shouldn't we put campaigns into the s25maps repo as well?

I'd say it makes sense to have "our" campaigns here. At least the lua files. Otherwise we'd need another submodule and as LUA files are text they don't hurt to have in this repo. Only binary files like maps should be separate. At least that was the intention for the maps repo IIRC.

Spikeone:
> > @Flamefire shouldn't we put campaigns into the s25maps repo as well?
> 
> I'd say it makes sense to have "our" campaigns here. At least the lua files. Otherwise we'd need another submodule and as LUA files are text they don't hurt to have in this repo. Only binary files like maps should be separate. At least that was the intention for the maps repo IIRC.

So, since I have a custom campaign which was made by a fan, we then init that one? Fine for me, was just thinking about easy PRs, even for the campaign maps

Flamefire:
> So, since I have a custom campaign which was made by a fan, we then init that one? Fine for me, was just thinking about easy PRs, even for the campaign maps

Makes sense I guess. No need to run the full CI suite for campaign PRs. On the other hand: We need a PR here for the submodule update then which results in the same. But with the (potential) maps it makes sense to pull it out. I guess we'll need a new repo for that. @Flow86 ?

However I wouldn't block the PR for that: Let's get this in first, moving the text files into a subproject can be done afterwards.

libs/s25main/desktops/dskCampaignSelection.cpp
```diff
@@ -0,0 +1,242 @@
...
+{
+    ID_CAMPAIGN_LABEL,
+    ID_CAMPAIGN_TABLE,
+    ID_CAMPAIGN_PREVIEW_TITLE,
+    ID_CAMPAIGN_PREVIEW_DESCRIPTION,
+    ID_BACK,
+    ID_NEXT
+};
+
+int startOffsetY = 20;
```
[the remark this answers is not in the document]
ottml: Done

libs/s25main/desktops/dskCampaignSelection.cpp
```diff
@@ -0,0 +1,242 @@
...
+int padding = 10;
+int firstColumnExtentX = 480;
+int secondColumnExtentX = 280;
+int columnExtentY = 500;
+int scecondColumnOffsetX = padding * 2 + firstColumnExtentX;
```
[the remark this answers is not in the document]
ottml: Done

Flamefire:
Please don't force-push during a review phase. It is difficult to find what you have changed, i.e. what parts of the review comments were adressed, after a force-push. You can commit with `git commit --fixup=<original commit sha>` to note that this previous commit should be fixed/changed and do that automatically after the review/before merge with `git rebase --autosquash`

ottml:
> Please don't force-push during a review phase. It is difficult to find what you have changed, i.e. what parts of the review comments were adressed, after a force-push. You can commit with `git commit --fixup=<original commit sha>` to note that this previous commit should be fixed/changed and do that automatically after the review/before merge with `git rebase --autosquash`

Thanks for the  hint. I will do this from now on. I do not  know that this is possible in such an esay was.

libs/s25main/QuickStartGame.cpp
```diff
@@ -87,8 +87,10 @@ bool QuickStartGame(const boost::filesystem::path& mapOrReplayPath, const std::v
 
     WINDOWMANAGER.Switch(std::make_unique<dskSelectMap>(csi));
 
-    if((extension == ".sav" && GAMECLIENT.HostGame(csi, mapOrReplayPath, MapType::Savegame))
-       || ((extension == ".swd" || extension == ".wld") && GAMECLIENT.HostGame(csi, mapOrReplayPath, MapType::OldMap)))
+    if((extension == ".sav"
+        && GAMECLIENT.HostGame(csi, mapOrReplayPath, boost::optional<boost::filesystem::path>(), MapType::Savegame))
+       || ((extension == ".swd" || extension == ".wld")
+           && GAMECLIENT.HostGame(csi, mapOrReplayPath, boost::optional<boost::filesystem::path>(), MapType::OldMap)))
```
Flamefire, line 93: Use `boost::none` for an empty optional which is much easier to read:

```suggestion
           && GAMECLIENT.HostGame(csi, mapOrReplayPath, boost::none, MapType::OldMap)))
```

Flamefire:
> Thanks for the hint. I will do this from now on. I do not know that this is possible in such an easy way.

No worries, that seems to be something new to git as I learned about that only recently. Previously I did that manually with stuff like `git commit -m 'f: Part of another commit title'` and then do the fixups with an interactive rebase manually (`git rebase -i`)

And with that change from GitHub showing the diff of a force-push it actually got easier to even track those.

libs/rttrConfig/src/files.h
Flow86: perhaps load both?
the first one in `RTTR_RTTR` for our campaigns (aka romans, ...) and for users campaigns one in `RTTR_USERDATA` - perhaps like we have with the map selection with categories?

data/RTTR/CAMPAIGNS/roman/MISS200.LUA
```diff
@@ -0,0 +1,517 @@
...
+    MissionEvent(15)
+end
+
+function onOccupied(p, x, y)
+    -- only check human player
+    if(p ~= 0) then
+        return
+    end
+
+    if(     (x == 34) and (y == 28) ) then MissionEvent(4) -- RttR calls onOccupied once a pole is on that spot, SettlersII called the function once the spot is really owned
```
Flow86, line 382: > RttR calls onOccupied once a pole is on that spot, SettlersII called the function once the spot is really owned

what do you mean by that? whats exactly the difference?

data/RTTR/CAMPAIGNS/roman/campaign.lua
```diff
@@ -0,0 +1,29 @@
+rttr:RegisterTranslations(
```
Flow86, line 1: do we have something like `getRequiredLuaVersion` for campaign luas too? that would be good to add something like that from the start, so we can add newer features later easily

libs/s25main/network/GameClient.cpp
```diff
@@ -54,6 +54,23 @@
 #include <helpers/chronoIO.h>
 #include <memory>
 
+namespace {
+void copyFileIfPathDifferent(const boost::filesystem::path& src_path, const boost::filesystem::path& dst_path)
```
Flow86, line 58: we should put this function somewhere like libutil or so

libs/s25main/network/GameClient.h
```diff
@@ -80,7 +81,8 @@ class GameClient final :
     bool Connect(const std::string& server, const std::string& password, ServerType servertyp, unsigned short port,
                  bool host, bool use_ipv6);
     /// Start the server and connect to it
-    bool HostGame(const CreateServerInfo& csi, const boost::filesystem::path& map_path, MapType map_type);
+    bool HostGame(const CreateServerInfo& csi, const boost::filesystem::path& map_path,
```
Flow86, line 84: somehow having so many parameters in the mean time it should probably be best to use a struct for it:
```suggestion
struct MapInfo
{
boost::filesystem::path map_path;
MapType map_type;
boost::optional<boost::filesystem::path lua_path;
};
bool HostGame(const CreateServerInfo& csi, const MapInfo& map);
```

so we don't need to use `boost::none` but can use struct initialization.

libs/s25main/network/GameServer.cpp
```diff
@@ -185,7 +186,7 @@ bool GameServer::Start(const CreateServerInfo& csi, const boost::filesystem::pat
     if(!mapinfo.mapData.CompressFromFile(mapinfo.filepath, &mapinfo.mapChecksum))
         return false;
 
-    bfs::path luaFilePath = bfs::path(mapinfo.filepath).replace_extension("lua");
+    bfs::path luaFilePath = lua_path.get_value_or(bfs::path(mapinfo.filepath).replace_extension("lua"));
     if(bfs::is_regular_file(luaFilePath))
```
Flow86, line 190: shouldn't it be an error if we explicitly want to load a lua file, but it does not exist? (e.g `lua_path` *not* empty but file does not exist)

Flow86 (changes requested):

data/RTTR/CAMPAIGNS/roman/MISS200.LUA
ottml: I copied the lua files from @Spikeone repositiory. The difference is here when the border is on the position the mission is won and in the original settler the mission is won when the gate in witgin the teritory. I don't think this is a problem for now.

libs/libGamedata/lua/CampaignDataLoader.cpp
```diff
@@ -38,14 +40,7 @@ bool CampaignDataLoader::Load()
...
 void CampaignDataLoader::Register(kaguya::State& state)
 {
-    state["RTTRCampaignData"].setClass(kaguya::UserdataMetatable<CampaignDataLoader, LuaInterfaceBase>().addFunction(
-      "AddCampaign", &CampaignDataLoader::AddCampaign));
-}
-
-void CampaignDataLoader::AddCampaign(const kaguya::LuaTable& data)
-{
-    CampaignDescription(data, campaignDesc_);
+    state["RTTRCampaignData"].setClass(kaguya::UserdataMetatable<CampaignDataLoader, LuaInterfaceBase>());
```
Flamefire, line 45: We don't need that Register now, do we?

libs/libGamedata/gameData/CampaignDescription.cpp
```diff
@@ -1,28 +1,42 @@
...
 #include "CampaignDescription.h"
+#include "helpers/format.hpp"
 #include "lua/CheckedLuaTable.h"
 #include "lua/LuaHelpers.h"
+#include "mygettext/mygettext.h"
 
-CampaignDescription::CampaignDescription() = default;
-
-CampaignDescription::CampaignDescription(CheckedLuaTable luaData, CampaignDescription& campaignDesc)
+CampaignDescription initFromLuaTable(kaguya::LuaTable table)
```
Flamefire, line 11: This now looks like it can become a regular (and `explicit`) constructor, doesn't it? I mean you create and return a new instance which is basically what constructors are for. Might be even cleaner as you can use an initializer list and only do the validity-checks on the body.

libs/libGamedata/lua/CampaignDataLoader.cpp
```diff
@@ -28,8 +28,10 @@ bool CampaignDataLoader::Load()
     auto curFile_ = basePath_ / "campaign.lua";
     try
     {
-        if(!loadScript(curFile_))
+        if(!loadScript(curFile_) || lua["campaign"].isNilref() || lua["campaign"].type() != LUA_TTABLE)
```
Flamefire, line 31: Maybe just get the campaign as a lua ref once instead of 3 times? See e.g. https://github.com/Return-To-The-Roots/s25client/blob/29dc07bfb32d96c0e931ddffa5d4d4a39419549c/libs/libGamedata/lua/LuaInterfaceBase.cpp#L109-L110 and IIRC `type() == LUA_TTABLE` implies `!isNilref()`
You could even throw an exception (caught below) if the entry does not exist to log that `campaign` is missing

libs/libGamedata/lua/CampaignDataLoader.cpp
```diff
@@ -28,8 +28,10 @@ bool CampaignDataLoader::Load()
...
             return false;
+
+        campaignDesc_ = initFromLuaTable(lua["campaign"]);
```
Flamefire, line 34: With the above suggestion:

```suggestion
        campaignDesc_ = CampaignDescription(lua["campaign"]);
```

tests/s25Main/campaign/testCampaignLuaFile.cpp
```diff
@@ -1,16 +1,27 @@
...
 #include "rttr/test/LocaleResetter.hpp"
+#include "rttr/test/LogAccessor.hpp"
 #include "rttr/test/TmpFolder.hpp"
 #include <s25util/utf8.h>
 #include <boost/filesystem.hpp>
 #include <boost/nowide/fstream.hpp>
 #include <boost/test/unit_test.hpp>
 
+/// Require that the log contains "content"
+#define REQUIRE_LOG_CONTAINS(content)                                                                                  \
```
Flamefire, line 16: Don't we have the macro already defined somewhere?

Flamefire (changes requested):

libs/rttrConfig/src/files.h
ottml: Added the additional search path for user defined campaigns.

libs/libGamedata/lua/CampaignDataLoader.cpp
ottml: I needed this otherwise it does not work. But i am not a lua expert therefore i am not sure what this line is doing really. In my understanding this is needed to get the connection betweeen lua and our class. Any input is welcome :)

tests/s25Main/campaign/testCampaignLuaFile.cpp
ottml: We have only a macro defined, which checks the first line of the log and i have to check for content of multiple lines.

libs/s25main/network/GameClient.h
ottml: Use name MapDescription because MapInfo already exists for other parameters.

libs/libGamedata/lua/CampaignDataLoader.cpp
ottml: Changed

libs/libGamedata/lua/CampaignDataLoader.cpp
ottml: Done

libs/libGamedata/gameData/CampaignDescription.cpp
ottml: Changed to constructor

tests/s25Main/campaign/testCampaignLuaFile.cpp
```diff
@@ -0,0 +1,211 @@
...
+            maps = { \"dessert0.WLD\", \"dessert1.WLD\", \"dessert2.WLD\"}\
+        }";
+    }
+
+    CampaignDescription desc;
+    CampaignDataLoader loader(desc, tmp.get());
+    rttr::test::LogAccessor logAcc;
+    BOOST_TEST(!loader.Load());
+    REQUIRE_LOG_CONTAINS(
+      "Failed to load campaign data!\nReason: Failed to load game data: Required field 'luaFolder' not found");
```
Flamefire, line 146: Maybe just throw a `std::runtime_error` such that this reads:
```suggestion
    REQUIRE_LOG_CONTAINS("Failed to load campaign data!\nReason: Required field 'luaFolder' not found");
```

libs/s25main/network/GameClient.h
```diff
@@ -79,8 +80,16 @@ class GameClient final :
 
     bool Connect(const std::string& server, const std::string& password, ServerType servertyp, unsigned short port,
                  bool host, bool use_ipv6);
+
+    struct MapDescription
```
Flamefire, line 84: This should be used by the GameServer too, doesn't it? Maybe move it to a new file in `gameTypes`

ottml: Yes i thought about this. I can do it.

ottml: done

tests/s25Main/campaign/testCampaignLuaFile.cpp
ottml: Thats not possible because CheckedLuaTable throws this error internal.

libs/s25main/desktops/dskCampaignSelection.cpp
```diff
@@ -0,0 +1,242 @@
...
+                LOG.write(_("Campaign map lua file %1% does not exist.\n")) % luaFilepath;
+                brokenCampaignPaths.insert(folder);
+                continue;
+            }
+
+            libsiedler2::Archiv map;
+            if(int ec = libsiedler2::loader::LoadMAP(mapPath, map, true))
+            {
+                LOG.write(_("Failed to load map %1%: %2%\n")) % mapPath % libsiedler2::getErrorString(ec);
+                brokenMapPaths.insert(folder);
```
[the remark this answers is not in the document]
ottml: Remove brokenMapPaths completly.

libs/s25main/desktops/dskCampaignSelection.cpp
```diff
@@ -0,0 +1,242 @@
...
+}
+
+void dskCampaignSelection::Draw_()
+{
+    Desktop::Draw_();
+
+    // replace this by AddImageButton as soon as it can handle this sceneraio correctly
+    if(showCampaignInfo_ && campaignImageId_)
+    {
+        auto* image = LOADER.GetImageN(*campaignImageId_, 0);
```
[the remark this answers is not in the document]
ottml: If i remember correct the pull request #1536 is needed, to place the image correct on the ctrlImage button.

data/RTTR/CAMPAIGNS/roman/campaign.lua
ottml: Introduced

libs/s25main/desktops/dskCampaignSelection.cpp
```diff
@@ -0,0 +1,246 @@
...
+    int const mutlilineExtentY = columnExtentY - (multilineOffsetY - getColumnOffsetY());
+    AddMultiline(ID_PreviewDescription, DrawPoint(scecondColumnOffsetX, multilineOffsetY),
+                 Extent(secondColumnExtentX, mutlilineExtentY), TextureColor::Grey, NormalFont);
+
+    AddTextButton(ID_Back, DrawPoint(380, 560), Extent(200, 22), TextureColor::Red1, _("Back"), NormalFont);
+    AddTextButton(ID_Next, DrawPoint(590, 560), Extent(200, 22), TextureColor::Green2, _("Continue"), NormalFont);
+
+    showCampaignInfo(false);
+
+    AddTimer(ID_TimerFillCampaignTable, 1ms);
```
Flamefire, line 92: Why the timer instead of a direct call?

libs/s25main/desktops/dskCampaignSelection.cpp
```diff
@@ -0,0 +1,246 @@
...
+                if(int ec = libsiedler2::loader::LoadMAP(mapPath, map, true))
+                {
+                    LOG.write(_("Failed to load map %1%: %2%\n")) % mapPath % libsiedler2::getErrorString(ec);
+                    brokenCampaignPaths_.insert(folder);
+                    success = false;
+                    continue;
+                }
+            }
+
+            if(success)
```
Flamefire, line 221: This can never be false at this point, as you always `continue` where you set this to false, or am I missing anything?

Flamefire (changes requested):

libs/s25main/desktops/dskCampaignSelection.cpp
ottml: We need the timer here because otherwise the error message showing the campaigns not loaded succesful is not displayed. Because we are in the constructor of the class and the call is not processed.

libs/s25main/desktops/dskCampaignSelection.cpp
ottml: The continue effects only the inner loop over the maps. So if any error during check of the maps occours success will be false after the inner loop has finished. The continue is only there to avoid checking further stuff for an already failed. But i want to log all errors for all maps and not only the first failed map.

ottml:
> @ottml I updated the script files sligthly, maybe you can also update them here (https://github.com/Spikeone/RttR_Campaigns/tree/master/Campaign_OriginalRoman) - they now use your mappreview function (although this could have been a campaign setting, thinking about it)

Updated to your version.

libs/s25main/desktops/dskCampaignMissionSelection.cpp
```diff
@@ -0,0 +1,197 @@
...
+    ctrlGroup* group = AddGroup(ID_GroupStart + currentPage_);
+    Extent catBtSize = Extent(300, missionButtonHeightY);
+    DrawPoint curBtPos(250, getStartOffsetMissionButtonsY());
+    for(unsigned int i = 0; i < missionsPerPage_; i++)
+    {
+        unsigned int missionIndex = currentPage_ * missionsPerPage_ + i;
+        if(missionIndex >= settings_->mapNames.size())
+            break;
+
+        const bfs::path& mapFilePath = RTTRCONFIG.ExpandPath(settings_->mapFolder) / settings_->mapNames[missionIndex];
```
Flamefire, line 114: At least use the function (and get rid of the RTTRConfig include):
```suggestion
        const bfs::path& mapFilePath = settings_->getMapFilePath(settings_->mapNames[missionIndex]);
```

That's also why I thought about making `mapNames/luaFolder/mapFolder` private and provide only access to the map count and map/lua file path by index. Iterating over it (instead of mapNames) could be done by `for(auto idx: helpers::range(settings.getNumMaps()))` but here you already have indices

libs/s25main/desktops/dskCampaignSelection.cpp
```diff
@@ -0,0 +1,246 @@
+// Copyright (C) 2005 - 2023 Settlers Freaks (sf-team at siedler25.org)
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#include "dskCampaignSelection.h"
+#include "Desktop.h"
+#include "ListDir.h"
+#include "Loader.h"
+#include "RttrConfig.h"
+#include "Window.h"
```
Flamefire, line 10:
```suggestion
#include "ListDir.h"
#include "Loader.h"
#include "RttrConfig.h"
```

libs/s25main/desktops/dskCampaignSelection.cpp
```diff
@@ -0,0 +1,246 @@
...
+    ID_Next,
+    ID_TimerFillCampaignTable
+};
+
+constexpr int startOffsetY = 20;
+constexpr int padding = 10;
+constexpr int firstColumnExtentX = 480;
+constexpr int secondColumnExtentX = 280;
+constexpr int columnExtentY = 500;
+constexpr int scecondColumnOffsetX = padding * 2 + firstColumnExtentX;
```
Flamefire, line 53:
```suggestion
constexpr int secondColumnOffsetX = padding * 2 + firstColumnExtentX;
```

libs/s25main/desktops/dskCampaignSelection.cpp
```diff
@@ -0,0 +1,246 @@
...
+}
+
+boost::filesystem::path dskCampaignSelection::getSelectedCampaignPath()
+{
+    auto* table = GetCtrl<ctrlTable>(ID_Table);
+    const auto& selection = table->GetSelection();
+    boost::filesystem::path campaingPath = selection ? table->GetItemText(*selection, 5) : boost::filesystem::path();
+    if(campaingPath.empty())
+        throw std::runtime_error("Campaign path is empty.");
+    return campaingPath;
```
Flamefire, line 150: The column can never be empty, can it?
Maybe rather:
```suggestion
    if(!selection)
        throw std::runtime_error("Campaign path is empty.");
    boost::filesystem::path campaignPath = table->GetItemText(*selection, 5);
    Rttr_Assert(!campaignPath.empty());
    return campaignPath;
```

libs/s25main/desktops/dskCampaignSelection.h
```diff
@@ -0,0 +1,33 @@
...
+    void Msg_TableSelectItem(unsigned ctrl_id, const boost::optional<unsigned>& selection) override;
+    void Msg_ButtonClick(unsigned ctrl_id) override;
+    void Msg_Timer(unsigned ctrl_id) override;
+    void FillCampaignsTable();
+    void showCampaignInfo(bool show);
+    boost::filesystem::path getSelectedCampaignPath();
+    bool showCampaignInfo_;
+    CreateServerInfo csi_;
+    glArchivItem_Bitmap* campaignImage_;
+    /// Campaings that we already know are broken
```
Flamefire, line 31:
```suggestion
    /// Campaigns that we already know are broken
```

tests/s25Main/campaign/testCampaignLuaFile.cpp
```diff
@@ -0,0 +1,274 @@
...
+    BOOST_TEST(desc.version == "1");
+    BOOST_TEST(desc.author == "Max Meier");
+    BOOST_TEST(desc.name == "Meine Kampagne");
+    BOOST_TEST(desc.shortDescription == "Sehr kurze Beschreibung");
+    BOOST_TEST(desc.longDescription == "Das ist die lange Beschreibung");
+    BOOST_TEST(desc.image == "<RTTR_GAME>/GFX/PICS/WORLD.LBM");
+    BOOST_TEST(desc.maxHumanPlayers == 1u);
+    BOOST_TEST(desc.difficulty == "easy");
+    BOOST_TEST(desc.mapFolder == "<RTTR_GAME>/DATA/MAPS");
+    BOOST_TEST(desc.luaFolder == "<RTTR_GAME>/CAMPAINGS/ROMAN");
```
Flamefire, line 121:
```suggestion
    BOOST_TEST(desc.luaFolder == "<RTTR_GAME>/CAMPAIGNS/ROMAN");
```

libs/s25main/network/GameClient.h
```diff
@@ -13,12 +13,14 @@
...
 #include "gameTypes/AIInfo.h"
 #include "gameTypes/ChatDestination.h"
+#include "gameTypes/MapDescription.h"
 #include "gameTypes/MapInfo.h"
 #include "gameTypes/Nation.h"
 #include "gameTypes/ServerType.h"
 #include "gameTypes/TeamTypes.h"
 #include "gameTypes/VisualSettings.h"
 #include "s25util/Singleton.h"
+#include <boost/optional.hpp>
```
Flamefire, line 23:
```suggestion
```

libs/s25main/network/GameServer.h
```diff
@@ -10,11 +10,13 @@
 #include "GlobalGameSettings.h"
 #include "JoinPlayerInfo.h"
 #include "NWFInfo.h"
+#include "gameTypes/MapDescription.h"
 #include "gameTypes/MapInfo.h"
 #include "gameTypes/ServerType.h"
 #include "liblobby/LobbyInterface.h"
 #include "s25util/LANDiscoveryService.h"
 #include "s25util/Singleton.h"
+#include <boost/optional.hpp>
```
Flamefire, line 19:
```suggestion
```

Flamefire (changes requested):
Looking good, almost done. Found a couple minor things

The map/lua-path-by-index thing is strictly optional. The other things are mostly spelling issues/typos.

I expect this to be ready afterwards. Can you provide screenshots of the 2 new screens in this PR? @Spikeone please check them :)

Thanks again!

libs/s25main/desktops/dskCampaignMissionSelection.cpp
ottml: Potential i will implement this :)

ottml (commented):
Added suggestions.

doc/AddingCustomCampaign.md
```diff
@@ -0,0 +1,94 @@
...
+-->
+
+# Add a custom campaign
+
+You can have a look into the already existing original Settlers 2 campaigns in the subfolders `RTTR/CAMPAIGNS/roman` and `RTTR/CAMPAIGNS/continent` for an example.
+We will now create an example campaign `garden`.
+
+## Location for adding a new campaign
+
+To add a new custom campaign you have to create a new subfolder under `RTTR/CAMPAIGNS/` or under `<RTTR_USERDATA>/CAMPAIGNS/` with your campaign name. This subfolder will be filled with all the needed data for the campaign.
```
Flamefire, line 14: I'm afraid users won't understand this without an example:
```suggestion
To add a new custom campaign you have to create a new subfolder with your campaign name under `RTTR/CAMPAIGNS/` in the install directory or under `<RTTR_USERDATA>/CAMPAIGNS/`.
`<RTTR_USERDATA>` is the placeholder for the RttR folder in your user dir, e.g. `~/.s25rttr` on Linux or "Saved Games" on Windows.  
This subfolder will be filled with all the needed data for the campaign.
```

Flamefire (changes requested):
Just a minor thing, besides that it is good to go from my side :+1: 

Please rebase (with `--autosquash` to incorporate the fixups) and force push.   
You'll likely need to wait for https://github.com/Return-To-The-Roots/s25client/pull/1634 though for the fix of CI after something has changed in the runners.

libs/s25main/desktops/dskCampaignMissionSelection.cpp
ottml: Added index access and made mapNames/luaFolder/mapFolder private. And use it at calling side.

libs/libGamedata/gameData/CampaignDescription.cpp
```diff
@@ -37,12 +37,28 @@ CampaignDescription::CampaignDescription(const kaguya::LuaRef& table)
...
+    if(idx < mapNames.size())
+        return mapNames[idx];
+    return std::string();
+}
+
+boost::filesystem::path CampaignDescription::getLuaFilePath(const size_t idx) const
+{
+    if(idx < mapNames.size())
+        return (RTTRCONFIG.ExpandPath(luaFolder) / mapNames[idx]).replace_extension("lua");
+    return boost::filesystem::path();
```
Flamefire, line 56: Better throw an exception here and in the others

ottml:
campaign_final1
campaign_final2
campaign_final3
campaign_final4
campaign_final5

libs/libGamedata/gameData/CampaignDescription.cpp
ottml: Make no check anymore at use at(idx) on vector container so a exception is thrown automatically when out of range.

libs/s25main/network/GameClient.cpp
ottml: Can we do this in another merge request or is there a location where i can move the function without having to update a submodule. If i understand the submodules correct i have to make a PR for the submodule repo and then after this is merged back i have to change the pointer to the submodule in a PR of the s5client repo.

Flamefire:
Awesome, I think this is ready now!

@ottml I rebased the branch squashing the fixup-commits into their originals (i.e `git commit -i --autosquash master`) and added a minor improvement on top to avoid you having to do another change after you already did so much here. I hope that is ok for you, otherwise feel free to do the squashing-rebase yourself and force-push your branch again.   
Note: If you want/need to work on the updated (rebased) branch again you'll have the update your local branch: `git fetch <your-remote> && git reset --hard <your-remote>/campaign_screen_final`. But with your agreement I'll merge this as soon as CI has passed so you can also just delete that local branch and continue of the master after the merge.    

Please tell me if this is OK for you to merge this with that. It is your work so I'm waiting for your OK or update here. Thank you!

libs/s25main/network/GameClient.cpp
Flamefire: Yes, IMO we can do that after this PR

ottml:
> Awesome, I think this is ready now!
> 
> @ottml I rebased the branch squashing the fixup-commits into their originals (i.e `git commit -i --autosquash master`) and added a minor improvement on top to avoid you having to do another change after you already did so much here. I hope that is ok for you, otherwise feel free to do the squashing-rebase yourself and force-push your branch again. Note: If you want/need to work on the updated (rebased) branch again you'll have the update your local branch: `git fetch <your-remote> && git reset --hard <your-remote>/campaign_screen_final`. But with your agreement I'll merge this as soon as CI has passed so you can also just delete that local branch and continue of the master after the merge.
> 
> Please tell me if this is OK for you to merge this with that. It is your work so I'm waiting for your OK or update here. Thank you!

@Flamefire It's OK for me to merge it. Thanks for the fast and productive review :)

Flamefire:
@Flow86 Blocked by:

- https://github.com/Return-To-The-Roots/s25client/pull/1634

Flamefire (approved):

PR #1634, opened 2023-11-04
CI: Fix finding Boost on OSX

Flamefire:
The brew installed boost changed to a new version. Update the code to just always use the latest in that folder instead of a specific one.


Flow86 (approved):

Flamefire:
@Flow86 The required check "Linux (clang, macos-12, Debug, 1.82.0)" needs to be changed and I guess Jenkins got confused by the force-push and needs to be restarted. Can you merge this please?

PR #1641, opened 2023-11-16
Fix flakyness in Test_network on CI

Flamefire:
Add more checks and retries to avoid that often failing test on GHA with macos


Flamefire:
@Flow86 Can you merge this too please? That configuration caused failures in otherwise correct contributions multiple times now.

Issue #1520, opened 2022-04-10
savegame crashes


Spikeone:
This is a savegame of a scripted map - it always crashes after some GF. Not sure what the cause is, the console does not indicate a LUA error.

[4.zip](https://github.com/Return-To-The-Roots/s25client/files/8459952/4.zip)
[WORLDS.zip](https://github.com/Return-To-The-Roots/s25client/files/8459953/WORLDS.zip)

Spikeone:
Replay is also crashing:
[2022-04-10_23-33-01.zip](https://github.com/Return-To-The-Roots/s25client/files/8459979/2022-04-10_23-33-01.zip)

wichern:
I think what happens here, is that there is a flag at 88,7.
When loading the savegame, `MissionEvent(e=1)` is triggered, which adds a static object of size 2.
`rttr:GetWorld():AddStaticObject(89, 8, 560, 0xFFFF, 2)`

This causes the removal of the flag.
Unfortunately, at that moment there is already a worker carrying something to that flag. When he reaches it and wants to place his goods, we have a nullptr dereference.

The lua doc says `e=1` is an initial event.
I guess this should only be called when the game starts and not on every loading of a savegame.

Spikeone:
Good find @wichern I guess - never understood the stack trace correctly.

I'm still confused since when having a look at the script, from the very beginning ```MissionEvent(1)``` has been called. That means ```rttr:GetWorld():AddStaticObject(89, 8, 560, 0xFFFF, 2)``` should work just fine - since an object should block all the space it requires. Although the ```MissionEvent(99)``` calls ```rttr:GetWorld():AddStaticObject(89, 8, 561, 0xFFFF, 2)``` as well, this shouldn't be a problem.

So maybe for this bug here, the root cause then is #1516 - maybe roads are removed when placing the object, but spots are not blocked correctly, thus leading to the crash.

wichern:
The build locations for portals are strange. And they get even stranger, when building/removing stuff around them. Seems they bend time and space :)
![Screenshot from 2024-02-27 18-51-15](https://github.com/Return-To-The-Roots/s25client/assets/47030768/fd265f16-7349-467e-821c-8683d907b50c)
![Screenshot from 2024-02-27 18-51-20](https://github.com/Return-To-The-Roots/s25client/assets/47030768/07d6c605-fca6-4c56-aa96-3228e3c3e45d)
[...]

Issue #1516, opened 2022-03-29
Gate object gets wrong size


Spikeone:
If you place a gate on the map it should be a large building. Currently it's not (closed gate).

Also spawning an object by using ```rttr:GetWorld():AddStaticObject(72, 28, 561, 0xFFFF, 2)``` doesn't result in a large building:
![grafik](https://user-images.githubusercontent.com/1276342/160649336-9b3fae32-a96b-4572-912d-e86fa64b78cb.png)

wichern:
I cannot reproduce this issue.
![image](https://github.com/Return-To-The-Roots/s25client/assets/47030768/fd2236a6-5409-4813-83a4-dfe0c790e3e7)
Do you have a savegame?

Spikeone:
fixed by #1647 - thanks to @wichern

PR #1647, opened 2024-02-29
Fix recalculating BQ around big static objects

wichern:
The recalculation around big static objects did not work properly.

Before this patch:
Screenshot from 2024-02-27 18-51-15
Screenshot from 2024-02-27 18-51-20
Screenshot from 2024-02-27 18-51-25

After this patch:
Screenshot from 2024-02-29 11-50-50

Maybe cause of #1516 and #1520

### DIFF
--- a/libs/s25main/lua/LuaWorld.cpp
+++ b/libs/s25main/lua/LuaWorld.cpp
@@ -84,7 +84,10 @@ bool LuaWorld::AddStaticObject(int x, int y, unsigned id, unsigned file /* = 0xF
 
     gw.DestroyNO(pt, false);
     gw.SetNO(pt, new noStaticObject(pt, id, file, size));
-    gw.RecalcBQAroundPoint(pt);
+    if(size > 1)
+        gw.RecalcBQAroundPointBig(pt);
+    else if(size == 1)
+        gw.RecalcBQAroundPoint(pt);
     return true;
 }
 


libs/s25main/lua/LuaWorld.cpp
```diff
@@ -84,7 +84,10 @@ bool LuaWorld::AddStaticObject(int x, int y, unsigned id, unsigned file /* = 0xF
 
     gw.DestroyNO(pt, false);
     gw.SetNO(pt, new noStaticObject(pt, id, file, size));
-    gw.RecalcBQAroundPoint(pt);
+    if(2 == size)
```
Flamefire, line 87: Please avoid "Yoda-conditions" --> `if(size == 2)`

Flamefire (changes requested):
Maybe we can make this even better: `size >1 -> RecalcBQAroundPointBig`, `size=1 -> RecalcBQAroundPoint`, else nothing.

Could you do that?

libs/s25main/lua/LuaWorld.cpp
wichern: Thank you for your suggestions. I've updated the PR.

I have two general questions:
1. Do you use a pre-commit hook in this repo? I haven't found one.
2. Is it ok to force-push in order to keep the amount of commits low? I noticed that changes are not squashed.

Flamefire: 1. We don't. But we do have a build target `make clangFormat` (IIRC) that is available when (the right version of) clang-format is found during configure/cmake that can be used to format (changed) files explicitely.

2. It depends ;-) If it is a single commit with a single small change then I'd say it is ok. Especially as GitHub now shows force-pushes and allows to compare the difference. Otherwise you'd better use additional commits as long as the review is in progress and rebase and squash before merge when review is finished. You can use `git commit --fixup` to squash automatically on a rebase (see the man page). But in the end having those extra commits isn't a real problem. E.g. in the case here the history would have shown the "sub-optimal" solution and the conscious change to the current solution which is often a good thing.    
So in summary: If the additional commit might be helpful (e.g. for review or future reference) or if in doubt keep it.

Flamefire (approved):

Issue #1621, opened 2023-08-23
Issue opening game on macOS 3/4 of window is red


diwilli:
under a fully updated macOS 13.5

This happens on both latest stable 0.9.5 and latest nightly s25rttr_20230814-e37a322209cffb00fda4afd50c87b8f7e8261083-darwin.x86_64.tar.bz2

Version 0.8.1 doesn't have this issue and launches correctly.

<img width="1303" alt="Screenshot 2023-08-23 at 13 43 23" src="https://github.com/Return-To-The-Roots/s25client/assets/14251725/c360dc7b-1740-492c-837e-0e469bf9e822">

Happy to test updates as they're available.

diwilli:
I'm intending to git bisect this, and try and work out where it breaks.

The only version I can get to build is v0.9.5, trying to build anything before that gives me the following errors (I can't seem to workout where to set the -Wno-deprecated-copy flag to avoid this (I'm not a C/C++ developer):

`In file included from /Users/dan/rttr/s25client/libs/common/include/Rect.h:7:
/Users/dan/rttr/s25client/libs/common/include/Point.h:27:15: error: definition of implicit copy assignment operator for 'Point<float>' is deprecated because it has a user-declared copy constructor [-Werror,-Wdeprecated-copy]
    constexpr Point(const Point&) = default;
              ^
/Users/dan/rttr/s25client/libs/libGamedata/gameData/TerrainDesc.cpp:147:24: note: in implicit copy assignment operator for 'Point<float>' first required here
            result.tip = middleBottom;
                       ^
1 error generated.
make[2]: *** [libs/libGamedata/CMakeFiles/gamedata.dir/gameData/TerrainDesc.cpp.o] Error 1
make[1]: *** [libs/libGamedata/CMakeFiles/gamedata.dir/all] Error 2
make: *** [all] Error 2`

Any pointers?

diwilli:
I've found -DRTTR_ENABLE_WERROR=ON, setting this to -DRTTR_ENABLE_WERROR=OFF seems to do the trick :)

Flamefire:
> Which feels like an actual code issue, and I'm sorry I don't know how to solve that one.

When you checkout different commits (e.g. when bisecting) you need to also checkout the submodules used at that version, i.e. `git submodule update`. What you see is basically: The submodule was changed removing a function and you have a version of the main code still using the function or the other way round: Main code using a function introduced in a later version of the submodule.

[...]

PR #1648, opened 2024-03-03
Fix high DPI issue on macOS

tiennou:
The bug is described in #1621, and is about the game using only the bottom-left corner of the window to draw, and the mouse cursor bounding box not being correct (so you can't access Settings > Graphics, as it gets clamped by the bottom of the screen).

The first commit fixes the rendering to be full-screen again, but the mouse cursor is still "locked" into the top-left corner (but the buttons work).
The second make the mouse cursor move properly by scaling its position with the DPI, but that makes the buttons active area all wrong, and I can't see where the hit-testing is actually happening.

### DIFF
--- a/libs/driver/src/VideoDriver.cpp
+++ b/libs/driver/src/VideoDriver.cpp
@@ -27,7 +27,7 @@ VideoDriver::VideoDriver(VideoDriverLoaderInterface* CallBack)
 
 Position VideoDriver::GetMousePos() const
 {
-    return mouse_xy.pos;
+    return mouse_xy.pos * (int)dpiScale_;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


Flamefire:
> The second make the mouse cursor move properly by scaling its position with the DPI, but that makes the buttons active area all wrong, and I can't see where the hit-testing is actually happening.

The problem here is that you are not correcting the mouse-move/click/drag operations. See e.g. https://github.com/Return-To-The-Roots/s25client/blob/73acea701915e0b857482af6006cfb9342925032/extras/videoDrivers/SDL2/VideoSDL2.cpp#L401 or search this file for `mouse_xy.pos = `

I suspect you don't need the 2nd commit here and just can remove the `getGuiScale().screenToView` from VideoSDL2, but haven't fully checked that. It just looks like you basically undo the scaling by `screenToView` with `* dpiScale` which sounds like the scaling is wrong in the first place.

PR #1651, opened 2024-03-09
fix AI interaction with granite mines with enabled Inexhaustible gran…

desofity:
Hello!

Currently if addon "Inexhaustible granite mines" is enabled, AI place granite mine on any available mine place. (Because method "SimpleFindPosition" is used instead of "FindBestPosition") Then after building of mine is complete, AI reailizes that there is no resource available and demolishes granite mine.

The code comment says that "inexhaustible granite mines do not require granite" but i can't find any code line that proves that.
If that was originally intended to allow building granite mine on any mine place if addon is enabled, we need to add some code in resource map generator, probably. But I decided that this was not intended and just removed condition.

### DIFF
--- a/libs/s25main/ai/aijh/AIPlayerJH.cpp
+++ b/libs/s25main/ai/aijh/AIPlayerJH.cpp
@@ -977,7 +977,7 @@ MapPoint AIPlayerJH::FindPositionForBuildingAround(BuildingType type, const MapP
             foundPos = FindBestPosition(around, AIResource::Ironore, BuildingQuality::Mine, searchRadius);
             break;
         case BuildingType::GraniteMine:
-			foundPos = FindBestPosition(around, AIResource::Granite, BuildingQuality::Mine, searchRadius);
+          foundPos = FindBestPosition(around, AIResource::Granite, BuildingQuality::Mine, searchRadius);
             break;
         case BuildingType::Fishery:
             foundPos = FindBestPosition(around, AIResource::Fish, BUILDING_SIZE[type], searchRadius);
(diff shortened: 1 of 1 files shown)


Flow86 (dismissed):
you're totally right, inexhaustible granite mines only says that a valid granite mine never runs out of stone, not that you can place it everywhere.

Spikeone:
well @Flow86 and @desofity - the AI currently correclty implements the addon - since thats how I implemented the addon once (just place graninte mines anywhere). At some point @Flamefire 'fixed' it since the description wasn't clear and it sounded like it is a minimal version of the inexhaustible mines addon. But yeah, the fix is probably fine for now, I need to add my settlers IV mining addons which also solves that addon issue :D

libs/s25main/ai/aijh/AIPlayerJH.cpp
```diff
@@ -977,13 +977,8 @@ MapPoint AIPlayerJH::FindPositionForBuildingAround(BuildingType type, const MapP
             foundPos = FindBestPosition(around, AIResource::Ironore, BuildingQuality::Mine, searchRadius);
             break;
         case BuildingType::GraniteMine:
-            if(!ggs.isEnabled(
-                 AddonId::INEXHAUSTIBLE_GRANITEMINES)) // inexhaustible granite mines do not require granite
-                foundPos = FindBestPosition(around, AIResource::Granite, BuildingQuality::Mine, searchRadius);
-            else
-                foundPos = SimpleFindPosition(around, BuildingQuality::Mine, searchRadius);
+			foundPos = FindBestPosition(around, AIResource::Granite, BuildingQuality::Mine, searchRadius);
```
Flamefire, line 980: Only the indentation is off:
```suggestion
          foundPos = FindBestPosition(around, AIResource::Granite, BuildingQuality::Mine, searchRadius);
```

Flamefire (changes requested):

Flamefire (approved):
Thanks!

Flow86 (approved):

PR #1652, opened 2024-03-10
Add ai-battle cli

wichern:
Adds an optional binary that runs headless (no UI) AI battles.
This can be useful when creating training data or testing new AIs.

image

I'm not sure if this is placed correctly or whether it should even be part of the master branch.
In any case, I would like your feedback on this tool. It aims to reuse as much of s25main as possible in order to create accurate replays.


Flamefire (approved):

extras/ai-battle/HeadlessGame.cpp
```diff
@@ -0,0 +1,224 @@
...
+}
+
+void HeadlessGame::Close()
+{
+    printf("\n");
+
+    if(replay_.IsRecording())
+    {
+        replay_.StopRecording();
+        printf("Replay written to %s\n", canonical(replayPath_).c_str());
```
Flamefire, line 108:
```suggestion
        bnw::cout << "Replay written to " << canonical(replayPath_) << '\n';
```

extras/ai-battle/HeadlessGame.cpp
```diff
@@ -0,0 +1,224 @@
...
+    Savegame save;
+    for(unsigned playerId = 0; playerId < world_.GetNumPlayers(); ++playerId)
+        save.AddPlayer(world_.GetPlayer(playerId));
+    save.ggs = game_.ggs_;
+    save.ggs.exploration = Exploration::Disabled; // no FOW
+    save.start_gf = em_.GetCurrentGF();
+    save.sgd.MakeSnapshot(game_);
+    save.Save(path, "AI Battle");
+
+    printf("Savegame written to %s\n", canonical(path).c_str());
```
Flamefire, line 148:
```suggestion
        bnw::cout << "Savegame written to " << canonical(path) << '\n';
```

extras/ai-battle/main.cpp
```diff
@@ -0,0 +1,140 @@
...
+#include "QuickStartGame.h"
+#include "RTTR_Version.h"
+#include "RttrConfig.h"
+#include "files.h"
+#include "random/Random.h"
+#include "s25util/System.h"
+
+#include <boost/filesystem.hpp>
+#include <boost/nowide/args.hpp>
+#include <boost/nowide/iostream.hpp>
```
Flamefire, line 16:
```suggestion
#include <boost/nowide/filesystem.hpp>
#include <boost/nowide/iostream.hpp>
```

extras/ai-battle/main.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+namespace bnw = boost::nowide;
+namespace bfs = boost::filesystem;
+namespace po = boost::program_options;
+
+std::atomic<bool> g_abort = false;
+
+int main(int argc, char** argv)
+{
+    bnw::args _(argc, argv);
```
Flamefire, line 29:
```suggestion
    bnw::nowide_filesystem();
    bnw::args _(argc, argv);
```

extras/ai-battle/HeadlessGame.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+#include <chrono>
+#include <sstream>
+
+std::vector<PlayerInfo> GeneratePlayerInfo(const std::vector<AI::Info>& ais);
+std::string ToString(const std::chrono::milliseconds& time);
+std::string HumanReadableNumber(unsigned num);
+
+namespace bfs = bfs;
+using bfs::canonical;
```
Flamefire, line 27:
```suggestion
#include <boost/nowide/iostream.hpp>
#include <chrono>
#include <sstream>

std::vector<PlayerInfo> GeneratePlayerInfo(const std::vector<AI::Info>& ais);
std::string ToString(const std::chrono::milliseconds& time);
std::string HumanReadableNumber(unsigned num);

namespace bfs = boost::filesystem;
namespace bnw = boost::nowide;
using bfs::canonical;
```

Flamefire (changes requested):
Few remarks to fix compilation

This will likely glitch on Windows though:
>        `printf("\x1b[%dA", 8 + world_.GetNumPlayers()); // Move cursor back up`

wichern:
> Few remarks to fix compilation
> 
> This will likely glitch on Windows though:
> 
> > ```
> >    `printf("\x1b[%dA", 8 + world_.GetNumPlayers()); // Move cursor back up`
> > ```

You are right.

It seems that I could fix this by calling `SetConsoleMode`.
https://learn.microsoft.com/en-us/windows/console/setconsolemode

I will have to set up a windows build machine to test it.

Flamefire:
> It seems that I could fix this by calling `SetConsoleMode`. https://learn.microsoft.com/en-us/windows/console/setconsolemode

There is a condition that `ENABLE_PROCESSED_OUTPUT` only works for "WriteFile or WriteConsole".
I suggest you use the C++ streams everywhere but `PrintState` and for that implement a wrapper instead of `printf` calling `vsprintf` and passing the result to either `std::puts` or `WriteConsole`.

Something like 
```
#ifdef WIN32
HANDLE setupStdOut()
{
  HANDLE h = GetStdHandle(STD_OUTPUT_HANDLE);
  SetConsoleMode(h, ENABLE_PROCESSED_OUTPUT);
  return h;
}
#endif

void printConsole(const char* fmt, ... )
{
   ...
   vsprintf(...)
#ifdef WIN32
  static auto h = setupStdOut();
  WriteConsole(h, formattedString);
#else
  std::puts(formattedString();
#endif
}
```

Core ideas:
- Use a local static variable to get the stdout handle and ensure the console mode is set exactly once
- Keep the code until `std::puts`/`WriteConsole` the same to be able to spot mistakes in formatting on both systems 
 
BTW: `#include <cstdio>` is missing although it seemingly gets pulled in by an other header

extras/ai-battle/HeadlessGame.cpp
```diff
@@ -0,0 +1,262 @@
...
+
+namespace bfs = boost::filesystem;
+namespace bnw = boost::nowide;
+using bfs::canonical;
+
+#ifdef WIN32
+HANDLE setupStdOut();
+void printConsole(const char* fmt, ...);
+#else
+void printConsole(const char* fmt, ...) __attribute__((format(gnu_printf, 1, 2)));
```
Flamefire, line 36: Try this:

```suggestion
#elif defined(__GNUC__)
void printConsole(const char* fmt, ...) __attribute__ ((format (printf, 1, 2)));
#else
void printConsole(const char* fmt, ...);
```

extras/ai-battle/HeadlessGame.cpp
```diff
@@ -0,0 +1,262 @@
...
+    return h;
+}
+#endif
+
+void printConsole(const char* fmt, ...)
+{
+    static char buffer[512];
+    va_list args;
+    va_start(args, fmt);
+    int len = vsnprintf(buffer, sizeof(buffer), fmt, args);
```
Flamefire, line 251:
```suggestion
    const int len = vsnprintf(buffer, sizeof(buffer), fmt, args);
```

extras/ai-battle/HeadlessGame.cpp
```diff
@@ -0,0 +1,264 @@
...
+using bfs::canonical;
+
+#ifdef WIN32
+HANDLE setupStdOut();
+void printConsole(const char* fmt, ...);
+#elif defined(__GNUC__)
+void printConsole(const char* fmt, ...) __attribute__((format(printf, 1, 2)));
+#else
+void printConsole(const char* fmt, ...);
+#endif
```
Flamefire, line 39: Only Jenkins errors with the warning:
> /source/extras/ai-battle/HeadlessGame.cpp:253:64: error: function 'void printConsole(const char*, ...)' might be a candidate for 'gnu_printf' format attribute [-Werror=suggest-attribute=format]

We are using MinGW as the cross-compiler and hence GCC, so this can be reproduced on godbolt: https://godbolt.org/z/sjx4ooMGW

This should work:
```suggestion
#ifdef WIN32
HANDLE setupStdOut();
#endif

#ifdef __GNUC__
void printConsole(const char* fmt, ...) __attribute__((format(printf, 1, 2)));
#else
void printConsole(const char* fmt, ...);
#endif
```

extras/ai-battle/HeadlessGame.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+    bnw::cout << "Savegame written to " << canonical(path) << '\n';
+}
+
+std::string ToString(const std::chrono::milliseconds& time)
+{
+    char buffer[90];
+    auto hours = std::chrono::duration_cast<std::chrono::hours>(time);
+    auto minutes = std::chrono::duration_cast<std::chrono::minutes>(time % std::chrono::hours(1));
+    auto seconds = std::chrono::duration_cast<std::chrono::seconds>(time % std::chrono::minutes(1));
```
Flamefire, line 168: Always const, makes reasoning easier:
```suggestion
    const auto hours = std::chrono::duration_cast<std::chrono::hours>(time);
    const auto minutes = std::chrono::duration_cast<std::chrono::minutes>(time % std::chrono::hours(1));
    const auto seconds = std::chrono::duration_cast<std::chrono::seconds>(time % std::chrono::minutes(1));
```

extras/ai-battle/HeadlessGame.cpp
```diff
@@ -0,0 +1,264 @@
...
+    snprintf(buffer, 90, "%03ld:%02ld:%02ld", static_cast<long int>(hours.count()),
```
Flamefire, line 169: No magic numbers if possible:
```suggestion
    snprintf(buffer, std::size(buffer), "%03ld:%02ld:%02ld", static_cast<long int>(hours.count()),
```

extras/ai-battle/HeadlessGame.h
```diff
@@ -0,0 +1,49 @@
...
+
+#pragma once
+
+#include "Game.h"
+#include "Replay.h"
+#include "ai/AIPlayer.h"
+#include "gameTypes/AIInfo.h"
+
+#include <boost/filesystem.hpp>
+
```
Flamefire, line 13: Newlines not required
```suggestion
#include <boost/filesystem.hpp>
```

extras/ai-battle/HeadlessGame.cpp
```diff
@@ -0,0 +1,264 @@
...
+    if(replay_.IsRecording())
+    {
+        replay_.StopRecording();
+        bnw::cout << "Replay written to " << canonical(replayPath_) << '\n';
+    }
+
+    replay_.Close();
+}
+
+void HeadlessGame::StartReplay(const bfs::path& path, unsigned random_init)
```
Flamefire, line 126: Better name as I thought this was for starting *from* a replay

```suggestion
void HeadlessGame::RecordReplay(const bfs::path& path, unsigned random_init)
```

extras/ai-battle/HeadlessGame.cpp
```diff
@@ -0,0 +1,264 @@
...
+    HANDLE h = GetStdHandle(STD_OUTPUT_HANDLE);
+    SetConsoleMode(h, ENABLE_PROCESSED_OUTPUT | ENABLE_VIRTUAL_TERMINAL_PROCESSING);
+    SetConsoleOutputCP(65001);
+    return h;
+}
+#endif
+
+void printConsole(const char* fmt, ...)
+{
+    static char buffer[512];
```
Flamefire, line 250: Any reason to have this static?

```suggestion
    char buffer[512];
```

extras/ai-battle/HeadlessGame.h
```diff
@@ -0,0 +1,49 @@
...
+    void Run(unsigned maxGF = std::numeric_limits<unsigned>::max());
+    void Close();
+
+    void StartReplay(const boost::filesystem::path& path, unsigned random_init);
+    void SaveGame(const boost::filesystem::path& path) const;
+
+private:
+    void PrintState();
+
+    const boost::filesystem::path& map_;
```
Flamefire, line 38: No need for a reference here is there?
```suggestion
    boost::filesystem::path map_;
```

extras/ai-battle/main.cpp
```diff
@@ -0,0 +1,148 @@
...
+#include <boost/program_options.hpp>
+
+#include <atomic>
+
+namespace bnw = boost::nowide;
+namespace bfs = boost::filesystem;
+namespace po = boost::program_options;
+
+std::atomic<bool> g_abort = false;
+
```
Flamefire, line 27: Unused?

```suggestion
```

extras/ai-battle/main.cpp
```diff
@@ -0,0 +1,148 @@
...
+#include "s25util/System.h"
+
+#include <boost/filesystem.hpp>
+#include <boost/nowide/args.hpp>
+#include <boost/nowide/filesystem.hpp>
+#include <boost/nowide/iostream.hpp>
+#include <boost/program_options.hpp>
+
+#include <atomic>
+
```
Flamefire, line 21: See below
```suggestion
```

extras/ai-battle/main.cpp
```diff
@@ -0,0 +1,148 @@
...
+{
+    bnw::nowide_filesystem();
+    bnw::args _(argc, argv);
+
+    po::options_description desc("Allowed options");
+    // clang-format off
+    desc.add_options()
+        ("help,h", "Show help")
+        ("map,m", po::value<std::string>(),"Map to load")
+        ("ai", po::value<std::vector<std::string>>(),"AI player(s) to add")
```
Flamefire, line 38:
```suggestion
        ("map,m", po::value<std::string>()->required(),"Map to load")
        ("ai", po::value<std::vector<std::string>>()->required(),"AI player(s) to add")
```

extras/ai-battle/main.cpp
```diff
@@ -0,0 +1,148 @@
...
+        bnw::cout << desc << std::endl;
+        return 0;
+    }
+    if(options.count("version"))
+    {
+        bnw::cout << rttr::version::GetTitle() << " v" << rttr::version::GetVersion() << "-"
+                  << rttr::version::GetRevision() << std::endl
+                  << "Compiled with " << System::getCompilerName() << " for " << System::getOSName() << std::endl;
+        return 0;
+    }
```
Flamefire, line 77: Move this (help/version handling) before the  `po::notify`

extras/ai-battle/main.cpp
```diff
@@ -0,0 +1,148 @@
...
+    if(options.count("map") == 0)
+    {
+        bnw::cerr << "No map specified" << std::endl;
+        return 1;
+    }
+    if(options.count("ai") == 0)
+    {
+        bnw::cerr << "No AI specified" << std::endl;
+        return 1;
+    }
```
Flamefire, line 87: Handled by `->required` which is less code

extras/ai-battle/main.cpp
```diff
@@ -0,0 +1,148 @@
...
+        RTTRCONFIG.Init();
+        RANDOM.Init(random_init);
+
+        const bfs::path mapPath = RTTRCONFIG.ExpandPath(options["map"].as<std::string>());
+        const std::vector<AI::Info> ais = ParseAIOptions(options["ai"].as<std::vector<std::string>>());
+
+        GlobalGameSettings ggs;
+        if(options.count("objective"))
+        {
+            std::string objective = options["objective"].as<std::string>();
```
Flamefire, line 111:
```suggestion
            const auto objective = options["objective"].as<std::string>();
```

extras/ai-battle/main.cpp
```diff
@@ -0,0 +1,148 @@
...
+            ggs.objective = GameObjective::TotalDomination;
+
+        ggs.objective = GameObjective::TotalDomination;
+        HeadlessGame game(ggs, mapPath, ais);
+        if(options.count("replay"))
+            game.StartReplay(options["replay"].as<std::string>(), random_init);
+
+        unsigned maxGF = std::numeric_limits<unsigned>::max();
+        if(options.count("maxGF"))
+            maxGF = options["maxGF"].as<unsigned>();
```
Flamefire, line 131: Optional:
I suggest to use `->default_value(std::numeric_limits<unsigned>::max())` in the `desc.add_options` part for this and the other values having defaults (objective and random_init)

Or `boost::optional` and `.value_or(...)` here

extras/ai-battle/main.cpp
```diff
@@ -0,0 +1,148 @@
...
+                bnw::cerr << "unknown objective: " << objective << std::endl;
+                return 1;
+            }
+        } else
+            ggs.objective = GameObjective::TotalDomination;
+
+        ggs.objective = GameObjective::TotalDomination;
+        HeadlessGame game(ggs, mapPath, ais);
+        if(options.count("replay"))
+            game.StartReplay(options["replay"].as<std::string>(), random_init);
```
Flamefire, line 127: Optional:
Can you use `boost::optional` for this (and savegame)? See https://www.boost.org/doc/libs/1_84_0/doc/html/program_options/howto.html#id-1.3.30.6.10

extras/ai-battle/main.cpp
```diff
@@ -0,0 +1,148 @@
...
+        bnw::cout << std::endl;
+
+        RTTRCONFIG.Init();
+        RANDOM.Init(random_init);
+
+        const bfs::path mapPath = RTTRCONFIG.ExpandPath(options["map"].as<std::string>());
+        const std::vector<AI::Info> ais = ParseAIOptions(options["ai"].as<std::vector<std::string>>());
+
+        GlobalGameSettings ggs;
+        if(options.count("objective"))
```
Flamefire, line 109: Optional:
Using `->default_value` avoids this if and the else branch

extras/ai-battle/main.cpp
```diff
@@ -0,0 +1,148 @@
...
+            game.SaveGame(options["save"].as<std::string>());
+    } catch(const std::exception& e)
+    {
+        bnw::cerr << e.what() << std::endl;
+        return 1;
+    } catch(...)
+    {
+        bnw::cerr << "An unknown exception occurred" << std::endl;
+        return 1;
+    }
```
Flamefire, line 145: Optional:
Why catch it at all? I'd think the error message by the system without this catch should have more information about this unknown error
```suggestion
    }
```

Flamefire (changes requested):
I marked a few comments as "Optional", you can skip those if you want. If you fix the others (and CI passes) we can merge this.

wichern:
> I marked a few comments as "Optional", you can skip those if you want. If you fix the others (and CI passes) we can merge this.

Thank's a lot. I will try to apply the other suggestions as well.

extras/ai-battle/main.cpp
wichern: done

extras/ai-battle/main.cpp
wichern: done

extras/ai-battle/main.cpp
wichern: done

extras/ai-battle/main.cpp
wichern: done

extras/ai-battle/main.cpp
wichern: done

extras/ai-battle/HeadlessGame.cpp
wichern: done

extras/ai-battle/main.cpp
```diff
@@ -0,0 +1,127 @@
...
+    po::options_description desc("Allowed options");
+    // clang-format off
+    desc.add_options()
+        ("help,h", "Show help")
+        ("map,m", po::value<std::string>()->required(),"Map to load")
+        ("ai", po::value<std::vector<std::string>>()->required(),"AI player(s) to add")
+        ("objective", po::value<std::string>()->default_value("domination"),"domination(default)|conquer")
+        ("replay", po::value(&replay_path),"Filename to write replay to (optional)")
+        ("save", po::value(&savegame_path),"Filename to write savegame to (optional)")
+        ("random_init", po::value<unsigned>()->default_value(std::chrono::high_resolution_clock::now().time_since_epoch().count()),"Seed value for the random number generator (optional)")
```
Flamefire, line 42: Compiler complains about signed->unsigned.
1. use an optional<unsigned> and `value_or(std::chrono...)`
2. Or: `const auto defaultSeed = static_cast<unsigned>(std::chrono::high_resolution_clock::now().time_since_epoch().count())` next to replay/savegame_path and use it here

extras/ai-battle/main.cpp
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        ggs.objective = GameObjective::TotalDomination;
+        HeadlessGame game(ggs, mapPath, ais);
+        if(replay_path)
+            game.RecordReplay(*replay_path, random_init);
+
+        unsigned maxGF = options["maxGF"].as<unsigned>();
+
+        game.Run(maxGF);
```
Flamefire, line 116: Either make it `const` or use it directly:
```suggestion
        game.Run(options["maxGF"].as<unsigned>());
```

extras/ai-battle/main.cpp
```diff
@@ -0,0 +1,127 @@
...
+    } catch(const std::exception& e)
+    {
+        bnw::cerr << "Error: " << e.what() << std::endl;
+        bnw::cerr << desc << std::endl;
+        return 1;
+    }
+
+    try
+    {
+        auto random_init = options["random_init"].as<unsigned>();
```
Flamefire, line 82:
```suggestion
        const auto random_init = options["random_init"].as<unsigned>();
```

Flamefire (changes requested):
ok, great!
There are 2 constness issues (Rule is: Use `const` always unless you really need to modify the variable/parameter)

CI fails due to 1 conversion issue and an unrelated issue with test replays on master which will be likely fixed later today or tomorrow.

extras/ai-battle/main.cpp
```diff
@@ -0,0 +1,124 @@
...
+namespace po = boost::program_options;
+
+int main(int argc, char** argv)
+{
+    bnw::nowide_filesystem();
+    bnw::args _(argc, argv);
+
+    boost::optional<std::string> replay_path;
+    boost::optional<std::string> savegame_path;
+    unsigned random_init = static_cast<unsigned>(std::chrono::high_resolution_clock::now().time_since_epoch().count());
```
wichern, line 32:
```suggestion
    const unsigned random_init = static_cast<unsigned>(std::chrono::high_resolution_clock::now().time_since_epoch().count());
```

forgot const.

wichern: done

extras/ai-battle/main.cpp
wichern: done

extras/ai-battle/main.cpp
wichern: done

extras/ai-battle/main.cpp
Flamefire: That was on purpose: The idea was to make this the default, i.e.: `po::value(&random_init)->default_value(defaultSeed)`
Or (I think) you can init  `random_init` here and use it as an implicit default. But in either case it must be non-const such that it can be set by the cmdline parser

Flamefire:
Once you are done and have squashed the changes the replay test will still fail (on appveyor) This will be fixed by https://github.com/Return-To-The-Roots/s25client/pull/1676
If that isn't merged yet when you are ready (and CI here succeeds except that one test) you could merge that branch into yours to make CI pass here so we can merge it.

Flamefire:
I rebased your branch onto current master and combined (squashed) some commits. I also fixed the (now) failing test on appveyor and merged the replay-fix branch again (instead of the previous merge)

As this was the only failing test and the diff of the rebased branch and yours is just the appveyor fix it should pass now. So I'll enable auto-merge.
Thanks a lot!

Flamefire (approved):

Flamefire (approved):

PR #1655, opened 2024-03-25
Add Arm optimization flags for Raspberry Pi 5 and include ARM 64 Bit

eltorrero:
Adds the optimization flags for Raspberry Pi 5 and also makes sure that 64 bit ARM is included, since CMAKE_SYSTEM_PROCESSOR is aarch64 in that case and not matching the condition "^arm".


cmake/Modules/ArmConfig.cmake
```diff
@@ -32,7 +34,7 @@ macro(get_arm_fpu result target)
         set(${result} vfp)
     elseif(target STREQUAL "RasPi2")
         set(${result} neon-vfpv4)
-    elseif(target MATCHES "RasPi3|RasPi4")
+    elseif(target MATCHES "RasPi3|RasPi4|RasPi5")
```
Flamefire, line 37: At this point we should use a range (same below):
```suggestion
    elseif(target MATCHES "RasPi[3-5]")
```

Flamefire (approved):
Thanks!

PR #1657, opened 2024-04-02
Don't allow roads in no-mans land or in border

### LINKED ISSUE #968: Road on Border
[2018-12-18_23-03-23.zip](https://github.com/Return-To-The-Roots/s25client/files/2692837/2018-12-18_23-03-23.zip)

Xellzul:
Roads are no longer allowed to be in no-mans land or in border

Fixes #968

The code is ugly so if you have any suggestions, feel free to fix it.

This should fix all problems, where roads were in places, where they shouldn't be.

If you need, I can provide few examples (savegames)


libs/s25main/world/GameWorld.cpp
```diff
@@ -413,43 +413,43 @@ void GameWorld::RecalcTerritory(const noBaseBuilding& building, TerritoryChangeR
             sizeChanges[oldOwner - 1]--;
     }
 
-    std::set<MapPoint, MapPointLess> ptsHandled;
+    const std::vector<MapPoint> ptsToHandled = GetAllNeighboursUnion(ptsWithChangedOwners);
```
Flamefire, line 416: Small typo, also needs changing use-sites:
```suggestion
    const std::vector<MapPoint> ptsToHandle = GetAllNeighboursUnion(ptsWithChangedOwners);
```

libs/s25main/world/GameWorld.cpp
```diff
@@ -413,43 +413,43 @@ void GameWorld::RecalcTerritory(const noBaseBuilding& building, TerritoryChangeR
...
+
     // Destroy everything from old player on all nodes where the owner has changed
-    for(const MapPoint& curMapPt : ptsWithChangedOwners)
+    for(const MapPoint& curMapPt : ptsToHandled)
     {
         // Destroy everything around this point as this is at best a border node where nothing should be around
         // Do not destroy the triggering building or its flag
```
Flamefire, line 422: Outdated comment, isn't it?

```suggestion
        // Destroy everything around this point as this is at best a border node where nothing should be around
```

libs/s25main/world/GameWorld.cpp
```diff
@@ -413,43 +413,43 @@ void GameWorld::RecalcTerritory(const noBaseBuilding& building, TerritoryChangeR
...
-        // TODO: What about this point?
         const uint8_t owner = GetNode(curMapPt).owner;
```
Flamefire, line 423: No need for the temp var:
```suggestion
```

libs/s25main/world/GameWorld.cpp
```diff
@@ -413,43 +413,43 @@ void GameWorld::RecalcTerritory(const noBaseBuilding& building, TerritoryChangeR
...
-        for(const MapPoint neighbourPt : GetNeighbours(curMapPt))
-        {
-            if(ptsHandled.insert(neighbourPt).second)
-                DestroyPlayerRests(neighbourPt, owner, &building);
-        }
+        DestroyPlayerRests(curMapPt, owner, &building);
```
Flamefire, line 424:
```suggestion
        DestroyPlayerRests(curMapPt, GetNode(curMapPt).owner, &building);
```

libs/s25main/world/GameWorld.cpp
```diff
@@ -413,43 +413,43 @@ void GameWorld::RecalcTerritory(const noBaseBuilding& building, TerritoryChangeR
...
         // (objects here are already destroyed and if the road ended there it would have been as well)
         if(GetNode(curMapPt).obj)
             continue;
         Direction dir;
         noFlag* flag = GetRoadFlag(curMapPt, dir);
-        if(!flag || flag->GetPlayer() + 1 == GetNode(curMapPt).owner)
+
+        if(!flag)
             continue;
+        if(flag->GetPlayer() + 1 == GetNode(curMapPt).owner && !IsBorderNode(curMapPt, GetNode(curMapPt).owner))
```
Flamefire, line 442: Double use of potentially expensive `GetNode`, store:
```suggestion
        const uint8_t owner = GetNode(curMapPt).owner;
        if(flag->GetPlayer() + 1 == owner && !IsBorderNode(curMapPt, owner))
```

libs/s25main/world/MapBase.cpp
```diff
@@ -87,6 +88,22 @@ MapPoint MapBase::GetNeighbour2(const MapPoint pt, unsigned dir) const
...
+{
+    std::set<MapPoint, MapPointLess> ptsToReturn;
+
+    for(const MapPoint& curMapPt : points)
+    {
+        ptsToReturn.insert(curMapPt);
+        for(const MapPoint neighbourPt : GetNeighbours(curMapPt))
+        {
+            ptsToReturn.insert(neighbourPt);
+        }
```
Flamefire, line 101: You can (optionally) make it more succinct:
```suggestion
        const auto neighbours = GetNeighbours(curMapPt);
        ptsToReturn.insert(neighbours.begin(), neighbours.end());
```

Or leave the braces:
```suggestion
        for(const MapPoint neighbourPt : GetNeighbours(curMapPt))
            ptsToReturn.insert(neighbourPt);
```

libs/s25main/world/MapBase.h
```diff
@@ -60,6 +60,10 @@ class MapBase
     MapCoord GetXA(MapPoint pt, Direction dir) const;
     // Gets all neighbors (in all directions) for given position
     helpers::EnumArray<MapPoint, Direction> GetNeighbours(MapPoint pt) const;
+    // Gets union of all neighbors including original set
+    // For one point it returns itself and all neighbours
+    // Returned set has no duplicities
```
Flamefire, line 65: Maybe:
```suggestion
    // Gets union of given points and all their neighbors without duplicates
    // For a single point it returns itself and all neighbours
```

Flamefire (changes requested):
Very good actually, thanks!   
For completeness maybe a test for the new function (at https://github.com/Return-To-The-Roots/s25client/blob/master/tests/s25Main/simple/testMapBase.cpp) would be good. I guess just 2 hard-coded, neighbouring points, first each one separately then their union without the duplicates, again with a hard-coded list of the expected result, should be easy enough and show what is done.

It also needs recreating the test replays (the test is only run when built in release mode) because the point order is now different. See https://github.com/Return-To-The-Roots/s25client/blob/fd16215cfb8b6b3cb35fb002578358d942600757/tests/s25Main/autoplay/main.cpp#L105-L123

Basically: Start the game normally with the described configuration, fast forward/skip to 200/300k GF and then copy the created replays into the test folder: https://github.com/Return-To-The-Roots/s25client/tree/fd16215cfb8b6b3cb35fb002578358d942600757/tests/testData

Xellzul:
@Flamefire Everything except the test should be addressed.

Flamefire:
> @Flamefire Everything except the test should be addressed.

I see perfect thanks! Are you going to fix the test too or do you need help with that?

Xellzul:
> > @Flamefire Everything except the test should be addressed.
> 
> 
> 
> I see perfect thanks! Are you going to fix the test too or do you need help with that?

i will try to add the test on my own later today or tomorow

tests/s25Main/simple/testMapBase.cpp
```diff
@@ -14,6 +14,43 @@
...
+
+    const auto resultPoints = world.GetAllNeighboursUnion(testPoints);
+
+    BOOST_TEST_REQUIRE(resultPoints.size() == expectedResultPoints.size());
+
+    for(auto pointIt = expectedResultPoints.begin(); pointIt != expectedResultPoints.end(); ++pointIt)
+    {
+        BOOST_REQUIRE_MESSAGE(std::find(resultPoints.begin(), resultPoints.end(), *pointIt) != resultPoints.end(),
+                              "Failed searching for: " << *pointIt);
+    }
```
Flamefire, line 51: Boost.Test already does that:
```suggestion
    BOOST_TEST(resultPoints == expectedResultPoints, boost::test_tools::per_element());
```

tests/s25Main/simple/testMapBase.cpp
```diff
@@ -14,6 +14,43 @@
...
+    // No points -> empty response
+    std::vector<MapPoint> testPointsEmpty{};
+
+    auto replyEmpty = world.GetAllNeighboursUnion(testPointsEmpty);
+    BOOST_TEST_REQUIRE(replyEmpty.empty());
+
+    // Two compontents of 1 and 2 vertices
+    std::vector<MapPoint> testPoints{MapPoint(1, 1), MapPoint(10, 10), MapPoint(10, 11)};
+    // ((center + 6 connecting hexes) * 3 points in test data) - 4 common vectors for the bigger component = 17 points
+    std::vector<MapPoint> expectedResultPoints{
```
Flamefire, line 30:
```suggestion
    const std::vector<MapPoint> expectedResultPoints{
```

tests/s25Main/simple/testMapBase.cpp
```diff
@@ -14,6 +14,43 @@
...
+    MapBase world;
+
+    // No points -> empty response
+    std::vector<MapPoint> testPointsEmpty{};
+
+    auto replyEmpty = world.GetAllNeighboursUnion(testPointsEmpty);
+    BOOST_TEST_REQUIRE(replyEmpty.empty());
+
+    // Two compontents of 1 and 2 vertices
+    std::vector<MapPoint> testPoints{MapPoint(1, 1), MapPoint(10, 10), MapPoint(10, 11)};
```
Flamefire, line 28:
```suggestion
    const std::vector<MapPoint> testPoints{MapPoint(1, 1), MapPoint(10, 10), MapPoint(10, 11)};
```

Flamefire (changes requested):
Well done test, only a few minor improvements!

Check them and note that you can apply each suggestion from this page on Github directly without having to apply them locally.

tests/s25Main/simple/testMapBase.cpp
```diff
@@ -14,6 +14,43 @@
...
+{
+    MapBase world;
+
+    // No points -> empty response
+    std::vector<MapPoint> testPointsEmpty{};
+
+    auto replyEmpty = world.GetAllNeighboursUnion(testPointsEmpty);
+    BOOST_TEST_REQUIRE(replyEmpty.empty());
+
+    // Two compontents of 1 and 2 vertices
```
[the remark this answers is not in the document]
Xellzul: Well, best explained here:
https://en.wikipedia.org/wiki/Component_(graph_theory)
https://en.wikipedia.org/wiki/Vertex_(graph_theory)

vertex is basically point
components are all connected points - two road networks in S2 are two components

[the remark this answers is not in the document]
Flamefire: I meant what you refer with "vertex" here. Because the function takes as inputs a list of points. For the purpose of this function the points are not connected, i.e. there is no vertex between (10,10) and (10,11) here, they are just neighbours. Of course you could refer to the points as nodes in a graph which basically is the map and vertices would be the relationships, i.e. what makes points neighbours. So there is a vertex between (10,10) and (10,11)  but none of either to (1,1) (unless the map is small)

But I wouldn't get to technical in the comment here. So maybe:
```suggestion
    // 3 points where the last 2 share neighbours and are neighbours to each other
```

Because that is the important thing for the tested function: We get the neighbours but remove and duplicates and additionally also the points itself, but also avoid duplicates if one is the neighbour of another given point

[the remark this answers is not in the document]
Xellzul: Yea, your comment works better.

Flamefire (approved):

Xellzul:
@Flamefire 
So I did some more testing and found out one new bug.
Border is considered as player territory.. 
Should be fixed in last commit

Screenshot 2024-04-10 151844

tests/s25Main/integration/testGamePlayer.cpp
```diff
@@ -65,7 +65,7 @@ BOOST_FIXTURE_TEST_CASE(ProductivityStats, WorldFixtureEmpty1P)
     MapPoint curPos(0, 0);
     for(const auto bldType : helpers::EnumRange<BuildingType>{})
     {
-        if(!BuildingProperties::IsValid(bldType))
+        if(!BuildingProperties::IsValid(bldType) || bldType == BuildingType::HarborBuilding)
```
Xellzul, line 68: Not sure what exactly is the failing test 'ProductivityStats' testing.
However it probably fails, because last created building Harbour is recalculating border and destroying buildings. 

Just skipping the creation of harbour will fix the test

Xellzul:
`        
if(!BuildingProperties::IsValid(bldType) || bldType == BuildingType::HarborBuilding)
    continue;
`

tests/s25Main/integration/testGamePlayer.cpp
Flamefire: That would make sense, yes. Maybe we need a comment why this is OK though:
```suggestion
        if(!BuildingProperties::IsValid(bldType))
            continue;
        if(bldType == BuildingType::HarborBuilding)
        {
            // Creating a harbor would recalculate the territory and destroy some buildings already added in the calculation, but we don't have productivity for it so no need to test
            BOOST_TEST_REQUIRE(!helpers::contains(iwBuildingProductivities::icons, bldType))
            continue
        }
```

However I don't feel so good about excluding (real) buildings here as we might miss bugs. Maybe a better solution would be to convert `helpers::EnumRange<BuildingType>` into a vector and sort such that all "usual" buildings come last, i.e. `BuildingProperties::IsUsual`

This way territory changes only happen at the start

libs/s25main/world/GameWorld.cpp
```diff
@@ -413,43 +413,43 @@ void GameWorld::RecalcTerritory(const noBaseBuilding& building, TerritoryChangeR
...
             continue;
         Direction dir;
         noFlag* flag = GetRoadFlag(curMapPt, dir);
-        if(!flag || flag->GetPlayer() + 1 == GetNode(curMapPt).owner)
+
+        if(!flag)
+            continue;
+
+        const uint8_t owner = GetNode(curMapPt).owner;
+        if(flag->GetPlayer() + 1 == owner && !IsBorderNode(curMapPt, owner))
```
Flamefire, line 442: Even better (and faster):
```suggestion
        if(flag->GetPlayer() + 1 == owner && IsPlayerTerritory(curMapPt, owner))
```

`IsBorderNode = GetNode(pt).owner == owner && !IsPlayerTerritory(pt, owner))`   
--> `!IsBorderNode = GetNode(pt).owner != owner || IsPlayerTerritory(pt, owner))`   and the first part can never be true here

tests/s25Main/integration/testGamePlayer.cpp
```diff
@@ -62,8 +62,22 @@ BOOST_FIXTURE_TEST_CASE(ProductivityStats, WorldFixtureEmpty1P)
     BOOST_TEST(buildingRegister.CalcProductivities() == expectedProductivity, per_element());
     BOOST_TEST(buildingRegister.CalcAverageProductivity() == 0u);
 
+    std::vector<BuildingType> buildingTypes;
+    const auto buildingTypesEnum = helpers::EnumRange<BuildingType>{};
+
+    for(auto it = buildingTypesEnum.begin(); it != buildingTypesEnum.end(); ++it)
+    {
+        buildingTypes.push_back(*it);
+    }
```
Flamefire, line 71: We use C++11/14 so you never need this `for(it=..., ++it)` but can always use the range-based-for: `for(auto t: buildingTypesEnum)`

However here you can just use the range constructor:
```suggestion
    const auto buildingTypesEnum = helpers::EnumRange<BuildingType>{};
    std::vector<BuildingType> buildingTypes(buildingTypesEnum.begin(), buildingTypesEnum.end());
```

tests/s25Main/integration/testGamePlayer.cpp
```diff
@@ -62,8 +62,22 @@ BOOST_FIXTURE_TEST_CASE(ProductivityStats, WorldFixtureEmpty1P)
...
+
+    // Sort buildings so military buildings are created first 
+    // and no buildings are destroyed when borders are recalculated
+    std::sort(buildingTypes.begin(), buildingTypes.end(), [](BuildingType a, BuildingType b) {
+        return !BuildingProperties::IsUsual(a) > !BuildingProperties::IsUsual(b);
+    });
```
Flamefire, line 77: Neat but a bit hard to read. I guess `!BuildingProperties::IsUsual(a) && BuildingProperties::IsUsual(b)` would also work?

But even better: You want all "!IsUsual" buildings first, correct? I.e. partition the vector:
```suggestion
    std::partition(buildingTypes.begin(), buildingTypes.end(), [](BuildingType bld) { return !BuildingProperties::IsUsual(bld); });
```

Flamefire (changes requested):

tests/s25Main/integration/testGamePlayer.cpp
Xellzul: This works well

tests/s25Main/integration/testGamePlayer.cpp
Xellzul: I looked into this, and didnt manage to make it work. 

believe you have to implement https://en.cppreference.com/w/cpp/iterator/iterator_tags 

`E0289	no instance of constructor "std::vector<_Ty, _Alloc>::vector [with _Ty=BuildingType, _Alloc=std::allocator<BuildingType>]" matches the argument list	Test_integration	\s25client\tests\s25Main\integration\testGamePlayer.cpp	73		
`

Flamefire (approved):

Flamefire:
I rebased your branch (fixing the conflict and squashing commits) and applied the remaining suggestion (see the last 3 commits)

Enabling automerge as CI should pass, if not I'll fix it ;-) Thanks a lot for hunting this down and fixing it!

PR #1658, opened 2024-04-06
Don't repeat all songs from playlist once

Flamefire:
When playing a song from the playlist `Play(1)` is called which means "1 repeat" not "play once".
This leads to #1480 which was reported on Discord to still happen. Fix is to remove the argument to default to *playing* only once.

Also add a test although it is actually a logic confusion.


Spikeone (approved):
👍 GTA Vice City war das letzte GTA das einen eigenen Player hatte, und wir haben bis heute einen ;)

PR #1659, opened 2024-04-07
Fix validity criterium for fish in AIResourceMap

Flamefire:
The resource map doesn't store where fish is but how good a fishery would be at this position.
Hence we need to check for buildable, not for water. See #933
Also slightly refactor the code and improve comments.

### DIFF
--- a/libs/s25main/ai/aijh/AIResourceMap.cpp
+++ b/libs/s25main/ai/aijh/AIResourceMap.cpp
@@ -43,23 +43,24 @@ void AIResourceMap::init()
     map.Resize(mapSize);
     // Calculate value for each point.
     // This is quite expensive so do just for the diminishable resources to sort out ones where there will never be
-    // anything, which allows an optimization when calculating the value which must always be done on demand
+    // anything, which allows to skip calculating the value when updating the resource map
     if(isDiminishableResource)
     {
+        ETerrain requiredTerrain;
+        if(res == AIResource::Fish || res == AIResource::Stones)
+            requiredTerrain = ETerrain::Buildable;
+        else
+        {
+            RTTR_Assert(helpers::contains(
+              std::vector<AIResource>{AIResource::Gold, AIResource::Ironore, AIResource::Coal, AIResource::Granite},
+              res));
+            requiredTerrain = ETerrain::Mineable;
+        }
         RTTR_FOREACH_PT(MapPoint, mapSize)
         {
-            bool isValid = true;
-            if(res == AIResource::Fish)
-            {
-                isValid =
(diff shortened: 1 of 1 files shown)


libs/s25main/ai/aijh/AIResourceMap.cpp
```diff
@@ -43,23 +43,24 @@ void AIResourceMap::init()
     map.Resize(mapSize);
     // Calculate value for each point.
     // This is quite expensive so do just for the diminishable resources to sort out ones where there will never be
-    // anything, which allows an optimization when calculating the value which must always be done on demand
+    // anything, which allows to skip calculating the value when updating the resource map
     if(isDiminishableResource)
     {
+        ETerrain requiredTerrain;
+        if(res == AIResource::Fish || res == AIResource::Stones)
+            requiredTerrain = ETerrain::Buildable;
```
Spikeone, line 51: Just asking if we have some styleguide on this - I'd expect the use of ```{ }``` even for 1-line if statements if the following ```else / else if``` uses them.

Spikeone (approved):

libs/s25main/ai/aijh/AIResourceMap.cpp
Flamefire: For short single-line statements I prefer to not use braces. We have clang-format with CI checks to enforce the indentation, hence logic mistakes due to that are basically impossible. So I prefer to save the vertical space to make it easier to read. Compare:
```
if(res == AIResource::Fish || res == AIResource::Stones)
    requiredTerrain = ETerrain::Buildable;
else
    requiredTerrain = ETerrain::Mineable;
```
to
```
if(res == AIResource::Fish || res == AIResource::Stones)
{
    requiredTerrain = ETerrain::Buildable;
} else
{
    requiredTerrain = ETerrain::Mineable;
}
```

Note that a common mistake like this is impossible to happen due to CI enforced clang-format:
```
if(res == AIResource::Fish || res == AIResource::Stones)
    requiredTerrain = ETerrain::Buildable;
    costs = 42;
doIt(requiredTerrain, costs);
```

What is gained by those braces? I even feel they are noise distracting from the logic.   
That is basically the point: The cost is additional typing and additional vertical space used (hence less code fits on the screen & more scrolling) and I don't see any gain/benefit. So to me the cost (as small as it may be perceived) isn't worth it.

Flow86: I prefer
```
if(res == AIResource::Fish || res == AIResource::Stones) {
    requiredTerrain = ETerrain::Buildable;
} else {
    requiredTerrain = ETerrain::Mineable;
}
```

though, but its okay as it is

PR #1662, opened 2024-05-05
Add map selection to campaign lua

ottml:
Add selection map screen to campaign lua and use it for world campaign.
Add documention.

### DIFF
--- a/libs/libGamedata/gameData/CampaignDescription.cpp
+++ b/libs/libGamedata/gameData/CampaignDescription.cpp
@@ -33,8 +33,13 @@ CampaignDescription::CampaignDescription(const kaguya::LuaRef& table)
     lua::validatePath(mapFolder);
     lua::validatePath(luaFolder);
     mapNames = luaData.getOrDefault("maps", std::vector<std::string>());
-
+    selectionMapData = luaData.getOptional<SelectionMapInputData>("selectionMap");
     luaData.checkUnused();
+}
+
+const std::optional<SelectionMapInputData>& CampaignDescription::getSelectionMapData() const
+{
+    return selectionMapData;
 }
 
 boost::filesystem::path CampaignDescription::getLuaFilePath(const size_t idx) const
--- a/tests/s25Main/campaign/testCampaignLuaFile.cpp
+++ b/tests/s25Main/campaign/testCampaignLuaFile.cpp
@@ -2,8 +2,10 @@
 //
 // SPDX-License-Identifier: GPL-2.0-or-later
 
+#include "PointOutput.h"
 #include "lua/CampaignDataLoader.h"
 #include "gameData/CampaignDescription.h"
+#include "gameData/SelectionMapInputData.h"
 #include "rttr/test/LocaleResetter.hpp"
 #include "rttr/test/LogAccessor.hpp"
 #include "rttr/test/TmpFolder.hpp"
@@ -58,6 +60,36 @@ BOOST_AUTO_TEST_CASE(ScriptVersion)
         CampaignDataLoader loader(desc, tmp.get());
         rttr::test::LogAccessor logAcc;
         BOOST_TEST_REQUIRE(!loader.Load());
+        BOOST_TEST_REQUIRE(loader.CheckScriptVersion());
+        logAcc.clearLog();
+    }
+
+    // Backwards compatibility: version 2 can load version 1
+    {
+        rttr::test::TmpFolder tmp;
+        {
+            bnw::ofstream file(tmp.get() / "campaign.lua");
+
+            file << "campaign ={\
+                version = \"1\",\
+                author = \"Max Meier\",\
+                name = \"Meine Kampagne\",\
+                shortDescription = \"Sehr kurze Beschreibung\",\
+                longDescription = \"Das ist die lange Beschreibung\",\
+                image = \"<RTTR_GAME>/GFX/PICS/WORLD.LBM\",\
+                maxHumanPlayers = 1,\
+                difficulty = \"easy\",\
+                mapFolder = \"<RTTR_GAME>/DATA/MAPS\",\
+                luaFolder = \"<RTTR_GAME>/CAMPAIGNS/ROMAN\",\
+                maps = { \"dessert0.WLD\", \"dessert1.WLD\", \"dessert2.WLD\"}\
+            }\
+            function getRequiredLuaVersion()\n return 1\n end";
+        }
+
+        CampaignDescription desc;
+        CampaignDataLoader loader(desc, tmp.get());
+        rttr::test::LogAccessor logAcc;
+        BOOST_TEST_REQUIRE(loader.Load());
         BOOST_TEST_REQUIRE(loader.CheckScriptVersion());
         logAcc.clearLog();
     }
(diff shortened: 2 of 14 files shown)


tests/s25Main/campaign/testCampaignLuaFile.cpp
```diff
@@ -89,7 +90,7 @@ BOOST_AUTO_TEST_CASE(LoadCampaignDescriptionWithoutTranslation)
         bnw::ofstream file(tmp.get() / "campaign.lua");
 
         file << "campaign ={\
-            version = \"1\",\
+            version = \"2\",\
```
Flamefire, line 93: Similar: Why? Where is this used and why do existing campaigns need to be upgraded if they don't use anything new and the new feature is optional?

tests/s25Main/campaign/testCampaignLuaFile.cpp
```diff
@@ -272,6 +273,96 @@ BOOST_AUTO_TEST_CASE(CampaignDescriptionLoadWithTranslation)
...
+    BOOST_TEST(selectionMap->map.filePath == "<RTTR_GAME>/GFX/PICS/WORLD.LBM");
+    BOOST_TEST(selectionMap->map.index == 0u);
+    BOOST_TEST(selectionMap->missionMapMask.filePath == "<RTTR_GAME>/GFX/PICS/WORLDMSK.LBM");
+    BOOST_TEST(selectionMap->missionMapMask.index == 0u);
+    BOOST_TEST(selectionMap->marker.filePath == "<RTTR_GAME>/DATA/IO/IO.DAT");
+    BOOST_TEST(selectionMap->marker.index == 231u);
+    BOOST_TEST(selectionMap->conquered.filePath == "<RTTR_GAME>/DATA/IO/IO.DAT");
+    BOOST_TEST(selectionMap->conquered.index == 232u);
+    BOOST_TEST(selectionMap->mapOffsetInBackground.x == 64);
+    BOOST_TEST(selectionMap->mapOffsetInBackground.y == 70);
```
Flamefire, line 354: Replace those (here and below) by single checks:

```suggestion
    BOOST_TEST(selectionMap->mapOffsetInBackground == Position(64, 70));
```

You might need to include `PointOutput.hpp`

Flamefire (changes requested):
I like how simple this is :)

Just make sure the changes are backwards compatible: Existing campaigns should continue to work without changes because the new feature is backwards compatible as it only introduces an optional new field. Hence the existing tests and campaigns shouldn't need changes.

tests/s25Main/campaign/testCampaignLuaFile.cpp
ottml: Yes i mixed up the lua version and the campaign version

Flamefire:
> @Flamefire Could it be that you changed the hash? The ci is failing with this error:

I did some work in one repo and didn't realize it might cause some fallout. So yes this is fixed now.

During review I was a bit confused about the `getRequiredLuaVersion` vs `campaign.version`. It seems we require the exact version match of the former which means old lua files won't work. The latter seems to be unused.

IIRC we increase `getRequiredLuaVersion` only when the change is a breaking change. This one isn't so I think we can keep the current version. Or am I missing anything?

libs/libGamedata/gameData/CampaignDescription.cpp
```diff
@@ -34,7 +34,16 @@ CampaignDescription::CampaignDescription(const kaguya::LuaRef& table)
     lua::validatePath(luaFolder);
     mapNames = luaData.getOrDefault("maps", std::vector<std::string>());
 
-    luaData.checkUnused();
+    kaguya::LuaRef entry = table["selectionMap"];
+    if(entry.type() == LUA_TTABLE)
+    {
+        selectionMapData = std::make_optional<SelectionMapInputData>(entry);
+    }
```
Flamefire, line 41: Why the removal of `checkUnused`? What about 
```suggestion
    selectionMapData = luaData.getOrDefault("selectionMap", std::optional<SelectionMapInputData>());
    luaData.checkUnused();
```

If that doesn't work out of the box we need a `getOptional<SelectionMapInputData>` for `CheckedLuaTable` based on `getOrDefault`

libs/libGamedata/gameData/CampaignDescription.cpp
```diff
@@ -34,7 +34,16 @@ CampaignDescription::CampaignDescription(const kaguya::LuaRef& table)
...
+}
+
+std::optional<SelectionMapInputData> CampaignDescription::getSelectionMapData() const
```
Flamefire, line 44: No need for copies as sometimes you just want to check IFF there is a value.
```suggestion
const std::optional<SelectionMapInputData>& CampaignDescription::getSelectionMapData() const
```

tests/s25Main/campaign/testCampaignLuaFile.cpp
```diff
@@ -62,6 +64,22 @@ BOOST_AUTO_TEST_CASE(ScriptVersion)
...
+        rttr::test::TmpFolder tmp;
+        {
+            bnw::ofstream file(tmp.get() / "campaign.lua");
+            file << boost::format("function getRequiredLuaVersion()\n return %1%\n end") % 1;
+        }
+
+        CampaignDescription desc;
+        CampaignDataLoader loader(desc, tmp.get());
+        rttr::test::LogAccessor logAcc;
+        BOOST_TEST_REQUIRE(!loader.Load());
```
Flamefire, line 78: > version 2 can load version 1

But asserting it fails to load?

tests/s25Main/campaign/testCampaignLuaFile.cpp
```diff
@@ -102,15 +103,15 @@ BOOST_AUTO_TEST_CASE(LoadCampaignDescriptionWithoutTranslation)
             maps = { \"dessert0.WLD\", \"dessert1.WLD\", \"dessert2.WLD\"}\
         }";
 
-        file << "function getRequiredLuaVersion() return 1 end";
+        file << "function getRequiredLuaVersion() return 2 end";
```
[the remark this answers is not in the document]
Flamefire: With the new check those changes can go, can't they?

tests/s25Main/campaign/testCampaignLuaFile.cpp
Flamefire: You seem to forgot this ;-)

tests/s25Main/campaign/testCampaignLuaFile.cpp
```diff
@@ -272,6 +290,93 @@ BOOST_AUTO_TEST_CASE(CampaignDescriptionLoadWithTranslation)
...
+    BOOST_TEST(selectionMap->missionMapMask.filePath == "<RTTR_GAME>/GFX/PICS/WORLDMSK.LBM");
+    BOOST_TEST(selectionMap->missionMapMask.index == 0u);
+    BOOST_TEST(selectionMap->marker.filePath == "<RTTR_GAME>/DATA/IO/IO.DAT");
+    BOOST_TEST(selectionMap->marker.index == 231u);
+    BOOST_TEST(selectionMap->conquered.filePath == "<RTTR_GAME>/DATA/IO/IO.DAT");
+    BOOST_TEST(selectionMap->conquered.index == 232u);
+    BOOST_TEST(selectionMap->mapOffsetInBackground.x == 64);
+    BOOST_TEST(selectionMap->mapOffsetInBackground.y == 70);
+    BOOST_TEST(selectionMap->disabledColor == 0x70000000u);
+    BOOST_TEST(selectionMap->missionSelectionInfos.size() == static_cast<size_t>(3));
```
Flamefire, line 373: I guess you were avoiding the signed-unsigned-mismatch warning? More readable:

```suggestion
    BOOST_TEST(selectionMap->missionSelectionInfos.size() == 3u);
```

Flamefire (changes requested):
> The former implementation for loading the campaingn description checked that all keys are read (luaData.checkUnused();) otherwise it failed. So the new format cannot be loaded by already existing builds.

Oh yes, already existing builds...

> We have to increase the version for the world campaign. And change the version check to <= current version. So the new build can also load older versions.

That logic isn't flexible enough either. If we do a breaking change we can not use `<=` although we can worry about that if we do that at one point. Then we can do something like '<major> * 100 + <minor>` and accept anything with the same major version and at least the same minor version.  

Again: For later. We can use the current approach for now

Some remarks left

libs/libGamedata/gameData/CampaignDescription.cpp
ottml: added getoptional

tests/s25Main/campaign/testCampaignLuaFile.cpp
ottml: Extended test

Flamefire:
Thanks a lot!

Issue #1660, opened 2024-04-22
[LUA] Some settings not correctly working in multiplayer


Spikeone:
Using this setting:
```
function getAllowedChanges()
    return  {
        ["general"]     = true,
        ["addonsAll"]   = false,
        ["addonsSome"]  = true,
        ["ownNation"]   = true,
        ["ownColor"]    = true,
        ["ownTeam"]     = false,
        ["aiNation"]    = false,
        ["aiColor"]     = true,
        ["aiTeam"]      = false,
        ["swapping"]    = true,
        ["playerState"] = false,
        ["aiNation"]    = false
    }
end
``` 
does work fine for the host player in multiplayer - sadly the ```ownTeam``` seems to not affect any other player than the host. Same goes for the
```
function isMapPreviewEnabled()
    return false;
end
```
which works fine for the host - but no client.

![grafik](https://github.com/Return-To-The-Roots/s25client/assets/1276342/c5366ffd-517c-4209-803f-b98482672c91)

PR #1663, opened 2024-05-16
Fix setting disabling by LUA for non-Host players 

Flamefire:
The lua script was only loaded for the host as many function are intended for the host, especially the lobbyController.
Load for all players and guard relevant parts for the host only.

Fixes https://github.com/Return-To-The-Roots/s25client/issues/1660


Flamefire:
@Flow86 Any idea what is wrong with Jenkins? It failed but the log is so large it doesn't even show properly

Flamefire:
@Flow86 Can you merge this? It needs a review approval which I can't do if I'm the author

PR #1664, opened 2024-05-17
Drop support for 32bit Windows

Flamefire:
It is unlikely anyone is still required to use that in 2024

And as it is now too much trouble to maintain (32bit builds on Appveyor started to fail)
just drop it, at least officially.
It might still work but we don't test it anymore.

See https://github.com/Return-To-The-Roots/s25client/pull/1663 where I tried to fix the 32bit appveyor build which has been failing for a long time now. I guess we could rebuild or remove the static MSVC libcurl library (CMake started to choose it over the shared one "recently") but I'd say it's not worth the time

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,11 +1,10 @@
-// Copyright (C) 2005 - 2021 Settlers Freaks <sf-team at siedler25.org>
+// Copyright (C) 2005 - 2024 Settlers Freaks <sf-team at siedler25.org>
 //
 // SPDX-License-Identifier: GPL-2.0-or-later
 
 dockerRegistry = "registry.mytrap.de/"
 dockerCredentials = "gitlab-Flow86"
 dockerImages = [
-    "windows.i686"    : "rttr/cross-compiler/mingw/mingw-w64-docker:master",
     "windows.x86_64"  : "rttr/cross-compiler/mingw/mingw-w64-docker:master",
     "linux.x86_64"    : "rttr/cross-compiler/linux/linux-amd64-docker:master",
     "apple.x86_64"    : "rttr/cross-compiler/apple/apple-docker:master"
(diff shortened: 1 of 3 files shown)


Spikeone:
Well, some people would like to use RttR on some old WinXP machines - but we are already missing some dlls on those systems I think so at this time, I'd say it's okay to drop it. But we should update the supported systems (if we havent).

Flamefire:
@Flow86 This is basically blocking all other PRs: It is failing CI so I can't accept & merge any PR or get mine into a state that you can merge it (you need to approve mine so in the end you need to merge them)

Flow86:
why not simply removing it from the appveyor? there are still people around using the 32bit variant though.

Flamefire:
> why not simply removing it from the appveyor? there are still people around using the 32bit variant though.

If we remove it (i.e. don't test it anymore), how can we be sure it works? Yes, there is Jenkins but the logs aren't readily available. Do you have any logs if people really do use/download the 32bit variant? Sounds odd in 2024 and I think it is mostly just by confusion which one works so people might continue using 32bit RttR out of habbit not of need.

Flow86:
I checked the update downloads:
In the last 30 days:
* windows: 2450
  * i386: 95 (~3%)
  * x86_64: 2355
* linux: 883
* darwin: 562

so still 3% of windows users using the nightly build use 32bit

Spikeone:
> so still 3% of windows users using the nightly build use 32bit, its not much, but still we need to perhaps at least first mention "hey we will remove the support", not directly doing it.

Can we make the 32-Bit version a bit less obvious to download - because I agree, that most people may not understand what they are actually downloading and are just happy, that it works fine. That way we may see a more significant drop in downloads. Sadly that would mean waiting some more time.

Flamefire:
Ok I found an alternative: just remove the static curl which wasn't used until recently: #1667

PR #1667, opened 2024-05-24
Fix build on 32bit Windows and reformat CML

Flamefire:
CMake recently finds the static libcurl (in s25update) which needs some defines on use.

Also replace tabs in CMake files and update the other submodules

Requires
- [x] https://github.com/Return-To-The-Roots/dev-tools/pull/6
- [x] https://github.com/Return-To-The-Roots/dev-tools/pull/5


Flow86 (dismissed):

Flamefire:
ok simply removing our prebuilt 32bit static curl works. Shouldn't affect anything as CMake only recently startet to prefer it over the import lib

Flow86:
weird why they are there at all,

Flamefire:
Maybe because it is easier to debug without also having/copying all DLLs to the right folder

Can you merge the other PR mentioned above (whitespace) too? Just wanted to do this once over all modules such that contributors (and autodetection from editors) don't get confused

Flow86 (approved):

Issue #1669, opened 2024-05-26
New Addon: Wine Buildings/Wares/Jobs


aztimh:
Finished working on graphics for a new addon that adds new wine buildings/wares/jobs as an optional addon.

https://github.com/aztimh/RTTR-Wine-Addon

I've included all the the documents for how it's all supposed to function and calculated the XY nation offsets for things like work windows, chimney smoke, stop production X signs.

Also made txt documents that give a quick reference for the graphics contained within the LST files.

I'm not a programmer (yet), so I am sharing this so someone can program it as an addon.

ottml:
<img width="1920" alt="wine_addon" src="https://github.com/Return-To-The-Roots/s25client/assets/33401986/cb8e708e-298b-4ce1-a87d-3a70500dfdca">

aztimh:
Oh wow! That's so awesome! I'm actually in the process of fixing some of the larger buildings which were a bit wide when built next to other ones so it will slightly alter some of the offsets. But I hope to have the last one finished tonight!

I'm not sure how to compile it actually, I've just been working on the pixel art and ideas mostly.

Thanks so much I'd love to test it out!

Spikeone:
@ottml just asking: If the addon is disabled, are all the icons (inventory, settings screen) hidden as well? If so, could you also implement it for the charburner as well, if not, could you implement that since using the default s2 settings should not show these icons so player can get an original feeling :)

But great work adding this so fast :) Thanks for your contribution(s)!

ottml:
> @ottml just asking: If the addon is disabled, are all the icons (inventory, settings screen) hidden as well? If so, could you also implement it for the charburner as well, if not, could you implement that since using the default s2 settings should not show these icons so player can get an original feeling :)
> 
> But great work adding this so fast :) Thanks for your contribution(s)!

No at the moment they are always shown. But when changing this for the wine addon this should not be a problem to do for the charburner too. The only problem is gui's use mainly hardcoded sizes at the moment for in game windows. :)

Another  question: Is there a possibility to trigger a build over the ci from a branch/Pullrequest to get a install package that another user can download and use for testing? @Flamefire

ottml:
[...]

PR #1672, opened 2024-06-01
Consistently use GetWareTex  and GetWareTexStack instead of GetMapTexture

ottml:
Use GetWareTex and GetWareTexStack texture interface instead of GetMapTexture.

- Adding new textures for additional wares is much easier because we have only one place to add a lookup for the new texture. Otherwise we have to add the code at each calling place. This is needed for the new wine addon (#1669) to add the new goods wine and grapes.

### DIFF
--- a/libs/s25main/buildings/noBuildingSite.cpp
+++ b/libs/s25main/buildings/noBuildingSite.cpp
@@ -192,12 +192,10 @@ void noBuildingSite::Draw(DrawPoint drawPt)
         // Bretter
         DrawPoint doorPos = drawPt + DrawPoint(GetDoorPointX(), GetDoorPointY());
         for(unsigned char i = 0; i < boards; ++i)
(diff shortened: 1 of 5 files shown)


Flamefire:
> Use abstract texture interface in all places. Needed to implement new addons more easy

Can you explain a bit more? Because although the changes look good and correct they don't match neither the title nor the description of the commit and PR:   
All of the changes are like:
```
- LOADER.GetMapTexture(WARES_TEX_MAP_OFFSET + rttr::enum_cast(good))
+ LOADER.GetWareTex(good);
```

In both cases the "abstract texture interface", i.e. `ITexture*` is used. Hence I would have expected a title like "Consistently use GetWare(Stack)Tex instead of GetMapTexture`.

The description (at least of the PR) should have a sentence on WHY. Or e.g. "Makes implementing addons such as/doing X easier due to Y"

Flamefire (approved):

PR #1673, opened 2024-06-03
Wine addon

### LINKED ISSUE #1669: New Addon: Wine Buildings/Wares/Jobs

ottml:
Fixes #1669


libs/s25main/figures/nofTempleServant.cpp
```diff
@@ -0,0 +1,85 @@
...
+#include "SoundManager.h"
+#include "WineLoader.h"
+#include "buildings/nobUsual.h"
+#include "network/GameClient.h"
+#include "ogl/glArchivItem_Bitmap_Player.h"
+#include "random/Random.h"
+#include "world/GameWorld.h"
+
+namespace {
+int producedGood = 0;
```
Flamefire, line 17: That has to be an instance variable doesn't it? In case you have multiple servants, which is very likely.

libs/s25main/gameData/GoodConsts.cpp
```diff
@@ -41,4 +41,6 @@ const helpers::EnumArray<std::string, GoodType> WARE_NAMES = {{
   /* 32 */ gettext_noop("Meat"),         // Fleisch
   /* 33 */ gettext_noop("Ham"),          // Schinken ( Schwein )
   /* 34 */ "",                           // Schild
+  /* 35 */ gettext_noop("Grapes"),       // Trauben
+  /* 36 */ gettext_noop("Wine"),         // Wein
```
Flamefire, line 45: We actually should get rid of all the comments that are simply translations to German
```suggestion
  /* 35 */ gettext_noop("Grapes"),
  /* 36 */ gettext_noop("Wine"),
```

libs/s25main/nodeObjs/noGrapefield.h
```diff
@@ -0,0 +1,60 @@
...
+
+#include "noCoordBase.h"
+
+class SerializedGameData;
+class GameEvent;
+
+class noGrapefield : public noCoordBase
+{
+private:
+    /// Typ of the grape field (2 different types)
```
Flamefire, line 15: Better something like:
```suggestion
    /// Type of the grape field (for drawing only, currently 0-1)
```

Maybe even use an enum or bool for this (

libs/s25main/WineLoader.h
```diff
@@ -0,0 +1,101 @@
...
+
+extern std::map<std::pair<Nation, BuildingType>, BuildingImages> buildings;
+using BobEntryRange = std::pair<unsigned, unsigned>;
+struct BobEntry
+{
+    BobEntryRange bobEntryRange;
+    std::string description;
+};
+
+enum BobTypes
```
Flamefire, line 47: Please only typed enums:
```suggestion
enum class BobTypes
```

libs/s25main/WineLoader.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+///////////////////////////////////////////////////////////////////////////////
+
+#include "WineLoader.h"
+#include "Loader.h"
+
+// Wine Addon animation and images
+
+constexpr unsigned NOT_AVAILABLE = 0;
+constexpr unsigned EMPTY = 0;
```
Flamefire, line 13: Why are they both the same?

libs/s25main/WineLoader.h
```diff
@@ -0,0 +1,101 @@
...
+bool isWineAddonBuildingType(BuildingType bld);
+bool isWineAddonGoodType(GoodType good);
+bool isWineAddonJobType(Job job);
+
+extern std::map<std::pair<Nation, BuildingType>, BuildingImages> buildings;
+using BobEntryRange = std::pair<unsigned, unsigned>;
+struct BobEntry
+{
+    BobEntryRange bobEntryRange;
+    std::string description;
```
Flamefire, line 44: Whats the use of this?

libs/s25main/WineLoader.cpp
```diff
@@ -0,0 +1,164 @@
...
+  {{Nation::Vikings, BuildingType::Winery}, {83, 84, 85, 86, 87, 88, 89, 90, 91}},
+  {{Nation::Vikings, BuildingType::Temple}, {92, 93, 94, 95, 96, 97, 98, 99, 100}},
+  {{Nation::Babylonians, BuildingType::Vineyard}, {101, 102, 103, 104, 105, 106, NOT_AVAILABLE, 107, 108}},
+  {{Nation::Babylonians, BuildingType::Winery}, {109, 110, 111, 112, 113, 114, NOT_AVAILABLE, 115, 116}},
+  {{Nation::Babylonians, BuildingType::Temple}, {117, 118, 119, 120, EMPTY, 122, NOT_AVAILABLE, EMPTY, 124}}};
+
+std::map<BobTypes, BobEntry> bobs = {
+  {BobTypes::WINEGROWER_DIGGING_ANIMATION, {{1, 15}, "Winegrower digging animation (setup grape field, loop)"}},
+  {BobTypes::WINEGROWER_PLANTING_ANIMATION,
+   {{17, 20}, "Winegrower planting animation (setup field after digging, loop)"}},
```
Flamefire, line 36: Why is there a gap (1-15, 17-20)?

libs/s25main/WineLoader.cpp
```diff
@@ -0,0 +1,164 @@
...
+#include "Loader.h"
+
+// Wine Addon animation and images
+
+constexpr unsigned NOT_AVAILABLE = 0;
+constexpr unsigned EMPTY = 0;
+
+namespace wineaddon {
+std::map<std::pair<Nation, BuildingType>, BuildingImages> buildings{
+  {{Nation::Africans, BuildingType::Vineyard}, {1, 2, 3, 4, 5, 6, NOT_AVAILABLE, 7, 8}},
```
Flamefire, line 17: Is it possible to keep the number of images constant, i.e. 9 images per building, 3 buildings per nation, 27 images per nation. That would make this redundant as a simple calculation can be used.

For LSTs there is the option to add an "empty" entry (in the file) which should work here.

Flamefire (commented):
Just took a quick look for ideas and potential issues and improvements

libs/s25main/WineLoader.cpp
aztimh: When numbering the images for the LST I tried to follow the same conventions as the charburner_bob LST which leaves a gap between different animations: carrying different wares, separate animation loops, etc. I made a text file that shows the number ranges of all the sprites and what animations they are for.

libs/s25main/WineLoader.cpp
aztimh: I can fix this. The only reason some of them have 9 instead of 8 is because some of the winterworld buildings need a separate shadow image for the building as a plant or sand missing means the shadow needs to be different.

libs/s25main/WineLoader.cpp
Flamefire: Not sure why those were there for the charburner. Might have been an artifact of the changes at that time.

libs/s25main/WineLoader.cpp
Flamefire: Sure I don't mind the empty ones. It is just about the bigish tables of pretty much only successive numbers in the code which could be a simple formula

aztimh: Would it be easier if I duplicate the shadow images that are unchanged for winterworld buildings so they can all be called for those buildings as the winterworld specific shadows are? Saving having to do an override for the few buildings that have a specific winter shadow, but rather all the winter shadows are called to the same multiple of 9?

aztimh: Otherwise I've just put an <empty> for buildings that don't have a winterworld shadow so each building is a group of 9 images.

ottml: Done

libs/s25main/WineLoader.cpp
ottml: Removed

libs/s25main/figures/nofTempleServant.cpp
ottml: Refactored and solved

libs/s25main/WineLoader.h
ottml: Add this as documentation for the entries. Removed it because the enums are descriptive enough.

libs/s25main/WineLoader.h
```diff
@@ -0,0 +1,101 @@
...
+
+bool isWineAddonBuildingType(BuildingType bld);
+bool isWineAddonGoodType(GoodType good);
+bool isWineAddonJobType(Job job);
+
+extern std::map<std::pair<Nation, BuildingType>, BuildingImages> buildings;
+using BobEntryRange = std::pair<unsigned, unsigned>;
+struct BobEntry
+{
+    BobEntryRange bobEntryRange;
```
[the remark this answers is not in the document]
ottml: No the value is not used. Removed it. Added it for first implmentation but it was not needed.

ottml:
I think the implementation is now feature complete. I will provide a new build in the next days for testing.
The only open task is:

> just asking: If the addon is disabled, are all the icons (inventory, settings screen) hidden as well? If so, could you also implement it for the charburner as well, if not, could you implement that since using the default s2 settings should not show these icons so player can get an original feeling :)

Flamefire:
> The only open task is:
> 
> > just asking: If the addon is disabled, are all the icons (inventory, settings screen) hidden as well? If so, could you also implement it for the charburner as well, if not, could you implement that since using the default s2 settings should not show these icons so player can get an original feeling :)

That would have been my question too: When the addon is disabled everything should look and work as before. I guess having the wares in the internal lists such as distribution is fine as it won't affect gameplay without those wares but for inventory screen, settings screen etc. it might matter.

libs/s25main/figures/nofWinegrower.cpp
```diff
@@ -0,0 +1,208 @@
...
+void nofWinegrower::DrawWalkingWithWare(DrawPoint drawPt)
+{
+    DrawWalking(drawPt, "wine_bobs", bobIndex[BobTypes::WINEGROWER_WALKING_WITH_FULL_BASKET]);
+}
+
+/// Draws the winegrower while walking
+/// (overriding standard method of nofFarmhand)
+void nofWinegrower::DrawOtherStates(DrawPoint drawPt)
+{
+    switch(state)
```
Flamefire, line 193: This switch is just an `if`, isn't it?

libs/s25main/gameTypes/TempleProductionMode.h
```diff
@@ -0,0 +1,25 @@
...
+    Coal,
+    Stone
+};
+
+constexpr auto maxEnumValue(ProductionMode)
+{
+    return ProductionMode::Stone;
+}
+
+const helpers::EnumArray<ProductionMode, ProductionMode> TRANSITIONS{
```
Flamefire, line 24:
```suggestion
constexpr helpers::EnumArray<ProductionMode, ProductionMode> TRANSITIONS{
```

libs/s25main/nodeObjs/noGrapefield.cpp
```diff
@@ -0,0 +1,153 @@
...
+#include "ogl/glArchivItem_Bitmap.h"
+#include "ogl/glSmartBitmap.h"
+#include "random/Random.h"
+#include "world/GameWorld.h"
+#include "s25util/colors.h"
+
+/// Duration of the grow waiting phase
+const unsigned GROWING_WAITING_LENGTH = 1100;
+/// Duration of grow phase
+const unsigned GROWING_LENGTH = 16;
```
Flamefire, line 21:
```suggestion
constexpr unsigned GROWING_WAITING_LENGTH = 1100;
/// Duration of grow phase
constexpr unsigned GROWING_LENGTH = 16;
```

libs/s25main/nodeObjs/noGrapefield.h
```diff
@@ -0,0 +1,60 @@
...
+    /// Can we harvest it?
+    bool IsHarvestable() const { return size == 3 && state == State::Normal; }
+
+    /// Return the ID of the withered grape field in the wine_bobs
+    unsigned GetHarvestID() const;
+
+    /// Winegrower starts harvesting the field
+    void BeginHarvesting();
+    /// Winegrower is interrupted duriong harvesting
+    void EndHarvesting();
```
Flamefire, line 59:
```suggestion
    /// Winegrower is interrupted during harvesting
    void AbortHarvesting();
```

"EndHarvesting" sounds like it will always be called. Maybe also "StopHarvesting"

Flamefire (changes requested):
Awesome work implementing this feature! Just some minor remarks but looks fine in general!

libs/s25main/GlobalGameSettings.cpp
```diff
@@ -100,7 +100,8 @@ void GlobalGameSettings::registerAllAddons()
         AddonStatisticsVisibility,
         AddonToolOrdering,
         AddonTrade,
-        AddonAutoFlags
+        AddonAutoFlags,
+        AddonWine
```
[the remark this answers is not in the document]
ottml: Not possible because of template here

libs/s25main/WineLoader.cpp
```diff
@@ -0,0 +1,109 @@
...
+    return {buildingStartIndex,     buildingStartIndex + 1, buildingStartIndex + 2,
+            buildingStartIndex + 3, buildingStartIndex + 4, buildingStartIndex + 5,
+            buildingStartIndex + 6, buildingStartIndex + 7, buildingStartIndex + 8};
+}
+
+helpers::EnumArray<unsigned, BobTypes> bobIndex = {1,   17,  22,  27,  32,  37,  42,  48,  97,  146, 195, 204,
+                                                   253, 270, 282, 289, 295, 344, 361, 378, 395, 412, 461, 510,
+                                                   517, 525, 526, 527, 528, 529, 530, 532, 533, 535, 536};
+
+unsigned GetWareTex(const GoodType good)
```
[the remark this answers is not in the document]
ottml: Return ITexture* and remove function GetWineBobImage not needed anymore

libs/s25main/figures/nofWinegrower.h
```diff
@@ -0,0 +1,54 @@
...
+
+class nofWinegrower : public nofFarmhand
+{
+    /// Is he harvesting grapes (or planting?)
+    bool harvest;
+
+private:
+    /// Draw worker at work
+    void DrawWorking(DrawPoint drawPt) override;
+    /// Id in jobs.bob or carrier.bob when carrying a ware
```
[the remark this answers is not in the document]
ottml: removed

libs/s25main/Loader.cpp
```diff
@@ -847,10 +873,29 @@ void Loader::fillCaches()
...
-                      dynamic_cast<glArchivItem_Bitmap_Player*>(bob_carrier->getOverlay(id, fat, imgDir, ani_step)));
-                    bmp.addShadow(GetMapImage(900 + static_cast<unsigned>(imgDir) * 8 + ani_step));
+                    if(ware == GoodType::Grapes)
+                    {
+                        const unsigned bodyIdx = static_cast<unsigned>(imgDir) * 8 + ani_step;
+                        bmp.add(dynamic_cast<glArchivItem_Bitmap_Player*>(wine_bob_carrier.get(
+                          wineaddon::bobIndex[fat ? wineaddon::BobTypes::FAT_CARRIER_CARRYING_GRAPES :
+                                                    wineaddon::BobTypes::THIN_CARRIER_CARRYING_GRAPES]
+                          + bodyIdx)));
+                        bmp.addShadow(GetMapImage(900 + static_cast<unsigned>(imgDir) * 8 + ani_step));
```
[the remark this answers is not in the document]
ottml: Yes

libs/s25main/Loader.cpp
```diff
@@ -847,10 +873,29 @@ void Loader::fillCaches()
...
                     const libsiedler2::ImgDir imgDir = toImgDir(dir);
 
-                    bmp.add(dynamic_cast<glArchivItem_Bitmap_Player*>(bob_carrier->getBody(fat, imgDir, ani_step)));
-                    bmp.add(
-                      dynamic_cast<glArchivItem_Bitmap_Player*>(bob_carrier->getOverlay(id, fat, imgDir, ani_step)));
-                    bmp.addShadow(GetMapImage(900 + static_cast<unsigned>(imgDir) * 8 + ani_step));
+                    if(ware == GoodType::Grapes)
+                    {
+                        const unsigned bodyIdx = static_cast<unsigned>(imgDir) * 8 + ani_step;
+                        bmp.add(dynamic_cast<glArchivItem_Bitmap_Player*>(wine_bob_carrier.get(
```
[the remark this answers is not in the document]
ottml: Because the Grapes carrier is not set up of body and head. We have here the complete worker animation in one image.

libs/s25main/figures/noFigure.cpp
```diff
@@ -799,6 +800,15 @@ void noFigure::DrawWalking(DrawPoint drawPt)
...
         case Job::CharBurner: DrawWalking(drawPt, "charburner_bobs", 53); break;
+        case Job::Vintner:
+            DrawWalking(drawPt, "wine_bobs", wineaddon::bobIndex[wineaddon::BobTypes::VINTNER_WALKING]);
+            break;
+        case Job::Winegrower:
+            DrawWalking(drawPt, "wine_bobs", wineaddon::bobIndex[wineaddon::BobTypes::WINEGROWER_WALKING_WITH_SHOVEL]);
+            break;
+        case Job::TempleServant:
+            DrawWalking(drawPt, "wine_bobs", wineaddon::bobIndex[wineaddon::BobTypes::TEMPLESERVANT_WALKING]);
+            break;
```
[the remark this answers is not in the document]
ottml: Yes this is a good idea, i thought also about this solution. I made it virtual and moved all the code to the accoring classes.

libs/s25main/ingameWindows/iwTempleBuilding.h
```diff
@@ -0,0 +1,26 @@
...
+
+#pragma once
+
+#include "IngameWindow.h"
+
+class nobUsual;
+class GameWorldView;
+class GameCommandFactory;
+
+class iwTempleBuilding : public IngameWindow
```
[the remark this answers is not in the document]
ottml: Its a derived class now

libs/s25main/ingameWindows/iwTempleBuilding.cpp
```diff
@@ -0,0 +1,167 @@
...
+        {
+            LOADER.GetWareTex(bldWorkDesk.waresNeeded[i])
+              ->DrawFull(waresPos, (z < building->GetNumWares(i) ? COLOR_WHITE : 0xFF404040));
+            waresPos.x += 24;
+        }
+
+        std::stringstream text;
+        text << (unsigned)building->GetNumWares(i) << "/" << wares_count;
+        NormalFont->Draw(curPos + DrawPoint(0, 12), text.str(), FontStyle::CENTER | FontStyle::VCENTER);
+        curPos.y += 29;
```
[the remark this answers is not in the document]
ottml: Make it a derived class to avoid code duplication. The window is larger and therefore the buttons have other positions.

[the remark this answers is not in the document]
Flamefire: Still: At least this code is C&P from `iwBuilding`, isn't it? Maybe you can pass the window size as a param to `iwBuilding` defaulting to the current size and generalize it enough to work for this. E.g. the first few buttons/images are on the same position here and the last can likely be calculated from the window height.

Would be great to not have this duplication leading to potential issues when changing this later on.

I guess at worst you'll need to move some controls in the ctor of this class after they were placed by the ctor of `iwBuilding`.

If it isn't feasible then we have to bite the bullet, but please check.

libs/s25main/ingameWindows/iwTempleBuilding.h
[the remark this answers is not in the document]
Flamefire: Perfect, much smaller class now!

BTW: Please don't "resolve" conversations that aren't just applied code as suggested. Otherwise it is hard to find your answers

Flamefire:
The failing replay test indicates you need to a) increase the gamedataversion and b) add compatibility code to allow loading older save games.

From a quick look it seems that https://github.com/Return-To-The-Roots/s25client/blob/7bf8281a41e094ad1aa43d574da489a13afe1958/libs/s25main/GamePlayer.cpp#L280 is the issue as you increased the transportPrio array size. Above that is an example how to handle it.

libs/s25main/figures/nofWinegrower.cpp
ottml: Done. Resolve when ok.

libs/s25main/ingameWindows/iwTempleBuilding.cpp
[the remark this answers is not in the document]
ottml: Removed class and integrate it into iwbuilding. Resolve when ok.

Flamefire:
> Remove iwTempleBuilding completly and add stuff to iwBuilding

What was the reason for this? I think the combination of this with the previous is the best:
- Default size parameter for `iwBuilding`
- A subclass for the temple which passes the custom/larger size to `iwBuilding`
- subclass adds extra controls as required
- "Relative" coords for controls (as in the new commit)

This removes the need for the seemingly random extend parameter for the temple window.

I mean the file for the subclass is less than 50 lines with only the changes required for it which looks very clean

ottml:
> > Remove iwTempleBuilding completly and add stuff to iwBuilding
> 
> What was the reason for this? I think the combination of this with the previous is the best:
> 
>     * Default size parameter for `iwBuilding`
> 
>     * A subclass for the temple which passes the custom/larger size to `iwBuilding`
> 
>     * subclass adds extra controls as required
> 
>     * "Relative" coords for controls (as in the new commit)
> 
> 
> This removes the need for the seemingly random extend parameter for the temple window.
> 
> I mean the file for the subclass is less than 50 lines with only the changes required for it which looks very clean

This was a misunderstanding. I changed it accordingly :)

libs/s25main/ingameWindows/iwTempleBuilding.cpp
```diff
@@ -0,0 +1,35 @@
...
+#include "world/GameWorldView.h"
+
+iwTempleBuilding::iwTempleBuilding(GameWorldView& gwv, GameCommandFactory& gcFactory, nobUsual* const building)
+    : iwBuilding(gwv, gcFactory, building, Extent(226, 223))
+{
+    GetCtrl<Window>(1)->SetPos(DrawPoint(117, 160));
+    GetCtrl<ctrlImage>(3)->SetImage(
+      wineaddon::GetTempleProductionModeTex(static_cast<nobTemple*>(building)->GetProductionMode()));
+    AddImageButton(8, DrawPoint(130, 176), Extent(34, 32), TextureColor::Grey,
+                   wineaddon::GetTempleProductionModeTex(static_cast<nobTemple*>(building)->GetProductionMode()));
```
Flamefire, line 20: Do you really want an image and the button with the same image? If so I'd guess you need to update both of them on click

Or can the button maybe show the next mode? I.e. click the button to switch to that shown mode and the image shows the current mode

libs/s25main/ingameWindows/iwBuilding.cpp
```diff
@@ -6,7 +6,9 @@
 #include "GamePlayer.h"
 #include "Loader.h"
 #include "WindowManager.h"
+#include "WineLoader.h"
 #include "buildings/nobShipYard.h"
+#include "buildings/nobTemple.h"
```
Flamefire, line 11:
```suggestion
#include "buildings/nobShipYard.h"
```

libs/s25main/Loader.cpp
```diff
@@ -150,13 +151,19 @@ glArchivItem_Bitmap* Loader::GetNationIcon(Nation nation, BuildingType bld)
...
     else
         return convertChecked<glArchivItem_Bitmap*>(nationIcons_[nation]->get(rttr::enum_cast(bld)));
 }
 
-ITexture* Loader::GetNationTex(Nation nation, unsigned nr)
+ITexture* Loader::GetNationTex(Nation nation, BuildingType bld)
 {
-    return checkedCast<ITexture*>(GetNationImage(nation, nr));
+    if(bld == BuildingType::Charburner || wineaddon::isWineAddonBuildingType(bld))
+        return &LOADER.building_cache[nation][bld].building;
```
Flamefire, line 163: Can you fix this at least for the wine-addon buildings to return the building image without the shadow.   
Or does it maybe look better with the shadow? I don't think so but we do that already for the building image in the building dialog. Maybe @Spikeone has an opinion here (whether to only use shadows for buildings on the map or also for dialogs as is now) especially compared to the original. IIRC there was even an issue long time ago, wasn't there?

Flamefire (changes requested):
> This was a misunderstanding. I changed it accordingly :)

Alright. Could have been that the implementation turned out harder than I thought or blocked by something. Hence the question :)

libs/s25main/Loader.cpp
```diff
@@ -150,13 +151,19 @@ glArchivItem_Bitmap* Loader::GetNationIcon(Nation nation, BuildingType bld)
...
+    else
+        return checkedCast<ITexture*>(GetNationImage(nation, 250 + rttr::enum_cast(bld) * 5));
+    return nullptr;
 }
```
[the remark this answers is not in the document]
ottml: I now use the NothingX entries for the wineaddon and use the override mechanism for addons. Building order addon shows building without shadow. Yes it was a bug of the charburner not displayed in the build order addon, i fixed this.

libs/s25main/WineLoader.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+bool isWineAddonJobType(Job job)
+{
+    return job == Job::Winegrower || job == Job::Vintner || job == Job::TempleServant;
+}
+
+BuildingImages GetBuildingImages(Nation nation, BuildingType buildType)
+{
+    const unsigned buildingStartIndex =
+      1 + rttr::enum_cast(nation) * 27 + (rttr::enum_cast(buildType) - rttr::enum_cast(BuildingType::Vineyard)) * 9;
```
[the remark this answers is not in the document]
ottml: Resolved by using override mechanism.

libs/s25main/gameTypes/BuildingType.h
```diff
@@ -48,11 +48,14 @@ enum class BuildingType : uint8_t
     Farm,           // 37
     DonkeyBreeder,  // 38
     HarborBuilding, // 39
+    Vineyard,       // 40
```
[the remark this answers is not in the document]
ottml: Reuse of nothing entries implemented

libs/s25main/Loader.cpp
ottml: Changed to use image without shadow.

libs/s25main/ingameWindows/iwTempleBuilding.cpp
ottml: This was a copy and past error. The first image should never change and is fixed.

libs/s25main/Loader.cpp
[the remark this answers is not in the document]
ottml: Renamed GetNationTex

libs/s25main/ingameWindows/iwTempleBuilding.cpp
aztimh: If you like I can add a larger output icon for the top right of the window: the gold, iron ore, coal and stone. Then use the smaller one for the button and I can add smaller icons for iron ore, coal and stone for the button toggle at the bottom?

ottml:
> The failing replay test indicates you need to a) increase the gamedataversion and b) add compatibility code to allow loading older save games.
> 
> From a quick look it seems that
> 
> https://github.com/Return-To-The-Roots/s25client/blob/7bf8281a41e094ad1aa43d574da489a13afe1958/libs/s25main/GamePlayer.cpp#L280
> is the issue as you increased the transportPrio array size. Above that is an example how to handle it.

I added compatibility code for old savegames. But the replay test does not use this code. Does the replay not only record GameCommands? Is there also a possibility for compatibility for GameCommands? For example the ChangeDistribution Command fails.

Flamefire:
> I added compatibility code for old savegames. But the replay test does not use this code. Does the replay not only record GameCommands? Is there also a possibility for compatibility for GameCommands? For example the ChangeDistribution Command fails.

I don't see a way to do this, the GameCommands use the basic serializer because they are used in the game where versioning doesn't make sense. While we could change that such that a version is added for those (and only those) we'd need to add the version to the replay file. And I cant see where/how without breaking backwards compatibility. Maybe we can use the `lastGF` entry and make it negative until raising the replay version to denote that the replay sub-version follows.   
We better do that in a separate PR, I can look into that.

Otherwise we need to declare a new version rendering previous replays unusable.
https://github.com/Return-To-The-Roots/s25client/blob/7bf8281a41e094ad1aa43d574da489a13afe1958/libs/s25main/Replay.cpp#L23

ottml:
> > I added compatibility code for old savegames. But the replay test does not use this code. Does the replay not only record GameCommands? Is there also a possibility for compatibility for GameCommands? For example the ChangeDistribution Command fails.
> 
> I don't see a way to do this, the GameCommands use the basic serializer because they are used in the game where versioning doesn't make sense. While we could change that such that a version is added for those (and only those) we'd need to add the version to the replay file. And I cant see where/how without breaking backwards compatibility. Maybe we can use the `lastGF` entry and make it negative until raising the replay version to denote that the replay sub-version follows. We better do that in a separate PR, I can look into that.
> 
> Otherwise we need to declare a new version rendering previous replays unusable.
> 
> https://github.com/Return-To-The-Roots/s25client/blob/7bf8281a41e094ad1aa43d574da489a13afe1958/libs/s25main/Replay.cpp#L23

If you think it's worse the effort we can do this. I think we should simply increase the version and live with the breaking loading of old replays, as lat done in 2023.

Flamefire:
> If you think it's worse the effort we can do this. I think we should simply increase the version and live with the breaking loading of old replays, as lat done in 2023.

Yes I think it is worth it and I already have a good idea how to do that. One of the reasons is that there are still some open issues with replays attached which are required to reproduce the issue. Breaking compatibility will make that impossible

I'll work on it over the weekend

libs/s25main/WineLoader.cpp
```diff
@@ -0,0 +1,103 @@
...
+///////////////////////////////////////////////////////////////////////////////
+
+#include "WineLoader.h"
+#include "GlobalGameSettings.h"
+#include "Loader.h"
+#include "addons/const_addons.h"
+#include "ogl/glArchivItem_Bitmap.h"
+#include "ogl/glTexturePacker.h"
+#include "world/GameWorldBase.h"
+#include "world/GameWorldView.h"
```
Flamefire, line 14: Unused?
```suggestion
```

libs/s25main/ingameWindows/iwBuildingProductivities.cpp
```diff
@@ -25,9 +29,27 @@ const std::array<BuildingType, 24> iwBuildingProductivities::icons = {
...
+}
+
+void iwBuildingProductivities::removeUnusedBuildings()
+{
+    auto removeNotUsedBuilding = [=](BuildingType const& bts) {
+        return ((!wineaddon::isAddonActive(player.GetGameWorld()) && wineaddon::isWineAddonBuildingType(bts))
+                || (!player.GetGameWorld().GetGGS().isEnabled(AddonId::CHARBURNER) && bts == BuildingType::Charburner));
+    };
+    usedIcons.erase(std::remove_if(std::begin(usedIcons), std::end(usedIcons), removeNotUsedBuilding),
+                    std::end(usedIcons));
```
Flamefire, line 50: Combine and simplify
```suggestion
    usedIcons.assign(allIcons.begin(), allIcons.end()
    const auto isUnused = [&](BuildingType const& bld) {
        if (!wineaddon::isAddonActive(player.GetGameWorld() && wineaddon::isWineAddonBuildingType(bld))
            return false;
        if(!player.GetGameWorld().GetGGS().isEnabled(AddonId::CHARBURNER) && bld == BuildingType::Charburner));
            return false;
        return true;
    };
    helpers::erase_if(usedIcons, isUnused);
```

Basically:
- Rename condition lambda from action verb to adjective
- the lambda needs to capture by reference or you'll copy at least `player`
- Split visibly into the 3 cases (avoids the prior confusion, did clang-tidy suggest to combine those? in that case keep it of course. But I guess it was just the `else` in `if-return-else`)
- for most container ops requiring a range over the full container there is a helper to avoid mistakes when repeating the container name. Kind of like ranges.
- Single use function can be inlined, there is no other possible use for this as you'd always need to refill the building first. With the short lambda and `helpers::erase_if(usedIcons, isUnused);` it's pretty clear what this code does already so the function name in this case doesn't help much for readability.

libs/s25main/ingameWindows/iwBuildings.cpp
```diff
@@ -26,24 +31,38 @@
 #include "gameData/const_gui_ids.h"
 
 /// Reihenfolge der Gebäude
```
Flamefire, line 33:
```suggestion
```

libs/s25main/ingameWindows/iwBuildings.cpp
```diff
@@ -26,24 +31,38 @@
...
-  BuildingType::PigFarm,
-  BuildingType::Storehouse, // entry 21
-  BuildingType::Mill,          BuildingType::Bakery,     BuildingType::Sawmill,        BuildingType::Mint,
-  BuildingType::Well,          BuildingType::Shipyard,   BuildingType::Farm,           BuildingType::DonkeyBreeder,
-  BuildingType::Charburner,
-  BuildingType::HarborBuilding // entry 31
-};
+void iwBuildings::setBuildingOrder()
+{
+    bts = {
```
Flamefire, line 36:
```suggestion
    // Order of the buildings in which they will be shown
    bts = {
```

libs/s25main/ingameWindows/iwBuildings.cpp
```diff
@@ -52,24 +71,29 @@ iwBuildings::iwBuildings(GameWorldView& gwv, GameCommandFactory& gcFactory)
...
+    const unsigned rowCount = helpers::divCeil(bts.size(), 4);
+    Resize(iconSpacing * Extent(4, rowCount + 1) + bldContentOffset);
+
     const Nation playerNation = gwv.GetViewer().GetPlayer().nation;
     // Symbole für die einzelnen Gebäude erstellen
-    for(unsigned y = 0; y < bts.size() / 4 + (bts.size() % 4 > 0 ? 1 : 0); ++y)
+    for(unsigned y = 0; y < rowCount; y++)
     {
-        for(unsigned x = 0; x < 4; ++x)
+        for(unsigned x = 0; x < 4; x++)
```
Flamefire, line 82: Please don't do such changes (no need to revert though either)
- There is no benefit in this case but it adds to the diff to review
- having a habbit of preincrement avoids thinking about the type
- `for(const x: helpers::range(4)` would be clearer if you want to improve such loops while changing them

libs/s25main/ingameWindows/iwDistribution.cpp
```diff
@@ -191,6 +186,27 @@ void iwDistribution::CreateGroups()
...
+        else if(!gwv.GetWorld().GetGGS().isEnabled(AddonId::CHARBURNER) && buildingType == BuildingType::Charburner)
+            return true;
+        else
+            return false;
+    };
+    for(auto& group : groups)
+    {
+        auto& entries = group.entries;
+        entries.erase(std::remove_if(std::begin(entries), std::end(entries), removeNotUsedDistribution),
+                      std::end(entries));
```
Flamefire: Similar here

Flamefire (changes requested):
Just some style improvements :)

libs/s25main/ingameWindows/iwBuildingProductivities.cpp
ottml: Applied the changes at all places

libs/s25main/ingameWindows/iwDistribution.cpp
ottml: Changed

libs/s25main/ingameWindows/iwWares.cpp
```diff
@@ -59,16 +63,16 @@ static void addElement(ctrlGroup& page, const glFont* font, const DrawPoint btPo
     page.AddText(600 + idOffset, txtPos, "", COLOR_YELLOW, FontStyle::RIGHT | FontStyle::BOTTOM, font);
 }
 
-// 167, 416
-iwWares::iwWares(unsigned id, const DrawPoint& pos, const Extent& size, const std::string& title, bool allow_outhousing,
-                 const glFont* font, const Inventory& inventory, const GamePlayer& player)
-    : IngameWindow(id, pos, size, title, LOADER.GetImageN("io", 5)), inventory(inventory), player(player), numPages(0)
+iwWares::iwWares(unsigned id, const DrawPoint& pos, unsigned additionalYSpace, const std::string& title,
+                 bool allow_outhousing, const glFont* font, const Inventory& inventory, const GamePlayer& player)
+    : IngameWindow(id, pos, Extent(167, 416), title, LOADER.GetImageN("io", 5)), inventory(inventory), player(player),
```
Flamefire, line 68: Wasn't the (min) height 415 (415 constant and 455=415+40)? And although you are resizing later I think it would be clearer to add it here:
```suggestion
    : IngameWindow(id, pos, Extent(167, 415 + ), title, LOADER.GetImageN("io", 5)), inventory(inventory), player(player),
```
```suggestion
    : IngameWindow(id, pos, Extent(167, 415 + additionalYSpace), title, LOADER.GetImageN("io", 5)), inventory(inventory), player(player),
```

libs/s25main/ingameWindows/iwBuildOrder.cpp
```diff
@@ -4,25 +4,44 @@
...
 #include "controls/ctrlComboBox.h"
 #include "controls/ctrlImage.h"
 #include "controls/ctrlList.h"
 #include "network/GameClient.h"
+#include "world/GameWorldBase.h"
 #include "world/GameWorldViewer.h"
 #include "gameData/BuildingConsts.h"
 #include "gameData/const_gui_ids.h"
 
+void iwBuildOrder::fillBuildOrderAndRemoveUnusedBuildings()
```
Flamefire, line 21: The caller doesn't need to know about the removing, that's an internal detail of this function. So keep it short:
```suggestion
void iwBuildOrder::fillBuildOrder()
```

libs/s25main/figures/nofPassiveWorker.cpp
```diff
@@ -32,7 +35,29 @@ void nofPassiveWorker::AbrogateWorkplace() {}
...
+            DrawWalking(drawPt, "wine_bobs", wineaddon::bobIndex[wineaddon::BobTypes::VINTNER_WALKING]);
+            break;
+        case Job::Winegrower:
+            DrawWalking(drawPt, "wine_bobs", wineaddon::bobIndex[wineaddon::BobTypes::WINEGROWER_WALKING_WITH_SHOVEL]);
+            break;
+        case Job::TempleServant:
+            DrawWalking(drawPt, "wine_bobs", wineaddon::bobIndex[wineaddon::BobTypes::TEMPLESERVANT_WALKING]);
+            break;
+        default: DrawWalkingBobJobs(drawPt, job_); break;
+    }
```
Flamefire, line 60: Besides the new 3 cases, this is a copy of the code in `nofFigure::DrawWalking`. Why this change?

libs/s25main/ingameWindows/iwWares.cpp
ottml: We can do it here but then we have to precompute the number of rows needed here because the size depends on it.

libs/s25main/figures/nofPassiveWorker.cpp
Flamefire: Ok duplicating code is never a good idea especially as we now have here exactly the code we wanted to avoid to have in `noFigure::DrawWalking`   
So yes we need to move this code into `DrawWalking` and make it non-virtual again :-/

I would go so far and replace the `DrawWalkingBobJobs(drawPt, job_)` together with the comment in nofActive/PassiveSoldier by `DrawWalking`.   
Maybe even all other calls to `DrawWalkingBobJobs`, except for the one in nofScout where the job argument is not `job_`, as `DrawWalking(drawPos)` is cleaner than e.g. `DrawWalkingBobJobs(drawPt, Job::Geologist);` removing the question whether anything special is done there.

What do you think? Relevant for this PR is only the de-duplication. If you think it is worth it or have a better idea check that 2nd part.

libs/s25main/ingameWindows/iwWares.cpp
Flamefire: Yeah we can't reasonably do that. The idea was to show right here where `additionalYSpace` is used instead of just the magic number for the height

Flamefire (changes requested):

libs/s25main/figures/nofPassiveWorker.cpp
ottml: Reverted drawwalking virtual and remove code from nofPassiveWorker

ottml: Will not do the second part now

libs/s25main/ingameWindows/iwWares.cpp
```diff
@@ -155,6 +175,9 @@ iwWares::iwWares(unsigned id, const DrawPoint& pos, const Extent& size, const st
         }
     }
 
+    // compute the final window size
+    Resize(Extent(167, 21 + (y + 1) * 42 + 50 + additionalYSpace + 11));
```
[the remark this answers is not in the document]
ottml: Added constants

Flamefire (changes requested):

ottml:
> > If you think it's worse the effort we can do this. I think we should simply increase the version and live with the breaking loading of old replays, as lat done in 2023.
> 
> Yes I think it is worth it and I already have a good idea how to do that. One of the reasons is that there are still some open issues with replays attached which are required to reproduce the issue. Breaking compatibility will make that impossible
> 
> I'll work on it over the weekend

@Flamefire Are you working on this topic? Or do we increase the version for new and record the replays with the new version?

ottml:
The implementation is finished. Only the topic in the above comment is unresolved. After this is finished i will provide a last test version to @aztimh.  If everything is fine we can squash the commits into one and rebase the master.

Flamefire:
> @Flamefire Are you working on this topic? Or do we increase the version for now and record the replays with the new version?

Yes and almost done testing. See #1677 
I'm currently testing whether we can have your addon and old replays working. Ran into an async related to the RNG

libs/s25main/figures/nofWinegrower.cpp
```diff
@@ -0,0 +1,200 @@
...
+    {
+        if(world->GetSpecObj<noGrapefield>(pt)->IsHarvestable())
+            return PointQuality::Class1;
+        else
+            return PointQuality::NotPossible;
+    } else // or a free space, to place a new one
+    {
+        // Try to "plant" a new grapefield
+        // Still enough wares when starting new work (state = Waiting1)?
+        if(!workplace->WaresAvailable() && state == State::Waiting1)
```
Flamefire, line 130: Use the "cheap" condition first:
```suggestion
        if(state == State::Waiting1 && !workplace->WaresAvailable())
```

Flamefire:
After the versioned serializer was merged you should be able to get the replay tests passing with [this commit](https://github.com/Flamefire/s25client/commit/214173bcf4ca6ee90568024bef06afec5af7d0fa)

Note: Just pick this single commit or the [patch](https://github.com/Flamefire/s25client/commit/214173bcf4ca6ee90568024bef06afec5af7d0fa.patch), not the branch or C&P the changes from that commit

ottml:
@Flamefire The patch fixes the crash but during execution there is a mismatch in the checksum. Any idea what could be the problem here?

ottml:
Because we have now 3 jobs more the random number generator is not in sync anymore. This happens for example in the loop in BurnedWarehouse.cpp:76. Calling the random generator at BurnedWarehouse.cpp:90 three times more. The replay can be executed but a warning is shown. The test itself fails.

libs/s25main/gameData/DoorConsts.h
```diff
@@ -12,17 +12,17 @@
 /// Y-value for the point where the carrier disappears inside the building for each nation and building
 constexpr helpers::MultiEnumArray<int8_t, Nation, BuildingType> DOOR_CONSTS = {
   {// Nubier
-   {5, 10, 13, 0,  10, 0,  0,  0, 0,  6,  8, 8,  8,  8, 6,  0, 10, 10, 12, 14,
-    9, 5,  11, 19, 19, 12, 18, 0, -6, 19, 0, 12, 11, 6, 10, 0, 0,  -1, 4,  13},
+   {5, 10, 13, 0,  10, 0,  11, 19, 0,  6,  8, 8,  8,  8, 6,  0, 10, 10, 12, 14,
```
Flamefire, line 15: The nubian vinyard doesn't need an offset?

Flamefire:
Sorry for the delay, I've been on vacation. You are right with your evaluation and I found that too. I had already fixed that earlier and made a PR in your fork: https://github.com/ottml/s25client/pull/1

Your idea is good, but too specific to keep. My attempt basically skips every job type without people in the warehouse, just as yours but not restricted to the wine addon.

It also contains some improvements to the farmhand class which helps for this PR, e.g. avoiding a costly re-computation.

It does require a new test replay (not too bad given we broke that just recently) but should be a) faster and b) more stable going forward.

I'll rebase that once more (currently verifying the replays work), so check that PR and review if you like. It is large, because I translated and fixed many comments and docstrings. It also contains an important bugfix where a figure could leave a warehouse in an invalid direction.

libs/s25main/gameData/DoorConsts.h
ottml: Yes the nubian vinyard has no offset

Flamefire: Ok thanks.

Please don't resolve conversations you answered to unless it is just applying a suggested fix from the comment.  
That makes it much easier to keep track of things I already looked at and what are new answers. The person starting the conversation should resolve it when the topic of the discussion is finished. E.g. when the original question wasn't fully answered or the reviewer has a follow-up question which the other person might not be aware of.

libs/s25main/buildings/BurnedWarehouse.cpp
```diff
@@ -75,57 +73,49 @@ void BurnedWarehouse::HandleEvent(const unsigned /*id*/)
...
-
-            // Die Figuren schließlich rausschicken
-            for(unsigned z = 0; z < numPeopleInDir; ++z)
+            // Get current direction accounting for startIdx and hence possible wrap around
+            const unsigned idx = j + startIdx;
+            const Direction curDir = possibleDirs[idx < possibleDirs.size() ? idx : idx - possibleDirs.size()];
+            // Take all in last direction
+            const auto curNumPeople = (j + 1u < possibleDirs.size()) ? numPeoplePerDir : count;
+            count -= curNumPeople;
+            for(const auto z : helpers::range(curNumPeople))
```
Flamefire, line 100: Should have tested with GCC, sorry for that -->
```suggestion
            for([[maybe_unused]] const auto z : helpers::range(curNumPeople))
```

Flamefire (dismissed):
Should be fine with the warning suppression. Thanks again for your efforts!

Flamefire (approved):

ottml:
If all ci checks succeed i will do a last squashing of some commits to have a cleaner history.

ottml:
Squash finished. We can merge it as soon as ci is through.

Flamefire (approved):

Flamefire:
:partying_face: :tada:

aztimh:
Awesome work thanks so much! I'll get some screenshots and write up a news post for Spike to put on the RTTR website. I'll be sure to credit all the work you both did programming and refining the code :)

Thanks again,

- aztimh :)

Issue #1674, opened 2024-06-16
crash/breakpoint in world.cpp


stefson:
the backtrace: 

```
Thread 1 "s25client" received signal SIGTRAP, Trace/breakpoint trap.
0x000055555573b78e in World::ReduceResource (this=<optimized out>, pt=...)
    at /var/tmp/portage/games-strategy/s25rttr-9999-r6/work/s25rttr-9999/libs/s25main/world/World.cpp:164
warning: 164	/var/tmp/portage/games-strategy/s25rttr-9999-r6/work/s25rttr-9999/libs/s25main/world/World.cpp: Datei oder Verzeichnis nicht gefunden
(gdb) bt
#0  0x000055555573b78e in World::ReduceResource (this=<optimized out>, pt=...) at /var/tmp/portage/games-strategy/s25rttr-9999-r6/work/s25rttr-9999/libs/s25main/world/World.cpp:164
#1  0x000055555562f62b in nofFisher::WorkFinished (this=0x555558945240) at /var/tmp/portage/games-strategy/s25rttr-9999-r6/work/s25rttr-9999/libs/s25main/figures/nofFisher.cpp:117
#2  0x000055555562e843 in nofFarmhand::HandleDerivedEvent (this=0x555558945240)
    at /var/tmp/portage/games-strategy/s25rttr-9999-r6/work/s25rttr-9999/libs/s25main/figures/nofFarmhand.cpp:47
#3  0x0000555555742923 in EventManager::ExecuteEvents (this=this@entry=0x5555574d59d0, itEvents=...)
    at /var/tmp/portage/games-strategy/s25rttr-9999-r6/work/s25rttr-9999/libs/s25main/EventManager.cpp:141
#4  0x00005555557425bf in EventManager::ExecuteCurrentEvents (this=0x5555574d59d0)
    at /var/tmp/portage/games-strategy/s25rttr-9999-r6/work/s25rttr-9999/libs/s25main/EventManager.cpp:125
#5  EventManager::ExecuteNextGF (this=0x5555574d59d0) at /var/tmp/portage/games-strategy/s25rttr-9999-r6/work/s25rttr-9999/libs/s25main/EventManager.cpp:87
#6  0x0000555555744eee in Game::RunGF (this=0x55555635a080) at /var/tmp/portage/games-strategy/s25rttr-9999-r6/work/s25rttr-9999/libs/s25main/Game.cpp:77
[...]

PR #1675, opened 2024-06-17
Fix fisher not getting the fish he caught

Flamefire:
The code triggers an assertion when 2 fishers try to get the last fish
from the same spot because only the first can get it.
So remove the fish when starting to fish such that it is already
unavailable when the 2nd starts fishing.

Fixes https://github.com/Return-To-The-Roots/s25client/issues/1674

Also a fix for some tests crashing depending on the order they are executed. (TextControlWorks & TableSorting due to glFont, the other due to the guiScale change)

### DIFF
--- a/libs/s25main/figures/nofFarmhand.cpp
+++ b/libs/s25main/figures/nofFarmhand.cpp
@@ -204,28 +204,29 @@ bool nofFarmhand::IsPointAvailable(const MapPoint pt) const
 
 void nofFarmhand::WalkToWorkpoint()
 {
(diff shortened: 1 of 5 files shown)


Flow86 (approved):

Flamefire:
This unfortunately breaks replays as it seems to happen quite often that 2 fishers fish on the exact some spot and with this change the probability changes in a few of those cases. Hence we need to regenerate them or CI will keep failing --> #1676 (WIP)

PR #1677, opened 2024-07-01
Add version to replay and recorded game commands

Flamefire:
For e.g. #1673 we should have some backwards compatibility to allow loading older replays even if there are some changes to serialized game commands.

The main contribution to this is the `VersionedSerializer` class in libutil which is an adapter over the serializer adding a version. To be as uninvasive as possible it allows to be used only for parts. E.g. an instance being passed a `Serializer` instance can create a `VersionedSerializer` on top of it for the duration of the method and afterwards the caller can use the original `Serializer` without even knowing about this process. This is done by recording the position of the Serializer and advancing it by the amount of bytes read through the Adapter.

I also translated/updated the docstrings and while doing that I found some places for major improvements:

- The random seed needs to be set when recording a replay, so require passing it
- We don't need `Replay::IsValid` but only `IsRecording/IsReplaying`
- `ReadGF` can return a proper optional instead of a bool setting a pointer param
- Reading a replay command involved reading the type and calling the right function afterwards. To much detail on the caller -> Use a variant
- Boost.Variant2 is superior to both Boost.Variant and `std::variant` by being trivial in some cases and being never empty, so use it and add a typedef for easier usability (instead of `boost::variant2::variant`)
- Return `std::string` references for some accessors instead of copies
- `GetDataWritable + SetLength` is inconvenient as we always need them paired up, so do that in one function
- Allow a Deserializer to use an existing buffer, i.e. don't copy data when passed to the ctor of `Serializer`

### DIFF
--- a/libs/common/include/variant.h
+++ b/libs/common/include/variant.h
@@ -1,22 +1,20 @@
-// Copyright (C) 2005 - 2021 Settlers Freaks (sf-team at siedler25.org)
+// Copyright (C) 2005 - 2024 Settlers Freaks (sf-team at siedler25.org)
 //
 // SPDX-License-Identifier: GPL-2.0-or-later
 
 #pragma once
 
-#include <boost/variant.hpp>
-#include <type_traits>
+#include <boost/variant2/variant.hpp>
+
+/// Shortcurs to avoid typing out boost::variant2
+template<typename... T>
+using boost_variant2 = boost::variant2::variant<T...>;
+
+using boost::variant2::get;
+using boost::variant2::get_if;
+using boost::variant2::holds_alternative;
 
 namespace detail {
-template<typename...>
-struct indexOf;
-template<typename T, typename... Rest>
-struct indexOf<T, T, Rest...> : std::integral_constant<size_t, 0>
-{};
-template<typename T, typename U, typename... Rest>
-struct indexOf<T, U, Rest...> : std::integral_constant<size_t, 1 + indexOf<T, Rest...>::value>
-{};
-
 template<typename... Lambdas>
 struct lambda_visitor;
 
@@ -35,12 +33,6 @@ struct lambda_visitor<Lambda1> : public Lambda1
     lambda_visitor(Lambda1 l1) : Lambda1(l1) {}
 };
 } // namespace detail
-
-template<class T, class... Types>
-bool holds_alternative(const boost::variant<Types...>& v) noexcept
-{
-    return v.which() == detail::indexOf<T, Types...>::value;
-}
 
 template<class... Fs>
 auto composeVisitor(Fs&&... fs)
--- a/libs/common/include/variant.h
+++ b/libs/common/include/variant.h
@@ -6,7 +6,7 @@
 
 #include <boost/variant2/variant.hpp>
 
-/// Shortcurs to avoid typing out boost::variant2
+/// Shortcuts to avoid typing out boost::variant2
 template<typename... T>
 using boost_variant2 = boost::variant2::variant<T...>;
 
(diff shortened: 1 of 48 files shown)


Flamefire:
This is now ready

Flamefire:
@Flow86 Please review and merge. The Job requirement "`Linux (gcc-9, ubuntu-20.04, Debug, 3.16.3, 1.69.0)`" needs to be replaced to check for Boost 1.71, i.e. "`Linux (gcc-9, ubuntu-20.04, Debug, 3.16.3, 1.71.0)`"

libs/common/include/variant.h
```diff
@@ -1,22 +1,20 @@
...
-namespace detail {
-template<typename...>
-struct indexOf;
-template<typename T, typename... Rest>
-struct indexOf<T, T, Rest...> : std::integral_constant<size_t, 0>
-{};
-template<typename T, typename U, typename... Rest>
-struct indexOf<T, U, Rest...> : std::integral_constant<size_t, 1 + indexOf<T, Rest...>::value>
-{};
+/// Shortcurs to avoid typing out boost::variant2
```
Flow86, line 9: typo

libs/s25main/GameCommand.cpp
```diff
@@ -9,7 +9,12 @@
 
 namespace gc {
 
-GameCommandPtr GameCommand::Deserialize(Serializer& ser)
+unsigned Deserializer::getCurrentVersion()
+{
+    return 0;
```
Flow86, line 14: can we have an explanation here what the value does?

Flow86 (changes requested):

libs/s25main/GameCommand.cpp
Flamefire: Do we need that here again? The documentation is at [`VersionedDeserializer`](https://github.com/Return-To-The-Roots/libutil/compare/98d6ab5b21bf7f35a2d5ee76d6ea43c7d416f8a1...a40c2337e87bec26ae495d7186e002d6cb0bb8bf#diff-310d747bd13159787f152c327cee22ff62a4c0ce15ee504c638c36846242cde1R14-R21)

I'm also not sure what to write here. It would be something like "Current version of the game data", but most of that is already in the method name. Any suggestion?

BTW: One use example is in a branch I created on top of the wine addon PR: https://github.com/Flamefire/s25client/commit/214173bcf4ca6ee90568024bef06afec5af7d0fa

libs/common/include/variant.h
Flamefire:
```suggestion
/// Shortcuts to avoid typing out boost::variant2
```

libs/s25main/GameCommand.cpp
Flow86: > The documentation is at [VersionedDeserializer](https://github.com/Return-To-The-Roots/libutil/compare/98d6ab5b21bf7f35a2d5ee76d6ea43c7d416f8a1...a40c2337e87bec26ae495d7186e002d6cb0bb8bf#diff-310d747bd13159787f152c327cee22ff62a4c0ce15ee504c638c36846242cde1R14-R21)
maybe exactly this. That should be enough.

is 0 now the version? if so, please make it a named constant, so its even more clear what it is.

Flamefire: I added that as a docstring.

This already is a named constant. But I see that a function could do more than just returning the constant. I changed it to a static member constant to make that clear. The original intention was that a function returning a value leaves no doubt that its value cannot be modified, as opposed to a member variable where it was forgotten to make it constant.

Flow86 (approved):

Flamefire:
>  Linux (gcc-9, ubuntu-20.04, Debug, 3.16.3, 1.69.0) Expected — Waiting for status to be reported 

@Flow86 Can you change that expected job to 1.71.0 please such that this can be merged

PR #1679, opened 2024-07-23
Add all classic S2 cheats and a few more

kubaau:
- refactors existing cheat code ("winter" code, toggle human AI player, armageddon) into new Cheats and CheatCommandTracker classes
- adds classic (S2) cheats:
  - all-visible (F7) - [video](https://streamable.com/464uxc)
  - build headquarters - [video](https://streamable.com/1u9l86)
  - show enemy productivity overlay - [video](https://streamable.com/ihv4xl)
  - enable all buildings - [video](https://streamable.com/98prx7)
  - set speed by using ALT-1 .. ALT-6
- adds new (RttR) cheats:
  - reveal resources - [video](https://streamable.com/ehxa4o)
  - destroy buildings (kinda similar to armageddon) - [video](https://streamable.com/4ojiaj)

I would happily split these cheats into separate PR's, but they all rely on the refactoring, so at least the first few commits would have to be merged before any split can be done.

The all-visible and build HQ cheats are useful for testing campaign scripts, which I have also been working on. The remaining S2 cheats are there mostly for feature parity and setting speed does not work the same as it did in S2 (certainly not fast enough, probably steps are different too).

The reveal resources cheat can be useful not only in gameplay but in debugging as well and the destroy buildings cheat is useful for testing the world campaign progress where the player needs to achieve total domination but also not destroy themselves (as they would with the armageddon cheat).

Related to issue #72

### DIFF
--- a/libs/s25main/world/GameWorldView.cpp
+++ b/libs/s25main/world/GameWorldView.cpp
@@ -611,7 +611,7 @@ void GameWorldView::DrawResource(const MapPoint& pt, DrawPoint curPos, const Che
 
     if(auto* bm = LOADER.GetWareTex(gt))
     {
-        for(auto i : helpers::range(amount))
+        for([[maybe_unused]] const auto i : helpers::range(amount))
         {
             bm->DrawFull(curPos);
             curPos.y -= 4;


Flow86:
Wow, great effort.
I have one question: Probably some kind of `enable cheats` "addon"-setting should be necessary (especially for multiplayer) Is there already one in there, otherwise we definitly need to add one before we can merge this.

libs/s25main/Cheats.h
```diff
@@ -0,0 +1,130 @@
...
+    /** Checks if cheats can be turned on at all.
+     *
+     * @return true if if cheats can be turned on, false otherwise
+     */
+    bool canCheatModeBeOn() const;
+
+    std::unique_ptr<CheatCommandTracker> cheatCmdTracker_;
+    bool isCheatModeOn_ = false;
+    bool isAllVisible_ = false;
+    GameWorldBase& world_;
```
Flow86, line 128: we should use a `std::reference_wrapper` instead of a plain reference

libs/s25main/world/GameWorldBase.h
```diff
@@ -59,6 +60,7 @@ class GameWorldBase : public World
     EventManager& em;
     std::unique_ptr<SoundManager> soundManager;
     LuaInterfaceGame* lua;
+    std::unique_ptr<Cheats> cheats;
```
Flow86, line 63: don't know if this needs to be an unique_ptr? can't you simply call the ctor in GameWorldBase ctor?

libs/s25main/Cheats.cpp
```diff
@@ -0,0 +1,150 @@
...
+    PlayerIDSet ais;
+    for(auto i = 0u; i < world_.GetNumPlayers(); ++i)
+        if(!world_.GetPlayer(i).isHuman())
+            ais.insert(i);
+    destroyBuildings(ais);
+}
+
+bool Cheats::canCheatModeBeOn() const
+{
+    return world_.IsSinglePlayer();
```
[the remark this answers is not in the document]
kubaau: I could certainly create an "enable cheats in multiplayer" addon or something similar, but it would be useful to ask which of these cheats make sense in multiplayer. I don't really know a reason to enable most of them since they are so game-breaking.

On the other hand, I guess they could help in debugging, giving someone a handicap (experienced player sees FOW, newbie sees everything; experienced needs geologists, newbie sees resources, etc.) or eliminating the need to send geologists anywhere.

libs/s25main/world/GameWorldBase.h
kubaau: No reason other than to forward-declare instead of including. This can just be `Cheats cheats;` if this is preferred. I didn't want to recompile so much code when working on the Cheats header but I can see the benefits of removing the unique_ptr crap (clarity, unnecessary to include Cheats.h in some other places).

Spikeone:
I don't think we should enable cheats in multiplayer, even with addons, I don't see a use case. For balancing reasons we should rather add some real balancing options instead of cheats (which most people are not aware of I think)

kubaau:
I will place a PR for the first three commits of this one and then add one PR per cheat/feature. I didn't want to do this in the first place because the refactoring may have seemed pointless, but now that a couple of people have seen the results this can bring, I think the first couple of commits can be merged without much issue and then each cheat can be considered on a case-by-case basis instead of a big blog that is difficult to review.

libs/s25main/world/GameWorldBase.h
kubaau: I'll leave the unique_ptr. Including Cheats.h from GameWorldBase.h is just a pain and several other members of GameWorldBase are also unique_ptr's (presumably for a similar reason). Please unresolve if you disagree.

libs/s25main/Cheats.h
kubaau: I don't quite see what the benefit of `reference_wrapper` would be here? `world_` is literally meant as a run of the mill C++ reference. Also there is just one use of `reference_wrapper` anywhere in RTTR code (excluding turtle) and it's in s25util where it is inside a `std::optional`, where a plain reference cannot be.

libs/s25main/world/GameWorldBase.h
Flamefire: Ah yes, it's another cyclic dependency here. Moving it out of the world should resolve it.

Flamefire:
> I don't think we should enable cheats in multiplayer, even with addons, I don't see a use case. For balancing reasons we should rather add some real balancing options instead of cheats (which most people are not aware of I think)

I agree. It should only be for debugging/map debugging or some fun, not disturb multi-player games

Flamefire:
I rebased the branch and resolved the conflicts. When this passes CI I'll merge it.

extras/videoDrivers/SDL2/VideoSDL2.cpp
```diff
@@ -332,6 +332,12 @@ bool VideoSDL2::MessageLoop()
                         // Die 12 F-Tasten
                         if(ev.key.keysym.sym >= SDLK_F1 && ev.key.keysym.sym <= SDLK_F12)
                             ke.kt = static_cast<KeyType>(rttr::enum_cast(KeyType::F1) + ev.key.keysym.sym - SDLK_F1);
+
+                        if((SDL_GetModState() & KMOD_ALT) && isdigit(ev.key.keysym.sym))
+                        {
+                            ke.kt = KeyType::Char;
+                            ke.c = ev.key.keysym.sym;
+                        }
```
Flow86: do we need to implement this on the WinAPI too?

libs/s25main/factories/BuildingFactory.h
```diff
@@ -21,5 +21,5 @@ class BuildingFactory
     BuildingFactory() = delete;
 
     static noBuilding* CreateBuilding(GameWorldBase& world, BuildingType type, MapPoint pt, unsigned char player,
-                                      Nation nation);
+                                      Nation nation, bool isTent = false);
```
Flow86: do not use default values, better to fix the places explicitly where its getting used

libs/s25main/network/GameClient.cpp
```diff
@@ -1199,18 +1194,30 @@ void GameClient::DecreaseSpeed()
...
+    const bool debugMode =
+#ifndef NDEBUG
+      true;
+#else
+      false;
+#endif
+    const auto minSpeed = (replayMode || debugMode) ? MIN_SPEED_DEBUG : MIN_SPEED;
+    const auto maxSpeed = (replayMode || debugMode) ? MAX_SPEED_DEBUG : MAX_SPEED;
+    const auto oldSpeed = framesinfo.gfLengthReq;
+    framesinfo.gfLengthReq = helpers::clamp<FramesInfo::milliseconds32_t>(gfLength, maxSpeed, minSpeed);
```
Flow86: unrelated changes?

libs/s25main/world/GameWorldView.cpp
```diff
@@ -356,6 +363,11 @@ void GameWorldView::DrawGUI(const RoadBuildState& rb, const TerrainRenderer& ter
 
 void GameWorldView::DrawNameProductivityOverlay(const TerrainRenderer& terrainRenderer)
 {
+    const bool showEnemyProductivity =
+      GetWorld().GetGameInterface()->GI_GetCheats().shouldShowEnemyProductivityOverlay();
+    auto* attackAidImage =
+      (GetWorld().GetGGS().getSelection(AddonId::MILITARY_AID) == 2) ? LOADER.GetImageN("map_new", 20000) : nullptr;
```
Flow86, line 367: can we please name the `2` and the  `20000` here? we should try to avoid magic numbers as much as possible (at least for the "new" stuff)

libs/s25main/CheatCommandTracker.cpp
```diff
@@ -21,7 +21,9 @@ void CheatCommandTracker::onKeyEvent(const KeyEvent& ke)
     if(!cheats_.areCheatsAllowed())
         return;
 
-    if(ke.kt == KeyType::Char)
+    if(checkSpeedKeyEvent(ke))
+        lastChars_.clear();
```
Flow86: again unrelated?

libs/s25main/CheatCommandTracker.cpp
```diff
@@ -39,18 +41,48 @@ void CheatCommandTracker::onChatCommand(const std::string& cmd)
...
+{
+    const char32_t c = ke.c;
+    if(ke.alt && c >= '1' && c <= '6')
+    {
+        cheats_.setGameSpeed(c - '1');
+        return true;
+    }
+    return false;
+}
+
```
Flow86: do we really need this? I think we have a standard way to change the speed, right?

libs/s25main/Cheats.cpp
```diff
@@ -47,6 +50,49 @@ void Cheats::toggleAllBuildingsEnabled()
...
+
+void Cheats::placeCheatBuilding(const MapPoint& mp, const GamePlayer& player)
+{
+    if(!canPlaceCheatBuilding(mp))
+        return;
+
+    // The new HQ will have default resources.
+    // In the original game, new HQs created in the Roman campaign had no resources.
+    constexpr auto checkExists = false;
+    world_.DestroyNO(mp, checkExists); // if CanPlaceCheatBuilding is true then this must be safe to destroy
```
Flow86: I don't understand what the extra var is for?

libs/s25main/Cheats.cpp
```diff
@@ -56,18 +102,58 @@ void Cheats::toggleHumanAIPlayer()
     }
 }
 
-void Cheats::armageddon() const
+void Cheats::armageddon()
```
Flow86, line 97: why the const removal? its not editing `this`?

libs/s25main/Cheats.cpp
```diff
@@ -56,18 +102,58 @@ void Cheats::toggleHumanAIPlayer()
...
+}
+
+void Cheats::destroyBuildings(const PlayerIDSet& playerIds)
+{
+    if(!isCheatModeOn())
+        return;
+
+    RTTR_FOREACH_PT(MapPoint, world_.GetSize())
+        if(world_.GetNO(pt)->GetType() == NodalObjectType::Building && playerIds.count(world_.GetNode(pt).owner - 1))
+            world_.DestroyNO(pt);
```
Flow86, line 127: please add ` {}` around it, at least for the loop

libs/s25main/Cheats.cpp
```diff
@@ -56,18 +102,58 @@ void Cheats::toggleHumanAIPlayer()
...
+
+void Cheats::destroyAllAIBuildings()
+{
+    if(!isCheatModeOn())
+        return;
+
+    PlayerIDSet ais;
+    for(auto i = 0u; i < world_.GetNumPlayers(); ++i)
+        if(!world_.GetPlayer(i).isHuman())
+            ais.insert(i);
```
Flow86, line 140: same here, please add `{}` for the loop

libs/s25main/GamePlayer.cpp
```diff
@@ -431,8 +445,13 @@ void GamePlayer::AddBuilding(noBuilding* bld, BuildingType bldType)
...
             ship->NewHarborBuilt(static_cast<nobHarborBuilding*>(bld));
     } else if(bldType == BuildingType::Headquarters)
-        hqPos = bld->GetPos();
-    else if(BuildingProperties::IsMilitary(bldType))
+    {
+        // If there is more than one HQ, keep the original position.
+        if(!hqPos.isValid())
+        {
+            hqPos = bld->GetPos();
+        }
```
Flow86: interesting that you added `{}` here :-D

extras/videoDrivers/SDL2/VideoSDL2.cpp
Flamefire: Actually this can be an issue: ALT+0124 results in `|` which is why SDL suppresses it in the first place
Below I work around this by allowing ALT+ASCII letter as a char-key.

WinAPI doesn't has this issue, but probably a different one

libs/s25main/network/GameClient.cpp
Flamefire: `SetNewSpeed` is used by the cheat

libs/s25main/world/GameWorldView.cpp
Flamefire: That's just moved code and not reasonably possible: `GetImageN` is always used with indices and addonselection too or it would requiring including each addon instead of having the generic `getSelection`

We could change that for all places in a separate PR

libs/s25main/CheatCommandTracker.cpp
Flamefire: `checkSpeedKeyEvent` is new. So no

libs/s25main/Cheats.cpp
Flamefire: I guess to give the param a name. I don't mind

libs/s25main/Cheats.cpp
Flamefire: Because it is semantically non-const similar to the other cheat functions

Flamefire:
From [here](https://github.com/Return-To-The-Roots/s25client/pull/1679#discussion_r2712648609)

> do we really need this? I think we have a standard way to change the speed, right?

I agree that we don't really need the ALT+N cheat. As this causes issues with SDL2, or actually entering characters by their number code, I'd remove it.    
To explain: "ALT + 01224" results in `È`. With the "fix" here it would result in `01224È`   
I'm not sure if and how WinAPI is affected. Maybe it doesn't support the ALT+X combos at all. 

Hence I'd remove that.   
I factored out changing speed in fixed steps by using the calculation and clamping in a single function because I think it looks cleaner

libs/s25main/Cheats.cpp
Flow86: semantically doesnt matter - it doenst change the this, so it should be const.

libs/s25main/Cheats.cpp
```diff
@@ -83,7 +83,7 @@ void Cheats::placeCheatBuilding(const MapPoint& mp, const GamePlayer& player)
                                     player.IsHQTent());
 }
 
-void Cheats::setGameSpeed(uint8_t speedIndex)
+void Cheats::setGameSpeed(uint8_t speedIndex) // NOLINT(readability-make-member-function-const)
```
Flow86: please make it const instead of a NOLINT. it doesnt matter that its "semantically" non-const, but technically it should be.

libs/s25main/Cheats.cpp
Flow86: the comment already explains everything, so simplify it

libs/s25main/Cheats.cpp
Flamefire: Why?

It could only technically be made const because of a design flaw. (Almost) all cheats act on `world` and naturally won't work in const methods because they do modify the referenced object. Using cheats is a modifying operation and using const methods for queries only helps catching bugs. Only here we happen to use a global, at least now.

We could store a `GameClient&` then we don't need the NOLINT but IMO don't gain anything.

Flamefire:
About the reveal resources: Currently it is additive: When showing water it will also show fish and ores.

To me it looks better code-wise if it really toggles, i.e. shows ores OR water OR fish.   
But doesn't matter much.

Second ideas? @Flow86 @Spikeone

Flow86:
in the original it also everything was visible with the cheat, wasnt it? so incrementally adding it is also fine

Flamefire:
I think the original does not have that cheat
At least it is listed as new.

I'm inclined to keep it as the author intended. Maybe rename the enum to reflect it, like "Ore" and "OreWaterFish" to avoid confusion reading only one part

Issue #1177, opened 2020-01-14
[LUA] Missing features for the original campaign


Spikeone:
This issue links all issues / features that need to be fixed for the original campaign
- [ ] #743 - AI agression level
- [ ] #568 - Campaign Screen
- [ ] #1175 - MET_LAND_SIZE
- [ ] #1176 - Change agression level on attack

Spikeone:
I've uploaded the updated (and original) map files and the lua scripts here: https://github.com/Spikeone/RttR_Campaigns

Issue #743, opened 2017-11-10
[LUA] one Sided alliances


Spikeone:
The original game allowed one sided alliances. 

What does it mean? Well, it disallows one player to attack an other player as they are allied. While player A can't attack player B, player B can attack player A. Also catapults do not break this but I think that was a bug in the original.

![image](https://user-images.githubusercontent.com/1276342/32668920-2c1fc8c0-c63f-11e7-8f5e-ac977b1d16d4.png)

So in this example:
player A = Player, can attack player B and C
player B = Computer, can attack player C
player C = Computer, can't attack anyone.

What is this needed for? Well for the original missions. The AI could expand as far as the AI liked to but could not attack the player unless the player attacked the AI first. That way an AI was just blocking but not agressive. 

I do not think we should implement the original bug, that you could use catapults without breaking the one sided pact as attacking results in a broken alliance, catapults did not.

Spikeone:
Well, it's not that important how it works internally, but the player should be able to attack to cancel the pact for the AI, maybe an other version as the non agressive pact which allows attacking to cancel?

Flamefire:
Wait. A pact that disallows attacking unless someone is attacking?
Nope...

Spikeone:
Well, for one player - but if the AI coulld be non agressive towards the player until the AI is atfacked, that would solve the problem as well

Flamefire:
Fixed in #773

Spikeone:
@Flamefire this still does not allow the following:
In the 2nd Mission, the AI player is friendly towards the player unless he attacks. This gives the player time to do whatever he wants and attack whenever he wants to.
Regarding the pact logic we currently have, it is not possible to do so as well, is it?

I think the main idea behind those pacts aren't infact pacts, but merely a "do not attack until attacked". Can we implement such option for the AI? So we may set right at the start rttr:GetPlayer(1):SetDefensiveLevel(1)
Where 0 = normal (as is), 1 = don't attack until attack (1 sided pacts) and 2 = do not attack?

That way we can leave pacts as they are but have the same outcome.

Flamefire:
Well the sequence would be: Player and AI have pact so none can attack. Player sends cancel pact to AI, AI is set to accept, player attacks AI. This is mostly the same, isn't it? Only difference is that the trigger is a cancel pact request instead of an attack.

If you'd really want that nicer behavior we'd need better names. And I'd suggest a target player to:

[...]

PR #1680, opened 2024-07-23
Add support for one-sided alliances

kubaau:
Adds supported for one-sided alliances exactly (I hope) how they worked in S2. This makes it possible to achieve the same alliance logic as in the campaigns designed for S2 (tested with the Roman campaign and the FANpaign). FANpaign in particular makes almost no sense to play without such alliances being possible. See [video1](https://streamable.com/tjl4t7) and [video2](https://streamable.com/a1ahz8) for a short demonstration.

Adds an addon for catapults to attack allies, which was the case in S2 (defaults to disabled in RttR). This also makes old campaigns more viable since many scenarios rely on a slow catapult creep without starting a war.

These should probably be split into 2 PR's, but I am modifying campaign scripts in both so I worked on one branch and the addon does not make much sense without modifying campaign alliances.

Related to issues #743 and #1177


Spikeone:
Great work! Just two questions: Does this work in multiplayer? Is it possible to break such pact via diplomacy window as well?

kubaau:
> Great work! Just two questions: Does this work in multiplayer? Is it possible to break such pact via diplomacy window as well?

This type of pact does not appear in the diplomacy window, so you cannot suggest or break such an alliance this way.

I was designing this with only single player in mind, thus the human can always attack (even their allies), as it was in S2. As it stands the pact does not make sense in multiplayer unless AIs are also involved and can only be made through scripts. I think it could be made to work with multiplayer by adding one additional if statement in `bool GamePlayer::CanAttack(const unsigned char otherPlayerId) const`, I just didn't think it through for multiplayer other than disallowing suggesting such a pact.

tests/s25Main/integration/testPacts.cpp
```diff
@@ -261,4 +261,196 @@ BOOST_FIXTURE_TEST_CASE(SuggestNewPactAfterExpiration, PactCreatedFixture,
...
+    // alliance from p0 to p2 remains
+    BOOST_TEST_REQUIRE(p0.CanAttack(1) == true);
+    BOOST_TEST_REQUIRE(p0.CanAttack(2) == false);
+    BOOST_TEST_REQUIRE(p1.CanAttack(0) == false);
+    BOOST_TEST_REQUIRE(p1.CanAttack(2) == true);
+    BOOST_TEST_REQUIRE(p2.CanAttack(0) == false);
+    BOOST_TEST_REQUIRE(p2.CanAttack(1) == true);
+}
+
+BOOST_FIXTURE_TEST_CASE(OneSidedAlliances_BeingAttacked_DoesNotAlsoBreakAlliancesForPlayers_WhoWeAreNotAlliedTo,
```
Flamefire, line 422: I guess this needs resolving the definition of "allied to".

Say player X attacks player Y. Then of course player Y breaks the pact to player X.

And I guess every player Z allied to player Y should also break the pact to player X. Only question is whether this Z->Y should include one-sided pacts and if so which way: Z->Y, Y->Z, both or either. I would argue only Z->Y should be checked as Z "doesn't know" about Y->Z so why should it care about Y being attacked in that case?

Might also be worth merging with the above test case to focus on the differences and add the "DoesNotAlsoBreakAlliancesForPlayers_WhoWeAreNotAlliedTo" as a comment instead

Flamefire (changes requested):
This might be a bit harder than it looks. Especially the definition of "allied to" isn't that clear any more. We might need to check where and how this is used.

Issue #1639, opened 2023-11-14
Add support for campaign status


ottml:
- [ ] Store/Load campaign status in/from settings (mission playable/finished)
- [ ] Api for querying and setting status of a campaign
`isFinished(UID,missionId)
setFinished(UID,missionId)`
- [ ] Add unique identifier for each campaign to identify it in the api uniquely
- [ ] Add unit-tests

Spikeone:
Just thinking, isn't the UID (guess thats the campaign ID?) the folder name? Since they are unique, otherwise you'd overwrite an existing campaign. That'd also allows easy manual unlocking of missions (e.g. you don't want to play mission 3 of the roman campaign).

I think we should allow campaign creators to unlock missions by Id (as you mentioned it I guess), since that'd allow branching stories, e.g. if you decide to defeat player 2 in mission 2, you unlock mission 4 instead of 3, while defeating player 3 would unlock mission 3. Alas this is currently not required and maybe is never, I still like having the option :)

Flamefire:
The API looks good. Question is only how to get the UID. I guess as you only need them from within a campaign it would be great to not having to specify it at all but make the C++ code calculate it, e.g. a hash of the path to the campaign description. But I'm not sure how'd we be able to get that from inside a running game or a savegame.

So "missionId" = index of the mission in the campaign, and UID is what we need to determine somehow. If it needs to be passed by the LUA script, it can't be the folder name as a campaign might reside in different folders (for different users) and still be the same (e.g. "world" and "world-testing-for-dev"). So how would a lua script know about it?

Spikeone:
> So "missionId" = index of the mission in the campaign, and UID is what we need to determine somehow. If it needs to be passed by the LUA script, it can't be the folder name as a campaign might reside in different folders (for different users) and still be the same (e.g. "world" and "world-testing-for-dev"). So how would a lua script know about it?

But does that really matter? Since they are two different campaigns at all, they should have different IDs as well or am I missing something?

Flamefire:
That is exactly my point. The lua script would call `isFinished("world", 2)` not knowing it is modifying a potentially different campaign status

ottml:
> Just thinking, isn't the UID (guess thats the campaign ID?) the folder name? Since they are unique, otherwise you'd overwrite an existing campaign. That'd also allows easy manual unlocking of missions (e.g. you don't want to play mission 3 of the roman campaign).
> 
[...]

PR #1681, opened 2024-07-23
Add support for campaign status

kubaau:
Adds support for Roman-style ([video](https://streamable.com/6zo42j)) campaigns and the world campaign ([video](https://streamable.com/4ojiaj)).

Fixed world campaign scripts to unlock the same continents as it was in S2.

Related to issues #1639 and #1177

### DIFF
--- a/libs/libGamedata/gameData/CampaignDescription.cpp
+++ b/libs/libGamedata/gameData/CampaignDescription.cpp
@@ -3,6 +3,7 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 
 #include "CampaignDescription.h"
+#include "CampaignSaveCodes.h"
 #include "RttrConfig.h"
 #include "helpers/format.hpp"
 #include "lua/CheckedLuaTable.h"
@@ -12,6 +13,7 @@
 CampaignDescription::CampaignDescription(const kaguya::LuaRef& table)
 {
     CheckedLuaTable luaData(table);
+    luaData.getOrThrow(uid, "uid");
     luaData.getOrThrow(version, "version");
     luaData.getOrThrow(author, "author");
     luaData.getOrThrow(name, "name");
@@ -33,6 +35,8 @@ CampaignDescription::CampaignDescription(const kaguya::LuaRef& table)
     lua::validatePath(mapFolder);
     lua::validatePath(luaFolder);
     mapNames = luaData.getOrDefault("maps", std::vector<std::string>());
+    defaultChaptersEnabled =
+      luaData.getOrDefault("defaultChaptersEnabled", std::string{CampaignSaveCodes::defaultChaptersEnabled});
     selectionMapData = luaData.getOptional<SelectionMapInputData>("selectionMap");
     luaData.checkUnused();
 }
--- a/libs/s25main/desktops/dskCampaignVictory.cpp
+++ b/libs/s25main/desktops/dskCampaignVictory.cpp
@@ -1,0 +1,39 @@
+// Copyright (C) 2024 Settlers Freaks (sf-team at siedler25.org)
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#include "dskCampaignVictory.h"
+#include "Loader.h"
+#include "WindowManager.h"
+#include "desktops/dskMainMenu.h"
+#include "network/GameClient.h"
+
+dskCampaignVictory::dskCampaignVictory(unsigned char chapter)
+    : Desktop(LOADER.GetImageN(ResourceId{chapter ? "setup896" : "setup895"}, 0))
+{
+    GAMECLIENT.SetCampaignChapterCompleted(0);
+    GAMECLIENT.SetCampaignCompleted(false);
+
+    if(!chapter)
+        return;
+
+    AddText(10, DrawPoint{800 / 2, 600 - 50},
+            _("You have successfully completed chapter") + std::string{" "} + std::to_string(chapter) + ".",
+            COLOR_YELLOW, FontStyle::CENTER, LargeFont);
+}
+
+bool dskCampaignVictory::Msg_LeftDown(const MouseCoords&)
+{
+    return ShowMenu();
+}
+
+bool dskCampaignVictory::Msg_KeyDown(const KeyEvent&)
+{
+    return ShowMenu();
+}
+
+bool dskCampaignVictory::ShowMenu()
+{
+    WINDOWMANAGER.Switch(std::make_unique<dskMainMenu>());
+    return true;
+}
--- a/libs/s25main/lua/LuaInterfaceGame.cpp
+++ b/libs/s25main/lua/LuaInterfaceGame.cpp
@@ -5,6 +5,7 @@
 #include "LuaInterfaceGame.h"
 #include "EventManager.h"
 #include "Game.h"
+#include "Settings.h"
 #include "WindowManager.h"
 #include "ai/AIInterface.h"
 #include "ai/AIPlayer.h"
@@ -12,9 +13,11 @@
 #include "lua/LuaHelpers.h"
 #include "lua/LuaPlayer.h"
 #include "lua/LuaWorld.h"
+#include "network/GameClient.h"
 #include "postSystem/PostMsg.h"
 #include "world/GameWorld.h"
 #include "gameTypes/Resource.h"
+#include "gameData/CampaignSaveCodes.h"
 #include "s25util/Serializer.h"
 #include "s25util/strAlgos.h"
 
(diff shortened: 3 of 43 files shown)


Spikeone:
Either I'm blind or the implementation is somewhere else? What is ```rttr:SetCampaignChapterCompleted``` good for?

kubaau:
> Either I'm blind or the implementation is somewhere else? What is `rttr:SetCampaignChapterCompleted` good for?

It doesn't make a difference in Roman-style campaigns, but it marks a continent as conquered in the world campaign. It could also be used for other campaign types as well if we wanted to show what was completed, especially if someone designs a campaign whose progression is not linear.

libs/s25main/lua/LuaInterfaceGame.cpp
```diff
@@ -313,6 +319,37 @@ void LuaInterfaceGame::PostMessageWithLocation(int playerIdx, const std::string&
...
+    if(saveData.length() < chapter)
+        saveData.resize(chapter);
+
+    saveData[chapter - 1] = code;
+}
+} // namespace
+
+void LuaInterfaceGame::EnableCampaignChapter(const std::string& campaignUid, unsigned char chapter)
+{
+    MarkCampaignChapter(campaignUid, chapter, CampaignSaveCodes::chapterEnabled);
```
kubaau, line 339: This should check if the campaign is not already completed. Otherwise a continent that is conquered can be marked as just enabled (not conquered) by conquering a different continent. Probably this check should be in `MarkCampaignChapter` (cannot set status to "less" than before).

libs/s25main/desktops/dskCampaignVictory.cpp
```diff
@@ -0,0 +1,39 @@
...
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#include "dskCampaignVictory.h"
+#include "Loader.h"
+#include "WindowManager.h"
+#include "desktops/dskMainMenu.h"
+#include "network/GameClient.h"
+
+dskCampaignVictory::dskCampaignVictory(unsigned char chapter)
+    : Desktop(LOADER.GetImageN(ResourceId{chapter ? "setup896" : "setup895"}, 0))
```
Flamefire, line 12: What is zero? It should never be zero shouldn't it? If it can please add a comment here or at the return below

libs/s25main/lua/LuaInterfaceGame.cpp
```diff
@@ -313,6 +319,42 @@ void LuaInterfaceGame::PostMessageWithLocation(int playerIdx, const std::string&
                                                       gw.MakeMapPoint(Position(x, y))));
 }
 
+namespace {
+void MarkCampaignChapter(const std::string& campaignUid, unsigned char chapter, char code)
```
Flamefire, line 323: Looking at this here the "code" should really be an enum to make it clearer

libs/s25main/lua/LuaInterfaceGame.cpp
```diff
@@ -313,6 +319,42 @@ void LuaInterfaceGame::PostMessageWithLocation(int playerIdx, const std::string&
...
+
+    if(saveData.length() < chapter)
+        saveData.resize(chapter);
+
+    auto& chapterSaveData = saveData[chapter - 1];
+
+    if(chapterSaveData == CampaignSaveCodes::chapterCompleted)
+        return;
+
+    chapterSaveData = code;
```
Flamefire, line 338: This is more clear than an "early" return

```suggestion
    if(chapterSaveData != CampaignSaveCodes::chapterCompleted)
        chapterSaveData = code;
```

tests/s25Main/campaign/testCampaignSaveData.cpp
```diff
@@ -0,0 +1,102 @@
...
+    saveData[uid] = "011";
+    BOOST_TEST_REQUIRE(isChapterEnabled(desc, 0) == false);
+    BOOST_TEST_REQUIRE(isChapterEnabled(desc, 1) == true);
+    BOOST_TEST_REQUIRE(isChapterEnabled(desc, 2) == true);
+    BOOST_TEST_REQUIRE(isChapterEnabled(desc, 9) == false);
+}
+
+BOOST_FIXTURE_TEST_CASE(Chapters0and1AreEnabledByDefault, CampaignSaveDataFixture)
+{
+    saveData[uid] = CampaignSaveCodes::defaultChaptersEnabled;
```
Flamefire, line 48: So the default is to enable the first 2 chapters? Why that? Wouldn't (only) the first one be a more sensible default?
I'm also a bit confused as you seem to be using indices **and** 1-based values for chapters

Maybe stick to indices and use an optional instead of zero

tests/s25Main/campaign/testCampaignSaveData.cpp
```diff
@@ -0,0 +1,102 @@
...
+    BOOST_TEST_REQUIRE(isChapterEnabled(desc, 9) == false);
+}
+
+BOOST_FIXTURE_TEST_CASE(DefaultChaptersCanBeSetPerCampaign, CampaignSaveDataFixture)
+{
+    desc.defaultChaptersEnabled = "100";
+    BOOST_TEST_REQUIRE(isChapterEnabled(desc, 0) == true);
+    BOOST_TEST_REQUIRE(isChapterEnabled(desc, 1) == false);
+    BOOST_TEST_REQUIRE(isChapterEnabled(desc, 2) == false);
+    BOOST_TEST_REQUIRE(isChapterEnabled(desc, 9) == false);
```
Flamefire, line 61: This should also include a change to `saveData` to check that the default really is only a default and overwritten by an explicit value.

tests/s25Main/campaign/testCampaignSaveData.cpp
```diff
@@ -0,0 +1,102 @@
...
+            mapFolder = \"<RTTR_GAME>/DATA/MAPS\",\
+            luaFolder = \"<RTTR_GAME>/CAMPAIGNS/ROMAN\",\
+            maps = { \"dessert0.WLD\", \"dessert1.WLD\", \"dessert2.WLD\"}\
+        }";
+
+        file << "function getRequiredLuaVersion() return 1 end";
+    }
+
+    CampaignDescription dsc;
+    CampaignDataLoader loader{dsc, tmp};
```
Flamefire, line 91: Why can't this just use `desc` from the fixture?

tests/s25Main/lua/testLua.cpp
```diff
@@ -843,6 +844,21 @@ BOOST_AUTO_TEST_CASE(onExplored)
...
+    initWorld();
+
+    SETTINGS.campaigns.saveData["campaign_id"] = "110";
+    executeLua("rttr:SetCampaignChapterCompleted('campaign_id', 2)");
+    executeLua("rttr:SetCampaignChapterCompleted('campaign_id', 4)");
+    executeLua("rttr:EnableCampaignChapter('campaign_id', 5)");
+    executeLua("rttr:SetCampaignChapterCompleted('campaign_id', 0)"); // noop - chapters start from 1
+    executeLua("rttr:EnableCampaignChapter('campaign_id', 0)");       // noop - chapters start from 1
+    executeLua("rttr:EnableCampaignChapter('campaign_id', 2)");       // noop - already completed
+    game.executeAICommands();
```
Flamefire, line 858: What is this line for?

libs/libGamedata/gameData/CampaignDescription.cpp
```diff
@@ -12,6 +13,7 @@
 CampaignDescription::CampaignDescription(const kaguya::LuaRef& table)
 {
     CheckedLuaTable luaData(table);
+    luaData.getOrThrow(uid, "uid");
```
Flamefire, line 16: For this you need to  Increase `CampaignDataLoader::GetVersion()`

I'd try to make this backwards compatible: and make the uid optional. Or check the version here (again) and make sure the uid is empty for older versions and non-empty otherwise.

Then you can revert the changes to existing tests

kubaau:
Rework in progress - converting to draft, will notify when ready for review. Resolving some conversations since they will not be relevant.

tests/s25Main/lua/testLua.cpp
kubaau: Neither this not initWorld are needed, some leftover garbage most likely.

tests/s25Main/campaign/testCampaignSaveData.cpp
Hirotaro: In the original S2, I also had first 2 missions unlocked at start. Maybe because 1st one is actually a tutorial without AI opponent?

Just rechecked with GOG S2 Gold Edition. 
obraz

PR #1682, opened 2024-07-23
Add classic campaigns straight to single player screen

kubaau:
Adds the Roman and world campaigns straight to the single player screen in order not to have to go through the campaign selection screen. Useful for testing and provides a look that is almost identical to what the S2 main screen offered.

Zrzut ekranu 2024-07-23 124955
Zrzut ekranu 2024-07-23 130747

### DIFF
--- a/libs/s25main/QuickStartGame.cpp
+++ b/libs/s25main/QuickStartGame.cpp
@@ -80,7 +80,7 @@ bool QuickStartGame(const boost::filesystem::path& mapOrReplayPath, const std::v
     }
 
     const CreateServerInfo csi(isSinglePlayer ? ServerType::Local : ServerType::Direct, SETTINGS.server.localPort,
-                               _("Unlimited Play"));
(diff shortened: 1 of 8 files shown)


Flow86:
Hi, 
I appreciate the effort,
but:
to be honest, I'm unsure if I like this change or not. 

On one side we want to be "a clone" of the old game, but on the other hand we do want to support (all) the campaign out there aswell. 

It even makes it a bit hard to distinguish where to start a campaign etc. I think that was the idea of the campaign screen. 

I do not have a good idea to clean the menu up in a way I assume "its right" though. 

Lets check what other say (e.g @Flamefire )

kubaau:
I'm not surprised. I'm not sure about this change either :D To be honest the original campaign suck when compared to what others did and what could theoretically be possible, so it doesn't make much sense to highlight them as if they are something special. We can drop this PR, I'll just keep it in my fork because it saves me some frustration while testing the campaigns.

PR #1683, opened 2024-07-23
Fix map "The snake" and enable overriding crucial map settings from LUA

kubaau:
Fixes map "The snake" WITHOUT changing the map. You can now use the original S2 map, but I needed to implement some additional features, like placing HQs scripts and overriding the number of players from LUA scripts. For example Roman chapter 1 does not show 7 players anymore and chapter 7 has the player 3 HQ set from LUA ([video](https://streamable.com/1m826i)). This also provides makes it possible to fix all other maps where the positions are not set in the map but only in RTX files (like half the FANpaign maps - [example video](https://streamable.com/9m3a82)).

### DIFF
--- a/data/RTTR/campaigns/roman/MISS206.lua
+++ b/data/RTTR/campaigns/roman/MISS206.lua
@@ -218,9 +218,12 @@ function addPlayerBld(p, onLoad)
             nil, nil
         )
 
-        if(p == 2) then
-            if onLoad then return end
+        if onLoad then return end
 
+        if(p == 1) then
+            rttr:GetPlayer(p):PlaceHQ(112, 82)  -- !SET_HOUSE 24, 112, 82
+        elseif(p == 2) then
+            rttr:GetPlayer(p):PlaceHQ(40, 54)   -- !SET_HOUSE 24, 40, 54
             rttr:GetPlayer(p):AIConstructionOrder(43, 59, BLD_FORTRESS)
         end
     end
--- a/libs/s25main/GamePlayer.cpp
+++ b/libs/s25main/GamePlayer.cpp
@@ -13,6 +13,7 @@
 #include "Ware.h"
 #include "addons/const_addons.h"
 #include "buildings/noBuildingSite.h"
+#include "buildings/nobHQ.h"
 #include "buildings/nobHarborBuilding.h"
 #include "buildings/nobMilitary.h"
 #include "buildings/nobUsual.h"
@@ -374,6 +375,19 @@ void GamePlayer::RemoveBuildingSite(noBuildingSite* bldSite)
 {
     RTTR_Assert(bldSite->GetPlayer() == GetPlayerId());
     buildings.Remove(bldSite);
+}
+
+bool GamePlayer::IsHQTent() const
+{
+    if(const nobHQ* hq = GetHQ())
+        return hq->IsTent();
+    return false;
+}
+
+void GamePlayer::SetHQIsTent(bool isTent)
+{
+    if(nobHQ* hq = GetHQ())
+        hq->SetIsTent(isTent);
 }
 
 void GamePlayer::AddBuilding(noBuilding* bld, BuildingType bldType)
@@ -1363,6 +1377,12 @@ void GamePlayer::TestDefeat()
     // Keine Militärgebäude, keine Lagerhäuser (HQ,Häfen) -> kein Land --> verloren
     if(!isDefeated && buildings.GetMilitaryBuildings().empty() && buildings.GetStorehouses().empty())
         Surrender();
+}
+
+nobHQ* GamePlayer::GetHQ() const
+{
+    const MapPoint& hqPos = GetHQPos();
+    return const_cast<nobHQ*>(hqPos.isValid() ? GetGameWorld().GetSpecObj<nobHQ>(hqPos) : nullptr);
 }
 
 void GamePlayer::Surrender()
--- a/libs/s25main/desktops/dskGameLobby.cpp
+++ b/libs/s25main/desktops/dskGameLobby.cpp
@@ -32,6 +32,7 @@
 #include "ingameWindows/iwMsgbox.h"
 #include "lua/LuaInterfaceSettings.h"
 #include "network/GameClient.h"
+#include "network/GameServer.h"
 #include "ogl/FontStyle.h"
 #include "gameData/GameConsts.h"
 #include "gameData/const_gui_ids.h"
@@ -99,6 +100,13 @@ dskGameLobby::dskGameLobby(ServerType serverType, std::shared_ptr<GameLobby> gam
               gameLobby_->isHost()); // This should be done first for the host so others won't even see the script
             LOG.write(_("Lua was disabled by the script itself\n"));
             lua.reset();
+        } else
+        {
+            if(const auto num = lua->GetNumPlayersFromScript())
+            {
+                GAMESERVER.SetNumPlayers(num);
+                gameLobby_->setNumPlayers(num);
+            }
         }
         if(!lua)
             lobbyHostController->RemoveLuaScript();
--- a/libs/s25main/lua/LuaInterfaceSettings.cpp
+++ b/libs/s25main/lua/LuaInterfaceSettings.cpp
@@ -247,3 +247,13 @@ bool LuaInterfaceSettings::IsMapPreviewEnabled()
     }
     return true;
 }
+
+unsigned LuaInterfaceSettings::GetNumPlayersFromScript()
+{
+    kaguya::LuaRef func = lua["getNumPlayers"];
+    if(func.type() == LUA_TFUNCTION)
+    {
+        return func.call<unsigned>();
+    }
+    return 0;
+}
--- a/libs/s25main/lua/LuaInterfaceSettings.h
+++ b/libs/s25main/lua/LuaInterfaceSettings.h
@@ -36,6 +36,7 @@ public:
     /// Get addons that are allowed to be changed
     std::vector<AddonId> GetAllowedAddons();
     bool IsMapPreviewEnabled();
+    unsigned GetNumPlayersFromScript();
 
 private:
     IGameLobbyController& lobbyServerController_;
--- a/libs/s25main/lua/LuaPlayer.cpp
+++ b/libs/s25main/lua/LuaPlayer.cpp
@@ -9,6 +9,7 @@
 #include "ai/AIPlayer.h"
 #include "buildings/nobBaseWarehouse.h"
 #include "buildings/nobHQ.h"
+#include "factories/BuildingFactory.h"
 #include "helpers/EnumRange.h"
 #include "helpers/toString.h"
 #include "lua/LuaHelpers.h"
@@ -46,6 +47,7 @@ void LuaPlayer::Register(kaguya::State& state)
                                .addFunction("GetNumPeople", &LuaPlayer::GetNumPeople)
                                .addFunction("GetStatisticsValue", &LuaPlayer::GetStatisticsValue)
                                .addFunction("AIConstructionOrder", &LuaPlayer::AIConstructionOrder)
+                               .addFunction("PlaceHQ", &LuaPlayer::PlaceHQ)
                                .addFunction("ModifyHQ", &LuaPlayer::ModifyHQ)
                                .addFunction("GetHQPos", &LuaPlayer::GetHQPos)
                                .addFunction("IsDefeated", &LuaPlayer::IsDefeated)
@@ -259,15 +261,22 @@ bool LuaPlayer::AIConstructionOrder(unsigned x, unsigned y, lua::SafeEnum<Buildi
     return true;
 }
 
+void LuaPlayer::PlaceHQ(MapCoord x, MapCoord y)
+{
+    // Ignore if there is an HQ set in the map file.
+    if(player.GetHQPos().isValid())
+        return;
+
+    GameWorld& world = player.GetGameWorld();
+    const MapPoint mp{x, y};
+    constexpr auto checkExists = false;
+    world.DestroyNO(mp, checkExists);
+    BuildingFactory::CreateBuilding(world, BuildingType::Headquarters, mp, player.GetPlayerId(), player.nation);
+}
+
 void LuaPlayer::ModifyHQ(bool isTent)
 {
-    const MapPoint hqPos = player.GetHQPos();
-    if(hqPos.isValid())
-    {
-        auto* hq = player.GetGameWorld().GetSpecObj<nobHQ>(hqPos);
-        if(hq)
-            hq->SetIsTent(isTent);
-    }
+    player.SetHQIsTent(isTent);
 }
 
 bool LuaPlayer::IsDefeated() const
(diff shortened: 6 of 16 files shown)


data/RTTR/campaigns/roman/MISS206.lua
```diff
@@ -218,9 +218,12 @@ function addPlayerBld(p, onLoad)
             nil, nil
         )
 
-        if(p == 2) then
-            if onLoad then return end
+        if onLoad then return end
 
+        if(p == 1) then
+            rttr:GetPlayer(p):PlaceHQ(112, 82)  -- !SET_HOUSE 24, 112, 82
```
Spikeone, line 224: just to make sure, if someone calls this multiple times for the same player, or even if loading, the game doesn't break does it?

Spikeone: And another question, does this work in multiplayer as well? Since I really love this feature - one could create a map with 10 possible starting points, assigning them "randomly" (not random but by some kind of algorithm to use a spot for a player) or in a campaign where enemies can come from different sides, each time you play the mission

Spikeone:
This is gerat, but I'd still like to have some feature to "fix" map 6, where they failed when converting from WORLD.DAT to WLD :D

data/RTTR/campaigns/roman/MISS206.lua
kubaau: The PlaceHQ call is ignored if there is already an HQ. The map data takes precedence because it is loaded before any of these PlaceHQ calls are made. If you put two HQs for a player from LUA then only the first one will be placed. This needs documenting.

I didn't test multiplayer, but it would only work with "broken" maps, because it is designed for "broken" maps :) You would have to create a map with missing player positions and then it should work, but I didn't check. For the use case you mentioned I think a different function should be designed, such as `AssignHQToPlayer` or something. Then the map would (correctly) define positions, but LUA would shuffle them around players.

kubaau:
> This is gerat, but I'd still like to have some feature to "fix" map 6, where they failed when converting from WORLD.DAT to WLD :D

What do you mean by this?

Spikeone:
> > This is gerat, but I'd still like to have some feature to "fix" map 6, where they failed when converting from WORLD.DAT to WLD :D
> 
> What do you mean by this?

On map IX they changed the map between the original version and gold/mission CD. For one instance they removed the harbors completly, this means there is no more seafaring. They also changed the terrain for the second AI, which had nearly no way of farming or producing wood since they used a non farmable terrain which simply stopped that AI from doing anything usefull.

You can compare that version: [Campaign_OriginalRoman/MISS208.WLD](https://github.com/Spikeone/RttR_Campaigns/blob/master/Campaign_OriginalRoman/MISS208.WLD) with the one in your game directory, I've created it by reading the WORLD.DAT file of the original campaign and replacing the terrain with the correct one and re adding the harbors

kubaau:
> On map IX

Ah yes, that one. Yeah this one is fucked in mission CD/gold. The RTX file is also wrong in a couple of places resulting in `NO STRING WITH THIS ID` errors and instead of placing an activated gate in the right place they place a rock in the middle of water north of your HQ... I think the only way to handle being given the original file by the user is to add code that detects that this is indeed The Gray Island and adds/modifies some data it reads internally (I don't know if we can add such a modified WLD file in the game files itself?).

If you create a ticket for this I would very much like to take a look at this once I finish all the stuff I currently have open. While this does actually change an original map (which I don't like in principle), I think it is safe to assume the map was not supposed to be played as it was post mission CD.

libs/s25main/network/GameClient.cpp
```diff
@@ -323,8 +323,8 @@ void GameClient::StartGame(const unsigned random_init)
             gameWorld.GetPlayer(i).MakeStartPacts();
 
         MapLoader loader(gameWorld);
-        if(!loader.Load(mapinfo.filepath)
-           || (!mapinfo.luaFilepath.empty() && !loader.LoadLuaScript(*game, *this, mapinfo.luaFilepath)))
+        if((!mapinfo.luaFilepath.empty() && !loader.LoadLuaScript(*game, *this, mapinfo.luaFilepath))
+           || !loader.Load(mapinfo.filepath)) // do not reorder: load lua first, load map second
```
Flamefire, line 327: Can you include the reason  in the comment and not only what you did?

libs/s25main/world/MapLoader.cpp
```diff
@@ -421,8 +421,11 @@ bool MapLoader::PlaceHQs(GameWorldBase& world, std::vector<MapPoint> hqPositions
         // Does the HQ have a position?
         if(i >= hqPositions.size() || !hqPositions[i].isValid())
         {
-            LOG.write(_("Player %u does not have a valid start position!")) % i;
-            return false;
+            LOG.write(_("Player %u does not have a valid start position!\n")) % i;
+            if(world.HasLua()) // maybe the HQ is placed in the script?
+                continue;
+            else
+                return false;
```
Flamefire, line 428:
```suggestion
            if(!world.HasLua()) // HQ can be placed in the script, so don't error
                return false;
```

Flamefire (changes requested):
I hope this does not break something later on, but looks good for now :)

> If you put two HQs for a player from LUA then only the first one will be placed. This needs documenting.

Indeed.

libs/s25main/GamePlayer.cpp
```diff
@@ -1365,6 +1379,12 @@ void GamePlayer::TestDefeat()
         Surrender();
 }
 
+nobHQ* GamePlayer::GetHQ() const
+{
+    const MapPoint& hqPos = GetHQPos();
+    return const_cast<nobHQ*>(hqPos.isValid() ? GetGameWorld().GetSpecObj<nobHQ>(hqPos) : nullptr);
```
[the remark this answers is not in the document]
kubaau: Moved the const_cast to the client requiring this.

libs/s25main/GamePlayer.cpp
```diff
@@ -376,6 +377,19 @@ void GamePlayer::RemoveBuildingSite(noBuildingSite* bldSite)
     buildings.Remove(bldSite);
 }
 
+bool GamePlayer::IsHQTent() const
```
Flamefire, line 380: What is the reason for adding this here? It isn't used (yet)

libs/s25main/GamePlayer.cpp
```diff
@@ -376,6 +377,19 @@ void GamePlayer::RemoveBuildingSite(noBuildingSite* bldSite)
...
+{
+    if(const nobHQ* hq = GetHQ())
+        return hq->IsTent();
+    return false;
+}
+
+void GamePlayer::SetHQIsTent(bool isTent)
```
Flamefire, line 387: This and `GetHQ` is basically the code moved from `ModifyHQ`. What was the motivation?

I'm thinking about the case where there are multiple HQs (e.g. the S2 cheats had it) and one might want to specify which HQ to change. Putting the code in the caller uses the world to get a specific HQ and changes that.

Although we already have `GetHQPos` which returns the position of "a" HQ, IIRC the last one added that still exists.

libs/s25main/GamePlayer.cpp
```diff
@@ -376,6 +377,19 @@ void GamePlayer::RemoveBuildingSite(noBuildingSite* bldSite)
...
+{
+    if(nobHQ* hq = const_cast<nobHQ*>(GetHQ()))
```
Flamefire, line 389: I missed that `GetSpecObj` (in `GetHQ`) can return a non-const pointer. So we don't need the const cast at all.

If you make `GetHQ` non-const then this can be removed

libs/s25main/desktops/dskGameLobby.cpp
```diff
@@ -99,6 +100,13 @@ dskGameLobby::dskGameLobby(ServerType serverType, std::shared_ptr<GameLobby> gam
               gameLobby_->isHost()); // This should be done first for the host so others won't even see the script
             LOG.write(_("Lua was disabled by the script itself\n"));
             lua.reset();
+        } else
+        {
+            if(const auto num = lua->GetNumPlayersFromScript())
+            {
+                GAMESERVER.SetNumPlayers(num);
```
Flamefire, line 107: This won't work for savegames, will it? Because I think for a savegame the player needs to be already initialized.  
Or does that work still?

libs/s25main/lua/LuaInterfaceSettings.cpp
```diff
@@ -247,3 +247,13 @@ bool LuaInterfaceSettings::IsMapPreviewEnabled()
     }
     return true;
 }
+
+unsigned LuaInterfaceSettings::GetNumPlayersFromScript()
+{
+    kaguya::LuaRef func = lua["getNumPlayers"];
```
Flamefire, line 253: Please add documentation for this new function.

libs/s25main/lua/LuaInterfaceSettings.h
```diff
@@ -36,6 +36,7 @@ class LuaInterfaceSettings : public LuaInterfaceGameBase
     /// Get addons that are allowed to be changed
     std::vector<AddonId> GetAllowedAddons();
     bool IsMapPreviewEnabled();
+    unsigned GetNumPlayersFromScript();
```
Flamefire, line 39:
```suggestion
    /// Get number of players as set in the script. Return zero to use the value set in the map/save file.
    unsigned GetNumPlayersFromScript();
```

Or maybe returning an optional is more clear and disallow zero for lua files.

libs/s25main/lua/LuaPlayer.cpp
```diff
@@ -259,15 +261,22 @@ bool LuaPlayer::AIConstructionOrder(unsigned x, unsigned y, lua::SafeEnum<Buildi
     return true;
 }
 
+void LuaPlayer::PlaceHQ(MapCoord x, MapCoord y)
+{
+    // Ignore if there is an HQ set in the map file.
```
Flamefire, line 266: This doesn't actually care about the map file directly. A script might (attempt to) set the HQ multiple times. Or re-add it if destroyed.

```suggestion
    // Ignore if an HQ already exists
```

doc/lua/functions.md
```diff
@@ -343,6 +343,10 @@ Ignored if the current player is not controlled by AI.
 **GetHQPos()**  
 Return x,y of the players HQ
 
+**PlaceHQ(x, y)**  
+Places this player's headquarters at map location {x, y}.  
+Does nothing if the player already has a valid headquarters position (e.g. set by the map or using a previous PlaceHQ call).
```
Flamefire, line 348: See below.

```suggestion
Does nothing if the player already has an existing headquarter (e.g. set by the map or using a previous PlaceHQ call).
```

Maybe even remove the part in the parentheses

libs/s25main/lua/LuaPlayer.cpp
```diff
@@ -259,15 +261,22 @@ bool LuaPlayer::AIConstructionOrder(unsigned x, unsigned y, lua::SafeEnum<Buildi
...
+    if(player.GetHQPos().isValid())
+        return;
+
+    GameWorld& world = player.GetGameWorld();
+    const MapPoint mp{x, y};
+    constexpr auto checkExists = false;
+    world.DestroyNO(mp, checkExists);
+    BuildingFactory::CreateBuilding(world, BuildingType::Headquarters, mp, player.GetPlayerId(), player.nation);
```
Flamefire, line 274: Nit: Better use `pos(ition)` as in other places

```suggestion
    const MapPoint pos{x, y};
    constexpr auto checkExists = false;
    world.DestroyNO(pos, checkExists);
    BuildingFactory::CreateBuilding(world, BuildingType::Headquarters, pos, player.GetPlayerId(), player.nation);
```

libs/s25main/network/GameClient.cpp
```diff
@@ -323,8 +323,10 @@ void GameClient::StartGame(const unsigned random_init)
             gameWorld.GetPlayer(i).MakeStartPacts();
 
         MapLoader loader(gameWorld);
-        if(!loader.Load(mapinfo.filepath)
-           || (!mapinfo.luaFilepath.empty() && !loader.LoadLuaScript(*game, *this, mapinfo.luaFilepath)))
+        if((!mapinfo.luaFilepath.empty() && !loader.LoadLuaScript(*game, *this, mapinfo.luaFilepath))
+           || !loader.Load(mapinfo.filepath)) // Do not reorder: load lua first, load map second.
+                                              // If the map is loaded first and it does not have a player HQ set, it
+                                              // will not load correctly, even though the HQ may be set using LUA.
```
Flamefire, line 329: I still find it hard to understand why it will "not load correctly" and what loading the script first changes. What about:

```suggestion
        // Load (and verify) Lua script before the map.
        // If a valid script exists loading the map will not abort when HQ positions in the map are missing
        // as HQs might be placed by the script when starting the game.
        if((!mapinfo.luaFilepath.empty() && !loader.LoadLuaScript(*game, *this, mapinfo.luaFilepath))
           || !loader.Load(mapinfo.filepath))
```

libs/s25main/network/GameServer.cpp
```diff
@@ -526,6 +526,11 @@ bool GameServer::assignPlayersOfRandomTeams(std::vector<JoinPlayerInfo>& playerI
     return playerWasAssigned;
 }
 
+void GameServer::SetNumPlayers(unsigned num)
+{
+    playerInfos.resize(num);
```
Flamefire, line 531: We should add a check `(state == ServerState::Config)` before doing this. Maybe:
```suggestion
    if(state != ServerState::Config)
        throw std::logic_error("Changing the number of players must only be done during game configuration");
    if(num < 1 || num > MAX_PLAYERS)
        throw std::invalid_argument("Invalid number of players");
    playerInfos.resize(num);
```

Not sure about the 2nd check as the value is likely from a lua script. So maybe rather return false here and abort the game in the caller.

libs/s25main/network/GameServer.h
```diff
@@ -52,6 +52,8 @@ class GameServer :
     /// Assign players that do not have a fixed team, return true if any player was assigned.
     static bool assignPlayersOfRandomTeams(std::vector<JoinPlayerInfo>& playerInfos);
 
+    void SetNumPlayers(unsigned num);
```
Flamefire, line 55: Maybe document as something like:
```suggestion
    /// Change number of players (after loading the map but before starting the game)
    void SetNumPlayers(unsigned num);
```

libs/s25main/world/MapLoader.cpp
```diff
@@ -421,8 +421,9 @@ bool MapLoader::PlaceHQs(GameWorldBase& world, std::vector<MapPoint> hqPositions
         // Does the HQ have a position?
         if(i >= hqPositions.size() || !hqPositions[i].isValid())
         {
-            LOG.write(_("Player %u does not have a valid start position!")) % i;
-            return false;
+            LOG.write(_("Player %u does not have a valid start position!\n")) % i;
```
Flamefire, line 424: This needs an update to the translations. I used a quick regex-search-replace in that repo. Please update the submodule external/languages to the current master (revision fa7f546)

Flamefire (changes requested):
I added some comments. 

We also need to increase the version in `unsigned LuaInterfaceGameBase::GetFeatureLevel()` due to the added functions such that scripts can check for that. Also the 2 changed mission scripts should require that new version too

PR #1684, opened 2024-07-23
Add SetComputerBarrier

kubaau:
Adds `SetComputerBarrier` LUA function which works like `!GLOBAL_SET_COMPUTER_BARRIER` did in S2 RTX scripts. This makes it possible to script old campaigns much more simply than with the tools previously available.


Flow86 (dismissed):

Flow86:
> Force-pushed clang-tidy fix:
> 
> ```
> -for(auto bldSite : bldSites)
> +for(noBuildingSite* bldSite : bldSites)
> ```

you still can use `auto` - but you have to use `auto* bldSite : bldSites)` to make tidy happy

Flow86 (approved):

kubaau:
> It'd be great to: have this in multiplayer (if that doesn't already work), for players and a "lift" function as well, that way one could "nicely" block a player in a custom campaign

This implementation only works for AIs. Implementing this for each player would not be difficult, but this particular implementation is only basically a mirror of S2 campaign AI behavior.

doc/lua/functions.md
```diff
@@ -379,6 +379,9 @@ rttr:GetWorld():AddAnimal(41, 44, SPEC_DUCK)
 
 [Back](#Lua-objects-and-their-methods)  
 
+**SetComputerBarrier(radius, x, y)**
```
Flamefire, line 382: How do you undo this? The current implementation and the mission scripts allowed to remove the restriction

libs/s25main/lua/LuaWorld.cpp
```diff
@@ -33,7 +33,8 @@ void LuaWorld::Register(kaguya::State& state)
     state["World"].setClass(kaguya::UserdataMetatable<LuaWorld>()
                               .addFunction("AddEnvObject", AddEnvObjectWrapper())
                               .addFunction("AddStaticObject", AddStaticObjectWrapper())
-                              .addFunction("AddAnimal", &LuaWorld::AddAnimal));
+                              .addFunction("AddAnimal", &LuaWorld::AddAnimal)
+                              .addFunction("SetComputerBarrier", &LuaWorld::SetComputerBarrier));
```
Flamefire, line 37: This needs increasing the LUA interface version such that mission scripts can require it if using this new function

libs/s25main/world/GameWorldBase.cpp
```diff
@@ -696,3 +696,14 @@ void GameWorldBase::RecalcBQ(const MapPoint pt)
...
+
+void GameWorldBase::SetComputerBarrier(const MapPoint& pt, unsigned radius)
+{
+    for(const auto& pt : GetPointsInRadiusWithCenter(pt, radius))
+        ptsInsideComputerBarriers.insert(pt);
+}
+
+bool GameWorldBase::IsInsideComputerBarrier(const MapPoint& pt) const
+{
+    return ptsInsideComputerBarriers.count(pt);
```
Flamefire, line 708: Maybe rather use `helpers::contains` which is more readable and doesn't coerce a `std::size_t` to a bool (IIRC MSVC warns about that)

Have you evaluated performance? As this is called quite often when searching for a building spot it likely matters. I imagine a `flat_map` might help or just storing point-radius pairs and comparing the distance. A plain `std::set` might cause a larger hit due to memory accesses and `MapPoint` being small (32 bit) at least a flat_map might be beneficial

doc/lua/functions.md
kubaau: 1. You don't :) As mentioned earlier I didn't think this is supposed to mirror S2 behavior and  don't think the restrictions could be lifted in S2. Maybe @Spikeone can correct me on this. I think if more features (lifting, restricting for each player, etc.) are desired then they could be requested through an issue and with a separate PR. Unless they are considered a blocker for this PR of course.

2. If there is, it is not used in RttR. Basically everything is passed using built-in types, except enums.

libs/s25main/world/GameWorldBase.cpp
kubaau: Replaced count with helpers::contains.

Ran a quick performance check with VS + RelWithDebInfo. This function is irrelevant in the grand scheme of things, with either count or contains:
Zrzut ekranu 2024-08-05 183218
Zrzut ekranu 2024-08-05 182641
Zrzut ekranu 2024-08-05 182624

libs/s25main/lua/LuaWorld.cpp
kubaau: Bumped feature level from 4 to 5. Not sure if this is what you meant.

doc/lua/functions.md
Flamefire: 1. OK then.
2. I see. Then I'd just change the order to have the radius last to match the other functions of this class like `AddEnvObject(x, y, ...`

Spikeone: > 1. You don't :) As mentioned earlier this is supposed to mirror S2 behavior and I don't think the restrictions could be lifted in S2. Maybe @Spikeone can correct me on this. I think if more features (lifting, restricting for each player, etc.) are desired then they could be requested through an issue and with a separate PR. Unless they are considered a blocker for this PR of course.
> 
>     2. If there is, it is not used in RttR. Basically everything is passed using built-in types, except enums.

Regarding 1., I'd really like to have that feature - but I do see, that it is currently not required or used at all. So this is no blocker for this PR I guess. Just asking (Since I don't know how this works) can't you just add a RemoveComputerBarrier with same parameters which just does the opposite?

libs/s25main/world/GameWorldBase.cpp
Flamefire: > This function is irrelevant in the grand scheme of things, with either count or contains:

I was more thinking about the "death by a thousand cuts": If we don't at least think about this for anything with loops the sum of those functions will be slow.

There might be a flaw in your performance chart: It looks like those counts are "inclusive times", i.e. `GameManager.Run` includes the time of e.g. `GameClient.Run` and `GameServer.Run` so the percentages don't add up to 100% which leads to those many 0.01% times. Especially as the network and draw calls are included. I guess including only `AIPlayerJH.Run` and below is more expressive.

What is the highest number of barriers and how many points does that yield? And what is "typical"? Maybe it really is irrelevant...

doc/lua/functions.md
Flamefire: > can't you just add a RemoveComputerBarrier with same parameters which just does the opposite?

Currently not: What is stored are the affected points. There is no way to go backwards from that: Assume blocking 2 points close to each others with radius partially overlapping. If you want to remove one of the points you can't know if a currently blocked point belongs also to another points radius.

The alternative would be to store only the points and radius set, making them trivial to remove. In the check code the distance to each point would need to be calculated to decide if the currently checked point is blocked. This _might_ even be faster if there are few barrier points (those passed with radius) but with large radius.

Spikeone: > Currently not: What is stored are the affected points. There is no way to go backwards from that: Assume blocking 2 points close to each others with radius partially overlapping. If you want to remove one of the points you can't know if a currently blocked point belongs also to another points radius.

I think, that this, for an initial implementation would be fine - so if you have overlapping areas, well, you may need to remove one and re-create the other again - thats fine for me

Flamefire (approved):

PR #1703, opened 2024-09-29
Move Windows CI to GitHub Actions

Flamefire:
The time limit of 1h is not enough to build the release version.

### DIFF
--- a/libs/libGamedata/lua/CheckedLuaTable.h
+++ b/libs/libGamedata/lua/CheckedLuaTable.h
@@ -67,7 +67,7 @@ inline T CheckedLuaTable::getOrThrow(const std::string& fieldName)
     if(value.isNilref())
         throw GameDataLoadError("Required field '" + fieldName + "' not found");
     if(value.isConvertible<T>())
-        return value;
+        return value.get<T>();
     else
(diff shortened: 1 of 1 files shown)


.github/workflows/unit-tests.yml
```diff
@@ -27,6 +27,50 @@ env:
   ADDITIONAL_CMAKE_FLAGS: -DRTTR_ENABLE_BENCHMARKS=ON
 
 jobs:
+  Windows:
+    strategy:
+      matrix:
+        include:
+          - { os: windows-2019, generator: Visual Studio 16 2019, type: Debug,   platform: Win32}
```
Flow86, line 34: I honestly would skip 2019

Flow86 (approved):

.github/workflows/unit-tests.yml
Flamefire: I'd rather skip the Win32 builds, who still has 32 bit Windows? I found it useful to have at least 2 different compilers to shake out compiler issues we might run into. I.e. if it works for 2019 but fails in 2022 that would be a good information.

I'll merge this now and we can remove some later if we still want to.

PR #1706, opened 2024-10-13
Fix stuck soldiers leading to crashes in drawing code

### LINKED ISSUE #1668: Crash while drawing walking soldier
The game (and replay) crashes:
grafik

grafik

Although the replay seems to end at GF 469209, the replay crashes if you observe that spot early.
If you do not look at that spot (so just start the replay, jump to end, continue with P) the game doesn't crash. As soon as you jump to that spot the game will crash.

Version: 20240409

[2024-05-25_19-40-22.zip](https://github.com/Return-To-The-Roots/s25client/files/15446575/2024-05-25_19-40-22.zip)

### LINKED ISSUE #1695: River: Removing a road while a soldier is crossing sometimes hangs the game
Hi!

I'm lately into river games with my brother, i.e. I like those maps that have rivers which can only be crossed if a player can build a road across them. Part of the strategy around those rivers is to build and remove roads, sometimes as soldiers cross the water. In some cases, roughly maybe ~25% of the cases, when one removes a road while the soldier is crossing the game hangs.

Thanks for taking a look!

By the way: I'm playing this game weekly, and it's literally the first game I've ever played on a computer back when I was 10. Thanks to everyone who has worked so hard to bring this back as open source!

Flamefire:
Fixes #1695 
Fixes #1668

The cause is in both issues the same: 

One soldier involved in an upcoming free fight is waiting for the other. That aborts the free fight as the spot becomes unreachable due to a removed road over water (either manually or by a captured building)
The waiting soldier changes state but doesn't start moving and as he isn't moving already there is no event queued that would make it handle the new (moving) state.
This leads to it being stuck there forever and also crashes the draw code which doesn't expect a supposedly walking soldier not having a walk-event.

Implemented solution:

Split the `AbortFreeFight` method in 2:
- virtual method to determine the next state -> This ensures derived classes change state
- base method to reset the free fight state for both soldiers and let   them continue -> Actual abort-fight handling in one place

For the aborting soldier effectively nothing is changed.
The other soldier gets a simulated walk event if he was waiting to handle the new state.

As the 2 soldiers might get involved in another fight with each other both their "fight states" are cleaned up first to establish a valid state before the other soldier gets triggered.

As this was a bit tricky to find and create a test case to reproduce this I added quite a bit refactoring:

General:

- Method to jump to a specific position in the map-debug window and disable FoW when opening that window to avoid having to search & scroll to the spot where it crashes
- During tests I ended up accidentally creating a copy of a soldier which was very confusing and hard to find -> Avoid that at the root as much as possible
- Enhanced the `range` class start iterating not only from zero but any value
- Easier search for `DescIdx` matching conditions (useful especially for tests to quickly find e.g. water)
- `DescriptionVector` to have a `std::vector` but indexed only by `DescIdx`

To understand what is happening with (fighting) soldiers I translated and fixed comments and documentation and some smaller refactoring to make it easier to read. Especially use of an enum instead of `uint8` and references where NULL isn't allowed.


libs/s25client/s25client.cpp
```diff
@@ -570,12 +570,12 @@ int main(int argc, char** argv)
         result = e.code;
     } catch(const std::exception& e)
     {
-        bnw::cerr << "An exception occurred: " << e.what() << "\n\n";
+        bnw::cerr << "An exception occurred: " << e.what() << std::endl;
```
Flow86, line 573: I though one shouldn't use `std::endl` on non-buffered outputs (e.g `cerr`)

Flamefire: It is usually superflous as `std::sync` is implicit in unbuffered streams. However I noticed that it isn't for the Boost::nowide `cerr`, which is a bug I now reported.
What I noticed is that loading a broken replay (actually due to broken code) terminated the game without any output except for the crash dialog. This causes the flush.

Reason for not using `std::endl` is the degraded performance due to the superflous sync. However here we ware terminating anyway so performance doesn't matter and the potentially additional sync doesn't hurt us.

Flamefire: Turns out there was a bug in Boost::nowide that has since been fixed. We still pulled in a pre-Boost 1.77 version though (in my local tests) which lead to this.

I updated the fetched version to include the fix.

Flow86:
rebase necessary, but other than that, lgtm

Flamefire:
Done. IIRC you can approve which stays active after the update/rebase such that it can be merged right away.

Flamefire:
@Flow86

PR #1707, opened 2024-10-15
Adding hotkey to go to notification event

### LINKED ISSUE #1539: Missing hotkey for "Go to notification"
Hi,

In the old settlers I would often use hotkeys to navigate my notifications. I would click "n" to open the notification window, "enter" to go to the notification site, then "del" to remove the notification.

In Settlers 2: RttR the "enter"-key is understandably bound to the chat functionality. I haven't been able to find a replaced hotkey for "go to notifcation site" and according to a chat on Discord it seems there is no new hotkey for this function.

I was wondering if another button could be bound to this functionality. Let's say "g" for "go-to" or maybe "home" by the "del" button.

The ideal solution would be to let us map our own hotkeys, but this is a lot more work than just binding the functionality for a single button, at least as a temporary(hehe) solution.

This issue is adjacent to the issue posted here, but since this is a missing hotkey I thought I'd submit a new issue.
#887 

Best regards,
Endre

sjoblomj:
Pressing `g` will go to the position of the latest notification. If the notification window is open, pressing `g` will go to the position of the currently open notification.

Closes #1539.

### DIFF
--- a/libs/s25main/postSystem/PostBox.cpp
+++ b/libs/s25main/postSystem/PostBox.cpp
@@ -1,4 +1,4 @@
-// Copyright (C) 2005 - 2021 Settlers Freaks (sf-team at siedler25.org)
+// Copyright (C) 2005 - 2024 Settlers Freaks (sf-team at siedler25.org)
 //
 // SPDX-License-Identifier: GPL-2.0-or-later
 
@@ -61,6 +61,12 @@ const PostMsg* PostBox::GetMsg(unsigned idx) const
     if(idx >= numMessages)
         return nullptr;
     return messages[idx].get();
+}
+
+const PostMsg* PostBox::GetCurrentMsg() const
+{
+    unsigned idx = currMsgIdx >= 0 && currMsgIdx < (int) numMessages ? currMsgIdx : numMessages - 1;
+    return GetMsg(idx);
 }
 
 void PostBox::SetCurrentMissionGoal(const std::string& goal)
--- a/data/RTTR/texte/keyboardlayout.txt
+++ b/data/RTTR/texte/keyboardlayout.txt
@@ -2,19 +2,17 @@
 ------------------------------------------------------------------------
 Global keyboard layout in game:
 
-Bracketed keys do not work yet.
-
 P:.................... Pause
 
 H:.................... Go to headquarters
-B:.................... Go back to start position
+B:.................... Go back to previous position
 
 N:.................... Open post office
 L:.................... Show map
 M:.................... Main selection
 I:.................... Open inventory
 
-SPACE:................ Switch contruction aid function
+SPACE:................ Switch construction aid function
 C:.................... Check information about houses
 S:.................... Show status of houses
 
@@ -24,11 +22,10 @@ V:.................... Game acceleration
 Alt-W, Esc:........... Close active window
 Alt-Q:................ Exit from game
 
-F1:................... (Load game)
 F2:................... Save game
 F8:................... Readme "Keyboard layout"
 F9:................... Readme
-F11:.................. Musicplayer
+F11:.................. Music player
 F12:.................. Options window
 
 Post office:
@@ -39,12 +36,12 @@ G:.................... Go to site of event
 
 Arrow keys: Scroll the main map
 
-Additionally when in Replay-mode:
+Additionally, when in Replay-mode:
 
 D:.................... Discover/hide unexplored land
 J:.................... Go to specific GF
 P:.................... Pause/Resume (also after GF jump)
 1-7:.................. Change to other player
 ------------------------------------------------------------------------
-https://www.siedler25.org        Copyright (C) 2005-2022 Settlers Freaks
+https://www.siedler25.org        Copyright (C) 2005-2024 Settlers Freaks
 ------------------------------------------------------------------------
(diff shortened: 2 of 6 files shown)


libs/s25main/postSystem/PostBox.cpp
```diff
@@ -63,6 +63,12 @@ const PostMsg* PostBox::GetMsg(unsigned idx) const
     return messages[idx].get();
 }
 
+const PostMsg* PostBox::GetCurrentMsg() const
+{
+    unsigned idx = currMsgIdx >= 0 && currMsgIdx < (int) numMessages ? currMsgIdx : numMessages - 1;
```
Flamefire, line 68: Is the possible underflow for `numMessages == 0`  intended?

Maybe be explicit and do:
```
if(curMsgIdx < 0 || numMessages == 0)
    return nullptr;
```

Afterwards you can use `static_cast<unsigned>(currMsgIdx)` safely instead of the (C-style) cast to int.

sjoblomj: It is intended, since `GetMsg` checks that the index is within bounds and returns `nullptr` if not. But you're right that it's probably best to be explicit about it. 👍

sjoblomj: I added check to make sure the index is within bounds. Checking that `currMsgIdx < 0` - as you suggest - does not make sense though, since `currMsgIdx == -1` whenever the notification window is closed. In that case, we get the the most recent notification (`numMessages - 1`).

Flamefire:
Although the change looks good to me I'm wondering why the existing code doesn't work: https://github.com/Return-To-The-Roots/s25client/blob/35983cd4c81617d2acb37a22886f5630fc18e44a/libs/s25main/ingameWindows/iwPostWindow.cpp#L251-L253

That should suffice. If it doesn't then I guess the window isn't active anymore so it doesn't get the events. In that case I'm wondering if we should even react on the key event.

From the issue:
> In the old settlers [...] would click "n" to open the notification window, "enter" to go to the notification site, then "del" to remove the notification.

If 'g'  doesn't work then 'del'  wouldn't either and we'd need to fix that too. To handle those (kind of) window-specific events in the desktop is not really "clean" and might lead to issues when another window uses the same key for something. So we should really check if and why it currently does not work and rather fix that. If it is only that the window lost focus due to clicking somewhere else the workaround would simply be to press 'n' again (or click inside the window), but IIRC the window is supposed the get the event first before the desktop, but I'd need to check too.

sjoblomj:
> Although the change looks good to me I'm wondering why the existing code doesn't work:
> 
> https://github.com/Return-To-The-Roots/s25client/blob/35983cd4c81617d2acb37a22886f5630fc18e44a/libs/s25main/ingameWindows/iwPostWindow.cpp#L251-L253
> 
> That should suffice. If it doesn't then I guess the window isn't active anymore so it doesn't get the events. In that case I'm wondering if we should even react on the key event.
> 
> From the issue:
> 
> > In the old settlers [...] would click "n" to open the notification window, "enter" to go to the notification site, then "del" to remove the notification.
> 
> If 'g' doesn't work then 'del' wouldn't either and we'd need to fix that too. To handle those (kind of) window-specific events in the desktop is not really "clean" and might lead to issues when another window uses the same key for something. So we should really check if and why it currently does not work and rather fix that. If it is only that the window lost focus due to clicking somewhere else the workaround would simply be to press 'n' again (or click inside the window), but IIRC the window is supposed the get the event first before the desktop, but I'd need to check too.

Like you say, the notification window also has a hotkey for `g` which goes to the position of the currently open message. And like you say, the notification window must be focused for this to work; so if the player has clicked on the game interface or another window, this no longer works.

So, the bigger question would be if the game interface should also have this hotkey or not, regardless of whether the notification window is open, which is what this PR implements. I'm not too familiar with the code, but I suppose it could be hairy to 1) let unfocused windows receive key strokes first, and 2) make sure different unfocused windows don't have the same key binding for different things.

Regardless, the current behavior is a bit of an annoyance, where you'd think pressing `g` would take you to the position of the event, but it turns out the notification window has lost focus and nothing happens.

As for the question of the other keys: If the `del` key should also have a similar functionality as `g` in this PR, I'd argue that there should be checks to make sure the notification window is open (but possibly unfocused) - otherwise it seems like the player could delete messages by mistake, possibly without even realizing it.

Flamefire:
> So, the bigger question would be if the game interface should also have this hotkey or not, regardless of whether the notification window is open

I don't think it is a good idea to have that jump-to when the window is not open. It might seem quite random, especially as a newer message might have arrived so the jump position is not fully predictable. Of course that might be a feature: Jump to position of latest notification.

>  the notification window must be focused for this to work; so if the player has clicked on the game interface or another window, this no longer works.

In combination with the above I lean towards that this is supposed to work that way. Are people really having the post window open all the time while doing something else? Pressing "n" and then "g" should always work, so that would only be one additional key press.

Do we have any other windows with hotkeys that exist in the original S2? Then we could test how the behavior with respect to focused windows  worked there. Maybe @Spikeone knows more or has suggestions.

Flamefire:
Coming back to this:

I tested the original S2 and the behavior is the same as in RttR: When the notification window is focused the "g" key is handle, otherwise not.   
So reacting to the hotkey when the window is **not** focused would be an extension/change in RttR.   

The current implementation has still 2 issues:
- I'm note sure that introducing the concept of a "current message" to the Postbox is a good idea from a design point: The current message is a concept from the view/presentation layer so should IMO be kept there
- When a message without a location gets shown the "current message" doesn't get updated. This makes the concept a bit unclear and might seem random if the view jumps to the location of a message that isn't currently shown

Idea: On keypress in the desktop try to get the post window (`FindNonModalWindow(CGI_POST_OFFICE)`) and if there is one relay the whole message/event to that window. That will then handle the keypress as-if it was focused.   
That avoids any changes to the "data structure" and automatically makes the behavior of focused or unfocused post window consistent.

What do you think?

PR #1708, opened 2024-10-19
Fix inverted cursor is not saved in global settings

Farmer-Markus:
The inverted cursor setting never was applied when activating in settings. Just when activating ingame in advanced options. 

Just a very small fix.


Flamefire (dismissed):
Good catch, thanks!

Flow86 (approved):

PR #1709, opened 2024-10-21
Actually allow maps and lua files to be placed within the campaign folder

Flamefire:
Although this is documented it was not implemented.
Make both folders optional defaulting to the campaign folder.

I also restricted this to a single subfolder. If it contains anything else, i.e. a parent folder, it must be a (restricted) path as in the other path references in lua files, i.e. start with `<RTTR_...>` and do not contain upward references.

I also replaced the script-strings in the test by C++11 raw strings to avoid the many backslashes and quotes that make them rather hard to read.


Flamefire:
@Flow86 ping

Flow86 (approved):

PR #1710, opened 2024-10-21
Load campaign data only once

### LINKED ISSUE #1705: Console infinite loop with specific languages in campaign menu.
Version: 0.9.5 for Windows
What did you do/shall be done to reproduce the bug?
Change language to non German or English. In my case it was Finnish.
After the language is selected:

1. Go to Singleplayer (in Finnish it's "yksinpeli").
2. Select "Campaigns", it's in English for some reason. In Finnish it would be "kampanjoita".
3. Press either "Roman campaign" or "Original world campaign".
4. Open the console.

What is expected:
One error message about the error.

What actually happens:
It spams "Did not found translation for language 'fi_FI.UTF-8' in LUA file. Available translations:  de, en" infinite times.

That line of code is located in:
 https://github.com/Return-To-The-Roots/s25client/blob/master/libs/libGamedata/lua/LuaInterfaceBase.cpp
The error message that is being spammed is in the line 125.

Flamefire:
This fixes #1705 but is a larger rework to make campaign handling easier and more flexible.

The reason for the bug was that the campaign file is loaded repeatedly, especially in the draw call to determine whether to draw the image.

So load data for all campaigns when opening the capaign selection window.
We don't rescan the folders while the window is open so there is no need to load the lua files multiple times.

If we pass the parsed campaign to the mission selection screen we can also get rid of the invisible column for the filepath.
Finally we can load all resources (images) in one go.
The 2 mission selection classes share some code and essentially the same functionality and the difference in naming doesn't make it clear which is which.
Especially problematic is the reliance on the "mission map" in the campaign description which may lead to an empty screen if
`dskCampaignMissionMapSelection` is instantiated in absence of that.
So just merge them so the campaign selection doesn't need to decide which one to use and the screen can use the representation that is available, i.e. either a map to select an area or the list as a fallback. The result looks exactly the same.

I noticed that we can rather use `std::optional` in 2 places for improved semantics. When modifying the test I noticed that the whole project gets recompiled instead of only `libGamedata` (I moved the test to the corresponding test suite) which was due to some superflous dependencies in CMake.


doc/AddingCustomCampaign.md
```diff
@@ -86,7 +86,7 @@ If you want a field to be translated you have to add the translation as describe
   3. `name` the name of the campaign
   4. `shortDescription` Short description of the campaign (like a head line to get a rough imagination of the campaign)
   5. `longDescription` Extended description describing the campaign in detail. Will be shown in the campaign selection screen, when the campaign is selected.
-  6. `image` Path to an image displayed in the campaign selection screen. This can also be `image=""` if you do no want to provide an image.
+  6. `image` Path to an image displayed in the campaign selection screen. You can ommit this if you do no want to provide an image.
```
Flow86, line 89: typo `omit`

Flow86 (changes requested):

doc/AddingCustomCampaign.md
Flamefire: Fixed and rebased.

Flow86 (approved):

PR #1719, opened 2024-11-22
Fix wrong file-indices for vintner and templeservant graphics walking in/out with ware

### LINKED ISSUE #1718: Crash when vintner reenters building with ware because flag has no empty space
Load the save file and scroll to the right. The application crashes.

Original description from Discord:

> I can replicate the issue on my save. However it's also avoidable. It is happening only when I'm scrolling to a certain area of the map, where (may be a coincident) addon buildings are (temple, winyard, winery)

Replay:
[2024-11-21_22-04-01.zip](https://github.com/user-attachments/files/17876437/2024-11-21_22-04-01.zip)
Save:
[1112.zip](https://github.com/user-attachments/files/17876438/1112.zip)

On the screenshot you can see the flag in front of the vintner has no space for more wares.
grafik

ottml:
Fixes #1718

After finishing his work the winery servant walks out with the ware and if there is no space at flag in front of his house he walks back with the ware into the house. The crash happens only if you have a look at it, means the position is on your visible screen, because the part of the map is not rendered otherwise. It is only a rendering bug because the access to the graphic only happens if the servant is rendered. The logic itself is correct.

Because the in/out animation for the winery servant has only 8 frames for the NORTHWEST and SOUTHEAST stored in the lst archive and the render code assuming a full animation of 6 x 8 frames present an access to a not existing index happens. This crashes the application. To solve the problem we repack the archive to have 6x8 frames but only the needed ones have an image.


Flamefire:
Can you explain why/how this fixes the crash? I.e. what the issue is?

The code looks very suspicious and I suspect that the texture file is wrong instead: The `DrawWalking` function expects the ID in the file where the walking animations start, 8 frames for each of the 6 directions, EAST first

If `bobIndex` is supposed to hold the index, why is the adjustment required at all? I.e. it could just hold the correct value in the first place. And why different depending on the direction?

Can you fix the value in `bobIndex` and/or the order of the animations in the file? I assume only the SE and NW images are required and hence contained in the file which is fine. They just need to be at the right positions and 24 indices apart

Flamefire (approved):
Can't easily review the contents of the archive but looks like the issue has been addressed properly.

I would have likely used a more efficient spacing, i.e. when the first set is at 540, the next could be at 588, not 600 and after 600 would be 648

But the file format is pretty efficient in storing empty entries this is no big deal and we can use the simpler decimal-aligned offsets.

libs/s25main/figures/noFigure.cpp
```diff
@@ -772,30 +781,27 @@ void noFigure::DrawWalking(DrawPoint drawPt, glArchivItem_Bob* file, unsigned id
...
     {
         case Job::PackDonkey:
         {
             const unsigned ani_step = CalcWalkAnimationFrame();
             drawPt = InterpolateWalkDrawPos(drawPt);
 
-            // Esel
-            LOADER.GetMapTexture(2000 + rttr::enum_cast(GetCurMoveDir() + 3u) * 8 + ani_step)->DrawFull(drawPt);
-            // Schatten des Esels
+            LOADER.GetMapTexture(calcWalkFrameIndex(200, GetCurMoveDir(), ani_step))->DrawFull(drawPt);
```
ottml, line 803: Is the start Index here not 2000 instead of 200?

Flamefire: Indeed thanks for catching this! I'll test run this anyway a bit later on. Just wanted the CI to pass already to merge it fast

Flamefire (approved):

PR #1721, opened 2024-12-04
Fix & Update Github Actions CI

Flamefire:
The `macos-12` runner image is now removed. `macos-15` is available as preview, `macos-latest` is still at 14, so use that.

Also update the "latest GCC &Clang" compilers to the currently latest versions


Flamefire:
@Flow86 Please merge and update the required checks to require macos-14 instead

PR #1722, opened 2024-12-14
Add classic cheat to reveal map (F7)

kubaau:
Part of https://github.com/Return-To-The-Roots/s25client/pull/1679

### DIFF
--- a/libs/s25main/world/GameWorldBase.h
+++ b/libs/s25main/world/GameWorldBase.h
@@ -79,7 +79,7 @@ public:
     // Remaining initialization after loading (BQ...)
     void InitAfterLoad();
 
-    /// Setzt GameInterface
+    GameInterface* GetGameInterface() const { return gi; }
     void SetGameInterface(GameInterface* const gi) { this->gi = gi; }
 
     /// Get the economy mode handler if set.
--- a/libs/s25main/world/GameWorldBase.h
+++ b/libs/s25main/world/GameWorldBase.h
@@ -79,7 +79,8 @@ public:
     // Remaining initialization after loading (BQ...)
     void InitAfterLoad();
 
-    GameInterface* GetGameInterface() const { return gi; }
+    GameInterface* GetGameInterface() { return gi; }
+    const GameInterface* GetGameInterface() const { return gi; }
     void SetGameInterface(GameInterface* const gi) { this->gi = gi; }
 
     /// Get the economy mode handler if set.
--- a/tests/s25Main/integration/testCheats.cpp
+++ b/tests/s25Main/integration/testCheats.cpp
@@ -3,52 +3,37 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 
 #include "Cheats.h"
+#include "GamePlayer.h"
+#include "desktops/dskGameInterface.h"
 #include "worldFixtures/CreateEmptyWorld.h"
 #include "worldFixtures/WorldFixture.h"
+#include <turtle/mock.hpp>
 
 BOOST_AUTO_TEST_SUITE(CheatsTests)
 
 namespace {
-template<unsigned T_numPlayers>
-struct CheatsFixture : WorldFixture<CreateEmptyWorld, T_numPlayers>
+constexpr auto numPlayers = 1;
+constexpr auto worldWidth = 64;
+constexpr auto worldHeight = 64;
+struct CheatsFixture : WorldFixture<CreateEmptyWorld, numPlayers, worldWidth, worldHeight>
 {
-    Cheats cheats{this->world};
+    CheatsFixture()
+        : cheats{gameDesktop.GI_GetCheats()}, viewer{gameDesktop.GetView().GetViewer()}, p1HQPos{p1.GetHQPos()}
+    {}
+
+    dskGameInterface gameDesktop{game, nullptr, 0, false};
+    Cheats& cheats;
+    const GameWorldViewer& viewer;
+
+    GamePlayer& getPlayer(unsigned id) { return world.GetPlayer(id); }
+    GamePlayer& p1 = getPlayer(0);
+    const MapPoint p1HQPos;
 };
-using CheatsFixture1P = CheatsFixture<1>;
-using CheatsFixture2P = CheatsFixture<2>;
 } // namespace
 
-BOOST_FIXTURE_TEST_CASE(CheatsAreAllowed_InSinglePlayer, CheatsFixture1P)
+BOOST_FIXTURE_TEST_CASE(CanToggleCheatModeOnAndOffRepeatedly, CheatsFixture)
 {
-    BOOST_TEST_REQUIRE(cheats.areCheatsAllowed() == true);
-}
-
-BOOST_FIXTURE_TEST_CASE(CheatsAreNotAllowed_InMultiplayer, CheatsFixture2P)
-{
-    BOOST_TEST_REQUIRE(cheats.areCheatsAllowed() == false);
-}
-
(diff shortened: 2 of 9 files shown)


tests/s25Main/integration/testCheats.cpp
```diff
@@ -3,60 +3,94 @@
...
+    MOCK_METHOD(GI_UpdateMinimap, 1)
+    MOCK_METHOD(GI_FlagDestroyed, 1)
+    MOCK_METHOD(GI_TreatyOfAllianceChanged, 1)
+    MOCK_METHOD(GI_UpdateMapVisibility, 0)
+    MOCK_METHOD(GI_Winner, 1)
+    MOCK_METHOD(GI_TeamWinner, 1)
+    MOCK_METHOD(GI_StartRoadBuilding, 2)
+    MOCK_METHOD(GI_CancelRoadBuilding, 0)
+    MOCK_METHOD(GI_BuildRoad, 0)
+    MOCK_METHOD(GI_GetCheats, 0)
```
kubaau, line 56: `Unit tests / Linux (gcc-9, ubuntu-20.04, Debug, 3.16.3, 1.71.0) (pull_request)` gives some crap errors about `GI_GetCheats()` not being `noexcept`. I tried making it `noexcept` but I cannot mock a `noexcept` function properly with turtle. There are also some other warnings related to marking the mocked functions `override`, but I have no idea how to do that with turtle either.

Flamefire: The override stuff can be ignored. However the noexcept issue is a long standing feature request for turtle.

My suggestion: Fix the noexcept warnings and to avoid having to mock the `GI_GetCheats` method just pass the Cheats instance in the ctor of the mock class and use a "regular" method.

tests/s25Main/integration/testCheats.cpp
```diff
@@ -3,60 +3,100 @@
...
+    MOCK_METHOD(GI_CancelRoadBuilding, 0)
+    MOCK_METHOD(GI_BuildRoad, 0);
+
+    // can't mock noexcept function in turtle - working around it
+    MockGameInterface(Cheats & cheats) : cheats_{cheats} {}
+    Cheats & cheats_;
+    Cheats& GI_GetCheats() noexcept override
+    {
+        return cheats_;
+    }
```
kubaau, line 63: I think clang-format is getting confused here - the formatting becomes inconsistent with regards to `&` placement and the definition of `GI_GetCheats()` is formatted differently from how it is in `dskGameInterface`. I'm sure this can be fixed (e.g. by `// clang-format off` and manual formatting), but it's probably not worth it.

tests/s25Main/integration/testCheats.cpp
```diff
@@ -3,60 +3,100 @@
...
+    MOCK_METHOD(GI_CancelRoadBuilding, 0)
+    MOCK_METHOD(GI_BuildRoad, 0);
+
+    // can't mock noexcept function in turtle - working around it
+    // also clang-format is getting confused around this code - formatting manually
+    // clang-format off
+    MockGameInterface(Cheats& cheats) : cheats_{cheats} {}
+    Cheats& cheats_;
+    Cheats& GI_GetCheats() noexcept override { return cheats_; }
+    // clang-format on
```
kubaau, line 63: My clang-format (I think the one coming with Visual Studio) is getting confused here and keeps formatting the `&` placement and `GI_GetCheats()` definition inconsistently with the rest of the codebase.

Flamefire: I know of another place where this happens, IIRC in the `tmpFile` part of the test helpers. What does it do here?

kubaau: Both the & placement and definition not being on the same line get rejected by StyleAndFormatting.

Flamefire: I see, this is similar to the other place and fine to disable formatting here.

libs/s25main/GamePlayer.h
```diff
@@ -90,7 +90,7 @@ class GamePlayer : public GamePlayerInfo
     GameWorld& GetGameWorld() { return world; }
     const GameWorld& GetGameWorld() const { return world; }
 
-    const MapPoint& GetHQPos() const { return hqPos; }
+    const MapPoint& GetHQPos() const noexcept { return hqPos; }
```
kubaau, line 93: It's a little weird that the gcc-9 job complained about this. I didn't change this code (and `GameWorldView::GetViewer`) and yet I have to "fix" this? I don't even necessarily think this is a valid suggestion (keep marking things noexcept just because they don't throw) and if it is, then there are hundreds of other functions like this.

kubaau: Moved troublesome calls to initializer list.

Flow86 (dismissed):

Flamefire (commented):
Just one question, otherwise LGTM too

libs/s25main/world/GameWorldBase.h
```diff
@@ -79,7 +79,7 @@ class GameWorldBase : public World
     // Remaining initialization after loading (BQ...)
     void InitAfterLoad();
 
-    /// Setzt GameInterface
+    GameInterface* GetGameInterface() const { return gi; }
```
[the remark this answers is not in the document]
kubaau: In principle I would agree with this, but in practice there will be other classes that will want to `GetCheats()` but only have access to a `const GameWorldBase` (such as [GamePlayer](https://github.com/kubaau/s25client/commit/661d1f41b042ec830ba17017a6037d64fd63acb6?diff=unified#diff-52f6aea2a82a922f095f592a2f1b53effc939822718441624200437cbbf7544dR2260) and [GameWorldView](https://github.com/kubaau/s25client/commit/395596ff2985d239b9521f631d325bb18af01520?diff=unified#diff-ed4c289e07ca0a90c1cd9f11b1fc2e93327a11ea1c01fe53800d7a6540b7d071R377)). Creating two versions like this:
```
    GameInterface* GetGameInterface() { return gi; }
    const GameInterface* GetGameInterface() const { return gi; }
```
will also force to extend GameInterface like this:
```
    virtual const Cheats& GI_GetCheats() const = 0;
    virtual Cheats& GI_GetCheats() = 0;
```
or like this:
```
    virtual Cheats& GI_GetCheats() const = 0;
```
Where the first solution only provides a minor safety improvement and the second just moves the issue someplace else. It is simply convenient to have it defined like this.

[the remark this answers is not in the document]
Flamefire: > Creating two versions like this: [..] will also force to extend GameInterface like this:

Yes that is what I had in mind: `Cheats` is used to query some state (read-only) and modify state. Those with access to `const GameWorldBase` should not modify the state and both places you mentioned don't do that. So from a `const GameWorldBase` you only get a `const GameInterface` and from that only a `const Cheats`. That's how I understand "const correctness".

To avoid 2 virtuals, you can "cheat": 
```
const Cheats& GI_GetCheats() const = 0;
virtual Cheats& GI_GetCheats()  { const auto* me = this; return const_cast<Cheats&>(me->GI_GetCheats()); }
```

Not sure what you mean by "only minor safety improvement". E.g. `std::vector` and most other data structures work like this. And yes const correctness provides only a small advantage. But when it catches an unintentional access/modification it is quite useful, isn't it? And I'd rather do it consistently "right" for newly added code, see https://isocpp.org/wiki/faq/const-correctness#retrofitting-const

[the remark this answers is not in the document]
kubaau: Added what was requested, just implemented the virtual hack differently.

Flamefire (approved):
Thanks a lot!

PR #1731, opened 2025-01-11
Add Chinese font

DevJogger:
- Add Chinese font, including full-width punctuation and CJK symbols.
- Add Chinese config in `languages.ini`.
- Also created [a PR for the Chinese translation file update](https://github.com/Return-To-The-Roots/languages/pull/28).

_The font is "Source Han Sans", which is under the terms of the SIL Open Font License, Version 1.1
(ref: https://github.com/adobe-fonts/source-han-sans/tree/release)_

Screen shot:

image

image

image

### DIFF
--- a/data/RTTR/assets/base/languages.ini
+++ b/data/RTTR/assets/base/languages.ini
@@ -21,3 +21,4 @@ Norwegian Bokmal=nb
 Slovenian=sl
 Latin=la
 Portuguese=pt
+Chinese=zh_CN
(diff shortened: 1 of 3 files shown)


Flamefire (approved):
Thanks!

Flamefire (approved):
I updated the submodules such that your translations are included too once the PR is merged which happens right after CI finishes. Use `git pull` to update your local branch if necessary.

Thanks for the contribution!

Flamefire (approved):

PR #1737, opened 2025-01-19
Fix testCheats when running standalone

Flamefire:
The test creates a desktop instance that requires access to the VideoDriver and potentially resources (graphics) which are not available at this point. This leads to failures when the test is not started after a test installing a test-videodriver in the same process.

Manually create the required instances without that desktop.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -15,13 +15,15 @@ def transformIntoStep(architecture, dockerImage, buildScript) {
     return {
         timeout(120) {
             def wspwd = pwd()
+            def ccache_dir_host = "$HOME/.ccache-$architecture"
             dir("build-$architecture") {
                 sh 'touch .git-keep'
+                sh "mkdir -p '$ccache_dir_host'"
                 withDockerRegistry( registry: [credentialsId: dockerCredentials, url: 'https://'+dockerRegistry ] ) {
                     withDockerContainer(
                         image: dockerRegistry+dockerImage,
                         args: " \
-                            -v $HOME/.ccache:/.ccache \
+                            -v $ccache_dir_host:/.ccache \
                             -v $wspwd/source:/source:ro \
                             -v $wspwd/result:/result \
                         ") {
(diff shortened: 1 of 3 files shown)


Flow86 (commented):
needs rebase

Flamefire:
Resolved the conflict.

Flamefire:
@Flow86 Again having the issue with undefined references to symbols removed earlier. Not sure why this keeps popping up.
```
/usr/bin/ld: CMakeFiles/s25client.dir/s25client.cpp.o: in function `(anonymous namespace)::handleException(void*) [clone .constprop.0]':
s25client.cpp:(.text+0x18b2): undefined reference to `DebugInfo::SendStackTrace(std::vector<void*, std::allocator<void*> > const&)'
```

I added a commit to make it at least use architecture dependent folders as maybe it was mixing up the builds    
But I guess the real reason is that the obj files are kept and `git restore-mtime -c` might not correspond well enough with the date on the server leading t skipped rebuilds.    
I guess the `rm -r CMakeFiles` was indented to avoid this but there are more such folders in subfolders. I added a commit to delete them too.   
I guess that increases the build time as everything will be rebuild but better than building wrong combinations and ccache should be helpful enough.    
Or is there another solution?

Flow86 (approved):

Jenkinsfile
```diff
@@ -15,13 +15,15 @@ def transformIntoStep(architecture, dockerImage, buildScript) {
...
             def wspwd = pwd()
+            def ccache_dir_host = "$HOME/.ccache-$architecture"
             dir("build-$architecture") {
                 sh 'touch .git-keep'
+                sh "mkdir -p '$ccache_dir_host'"
                 withDockerRegistry( registry: [credentialsId: dockerCredentials, url: 'https://'+dockerRegistry ] ) {
                     withDockerContainer(
                         image: dockerRegistry+dockerImage,
                         args: " \
-                            -v $HOME/.ccache:/.ccache \
```
Flow86, line 24: it looks like apple does not use cache at all:

```
cache directory                     .ccache-apple.x86_64/
primary config                      .ccache-apple.x86_64//ccache.conf
secondary config (readonly)         /etc/ccache.conf
cache hit (direct)                     0
cache hit (preprocessed)               0
cache miss                             0
cache hit rate                      0.00 %
cleanups performed                     0
files in cache                         0
cache size                           0.0 kB
max cache size                       5.0 GB
```

I'll probably have too look if ccache-lipo is used properly in the runner image

PR #1738, opened 2025-01-21
"Show enemy productivity overlay" cheat

kubaau:


libs/s25main/world/GameWorldView.cpp
```diff
@@ -370,7 +372,10 @@ void GameWorldView::DrawNameProductivityOverlay(const TerrainRenderer& terrainRe
                     auto* attackAidImage = LOADER.GetImageN("map_new", 20000);
                     attackAidImage->DrawFull(curPos - DrawPoint(0, attackAidImage->getHeight()));
                 }
-                continue;
+                // Do not draw enemy productivity overlay unless the object is visible AND the related cheat is on
+                if(!(gwv.GetVisibility(pt) == Visibility::Visible
```
Flamefire, line 376: I'd rather put that at the top of both to make it clearer:
```
if != Visible -> continue
if addon -> draw military aid
// Skip drawing the name and productivity below if not enabled (or similar comment)
if !prod Overlay enabled -> continue
```

Although the first check isn't exactly there it is there implicitly as you cannot attack when you don't see the building. So !visible implies numSoldiers=0

This avoids the `if(!(conditions))` which is (slightly) more complex.

And while touching this code: Can you move the checks for the addon and cheat, and loading the image out of the loop please? This is a hot loop that is run very often (in times and iterations) so we can move those constants, that might not be obvious constants to the compiler, to a single place as const variables.   
Speaking of those: The visibility check has a few [constant conditions](https://github.com/Return-To-The-Roots/s25client/blob/35a55bb58559b0b7ea25ce594b7019c61915bc4c/libs/s25main/world/GameWorldViewer.cpp#L118-L128) Maybe we split that into 2 functions `IsAllVisible` and `CalcVisibility` (both called from the current function) but move a call to the first out of the loop to a constant and use the 2nd here as `if (!isAllVisible && !gwv.CalcVisibility(pt))` Although I don't expect that there are too many enemy buildings in the current view that there is a huge impact. So impact is low.    
But moving the 2 "constants" used here out of the loop is trivial enough to be worth it without measuring.

And possibly also the 2 lines at the top for coordinate transform below the no check directly on top of `curPos = `.

This isn't exactly part of the PR topic but I like to do such small cleanups when touching code ("Boy Scout Rule")

Flamefire (changes requested):
Looks good. Just some ideas for small improvements if you agree.

Flamefire (approved):

libs/s25main/world/GameWorldViewer.h
```diff
@@ -54,7 +54,10 @@ class GameWorldViewer
     /// Recalculates the visual BQ when a road part is build at that point
     void RecalcBQForRoad(const MapPoint& pt);
     /// Ermittelt Sichtbarkeit eines Punktes für den lokalen Spieler, berücksichtigt ggf. Teamkameraden
-    Visibility GetVisibility(MapPoint pt) const;
+    /// If checkAllVisible is false then `IsAllVisible` is assumed to return false, else it is called to check
+    Visibility GetVisibility(MapPoint pt, bool checkAllVisible = true) const;
+    /// Check if the whole map should be visible
```
Flow86, line 59: that comment does not make sense to me

Maybe `Returns true if the whole map is visible`?

Flamefire: I don't like "Returns true" type comments. What else do you expect from `IsSomething`? 

As for "should be" vs "is": The all-visible thing is a bit special as it is only visible for the purpose of the viewer but not for the game.

Flow86: I expect that "IsXYZ" is pretty dump. "Check" and "should" sound weird. "Is" is something else than "should"

Maybe you find a better comment wording

PR #1739, opened 2025-01-25
Make AIJH Actions Reproducible by Using the Games Random Seed

wichern:
Currently, the actions of AIJH are not reproducible because it uses a different random seed for each execution, resulting in different outcomes every time. I'm unsure whether this behavior is intentional.

This change ensures reproducibility by switching from the standard `rand()` function to `RANDOM.Rand()`, which allows for the use of a fixed random seed. With this update, users can replicate results consistently if desired.

If the lack of reproducibility was intentional, feel free to disregard this pull request.

### DIFF
--- a/libs/s25main/ai/aijh/AIConstruction.cpp
+++ b/libs/s25main/ai/aijh/AIConstruction.cpp
@@ -16,6 +16,7 @@
 #include "buildings/nobMilitary.h"
 #include "buildings/nobUsual.h"
 #include "helpers/containerUtils.h"
+#include "random/Random.h"
 #include "nodeObjs/noFlag.h"
 #include "nodeObjs/noRoadNode.h"
 #include "gameTypes/BuildingQuality.h"
@@ -465,10 +466,12 @@ helpers::OptionalEnum<BuildingType> AIConstruction::ChooseMilitaryBuilding(const
     const BuildingType biggestBld = GetBiggestAllowedMilBuilding().value();
 
     const Inventory& inventory = aii.GetInventory();
-    if(((rand() % 3) == 0 || inventory.people[Job::Private] < 15)
+    uint8_t playerId = aii.GetPlayerId();
+    if((RANDOM.Rand(RANDOM_CONTEXT2(playerId), 3) == 0 || inventory.people[Job::Private] < 15)
        && (inventory.goods[GoodType::Stones] > 6 || bldPlanner.GetNumBuildings(BuildingType::Quarry) > 0))
         bld = BuildingType::Guardhouse;
-    if(aijh.getAIInterface().isHarborPosClose(pt, 19) && rand() % 10 != 0 && aijh.ggs.isEnabled(AddonId::SEA_ATTACK))
+    if(aijh.getAIInterface().isHarborPosClose(pt, 19) && RANDOM.Rand(RANDOM_CONTEXT2(playerId), 10) != 0
+       && aijh.ggs.isEnabled(AddonId::SEA_ATTACK))
     {
         if(aii.CanBuildBuildingtype(BuildingType::Watchtower))
             return BuildingType::Watchtower;
@@ -478,13 +481,12 @@ helpers::OptionalEnum<BuildingType> AIConstruction::ChooseMilitaryBuilding(const
     {
         if(aijh.UpdateUpgradeBuilding() < 0 && bldPlanner.GetNumBuildingSites(biggestBld) < 1
            && (inventory.goods[GoodType::Stones] > 20 || bldPlanner.GetNumBuildings(BuildingType::Quarry) > 0)
-           && rand() % 10 != 0)
+           && RANDOM.Rand(RANDOM_CONTEXT2(playerId), 10) != 0)
         {
             return biggestBld;
         }
     }
 
-    uint8_t playerId = aii.GetPlayerId();
     sortedMilitaryBlds military = aii.gwb.LookForMilitaryBuildings(pt, 3);
     for(const nobBaseMilitary* milBld : military)
     {
@@ -493,7 +495,7 @@ helpers::OptionalEnum<BuildingType> AIConstruction::ChooseMilitaryBuilding(const
         // Prüfen ob Feind in der Nähe
         if(milBld->GetPlayer() != playerId && distance < 35)
         {
-            int randmil = rand();
+            int randmil = RANDOM.Rand(RANDOM_CONTEXT2(playerId), std::numeric_limits<int>::max());
             bool buildCatapult = randmil % 8 == 0 && aii.CanBuildCatapult()
                                  && bldPlanner.GetNumAdditionalBuildingsWanted(BuildingType::Catapult) > 0;
             // another catapult within "min" radius? ->dont build here!
--- a/libs/s25main/ai/aijh/AIPlayerJH.cpp
+++ b/libs/s25main/ai/aijh/AIPlayerJH.cpp
@@ -26,6 +26,7 @@
 #include "notifications/RoadNote.h"
 #include "notifications/ShipNote.h"
 #include "pathfinding/PathConditionRoad.h"
+#include "random/Random.h"
 #include "nodeObjs/noAnimal.h"
 #include "nodeObjs/noFlag.h"
 #include "nodeObjs/noShip.h"
@@ -344,7 +345,7 @@ void AIPlayerJH::PlanNewBuildings(const unsigned gf)
         DistributeGoodsByBlocking(GoodType::Boards, 30);
         DistributeGoodsByBlocking(GoodType::Stones, 50);
         // go to the picked random warehouse and try to build around it
-        int randomStore = rand() % (storehouses.size());
+        int randomStore = RANDOM.Rand(RANDOM_CONTEXT2(playerId), storehouses.size());
         auto it = storehouses.begin();
         std::advance(it, randomStore);
         const MapPoint whPos = (*it)->GetPos();
@@ -365,7 +366,7 @@ void AIPlayerJH::PlanNewBuildings(const unsigned gf)
     const std::list<nobMilitary*>& militaryBuildings = aii.GetMilitaryBuildings();
     if(militaryBuildings.empty())
         return;
-    int randomMiliBld = rand() % militaryBuildings.size();
+    int randomMiliBld = RANDOM.Rand(RANDOM_CONTEXT2(playerId), militaryBuildings.size());
     auto it2 = militaryBuildings.begin();
     std::advance(it2, randomMiliBld);
     MapPoint bldPos = (*it2)->GetPos();
@@ -1209,7 +1210,7 @@ void AIPlayerJH::HandleExpedition(const noShip* ship)
         aii.FoundColony(ship);
     else
     {
-        const unsigned offset = rand() % helpers::MaxEnumValue_v<ShipDirection>;
+        const unsigned offset = RANDOM.Rand(RANDOM_CONTEXT2(playerId), helpers::MaxEnumValue_v<ShipDirection>);
         for(auto dir : helpers::EnumRange<ShipDirection>{})
         {
             dir = ShipDirection((rttr::enum_cast(dir) + offset) % helpers::MaxEnumValue_v<ShipDirection>);
@@ -1254,9 +1255,7 @@ void AIPlayerJH::HandleTreeChopped(const MapPoint pt)
 
     UpdateNodesAround(pt, 3);
 
-    int random = rand();
-
-    if(random % 2 == 0)
+    if(RANDOM.Rand(RANDOM_CONTEXT2(playerId), 2) == 0)
         AddMilitaryBuildJob(pt);
     else // if (random % 12 == 0)
         AddBuildJob(BuildingType::Woodcutter, pt);
@@ -1538,7 +1537,7 @@ void AIPlayerJH::TryToAttack()
         // We skip the current building with a probability of limit/numMilBlds
         // -> For twice the number of blds as the limit we will most likely skip every 2nd building
         // This way we check roughly (at most) limit buildings but avoid any preference for one building over an other
-        if(rand() % numMilBlds > limit)
+        if(static_cast<unsigned>(RANDOM.Rand(RANDOM_CONTEXT2(playerId), static_cast<int>(numMilBlds))) > limit)
             continue;
 
         if(milBld->GetFrontierDistance() == FrontierDistance::Far) // inland building? -> skip it
@@ -1571,7 +1570,7 @@ void AIPlayerJH::TryToAttack()
 
     // shuffle everything but headquarters and harbors without any troops in them
     std::shuffle(potentialTargets.begin() + hq_or_harbor_without_soldiers, potentialTargets.end(),
-                 std::mt19937(std::random_device()()));
+                 std::mt19937(RANDOM.Rand(RANDOM_CONTEXT2(playerId), 2048)));
 
     // check for each potential attacking target the number of available attacking soldiers
     for(const nobBaseMilitary* target : potentialTargets)
@@ -1704,7 +1703,10 @@ void AIPlayerJH::TrySeaAttack()
     unsigned limit = 15;
     unsigned skip = 0;
     if(searcharoundharborspots.size() > 15)
-        skip = std::max<int>(rand() % (searcharoundharborspots.size() / 15 + 1) * 15, 1) - 1;
+        skip =
+          std::max<int>(
+            RANDOM.Rand(RANDOM_CONTEXT2(playerId), static_cast<int>(searcharoundharborspots.size() / 15 + 1) * 15), 1)
+          - 1;
     for(unsigned i = skip; i < searcharoundharborspots.size() && limit > 0; i++)
     {
         limit--;


Flow86 (dismissed):

Flamefire:
I'm not sure we need this and the deterministic RANDOM is a likely more expensive and would make the Async debugging much harder as it now contains logs from the AI which are not relevant to the game.

> which allows for the use of a fixed random seed. With this update, users can replicate results consistently if desired.

Which kind of results? I think the only place where this matters is the [AI battle](https://github.com/Return-To-The-Roots/s25client/blob/828eaad9ad1a766ed27a1bcf39c28b8a7fb96e3f/extras/ai-battle/main.cpp#L91)

If we do want that I'd rather have a separate RNG (could even be from `std::random`) for AI which is deterministically seeded as you propose. It should not record the async logs/invocations or at least not influence those from the main game.

Flamefire (changes requested):
I'm not sure we need this and the deterministic RANDOM is a likely more expensive and would make the Async debugging much harder as it now contains logs from the AI which are not relevant to the game.

> which allows for the use of a fixed random seed. With this update, users can replicate results consistently if desired.

Which kind of results? I think the only place where this matters is the [AI battle](https://github.com/Return-To-The-Roots/s25client/blob/828eaad9ad1a766ed27a1bcf39c28b8a7fb96e3f/extras/ai-battle/main.cpp#L91)

If we do want that I'd rather have a separate RNG (could even be from `std::random`) for AI which is deterministically seeded as you propose. It should not record the async logs/invocations or at least not influence those from the main game.

wichern:
Hi @Flamefire 

You are right, it only matters for AI battle. I am using them to analyze performance. It would be easier to do if the AI behaviour is reproducible. If you are ok with it, I would update [RunProgram()](https://github.com/Return-To-The-Roots/s25client/blob/541bc230f33a9f39f2ce23b914c70e70e15da1b7/libs/s25client/s25client.cpp#L454) to support a new command line parameter. Only in case someone wants to use a fixed value.

Flamefire:
For the AI-Battle `RunProgram` is not used, so that wouldn't help there.

And for regular playing `rand()` is not only used for AI but also for drawing, which was the initial intention:

```
// Zufallsgenerator initialisieren (Achtung: nur für Animations-Offsets interessant, für alles andere
    // (spielentscheidende) wird unser Generator verwendet)
```
i.e. 
> Init RNG (Attention: Only used for animation offsets...)

Which became outdated as it is also used for sounds and AI.

I suggest to add a global RNG for AI, and maybe even the functions from `helpers/random.hpp` directly in the AI namespace to forward them using the AI RNG. I.e. something like `T randomValue(...) { return helpers::randomValue(getAiRng(), ...); }

This can the be seeded by the AI battle CLI and probably also in [`StartGame`](https://github.com/Return-To-The-Roots/s25client/blob/541bc230f33a9f39f2ce23b914c70e70e15da1b7/libs/s25main/network/GameClient.cpp#L282)

The AI code then needs to be updated to use those new functions.

It is still possible that the results are not fully deterministic as the time when the AI acts might depend on e.g. network packet order. It wasn't written to be deterministic. But I guess it will still work, at least for the AI battle.

libs/s25main/ai/aijh/AIConstruction.cpp
```diff
@@ -493,7 +495,7 @@ helpers::OptionalEnum<BuildingType> AIConstruction::ChooseMilitaryBuilding(const
         // Prüfen ob Feind in der Nähe
         if(milBld->GetPlayer() != playerId && distance < 35)
         {
-            int randmil = rand();
+            int randmil = AI::randomValue<int>(0, std::numeric_limits<int>::max());
             bool buildCatapult = randmil % 8 == 0 && aii.CanBuildCatapult()
```
Flamefire, line 499:
```suggestion
            const auto randmil = AI::randomValue(1, 8);
            bool buildCatapult = randmil == 1 && aii.CanBuildCatapult()
```

libs/s25main/ai/aijh/AIPlayerJH.cpp
```diff
@@ -344,7 +344,7 @@ void AIPlayerJH::PlanNewBuildings(const unsigned gf)
         DistributeGoodsByBlocking(GoodType::Boards, 30);
         DistributeGoodsByBlocking(GoodType::Stones, 50);
         // go to the picked random warehouse and try to build around it
-        int randomStore = rand() % (storehouses.size());
+        int randomStore = AI::randomValue<int>(0, storehouses.size() - 1);
```
Flamefire, line 347:
```suggestion
        const auto randomStore = AI::randomValue<unsigned>(0u, storehouses.size() - 1u);
```

libs/s25main/ai/aijh/AIConstruction.cpp
```diff
@@ -465,10 +466,12 @@ helpers::OptionalEnum<BuildingType> AIConstruction::ChooseMilitaryBuilding(const
     const BuildingType biggestBld = GetBiggestAllowedMilBuilding().value();
 
     const Inventory& inventory = aii.GetInventory();
-    if(((rand() % 3) == 0 || inventory.people[Job::Private] < 15)
+    uint8_t playerId = aii.GetPlayerId();
+    if((AI::randomValue<int>(0, 2) == 0 || inventory.people[Job::Private] < 15)
```
Flamefire, line 470:
```suggestion
    if(inventory.people[Job::Private] < 15 || AI::randomValue(0, 2) == 0)
```

libs/s25main/ai/aijh/AIConstruction.cpp
```diff
@@ -465,10 +466,12 @@ helpers::OptionalEnum<BuildingType> AIConstruction::ChooseMilitaryBuilding(const
...
        && (inventory.goods[GoodType::Stones] > 6 || bldPlanner.GetNumBuildings(BuildingType::Quarry) > 0))
         bld = BuildingType::Guardhouse;
-    if(aijh.getAIInterface().isHarborPosClose(pt, 19) && rand() % 10 != 0 && aijh.ggs.isEnabled(AddonId::SEA_ATTACK))
+    if(aijh.getAIInterface().isHarborPosClose(pt, 19) && AI::randomValue<int>(0, 9) != 0
```
Flamefire, line 473:
```suggestion
    if(aijh.getAIInterface().isHarborPosClose(pt, 19) && AI::randomValue(0, 9) != 0
```

libs/s25main/ai/aijh/AIPlayerJH.cpp
```diff
@@ -1571,7 +1569,7 @@ void AIPlayerJH::TryToAttack()
 
     // shuffle everything but headquarters and harbors without any troops in them
     std::shuffle(potentialTargets.begin() + hq_or_harbor_without_soldiers, potentialTargets.end(),
-                 std::mt19937(std::random_device()()));
+                 std::mt19937(AI::randomValue<unsigned>(0, 2047)));
```
Flamefire, line 1572:
```suggestion
                 AI::getRandomGenerator());
```

libs/s25main/ai/aijh/AIPlayerJH.cpp
```diff
@@ -1704,7 +1702,8 @@ void AIPlayerJH::TrySeaAttack()
     unsigned limit = 15;
     unsigned skip = 0;
     if(searcharoundharborspots.size() > 15)
-        skip = std::max<int>(rand() % (searcharoundharborspots.size() / 15 + 1) * 15, 1) - 1;
+        skip =
+          std::max<int>(AI::randomValue<int>(0, static_cast<int>(searcharoundharborspots.size() / 15) * 15), 1) - 1;
```
Flamefire, line 1706: This isn't the same.
16: -> `std::max<int>(rand(0,1) * 15, 1) - 1; -> 0 or 14`
30: -> `std::max<int>(rand(0,2) * 15, 1) - 1; -> 0, 14, 29`

The logic doesn't seem to make sense to me. It seems it wants one of the chunks of size 15, i.e. `[0,14)`, `[15, 29)` etc.

Maybe this is intended:
```suggestion
        skip = AI::randomValue(0u, searcharoundharborspots.size() / 15u) * 15u;
```
so for 16-29 we get either 0 or 15 and for 30-44 we get 0, 15, or 30

libs/s25main/ai/random.h
```diff
@@ -0,0 +1,20 @@
...
+
+#include "helpers/random.h"
+#include <random>
+
+namespace AI {
+
+std::minstd_rand& getRandomGenerator();
+
+template<typename T>
+T randomValue(T min = std::numeric_limits<T>::min(), T max = std::numeric_limits<T>::max())
```
Flamefire, line 15: Maybe a `bool randomBool` and `size_t randomIndex` makes sense to simplify code using `rand() % 2 == 0` and `randomValue(0, xx.size() - 1)`

Or instead of `randomBool` a `bool random(unsigned total, unsigned chance=1) -> randomValue(0u,total - 1u) <= chance`.   
Not sure I got the names and values right but it should replace `rand() % 10 == 0` by `random(10)` and `rand() % 10 > 5` by `random(10, 5)` and similar. This is the most common use so for. I.e. a "chance in total" chance to do something. Needs an assertion that `total > 0` and `chance < total` (a 5 out of 5 chance is a certain event). Or just make those 2 cases return false and true respectively

libs/s25main/ai/random.cpp
```diff
@@ -0,0 +1,16 @@
...
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#include "ai/random.h"
+
+namespace AI {
+
+std::minstd_rand& getRandomGenerator()
+{
+    std::random_device rnd_dev;
+    static std::minstd_rand rng(rnd_dev());
```
Flamefire, line 12: This would construct a random_device even when not required. Not sure if this does anything but better be sure:
```suggestion
    static std::minstd_rand rng(std::random_device{}());
```

libs/s25main/ai/aijh/AIConstruction.cpp
```diff
@@ -465,10 +466,12 @@ helpers::OptionalEnum<BuildingType> AIConstruction::ChooseMilitaryBuilding(const
     const BuildingType biggestBld = GetBiggestAllowedMilBuilding().value();
 
     const Inventory& inventory = aii.GetInventory();
-    if(((rand() % 3) == 0 || inventory.people[Job::Private] < 15)
+    uint8_t playerId = aii.GetPlayerId();
```
Flamefire, line 469: Why this addition?
```suggestion
```

Flamefire (changes requested):
Looking good. Made a few suggestions. Can you check in general that the template param of `randomValue` is omitted if possible and the result a `const auto` if it isn't and `const` otherwise?

The suggested helper functions should make this much easier and the code more readable.

Suggesting this here because I think it is easier to verify in the original change than to verify this and then the change to the convenience functions again

libs/s25main/ai/aijh/AIConstruction.cpp
wichern: randmil is used again in below lines. That's why we use max() here.

libs/s25main/ai/random.h
```diff
@@ -0,0 +1,41 @@
...
+{
+    return helpers::randomValue(getRandomGenerator(), min, max);
+}
+
+// Return a random bool:
+//   random()           ... will return true|false with 50% chance each
+//   random(15)         ... will return true in 1/15 of the cases
```
Flamefire, line 22: So the intended semantic is: Return true in 1 out of N cases? Then for the 50% case you'd want `2` not `1`, don't you?

libs/s25main/ai/random.h
```diff
@@ -0,0 +1,41 @@
...
+inline bool random(unsigned total = 1u)
+{
+    return helpers::randomValue(getRandomGenerator(), 0u, total) == 0u;
```
Flamefire, line 25: If the intention is "1 out of `total` cases" you need 
```suggestion
    return randomValue(1u, total) == 1u;
```
(with less code duplication using the function above)

Remember that the parameters are `minValue, maxValue`, inclusive! Otherwise generation a `UINT_MAX` would be impossible.

And I'd suggest to use the below overload instead if that already means `chance out of total` as you simply have `chance == 1`. Hence the suggestion to use a default parameter instead of an overload.

Or is there an advantage in having 2?

libs/s25main/ai/random.h
```diff
@@ -0,0 +1,41 @@
...
+}
+
+//   random(20, 5)      ... will return true in ~3/4 of the cases (random(20) > 5)
+inline bool random(unsigned total, unsigned chance)
+{
+    return helpers::randomValue(getRandomGenerator(), 0u, total) > chance;
```
Flamefire, line 31: - For 20, 5 we get values 0, 1, 2, ..., 20, i.e. 21 values 
- We succeed for all but 6 (0..5)
- --> Chance of 15 out of 21

So basically the semantic isn't "succeed for chance out of total cases" but "fail for chance out of total cases" (with an off-by-one error). IMO this is hard to understand at the call site.

```suggestion
    return helpers::randomValue(getRandomGenerator(), 1u, total) <= chance;
```

- Values 1, 2, ... 20, i.e. 20 values
- Succeed for 1,2,3,4,5
- --> 5 out of 20

Flamefire (changes requested):

libs/s25main/ai/random.cpp
```diff
@@ -0,0 +1,15 @@
...
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#include "ai/random.h"
+
+namespace AI {
+
+std::minstd_rand& getRandomGenerator()
+{
+    static std::minstd_rand rng(std::random_device{}());
```
Flow86, line 11: are the std rngs defined to have the same randomeness between platforms now? Since for `rand()` that was implementation specific (that was the reason why we built our own)

libs/s25main/ai/aijh/AIConstruction.cpp
Flamefire: I'd be explicit here and use the "chance"-type calls. A little bit wasteful but a bit clearer, isn't it?

Otherwise just use `AI::randomValue<int>()`

libs/s25main/ai/random.cpp
Flamefire: The rng yes. There might be differences in the distribution implementations leading to slightly different results or something like that. But IMO we don't really need that here: It is reproducible when run on the same platform which is enough for the current use case, isn't it?

libs/s25main/ai/random.h
wichern: No, with the `<= chance` check there is no need for two functions anymore.

libs/s25main/ai/random.cpp
wichern: In my understanding `std::minstd_rand` is using a fixed deterministic algorithm which should produce the same values on all platforms. I did not test this though and it is not important for the AI in my opinion.

Flamefire: Correct. It is just the distribution classes which are implementation defined. Found a [SO post](https://stackoverflow.com/a/74032845/1930508) confirming that.

But still yes: Fine for this use case.

libs/s25main/ai/aijh/AIPlayerJH.cpp
```diff
@@ -344,9 +344,8 @@ void AIPlayerJH::PlanNewBuildings(const unsigned gf)
         DistributeGoodsByBlocking(GoodType::Boards, 30);
         DistributeGoodsByBlocking(GoodType::Stones, 50);
         // go to the picked random warehouse and try to build around it
-        int randomStore = rand() % (storehouses.size());
         auto it = storehouses.begin();
-        std::advance(it, randomStore);
+        std::advance(it, AI::randomIndex(storehouses));
```
Flamefire, line 348: If you want you can replace this using (a wrapper of) `helpers::getRandomElement`. But can be done later.

Flamefire (changes requested):

libs/s25main/ai/aijh/AIConstruction.cpp
```diff
@@ -478,7 +479,7 @@ helpers::OptionalEnum<BuildingType> AIConstruction::ChooseMilitaryBuilding(const
     {
         if(aijh.UpdateUpgradeBuilding() < 0 && bldPlanner.GetNumBuildingSites(biggestBld) < 1
            && (inventory.goods[GoodType::Stones] > 20 || bldPlanner.GetNumBuildings(BuildingType::Quarry) > 0)
-           && rand() % 10 != 0)
+           && !AI::random(9u))
```
wichern, line 482: Shouldn't this be random(10u)?

Flamefire: Yes. See my [previous comment](https://github.com/Return-To-The-Roots/s25client/pull/1739#discussion_r1956110989) which includes a "recipe" for translating those. Didn't want to annotate all of them to avoid flooding this PR too much ;-)

Flamefire (approved):
I made the small adjustments suggested before and double checked all replacements of `rand()` fixing ~2 places where it was off.
Will merge once CI passes

libs/s25main/ai/aijh/AIPlayerJH.cpp
```diff
@@ -344,10 +344,8 @@ void AIPlayerJH::PlanNewBuildings(const unsigned gf)
         DistributeGoodsByBlocking(GoodType::Boards, 30);
         DistributeGoodsByBlocking(GoodType::Stones, 50);
         // go to the picked random warehouse and try to build around it
-        int randomStore = rand() % (storehouses.size());
-        auto it = storehouses.begin();
-        std::advance(it, randomStore);
-        const MapPoint whPos = (*it)->GetPos();
+        const auto storehouse = AI::randomElement(storehouses);
```
Flow86: `const auto*` (tidy is complaining)

libs/s25main/ai/aijh/AIPlayerJH.cpp
```diff
@@ -1396,21 +1384,19 @@ void AIPlayerJH::HandleLostLand(const MapPoint pt)
...
+    const auto* upgradeBld = UpdateUpgradeBuilding();
+    unsigned count = 0;
     const std::list<nobMilitary*>& militaryBuildings = aii.GetMilitaryBuildings();
     for(const nobMilitary* milBld : militaryBuildings)
     {
-        if(count != upb) // not upgrade building
+        if(milBld != upgradeBld) // not upgrade building
         {
-            if(upb >= 0) // we do have an upgrade building
+            if(upgradeBld) // we do have an upgrade building
```
Flow86: this is weird for `upgradeBld == nullptr`?

Flamefire: Why? `upgradeBld == nullptr` is a valid case

Flow86: can't we combine both ifs?

Flamefire: We can pull out `if(!upgradeBld)` because that implies `milBld != upgradeBld`.
Also `if(milBld->GetFrontierDistance() != FrontierDistance::Far)` can be checked first to avoid checking it twice

Flamefire: The clang warning about `milBld` being NULL is annoying though. Let's see if this fixes it already.

Otherwise we can use `std::reference_wrapper` for the building register which would be cleanest for the stored elements which are never null. Or `gsl::non_null` but pulling in GSL means another dependency, even though header-only

Flamefire (approved):

Issue #1742, opened 2025-02-21
Ship register window issue - content displayed when window rolled up.


Hirotaro:
When we hide / roll-up Ship register window, the content is still being displayed. 

Example 1 - empty list:
![Image](https://github.com/user-attachments/assets/b3f3d715-2236-401f-973c-e7305befb9d9)
![Image](https://github.com/user-attachments/assets/b989ddbc-54f9-4934-bef0-dca3ff041e78)

Example 2 - with ship
![Image](https://github.com/user-attachments/assets/48db2e93-42cf-46f0-a4eb-922c472b091c)
![Image](https://github.com/user-attachments/assets/6ee303c4-bb76-486e-b405-b017305449d8)

PR #1745, opened 2025-02-21
Don't draw content of ship window if minimized 

Flamefire:
Fixes https://github.com/Return-To-The-Roots/s25client/issues/1742

To avoid this whole class of issues introduce `IngameWindow::DrawContent` to allow  subclasses to handle only drawing its own content not also
checking whether they actually should or the window is minimized.

Also move the logic out of iwObservate::Draw to a timer. No need to update every frame, every second is enough.

### DIFF
--- a/libs/s25main/ingameWindows/IngameWindow.h
+++ b/libs/s25main/ingameWindows/IngameWindow.h
@@ -104,6 +104,8 @@ public:
 
 protected:
     void Draw_() override;
+    /// Called when not minimized after the frame and background have been drawn
+    virtual void DrawContent() {}
 
     /// Verschiebt Fenster in die Bildschirmmitte
     void MoveToCenter();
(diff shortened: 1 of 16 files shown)


Flow86 (dismissed):

libs/s25main/ingameWindows/IngameWindow.h
```diff
@@ -104,6 +104,8 @@ class IngameWindow : public Window
 
 protected:
     void Draw_() override;
```
Flow86, line 106: should Draw_ be final here then?

Flamefire: I can still imagine some Windows wanting to override this hence I left it. But I just checked and we don't used that right now, so made final.

BTW: The Jenkins build didn't seem to start. Do you know why?

Flow86: the jenkins cloud plugin was changed I had no time yet to fix the configuration :-/

Flow86 (approved):

PR #1759, opened 2025-04-19
Fix failure to detect failures when loading (Lua) game data

Flamefire:
One Windows CI was failing with a strange error that I cannot reproduce locally.

After a lot investigation I conclude that it is an optimization issue. Example:

```
  main: include(foo)
    foo: include(bar)
      bar: error out
    foo stores error from including bar
  main stores (no) error from including foo
```

The error storing happens as `anyError |= !load(...)`. During optimization I assume it loads `anyError` before calling `load` which down the chain sets `anyError` to true but because `main` does not get an error from `foo` and preloaded the "false" we store false again overwriting the `true`.

But actually the error should propagate upwards: When main includes foo which includes bar and that aborts, then foo should also abort and not continue and same for main.

So the fix here is to throw an error if the `loadScript` on an `include` fails.   
I played a bit with having `loadScript` throw an exception instead of returning true/false to "accumulate" error description. However as Lua errors (including exceptions) are logged we'll already have that in the log:
``` 
Include file 'bar' cannot be included
Include file 'foo' cannot be included
```

The caller of `GameData::Load` can then decide what to do with the failed main just like with any other error, e.g. syntax errors.

Includes a consistency fix and adds a missing test for game data loading (an error case actually)

Closes #1762

### DIFF
--- a/libs/libGamedata/lua/GameDataLoader.cpp
+++ b/libs/libGamedata/lua/GameDataLoader.cpp
@@ -72,7 +72,7 @@ void GameDataLoader::Include(const std::string& filepath)
     {
         constexpr int maxIncludeDepth = 10;
         // Protect against cycles and stack overflows
-        if(++curIncludeDepth_ >= maxIncludeDepth)
+        if(curIncludeDepth_ >= maxIncludeDepth)
             throw LuaIncludeError(helpers::format("Maximum include depth of %1% is reached!", maxIncludeDepth));
         const auto isAllowedChar = [](const char c) {
             return (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9') || c == '_' || c == '/'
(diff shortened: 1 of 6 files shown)


tests/s25Main/network/testGameClient.cpp
```diff
@@ -48,19 +49,13 @@ MOCK_BASE_CLASS(MockClientInterface, ClientInterface)
     // LCOV_EXCL_STOP
 };
 
-class CustomUserMapFolderFixture
+class CustomUserMapFolderFixture : rttr::test::ConfigOverride
```
Flow86, line 52: this shouldnt be inheritance, better use composition

libs/libGamedata/lua/GameDataLoader.cpp
```diff
@@ -39,17 +38,7 @@ bool GameDataLoader::Load()
...
-    try
-    {
-        if(!loadScript(curFile_))
-            return false;
-    } catch(std::exception& e)
-    {
-        LOG.write("Failed to load game data!\nReason: %1%\nCurrent file being processed: %2%\n") % e.what() % curFile_;
-        return false;
-    }
-    return !errorInIncludeFile_;
```
Flow86, line 52: errorInIncludeFile_ is always false ?

Flamefire:
Ok it's working now. No idea why Jenkins hat issues with that done in one go.

Flamefire:
Close for #1764

PR #1762, opened 2025-04-23
Minor refactor of GameDataLoader

Flamefire:
Fix off-by-one in counting of includes

Remove a try-catch that cannot be reached.

Include refactoring commit from #1759 to reduce the size of that PR

### DIFF
--- a/libs/libGamedata/lua/GameDataLoader.cpp
+++ b/libs/libGamedata/lua/GameDataLoader.cpp
@@ -72,7 +72,7 @@ void GameDataLoader::Include(const std::string& filepath)
     {
         constexpr int maxIncludeDepth = 10;
         // Protect against cycles and stack overflows
-        if(++curIncludeDepth_ >= maxIncludeDepth)
+        if(curIncludeDepth_ >= maxIncludeDepth)
             throw LuaIncludeError(helpers::format("Maximum include depth of %1% is reached!", maxIncludeDepth));
         const auto isAllowedChar = [](const char c) {
             return (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9') || c == '_' || c == '/'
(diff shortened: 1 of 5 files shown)


libs/libGamedata/lua/GameDataLoader.cpp
```diff
@@ -40,15 +40,8 @@ bool GameDataLoader::Load()
...
-    {
-        if(!loadScript(curFile_))
-            return false;
-    } catch(const std::exception& e)
-    {
-        LOG.write("Failed to load game data!\nReason: %1%\nCurrent file being processed: %2%\n") % e.what() % curFile_;
+    if(!loadScript(curFile_))
         return false;
-    }
     return !errorInIncludeFile_;
```
Flow86, line 45: this is now always true?

Flamefire: This is a member variable which can be set by `Include()`. There is no logical change here except removing the try-catch which cannot be triggered because `loadScript` catches exceptions and returns False

Flamefire:
Again strange that Jenkins passes this. Anyway: Cause identified and fixed in #1764 which is based on this. As we have either Jenkins or GHA failing when using individual commits we have to ignore one of them if we don't want to merge all fixes at the same time. So my suggestion is to merge this PR even though it still fails on GHA with the current known failure.

libs/libGamedata/lua/GameDataLoader.cpp
Flow86: I still find it weird why its here in the first place if we reset it to false=noerror anyway, no matter what it was before.

Flamefire: The intention is:
- call `loadGameData`
- reset include-error
- run LUA code
- LUA code may include other files (calling into the C++ code to do that)
- Any of those includes at any level may fail but the remaining LUA code continues to execute because the including is itself done with `loadScript` which doesn't throw
- Eventually this `loadScript` returns but has no information whether any of the includes failed as it only reports `false` when an exception is thrown and caught.

Hence the need to set the flag in the `Include` function when including failed.

And yes there is an obvious issue: Why do we continue running the LUA when it failed the `include`? It might not do what it expected and if we abort it on error this function will return `false` and we don't need the flag. I've done this in #1764

PR #1764, opened 2025-04-23
Fix failure to detect failures when loading (Lua) game data

Flamefire:
One Windows CI was failing with a strange error that I cannot reproduce locally.

After a lot investigation I conclude that it is an optimization issue. Example:

```
  main: include(foo)
    foo: include(bar)
      bar: error out
    foo stores error from including bar
  main stores (no) error from including foo
```

The error storing happens as `anyError |= !load(...)`. During optimization I assume it loads `anyError` before calling `load` which down the chain sets `anyError` to true but because `main` does not get an error from `foo` and preloaded the "false" we store false again overwriting the `true`.

But actually the error should propagate upwards: When main includes foo which includes bar and that aborts, then foo should also abort and not continue and same for main.

So the fix here is to throw an error if the `loadScript` on an `include` fails.   
I played a bit with having `loadScript` throw an exception instead of returning true/false to "accumulate" error description. However as Lua errors (including exceptions) are logged we'll already have that in the log:
``` 
Include file 'bar' cannot be included
Include file 'foo' cannot be included
```

The caller of `GameData::Load` can then decide what to do with the failed main just like with any other error, e.g. syntax errors.

Includes a consistency fix and adds a missing test for game data loading (an error case actually)

@Flow86 if you want clean MRs the merge order would be:
- [x] #1762 
- [x] #1766 
- This

### DIFF
--- a/libs/libGamedata/lua/GameDataLoader.cpp
+++ b/libs/libGamedata/lua/GameDataLoader.cpp
@@ -84,11 +84,12 @@ void GameDataLoader::Include(const std::string& filepath)
         const auto oldCurFile = curFile_;
         curFile_ = absFilePath;
         ++curIncludeDepth_;
-        if(!loadScript(absFilePath))
-            throw std::runtime_error(helpers::format("Include file '%1%' cannot be included", filepath));
+        const bool fileLoaded = loadScript(absFilePath);
         curFile_ = oldCurFile;
         RTTR_Assert(curIncludeDepth_ > 0);
         --curIncludeDepth_;
+        if(!fileLoaded)
+            throw std::runtime_error(helpers::format("Include file '%1%' cannot be included", filepath));
     } catch(const LuaIncludeError& e)
     {
         throw std::runtime_error(helpers::format("Include file '%1%' cannot be included: %2%", filepath, e.what()));
--- a/libs/libGamedata/lua/GameDataLoader.cpp
+++ b/libs/libGamedata/lua/GameDataLoader.cpp
@@ -20,8 +20,7 @@
 namespace bfs = boost::filesystem;
 
 GameDataLoader::GameDataLoader(WorldDescription& worldDesc, const boost::filesystem::path& basePath)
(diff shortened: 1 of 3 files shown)


libs/libGamedata/lua/GameDataLoader.cpp
```diff
@@ -94,7 +83,9 @@ void GameDataLoader::Include(const std::string& filepath)
             throw LuaIncludeError("File is outside the lua data directory!");
         const auto oldCurFile = curFile_;
         curFile_ = absFilePath;
-        errorInIncludeFile_ |= !loadScript(absFilePath);
+        ++curIncludeDepth_;
+        if(!loadScript(absFilePath))
+            throw std::runtime_error(helpers::format("Include file '%1%' cannot be included", filepath));
```
Flow86, line 88: why is this not a LuaIncludeError?

Flamefire: Because I wanted a different error. `LuaIncludeError` here is more like `optional<string> = try_include()` But here there is no additional error message because `loadScript` or inside its Lua execution already does all reporting. So we just need to pass the error up.

Flamefire:
What is wrong with Jenkins? How could that fail now?

Flow86:
>  unknown location(0): error: in "GameClientTests/ClientFollowsConnectProtocol/_0": unexpected call: callbacks.MockClientInterface::CI_NextConnectState( ? )

I don't know what the tests are doing.

Is there a 32/64bit mismatch somewhere?

Flamefire:
> I don't know what the tests are doing.
> 
> Is there a 32/64bit mismatch somewhere?

I added output of the enum values passed to replace the "?" by the value. Turns out it was a kind of race: If the test ran first it wouldn't reuse a map downloaded in another test and pass. Otherwise it skips a connect step and fails. No idea why this only ever happend on Jenkins.

I added a fix commit to use a new temporary folder and a separate test path for the caching behavior which makes sense to test too. Bonus there is a much easier to read sequence of expected events as the "side-checks" are tested already right before

PR #1773, opened 2025-05-31
Clean up control ids in lobby desktop

tehKaiN:
Split from #1534. Further ids for the lobby desktop will be introduced in the portait PR. I've created separate PR to limit the diff.

The idea was also to have options desktop id cleanup, but I see that it was already made in different commit.


libs/s25main/desktops/dskGameLobby.cpp
```diff
@@ -48,15 +48,15 @@
...
-    ID_PLAYER_GROUP_START = 50,
-    ID_SWAP_BUTTON = 80,
-    ID_FIRST_FREE = ID_SWAP_BUTTON + MAX_PLAYERS,
-    ID_GAME_CHAT,
-    ID_LOBBY_CHAT,
-    ID_CHAT_INPUT,
-    ID_CHAT_TAB,
-    TAB_GAMECHAT,
-    TAB_LOBBYCHAT
+    ID_groupPlayerStart = 50,
```
Flamefire: Just for consistency to what we use in other places:
```suggestion
    ID_grpPlayerStart = 50,
```
If you add IDs for the text controls we don't need the magic number

libs/s25main/desktops/dskGameLobby.cpp
```diff
@@ -48,15 +48,15 @@
...
+    ID_btnSwap = 80,
+    ID_FIRST_FREE = ID_btnSwap + MAX_PLAYERS,
+    ID_chatGame,
+    ID_chatLobby,
+    ID_edtChatMsg,
+    ID_optChatTab,
+    ID_btChatGame,
+    ID_btChatLobby
```
Flamefire: Moving them around should be easier:
```suggestion
    ID_chatGame,
    ID_chatLobby,
    ID_edtChatMsg,
    ID_optChatTab,
    ID_btChatGame,
    ID_btChatLobby
    ID_btnSwap, // MAX_PLAYERS consecutive controls
```

libs/s25main/desktops/dskGameLobby.cpp
```diff
@@ -48,15 +48,55 @@
...
+    ID_chkReady = 6,
+    ID_txtPing = 7,
+    ID_cbMove = 8,
+    ID_mbLuaLoadError = 1,
+    ID_mbLuaVersionError = 1,
+    ID_mbMapLoadError = 0,
+    ID_mbError = 0,
+    ID_mbStartErrror = 10,
+    ID_mbQuestionEconomy = 10,
+    ID_mbQuestionPeaceful = 11,
```
tehKaiN: I've searched for magic numbers across the code and replaced them through the code, but I'm still not sure if I haven't missed anything. I'll remove them after checking them once again during week.

libs/s25main/desktops/dskGameLobby.cpp
```diff
@@ -301,9 +322,9 @@ dskGameLobby::dskGameLobby(ServerType serverType, std::shared_ptr<GameLobby> gam
     {
         for(unsigned i = 0; i < gameLobby_->getNumPlayers(); i++)
         {
-            DrawPoint rowPos = GetCtrl<Window>(ID_PLAYER_GROUP_START + i)->GetCtrl<Window>(1)->GetPos();
-            ctrlButton* bt = AddTextButton(ID_SWAP_BUTTON + i, DrawPoint(5, 0), Extent(22, 22), TextureColor::Red1,
-                                           _("-"), NormalFont);
+            DrawPoint rowPos = GetCtrl<Window>(ID_grpPlayerStart + i)->GetCtrl<Window>(1)->GetPos(); // 1??
```
tehKaiN, line 325: Not sure what's going on here - is that `1` a different id?

libs/s25main/desktops/dskGameLobby.cpp
```diff
@@ -386,22 +407,22 @@ void dskGameLobby::UpdatePlayerRow(const unsigned row)
...
 
     // Spielername, beim Hosts Spielerbuttons, aber nich beim ihm selber, er kann sich ja nich selber kicken!
     if(gameLobby_->isHost() && !player.isHost && IsChangeAllowed("playerState"))
-        group->AddTextButton(1, DrawPoint(30, cy), Extent(200, 22), tc, name, NormalFont);
+        group->AddTextButton(ID_btReady, DrawPoint(30, cy), Extent(200, 22), tc, name, NormalFont);
     else
-        group->AddTextDeepening(1, DrawPoint(30, cy), Extent(200, 22), tc, name, NormalFont, COLOR_YELLOW);
-    auto* text = group->GetCtrl<ctrlBaseText>(1);
+        group->AddTextDeepening(ID_btReady, DrawPoint(30, cy), Extent(200, 22), tc, name, NormalFont, COLOR_YELLOW);
+    auto* text = group->GetCtrl<ctrlBaseText>(1); // 1??
```
tehKaiN, line 425: ditto

tehKaiN:
There were 2 comments here, they're gone now. My brain is slow today, but I think I've got most of it right. I'll look at it again during the week though.

libs/s25main/desktops/dskGameLobby.cpp
```diff
@@ -222,10 +244,9 @@ dskGameLobby::dskGameLobby(ServerType serverType, std::shared_ptr<GameLobby> gam
...
     }
 
-    // "Geschwindigkeit"
-    AddText(33, DrawPoint(400, 315), _("Speed:"), COLOR_YELLOW, FontStyle{}, NormalFont);
-    combo =
-      AddComboBox(43, DrawPoint(600, 310), Extent(180, 20), TextureColor::Grey, NormalFont, 100, !gameLobby_->isHost());
+    AddText(ID_txtSpeed, DrawPoint(400, 315), _("Speed:"), COLOR_YELLOW, FontStyle{}, NormalFont);
+    combo = AddComboBox(ID_cbSpeed, DrawPoint(600, 310), Extent(180, 20), TextureColor::Grey, NormalFont, 100,
+                        !gameLobby_->isHost());
     combo->AddString(_("Very slow")); // Sehr Langsam
```
Flamefire: BTW: Feel free to remove such comments that literally state the same as the code while modifying this or very adjacent code.

libs/s25main/desktops/dskGameLobby.cpp
Flamefire: IDs are only unique to the parent and unrelated to those of the parents parent and hence could be reused. So this is `ID_btReady` and yes it is repeated for each group, i.e. each player row

libs/s25main/desktops/dskGameLobby.cpp
tehKaiN: I tried to do so, but some must've slipped through the fingers. Will do another pass through that.

Flamefire: No worries, it's not related to this PR, just wanted to make sure you know this is fine if done.

tehKaiN: Removed some comments, but some seem to be useful - it'd be great to have them translated at some point.

tehKaiN:
I've Force-pushed initial commit because I've found that for some strange reason I've pushed a stray keypress that prevented compiling.

I've went through the magic values again and it seems that I've got all of them. I've also briefly playtested the unlimited play / lan dialogs and they seem to work fine.

libs/s25main/desktops/dskGameLobby.cpp
```diff
@@ -48,15 +48,55 @@
...
+    ID_txtMapName,
+    ID_miniMap,
+    ID_btPlayerState,
+    ID_btNation,
+    ID_btColor,
+    ID_btTeam,
+    ID_chkReady,
+    ID_txtPing,
+    ID_cbMove,
+    ID_mbLuaLoadError = 100,
```
Flamefire: Any reason for this value?

tehKaiN: Damn, I was sure I removed it - I added it during partial value removal to make sure that they won't collide with anything.

Flamefire (approved):
Looks good to me, thanks!

PR #1778, opened 2025-06-11
libutil: use GetUserNameExW to query user name on windows

tehKaiN:
Updates libutil submodule with https://github.com/Return-To-The-Roots/libutil/pull/26/

I guess this needs further updating after libutil is merged so that it points to proper commit hash

### DIFF
--- a/tools/release/bundleWinFiles.sh.cmake
+++ b/tools/release/bundleWinFiles.sh.cmake
@@ -69,7 +69,7 @@ DLL_BLACKLIST=(
   GDI32.dll KERNEL32.dll msvcrt.dll OPENGL32.dll USER32.dll	ADVAPI32.dll
   IPHLPAPI.DLL ole32.dll SHELL32.dll WS2_32.dll WINMM.DLL CRYPT32.dll
   wldap32.dll IMM32.DLL OLEAUT32.dll VERSION.dll api-ms-win-core-synch-l1-2-0.dll
-  SETUPAPI.DLL
+  SETUPAPI.DLL secur32.dll
 )
 is_blacklisted_dll()
 {


Flow86:
Secur32.dll needs to be added to the [blacklist](https://github.com/Return-To-The-Roots/s25client/blob/e6663858b7e702fef31bb66f18cd94d5c92ce49b/tools/release/bundleWinFiles.sh.cmake#L70)

Flamefire:
> I guess this needs further updating after libutil is merged so that it points to proper commit hash

You are right. Let's get that one merged

Flow86:
ok rebase here on the master submodule, and its good to go

tehKaiN:
the test failures don't look good - perhaps github actions don't like using that api function? perhaps `System::getUserName()` should fall back to non-ex version?

Flamefire:
I've found hints that this fails if the VM isn't fully set up. I guess you already tested it locally where it works?

Other fall back to first and lastname: https://github.com/ubuntu/WSL/blob/cd4e23dfd0967745d46bb5d42302bf17f91e703c/DistroLauncher/WindowsUserInfo.cpp#L78-L101

I guess for us the old variant is good enough.

tehKaiN:
Using commit on my private fork once again, so can't be merged until https://github.com/Return-To-The-Roots/libutil/pull/36 gets merged and this PR updated. Let's see if new code helps.

> Other fall back to first and lastname: https://github.com/ubuntu/WSL/blob/cd4e23dfd0967745d46bb5d42302bf17f91e703c/DistroLauncher/WindowsUserInfo.cpp#L78-L101

I dunno, I get the feeling that if some environments don't support display names, then first/last names might be empty too. I think it's best to fall back to old stuff in case of failure.

I've split it to separate fns to make the `System::getUserName()` as least convoluted as possible - if you say that's too much I can remove the extra static fns and make the code more concise, but it won't be pretty.

Flamefire:
Agreed. Pr merged, can you update the submodule here?

PR #1782, opened 2025-06-25
Allow to disable statistics for eight player

ottml:
The statistics for the eight player can not be hidden by clicking the portrait of the player in the statistics window. This works for all other players

### DIFF
--- a/libs/s25main/ingameWindows/iwStatistics.cpp
+++ b/libs/s25main/ingameWindows/iwStatistics.cpp
@@ -172,7 +172,8 @@ void iwStatistics::Msg_ButtonClick(const unsigned ctrl_id)
         case 4:
         case 5:
         case 6:
-        case 7: // Spielerportraits
+        case 7:
+        case 8: // Spielerportraits
             activePlayers[ctrl_id - 1] = !activePlayers[ctrl_id - 1];
             break;
         case 25: // Hilfe


Flamefire (approved):
good catch!

libs/s25main/ingameWindows/iwStatistics.cpp
```diff
@@ -172,7 +172,8 @@ void iwStatistics::Msg_ButtonClick(const unsigned ctrl_id)
         case 4:
         case 5:
         case 6:
-        case 7: // Spielerportraits
+        case 7:
+        case 8: // Spielerportraits
```
Flow86, line 176: maybe do something like `if(ctrl_id - 1 < activePlayers.size())` ... (of course check for the unsigned overflow correctly etc)?
instead of the switch? so that it works for *any* player numbers?

Flamefire: We should also add an enum for the control IDs to this window as already introduced in others. Then we could first check the other IDs and then`if(ctrl_id - ID_firstPlayerPortrait < activePlayers.size())` which doesn't need an additional check if as unsigned arithmetic is defined as modulo 2^N, i.e. we would get a large value failing the check already

Flow86: yes sounds good

Issue #1788, opened 2025-08-03
Game crashes when S2 gamefiles not found


Farmer-Markus:
Using Archlinux, Gnome

When trying to start rttr I remember it to show an SDL Message box with a message to copy the original S2 files to the share/.../S2 folder.
But in the latest nightly the following happens:

```
Return To The Roots v20250726-4f7e8f93aa2f85183d40a017b2402ac23b7e0ce3
Compiled with GNU C++ version 10.2.1 20210110 for linux 64 Bit

Starting in "/home/markus/Downloads/s25rttr_20250726-4f7e8f93aa2f85183d40a017b2402ac23b7e0ce3-linux.x86_64/s25rttr_20250726"
Directory for user data (config etc.): "/home/markus/.s25rttr"
Suche nach Treibern in "/home/markus/Downloads/s25rttr_20250726-4f7e8f93aa2f85183d40a017b2402ac23b7e0ce3-linux.x86_64/s25rttr_20250726/lib/s25rttr/driver/video"
1 video drivers found!
Videotreiber "(SDL2) OpenGL via SDL2-Library" geladen
Fontconfig warning: using without calling FcInit()
That operation is not supported
OpenGL 4.6 wird unterstützt
Suche nach Treibern in "/home/markus/Downloads/s25rttr_20250726-4f7e8f93aa2f85183d40a017b2402ac23b7e0ce3-linux.x86_64/s25rttr_20250726/lib/s25rttr/driver/audio"
1 audio drivers found!
Audiotreiber "(SDL2) Audio via SDL2_mixer-Library" geladen

Starte Spiel
Lade "/home/markus/Downloads/s25rttr_20250726-4f7e8f93aa2f85183d40a017b2402ac23b7e0ce3-linux.x86_64/s25rttr_20250726/share/s25rttr/RTTR/assets/base/splash.bmp": fertig in 3ms
Exception caught: Datei oder Ordner nicht gefunden: "/home/markus/Downloads/s25rttr_20250726-4f7e8f93aa2f85183d40a017b2402ac23b7e0ce3-linux.x86_64/s25rttr_20250726/share/s25rttr/S2/GFX/PALETTE/PAL5.BBM"

Konnte "/home/markus/Downloads/s25rttr_20250726-4f7e8f93aa2f85183d40a017b2402ac23b7e0ce3-linux.x86_64/s25rttr_20250726/share/s25rttr/S2/GFX/PALETTE/PAL5.BBM" nicht laden
[...]

PR #1789, opened 2025-08-03
Fixed crash on linux

Farmer-Markus:
When RTTR failes to find the game files a SDL message box is created to ask the user to install the files.
But when passing a valid SDL_window to the message box, SDL tries to use tools like zenity or kdialog.
This libraries are not always installed on every linux distribution and the program just crashes.
Using nullptr instead of a window object makes SDL use an x11 backend to display the message box.
This should work for every x11 or wayland system.

At least it works for me (wayland)

Fixes https://github.com/Return-To-The-Roots/s25client/issues/1788

### DIFF
--- a/extras/videoDrivers/SDL2/VideoSDL2.cpp
+++ b/extras/videoDrivers/SDL2/VideoSDL2.cpp
@@ -232,7 +232,13 @@ void VideoSDL2::PrintError(const std::string& msg) const
 void VideoSDL2::ShowErrorMessage(const std::string& title, const std::string& message)
 {
     // window==nullptr is okay too ("no parent")
+#ifdef __linux__
+    // When using window, SDL will try to use a system tool like "zenity" which isn't always installed on every distro so rttr will crash
+    // But without a window it will use an x11 backend that should be available on x11 AND wayland for compatibility reasons
+    SDL_ShowSimpleMessageBox(SDL_MESSAGEBOX_ERROR, title.c_str(), message.c_str(), nullptr);
+#else
     SDL_ShowSimpleMessageBox(SDL_MESSAGEBOX_ERROR, title.c_str(), message.c_str(), window);
+#endif
 }
 
 void VideoSDL2::HandlePaste()
--- a/extras/videoDrivers/SDL2/VideoSDL2.cpp
+++ b/extras/videoDrivers/SDL2/VideoSDL2.cpp
@@ -233,8 +233,9 @@ void VideoSDL2::ShowErrorMessage(const std::string& title, const std::string& me
 {
     // window==nullptr is okay too ("no parent")
 #ifdef __linux__
-    // When using window, SDL will try to use a system tool like "zenity" which isn't always installed on every distro so rttr will crash
-    // But without a window it will use an x11 backend that should be available on x11 AND wayland for compatibility reasons
+    // When using window, SDL will try to use a system tool like "zenity" which isn't always installed on every distro
+    // so rttr will crash. But without a window it will use an x11 backend that should be available on x11 AND wayland
+    // for compatibility reasons
     SDL_ShowSimpleMessageBox(SDL_MESSAGEBOX_ERROR, title.c_str(), message.c_str(), nullptr);
 #else
     SDL_ShowSimpleMessageBox(SDL_MESSAGEBOX_ERROR, title.c_str(), message.c_str(), window);


Flamefire:
How exactly does the crash look like? Because using a nullpointer might open the window behind the main window causing confusion. So maybe we can detect the issue and use this as a fallback

Flamefire:
Can you update the submodule with the translation too? Just merged that

Farmer-Markus:
> How exactly does the crash look like? Because using a nullpointer might open the window behind the main window causing confusion. So maybe we can detect the issue and use this as a fallback

This one was the bug with the language stuff:
```
Starte Spiel
Lade "/home/markus/Projekte/s25client/build/temp/s25client/build/share/s25rttr/RTTR/assets/base/splash.bmp": fertig in 16ms
Exception caught: Datei oder Ordner nicht gefunden: "/home/markus/Projekte/s25client/build/temp/s25client/build/share/s25rttr/S2/GFX/PALETTE/PAL5.BBM"

Konnte "/home/markus/Projekte/s25client/build/temp/s25client/build/share/s25rttr/S2/GFX/PALETTE/PAL5.BBM" nicht laden
An exception occurred: boost::too_many_args: format-string referred to fewer arguments than were passed
RttR crashed. Backtrace:
0x559e23f69059
0x559e23f692e3
0x559e23f450d1
0x559e23f47f03
0x7f2051737675
0x7f2051737729
0x559e23f44a85
RttR ist abgestürzt. Bitte starten sie die Anwendung neu!


Press ENTER to close this window . . .
Getötet                   bin/s25client
```
And this with the error window:
```
Konnte "/home/markus/Projekte/s25client/build/temp/s25client/build/share/s25rttr/S2/GFX/PALETTE/PAL5.BBM" nicht laden


ERROR: Einige Dateien konnten nicht geladen werden.
Bitte prüfe ob Die Siedler II Gold-Edition in folgendem Verzeichnis installiert ist
/home/markus/Projekte/s25client/build/temp/s25client/build/share/s25rttr/S2
RttR crashed. Backtrace:
0x563a5bdad059
0x563a5bdad2e3
0x563a5bd890d1
0x563a5bd893aa
0x7f0ef224c540
0x7f0eef36a776
0x7f0eef36b8d8
0x7f0eef338723
0x7f0eef338a30
0x7f0ef209d89e
0x7f0ef20f6c14
0x563a5beb219c
0x563a5be9cf86
0x563a5be9cabf
0x563a5c180cfe
0x563a5c1403e2
0x563a5be03569
0x563a5bdb0cf6
0x563a5bd8b4b9
0x563a5bd8bc51
0x7f0ef2235675
0x7f0ef2235729
0x563a5bd88a85
RttR ist abgestürzt. Bitte starten sie die Anwendung neu!


Press ENTER to close this window . . .

^CDo you really want to terminate the program (y/n) : Do you really want to terminate the program (y/n) : y
y
Clearing textures: Finished in 0ms

```

Farmer-Markus:
The window will open fine if I had zenity installed. I've tested it

Flamefire (approved):
I don't understand why it would crash, but there doesn't seem to be an easy way out, so ok for that workaround

stefson:
I'm on linux with a self compiled binary, and have never hit this issue before the pullrequest was merged, but I also can't see any issue since I recompiled the git checkout with the pullrequest merged.

Farmer-Markus:
> I'm on linux with a self compiled binary, and have never hit this issue before the pullrequest was merged, but I also can't see any issue since I recompiled the git checkout with the pullrequest merged.

If you have a library installed to show this window nothing will crash! Most distros will have this preinstalled but for example archlinux (what I'm using) has noting preinstalled and this results in a crash

Flow86:
we maybe need zenity as a dependency, there are more people out there having a similar issue in different projects (e.g https://github.com/Keriew/augustus/issues/1153 )

Flamefire:
If this is the only place where its non-existance causes issues we shouldn't care much.

stefson:
> > I'm on linux with a self compiled binary, and have never hit this issue before the pullrequest was merged, but I also can't see any issue since I recompiled the git checkout with the pullrequest merged.
> 
> If you have a library installed to show this window nothing will crash! Most distros will have this preinstalled but for example archlinux (what I'm using) has noting preinstalled and this results in a crash

yeah, I briefly removed zenity for testing before I recompiled, and nothing happend. 

maybe its a better approach to strace the binary with the affected git checkout? but I'm not a unix wizzard (:

PR #1794, opened 2025-08-24
Fix lags in right-click scrolling for SDL2

Flamefire:
Avoid reporting duplicate mouse move events in SDL2

When warping the mouse a new event will be recorded with the SDL2
backend which might lead to handling that event multiple times.

For the right-click scrolling this leads to a potential infinite loop of
zero-sized movements.

Added an option to change the smart cursor setting in the ingame option window to test it which requires a tooltip for the checkbox due to the limited width.   
- Added support for that
- moved duplicated function used by that to `Window` class
- support chaining for some options to use a build-like pattern

### DIFF
--- a/libs/s25main/Window.h
+++ b/libs/s25main/Window.h
@@ -98,8 +98,10 @@ public:
     void LockRegion(Window* window, const Rect& rect);
     /// Gibt eine gesperrte Region wieder frei.
     void FreeRegion(Window* window);
-    /// Check if the gicen point is in a region locked by any window other than exception
+    /// Check if the given point is in a region locked by any window other than exception
     bool IsInLockedRegion(const Position& pos, const Window* exception = nullptr) const;
+    bool IsMouseOver() const;
+    bool IsMouseOver(const MouseCoords& mousePos) const;
 
     /// Set the position for the window
     void SetPos(const DrawPoint& newPos);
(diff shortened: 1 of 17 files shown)


libs/s25main/controls/ctrlCheck.h
```diff
@@ -5,21 +5,34 @@
 #pragma once
 
 #include "Window.h"
+#include "ctrlBaseTooltip.h"
+#include <string>
 class MouseCoords;
 class glFont;
 
-class ctrlCheck : public Window
+class ctrlCheck : public Window, ctrlBaseTooltip
```
Flow86: is `ctrlBaseTooltip` used at all? nothing is marked override of the new stuff?

Flamefire: It is used by the class similar to how it is used in others: On mouse move the tooltip is shown or hidden.

I removed the added method to change the tooltip and used the base method directly. I was just confused about how it worked... Enhanced the doc a bit

libs/s25main/Window.h
```diff
@@ -98,8 +98,10 @@ class Window
     void LockRegion(Window* window, const Rect& rect);
     /// Gibt eine gesperrte Region wieder frei.
     void FreeRegion(Window* window);
-    /// Check if the gicen point is in a region locked by any window other than exception
+    /// Check if the given point is in a region locked by any window other than exception
     bool IsInLockedRegion(const Position& pos, const Window* exception = nullptr) const;
+    bool IsMouseOver() const;
```
Flow86, line 104: can you add a comment on what those two methods are for?

Flamefire: Done. Also added a test for them

PR #1802, opened 2025-09-28
Update Kaguya tracking our fork

Flamefire:
This fixes problems with removed CMake version support and a missing include (cstdint)


Flow86:
is this something we can provide upstream?

Edit: ah you already have. well ...

PR #1804, opened 2025-10-06
Android build

Farmer-Markus:
I'm working on recreating my android port of rttr.
~~Currently the game starts but failes to find any drivers.~~

~~I would need an option to set a custom output path for the translation/data copying.~~
Also an enviroment variable to set the path to the drivers or the game data (or both)

Also another problem is openGl.
On android only openGl-Es is available and I've used a translation library ([gl4es](https://github.com/ptitSeb/gl4es)) that works pretty well. But the SDL videodriver would need: 1. to link against gl4es and 2. the SDL_gl_getprocaddress to be replaced by the gl4es equivalent.
This could also be used to support devices like the raspberry pi.

Also I would implement basic touch controls which should also work on touchscreen laptops (thanks to sdl)

I know these would be some pretty big changes...

### DIFF
--- a/libs/s25main/WindowManager.cpp
+++ b/libs/s25main/WindowManager.cpp
@@ -183,7 +183,8 @@ IngameWindow& WindowManager::DoShow(std::unique_ptr<IngameWindow> window, bool m
     SetActiveWindow(result);
 
     // Maus deaktivieren, bis sie losgelassen wurde (Fix des Switch-Anschließend-Drück-Bugs)
-    disable_mouse = mouse;
+    if(!VIDEODRIVER.IsTouch())
+        disable_mouse = mouse;
     return result;
 }
 


Farmer-Markus:
Changes:

- Added 'natural' mouse mode & changed invert mouse setting to mouseMode (0:default, 1:inverted, 2:natural)
- Changed invert mouse GUI setting (translations would need to be added/changed)
- Added 'GL4ES' to RTTR_OPENGL cmake options & VideoSDL2 driver to support OpenGL-ES
- Added 'RTTR_BUILD_LIB' cmake variable to build main executable as shared library (Needed to run on android)
- Added Touch controls to VideoSDL2 driver (moving map with one finger, two finger zooming, double tap to close window)
- Added allowing to control config paths using enviroment variables
- Some minor changes to fix compiling with newer versions of gcc and clang

Flow86:
Git History should be cleaned up (Some things should be squashed, so that some distinguish commits are there)

CMakeLists.txt
```diff
@@ -242,6 +242,7 @@ include(RttrCoverageCfg)
 include(EnableSanitizers)
 
 option(RTTR_USE_SYSTEM_LIBS "Default to using system-wide installed libs for external dependencies that have known system versions." OFF)
+option(RTTR_BUILD_LIB "Executables will be build as shared library." OFF)
```
Flow86, line 247: That's what I do not understand, why must the executable be build as shared library ?!?

copyDepsToBuildDir.cmake
```diff
@@ -5,7 +5,7 @@
 # Needs to be configured with @ONLY and called with -DCMAKE_BUILD_TYPE=...
 
 set(CMAKE_HOST_UNIX "@CMAKE_HOST_UNIX@")
-set(CMAKE_SOURCE_DIR "@CMAKE_SOURCE_DIR@")
+set(CMAKE_SOURCE_DIR "@CMAKE_CURRENT_SOURCE_DIR@")
```
Flow86: the variable(-name) should be replaced everywhere, otherwise this may lead to confusion

extras/videoDrivers/CMakeLists.txt
```diff
@@ -6,4 +6,4 @@ rttr_set_output_dir(RUNTIME ${RTTR_DRIVERDIR}/video)
 rttr_set_output_dir(LIBRARY ${RTTR_DRIVERDIR}/video)
 
 add_subdirectory(WinAPI)
-add_subdirectory(SDL2)
+add_subdirectory(SDL2)
```
Flow86: unrelated change

extras/videoDrivers/SDL2/CMakeLists.txt
```diff
@@ -20,4 +39,4 @@ if(SDL2_FOUND)
       LIBRARY DESTINATION ${RTTR_DRIVERDIR}/video
     )
     add_dependencies(drivers videoSDL2)
-endif()
+endif()
```
Flow86: unrelated change

extras/videoDrivers/SDL2/VideoSDL2.cpp
```diff
@@ -25,6 +26,9 @@
 #    include <windows.h> // Avoid "Windows headers require the default packing option" due to SDL2
 #    include <SDL_syswm.h>
 #endif // _WIN32
+#if RTTR_OGL_GL4ES == 1
```
Flow86: `RTTR_OGL_GL4ES` should be defined only if available, so `#ifdef RTTR_OGL_GL4ES` should be used?

extras/videoDrivers/SDL2/VideoSDL2.cpp
```diff
@@ -56,7 +60,11 @@ void FreeVideoInstance(IVideoDriver* driver)
 
 const char* GetDriverName()
 {
+#if RTTR_OGL_GL4ES == 1
```
Flow86: `RTTR_OGL_GL4ES` should be defined only if available, so `#ifdef RTTR_OGL_GL4ES` should be used?

extras/videoDrivers/SDL2/VideoSDL2.cpp
```diff
@@ -116,7 +126,7 @@ bool VideoSDL2::CreateScreen(const std::string& title, const VideoMode& size, bo
     CHECK_SDL(SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, RTTR_OGL_MAJOR));
     CHECK_SDL(SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, RTTR_OGL_MINOR));
     SDL_GLprofile profile;
-    if((RTTR_OGL_ES))
+    if((RTTR_OGL_ES) || (RTTR_OGL_GL4ES == 1))
```
Flow86: same here, even though that this is not a preprocessor define here.

@Flamefire  why?

extras/videoDrivers/SDL2/VideoSDL2.cpp
```diff
@@ -446,7 +500,11 @@ void VideoSDL2::ListVideoModes(std::vector<VideoMode>& video_modes) const
 
 OpenGL_Loader_Proc VideoSDL2::GetLoaderFunction() const
 {
+#if RTTR_OGL_GL4ES == 1
```
Flow86: `RTTR_OGL_GL4ES` should be defined only if available, so `#ifdef RTTR_OGL_GL4ES` should be used?

extras/videoDrivers/SDL2/VideoSDL2.cpp
```diff
@@ -482,7 +540,8 @@ void VideoSDL2::MoveWindowToCenter()
     SDL_Rect usableBounds;
     CHECK_SDL(SDL_GetDisplayUsableBounds(SDL_GetWindowDisplayIndex(window), &usableBounds));
     int top, left, bottom, right;
-    CHECK_SDL(SDL_GetWindowBordersSize(window, &top, &left, &bottom, &right));
+    if(SDL_GetWindowBordersSize(window, &top, &left, &bottom, &right) != 0)
```
Flow86: whats the difference to CHECK_SDL?

libs/driver/include/driver/MouseCoords.h
```diff
@@ -15,18 +15,19 @@
...
     {}
 
     Position pos;
-    bool ldown;     /// Linke Maustaste gedrückt
-    bool rdown;     /// Rechte Maustaste gedrückt
-    bool dbl_click; /// Linke Maustaste - Doppelklick
+    bool ldown;            /// Linke Maustaste gedrückt
+    bool rdown;            /// Rechte Maustaste gedrückt
+    bool dbl_click;        /// Linke Maustaste - Doppelklick
+    unsigned num_tfingers; /// Anzahl Finger auf dem Display
```
Flow86: when touching this, change the german comments to english ones directly please

libs/driver/include/driver/VideoDriver.h
```diff
@@ -27,6 +27,9 @@ class VideoDriver : public IVideoDriver
     /// Funktion zum Auslesen ob die Rechte Maustaste gedrückt ist.
     bool GetMouseStateR() const override;
 
+    /// Funktion zum Auslesen ob momentan touch verwendet wird.
```
Flow86: use english comment please

and for me its not really clear: what does "touch" mean?

libs/rttrConfig/src/RttrConfig.cpp
```diff
@@ -130,13 +130,25 @@ bool RttrConfig::Init()
...
+    pathMappings["DATA"] = getEnvOverride("DATA", RTTR_DATADIR);
+    pathMappings["GAME"] = getEnvOverride("GAME", RTTR_GAMEDIR);
+    pathMappings["LIB"] = getEnvOverride("LIB", RTTR_LIBDIR);
+    pathMappings["DRIVER"] = getEnvOverride("DRIVER", RTTR_DRIVERDIR);
+    pathMappings["RTTR"] = getEnvOverride("RTTR", RTTR_DATADIR "/RTTR");
+    pathMappings["USERDATA"] = getEnvOverride("USERDATA", RTTR_USERDATADIR);
     return true;
 }
+
+bfs::path RttrConfig::getEnvOverride(const std::string& id, const bfs::path& defaultPath)
```
Flow86, line 144: thats quiet a nice idea

libs/s25client/CMakeLists.txt
```diff
@@ -18,9 +18,20 @@ else()
     set(s25client_RC )
 endif()
 
-add_executable(s25client s25client.cpp commands.cpp ${s25client_RC})
-target_link_libraries(s25client PRIVATE s25Main Boost::program_options Boost::nowide rttr::vld)
-add_dependencies(s25client drivers)
+if(RTTR_BUILD_LIB)
```
Flow86, line 21: thats the point I do not understand? why does the application needs to be a shared library? O.o

libs/s25main/Settings.cpp
```diff
@@ -93,7 +93,11 @@ void Settings::LoadDefaults()
     }
     video.framerate = 0; // Special value for HW vsync
     video.vbo = true;
+#ifdef __ANDROID__
+    video.shared_textures = false;
+#else
     video.shared_textures = true;
+#endif
```
Flow86: put the ifdef around the value only, not the variable itself. this may otherwise lead to unwanted changes/behavior. (and yes its always ugly)

or make some constants in the beginning of the file in an anonymous namespace, e.g `SHARED_TEXTURES_DEFAULT` which is set to the appropriate value and used here

libs/s25main/Settings.cpp
```diff
@@ -141,7 +145,11 @@ void Settings::LoadDefaults()
     // interface
     // {
     interface.autosave_interval = 0;
-    interface.invertMouse = false;
+#ifdef __ANDROID__
+    interface.mouseMode = 2;
+#else
+    interface.mouseMode = 0;
+#endif
```
Flow86: the 2 and 0 are even used twice (see below)

make a named constant and use it. (we should have done that anyway already)

this also *f...ks* up existing configs. is there a way to have an proper migration path?

libs/s25main/Settings.cpp
```diff
@@ -307,7 +315,11 @@ void Settings::Load()
         // interface
         // {
         interface.autosave_interval = iniInterface->getIntValue("autosave_interval");
-        interface.invertMouse = iniInterface->getValue("invert_mouse", false);
+#ifdef __ANDROID__
+        interface.mouseMode = iniInterface->getValue("mouse_mode", 2);
+#else
+        interface.mouseMode = iniInterface->getValue("mouse_mode", 0);
+#endif
```
Flow86: the 2 and 0 are even used twice (see above)

make a named constant and use it. (we should have done that anyway already)

basically we should first try to get  `mouse_mode` and if it does not exist try to read `invert_mouse` instead.

libs/s25main/TerrainRenderer.cpp
```diff
@@ -219,6 +220,16 @@ void TerrainRenderer::UpdateVertexColor(const MapPoint pt, const GameWorldViewer
             // Unsichtbar -> schwarz
             GetVertex(pt).color = 0.0f;
             break;
+#if RTTR_OGL_GL4ES == 1
```
Flow86: `#ifdef`

can we define some constants instead duplicating a lot of code?

libs/s25main/TerrainRenderer.cpp
```diff
@@ -791,9 +803,14 @@ void TerrainRenderer::Draw(const Position& firstPt, const Position& lastPt, cons
         glColorPointer(3, GL_FLOAT, 0, &gl_colors.front());
     }
 
+#if RTTR_OGL_GL4ES == 1
+    // Gl4es does not like GL_COMBINE...
```
Flow86: is there a link/reason we can write down here?

libs/s25main/TerrainRenderer.cpp
```diff
@@ -791,9 +803,14 @@ void TerrainRenderer::Draw(const Position& firstPt, const Position& lastPt, cons
         glColorPointer(3, GL_FLOAT, 0, &gl_colors.front());
     }
 
+#if RTTR_OGL_GL4ES == 1
```
Flow86: `#ifdef`

libs/s25main/WindowManager.cpp
```diff
@@ -356,10 +356,22 @@ void WindowManager::Msg_LeftUp(MouseCoords mc)
...
+        else if(VIDEODRIVER.IsTouch()) // Fast unmöglich 2 mal auf den exakt selben punkt zu tippen
+        {
+            // Wenn doppeltippen -> fenster schließen
+            IngameWindow* window = FindWindowAtPos(mc.GetPos());
+            if(window && !window->IsPinned())
+                window->Close();
+        }
+    }
+
+    if(!mc.dbl_click)
```
Flow86: that logic is a bit weird. But I don't have a good idea here.

@Flamefire do you have an idea to improve this?

libs/s25main/desktops/dskGameInterface.cpp
```diff
@@ -692,37 +523,243 @@ bool dskGameInterface::Msg_LeftDown(const MouseCoords& mc)
...
+bool dskGameInterface::Msg_LeftUp(const MouseCoords& mc)
 {
     if(isScrolling)
     {
         StopScrolling();
         return true;
     }
+    // num_tfingers wird erst reduziert nachdem diese Funktion läuft :/
+    // Herausfinden ob kurz genug unten um context click auszuführen
+    if(mc.num_tfingers == 1 && (VIDEODRIVER.GetTickCount() - touchDuration) < 250)
```
Flow86: please create a named constant for the `250`

libs/s25main/desktops/dskGameInterface.cpp
```diff
@@ -692,37 +523,243 @@ bool dskGameInterface::Msg_LeftDown(const MouseCoords& mc)
...
-        return false;
+    {
+        if(mc.num_tfingers == 1)
+            Msg_RightDown(mc);
+        else
+            return false;
+    }
+
+    // Natural scrolling
+    if(SETTINGS.interface.mouseMode == 2)
```
Flow86: use a named constant for `2`, then the comment is superfluous

libs/s25main/desktops/dskGameInterface.cpp
```diff
@@ -692,37 +523,243 @@ bool dskGameInterface::Msg_LeftDown(const MouseCoords& mc)
...
+    {
+        const Position mapPos = gwv.ViewPosToMap(mc.GetPos());
+        gwv.MoveBy(-(mapPos - startScrollPt));
+        startScrollPt = mapPos;
+    } else
+    {
+        int acceleration = SETTINGS.global.smartCursor ? 2 : 3;
 
-    int acceleration = SETTINGS.global.smartCursor ? 2 : 3;
+        if(SETTINGS.interface.mouseMode == 1)
```
Flow86: same goes for the `1`

libs/s25main/desktops/dskGameInterface.cpp
```diff
@@ -692,37 +523,243 @@ bool dskGameInterface::Msg_LeftDown(const MouseCoords& mc)
...
     return true;
 }
 
 bool dskGameInterface::Msg_RightDown(const MouseCoords& mc)
 {
-    StartScrolling(mc.pos);
+    if(SETTINGS.interface.mouseMode == 2)
+        StartScrolling(gwv.ViewPosToMap(mc.GetPos()));
+    else
+        StartScrolling(mc.pos);
```
Flow86: what I do not understand is the difference - is `mc.pos` and `mc.GetPos` not the same?

Flow86 (changes requested):

extras/videoDrivers/SDL2/VideoSDL2.cpp
Flamefire: Actually I find `#if RTTR_OGL_GL4ES` better with it always being defined as 0 or 1.
This detects misspellings through the compiler warning.

In any case `==1` is superflous

extras/videoDrivers/SDL2/VideoSDL2.cpp
Flamefire: I can't remember. They are set in https://github.com/Return-To-The-Roots/s25client/blob/2d68497723d0c70db410cb7f945d881821cb4f9e/external/glad/openglCfg.hpp.cmake#L12 as defines.

Might have been prior to constexpr and to avoid the type name in each branch.

But don't see a particular reason and there won't be a difference in the binary eigher

extras/videoDrivers/SDL2/VideoSDL2.cpp
Flamefire: CHECK_SDL does `PrintError(SDL_GetError());`

I guess it should either throw an exception or return a boolean so one can do `if(!CHECK_SDL(...))`

libs/s25main/Settings.cpp
Flamefire: IMO for 1-line variables it is less ugly to repeat the name. Named constant might be good either way

libs/s25main/Settings.h
```diff
@@ -106,7 +106,7 @@ class Settings : public Singleton<Settings, SingletonPolicies::WithLongevity>
     struct
     {
         unsigned autosave_interval;
-        bool invertMouse;
+        unsigned mouseMode; // default/original:0 inverted:1 natural:2
```
Flamefire: This should be an enum then. Also what do those 3 setting mean?
> Map moves with your cursor when scrolling/panning

This sounds like just "inverted" (to original)

libs/s25main/TerrainRenderer.cpp
```diff
@@ -219,6 +220,16 @@ void TerrainRenderer::UpdateVertexColor(const MapPoint pt, const GameWorldViewer
...
             GetVertex(pt).color = 0.0f;
             break;
+#if RTTR_OGL_GL4ES == 1
+        case Visibility::FogOfWar:
+            // Fog of War -> abgedunkelt
+            GetVertex(pt).color = clr / 2.f;
+            break;
+        case Visibility::Visible:
+            // Normal sichtbar
+            GetVertex(pt).color = clr / 1.f;
```
Flamefire:
```suggestion
            GetVertex(pt).color = clr;
```

libs/s25main/WindowManager.cpp
```diff
@@ -356,10 +356,22 @@ void WindowManager::Msg_LeftUp(MouseCoords mc)
...
-    if(time_now - lastLeftClickTime < DOUBLE_CLICK_INTERVAL && mc.GetPos() == lastLeftClickPos)
+
+    // Sehr schwierig auf touch mit dem default dbl-click interval z.B. Mails zu löschen ohne das Fenster zu schließen
+    if(time_now - lastLeftClickTime < (VIDEODRIVER.IsTouch() ? DOUBLE_CLICK_INTERVAL / 3 : DOUBLE_CLICK_INTERVAL))
     {
-        mc.dbl_click = true;
-    } else
+        if(mc.GetPos() == lastLeftClickPos)
+            mc.dbl_click = true;
+        else if(VIDEODRIVER.IsTouch()) // Fast unmöglich 2 mal auf den exakt selben punkt zu tippen
```
Flamefire: So if you do hit the same point with touch it is a double click, else it is a right click which closes the window. This sounds surprising

libs/s25main/WindowManager.cpp
Flamefire: The previous logic not so much. The new one mixes double-click detection with window closing.

This should only be for double-click detection. For Touch you could have a "points-close" check. And in the window you could handle a double-click while touch as close command.

But the current solution stays weird. IIRC we have windows where you select an entry with double-click. Now this will close the window instead.

libs/s25main/Settings.h
Farmer-Markus: > This should be an enum then. Also what do those 3 setting mean?
> 
> > Map moves with your cursor when scrolling/panning
> 
> This sounds like just "inverted" (to original)

inverted is the same as the previous 'invert_mouse' setting (which I removed/replaced)

Natural is 1to1 moving the map with mouse like in most games
In rttr the map movement never was 'grab map and drag around'

libs/s25main/WindowManager.cpp
Farmer-Markus: > So if you do hit the same point with touch it is a double click, else it is a right click which closes the window. This sounds surprising

Aaah yeah
I'll work on that :D

libs/s25main/WindowManager.cpp
Farmer-Markus: > The previous logic not so much. The new one mixes double-click detection with window closing.
> 
> This should only be for double-click detection. For Touch you could have a "points-close" check. And in the window you could handle a double-click while touch as close command.
> 
> But the current solution stays weird. IIRC we have windows where you select an entry with double-click. Now this will close the window instead.

So what could I do?
Should the double click be handled in the window class and only close when IsTouch & DblClick is not needed to interact with that part of the window?
Which windows do even need a dbl click to work? Never seen one.

libs/s25client/CMakeLists.txt
Farmer-Markus: An android app can't just execute an ELF binary. (security reasons)
But you can use java to load a native library and then use the java jni(java native interface) to execute specific parts of the native library.
Luckily SDL2 already has a jni implementation in its SDL_main function which will start the main() function of the library.
Android java app calls -> SDL_main function which calls -> main() function in library (s25client entry point)

libs/s25main/desktops/dskGameInterface.cpp
Farmer-Markus: Yeah my bad

libs/driver/include/driver/VideoDriver.h
Farmer-Markus: > and for me its not really clear: what does "touch" mean?

If the current event was triggered by a touch event ( finger down, finger up, finder move )
This allows the game to be played with a touch screen and the switch over to mouse/keyboard without changing a setting or anything else. It just works fine for both variants

extras/videoDrivers/SDL2/VideoSDL2.cpp
Farmer-Markus: > same here, even though that this is not a preprocessor define here.

I can't use
`if((RTTR_OGL_ES) || (RTTR_OGL_GL4ES))`

This results in compiler errors (gcc and clang)

Farmer-Markus: Should I just change all of them to use proprocessor if statements?

libs/s25main/TerrainRenderer.cpp
Farmer-Markus: Not really. I could not find any similar problems with gl4es but trying around and testing lead to this solution.

libs/s25main/TerrainRenderer.cpp
Farmer-Markus: I could do something like this

```
namespace {
#if RTTR_OGL_GL4ES
constexpr float VISIBILITY_DIVISOR_FOG = 2.f;
constexpr float VISIBILITY_DIVISOR_VISIBLE 1.0f;
#else
constexpr float VISIBILITY_DIVISOR_FOG = 4.f;
constexpr float VISIBILITY_DIVISOR_VISIBLE = 2.f;
#endif
} // namespace
```

Is this better?

libs/s25main/Settings.h
Flamefire: But isn't inverted that grab map and move? Or does it mean it doesn't endless scroll as the cursor actually moves inside the window together with the map instead of staying where it is?
Maybe add a short description to the enum

Then maybe rename it to "mouseMapMoveMode" or something better. And if you can come up with enum names that convey that very succinct then it's perfect!

libs/s25main/WindowManager.cpp
Flamefire: > Which windows do even need a dbl click to work? Never seen one.

ctrlList, ctrlTable use it for selection.
Double-click on window titlebars maximizes/minimizes the window

So yes handling it in `IngameWindow` when not otherwise handled is the only sensible option.

IIRC the IngameWindow::MouseLeftUp won't be called if a control handled it. If that is the case you can add it there.

copyDepsToBuildDir.cmake
Flamefire: I don't get this change anyway. `CMAKE_CURRENT_SOURCE_DIR` might be something different depending on where this is configured from.

Maybe `RTTR_PROJECT_SOURCE_DIR` works?

Farmer-Markus: > Maybe `RTTR_PROJECT_SOURCE_DIR` works?

Great! I think that should work.

libs/s25main/Settings.h
Farmer-Markus: > But isn't inverted that grab map and move? Or does it mean it doesn't endless scroll as the cursor actually moves inside the window together with the map instead of staying where it is? Maybe add a short description to the enum
> 

Smart cursor: off
original:
    When moving mouse to left, the map moves to the right
inverted:
    When moving mouse to left, the map moves to the left
natural:
    When moving mouse to left, the map moves to the map moves to the left

Smart cursor on:
original, inverted:
    Same as above but mouse stays on the same position & I think the acceleration is higher
natural:
     Same as above
     
 Every mode moves the map faster than the cursor but natural lets the user drag the map
 
 > Then maybe rename it to "mouseMapMoveMode" or something better. And if you can come up with enum names that convey that very succinct then it's perfect!

Wouldn't "mouseMapMoveMode" be pretty long? this would result in mouse_map_move_mode in the config...
If you don't like mouseMode what do you think of mapScrollMode or something like this?

extras/videoDrivers/SDL2/VideoSDL2.cpp
Flamefire: > This results in compiler errors (gcc and clang)

Which ones? Can't see how

> Should I just change all of them to use proprocessor if statements?

might be easiest, yes.

As those are constants I'd declare them as such, like `constexpr SDL_GLprofile profile = ...;`

libs/driver/include/driver/VideoDriver.h
Flamefire: Then we need a better name. Because "Mouse touch" is weird. And for bools I'd use `Is` not `Get`. Although I see that you did it for consistency with the existing functions. Those should be `IsRightMouseBtnDown` or so...

Maybe `hasActiveTouchEvent`?

libs/rttrConfig/src/RttrConfig.cpp
```diff
@@ -130,13 +130,25 @@ bool RttrConfig::Init()
...
+{
+    bfs::path path = System::getPathFromEnvVar("RTTR_" + id + "_DIR");
+    if(!path.empty())
+    {
+        LOG.write("Note: %1% path manually set to %2%\n", LogTarget::Stdout) % id % path;
+    } else
+        return defaultPath;
+
+    return path;
```
Flamefire: nit: I'd try to avoid negated checks. Especially here it might be natural the other way round

```suggestion
    if(path.empty())
        return defaultPath;
    LOG.write("Note: %1% path manually set to %2%\n", LogTarget::Stdout) % id % path;
    return path;
```

libs/s25main/Settings.h
Flamefire: > this would result in mouse_map_move_mode in the config...

I don't have an issue with that at all. It has to be descriptive

mapScrollMode sounds ok. I have more of an issue with "natural" as that is highly ambiguous. The "normal" mode is "natural" to many (old) gamers.

Maybe something like "sticky", "stickToCursor" or so? Yeah, naming is hard :-/

Farmer-Markus: > This should be an enum then. Also what do those 3 setting mean?

An local enum or an global enum class? I'm not sure what would be better. And where would I define that enum class?
I could use an local enum in Settings.cpp and dskGameInterface.

libs/s25main/TerrainRenderer.cpp
Flamefire: Yes. And maybe explain why. Or better: This seems to be related to the MODULATE in the driver. In non GL4ES we have a factor of 2.

So rather make a constant for that and use it here. for GL4ES that would be 1. I.e. visible=clr/factor, FoW = clr/(factor*2)

libs/s25main/TerrainRenderer.cpp
Flamefire: Then at least explain the situation. "does not like" doesn't tell anything about the issue, i.e. how it manifests.

This is important if someone comes around with a solution for this issue. A short explanation helps in a) finding the issue and b) verifying it is fixed

extras/videoDrivers/SDL2/VideoSDL2.cpp
Farmer-Markus: > > This results in compiler errors (gcc and clang)
> 
> Which ones? Can't see how
>

```
/home/markus/Projekte/Android/s25client-android/extras/videoDrivers/SDL2/VideoSDL2.cpp:129:22: error: use of logical '||' with constant operand [-Werror,-Wconstant-logical-operand]
  129 |     if((RTTR_OGL_ES) || (RTTR_OGL_GL4ES))
      |                      ^  ~~~~~~~~~~~~~~~~
/home/markus/Projekte/Android/s25client-android/extras/videoDrivers/SDL2/VideoSDL2.cpp:129:22: note: use '|' for a bitwise operation
  129 |     if((RTTR_OGL_ES) || (RTTR_OGL_GL4ES))
      |                      ^~
      |                      |
1 error generated.
```

> As those are constants I'd declare them as such, like `constexpr SDL_GLprofile profile = ...;`
>
I'll work on that :+1:

libs/s25main/desktops/dskGameInterface.cpp
Flamefire: Indeed. And `mc.pos` should be used. I created a PR to avoid the confusion: It is a struct and having a method for getting the position is a relict of the time where we didn't have mc.pos being a Point: https://github.com/Return-To-The-Roots/s25client/pull/1811

libs/s25main/Settings.h
Flamefire: In `Settings.h` seems best as a global enum(class). Everyone using only does through this class/header anyway.

It can be global as `Settings::MapScrollMode::Original` is unnecessarily long and the enum class makes it "scoped" already

> Maybe something like "sticky", "stickToCursor" or so? Yeah, naming is hard :-/

A better idea: "Grab" or "GrabAndDrag" as that seems to describe it best: You don't scroll but you drag the map with your mouse/finger. Then maybe `MapMoveMode` for the enum and possibly `ScrollSame` & `ScrollOpposite` for map movement in the same direction as the mouse and the other way to be clearer than "original/inverted"   
A bit confusing: in Original mode the map moves in the *opposite* direction as the mouse so the *view* moves in the same direction as the mouse. While the "revert/invert" mode is the opposite.

As with "GrabAndDrag" we refer to dragging the map I guess `ScrollOpposite` would be the current "original"

@Flow86 Your thoughts?

libs/s25main/WindowManager.cpp
Farmer-Markus: > > Which windows do even need a dbl click to work? Never seen one.
> 
> ctrlList, ctrlTable use it for selection. Double-click on window titlebars maximizes/minimizes the window
>

maximize/minimize won't work for me.
I think the wrong config is being used... windowpinning for minimizing??
in IngameWindow.cpp
```
switch(btn)
            {
                case IwButton::Close: Close(); break;
                case IwButton::Title:
                    if(SETTINGS.interface.enableWindowPinning && mc.dbl_click)
                    {
                        SetMinimized(!IsMinimized());
                        LOADER.GetSoundN("sound", 113)->Play(255, false);
                    }
                    break;
```

But I could swear it worked some time ago. Is this wanted or should I correct it?
Could just do this:
```
switch(btn)
            {
                case IwButton::Close: Close(); break;
                case IwButton::Title:
                    if(!mc.dbl_click)
                        break;
                case IwButton::PinOrMinimize:
                    if(SETTINGS.interface.enableWindowPinning)
                    {
                        SetPinned(!IsPinned());
                        LOADER.GetSoundN("sound", 111)->Play(255, false);
                    } else
                    {
                        SetMinimized(!IsMinimized());
                        LOADER.GetSoundN("sound", 113)->Play(255, false);
                    }
                    break;
            }
```
>IIRC the IngameWindow::MouseLeftUp won't be called if a control handled it. If that is the case you can add it there.
>
I don't think this is necessary. Every window which makes sense to quick-close works with its mouseLeftUp function

Flamefire: > maximize/minimize won't work for me.
> I think the wrong config is being used... windowpinning for minimizing??

It is working as intended, see description of 5b33edebcdafe810736e8864c3dcb4f70047e1d7
> Add a "pin" function to in-game windows. The function replaces the minimize button and the window can instead be minimized by double-clicking the title bar.

Not fully obvious but reasons seems to be that the button is repurposed for the pinning.

>> IIRC the IngameWindow::MouseLeftUp won't be called if a control handled it. If that is the case you can add it there. 
>
>I don't think this is necessary. Every window which makes sense to quick-close works with its mouseLeftUp function

My point was: You can add the close-functionality to  IngameWindow::MouseLeftUp if and only if it doesn't get called if a control already handled it. Otherwise the window might be closed when it shouldn't.   
Hence you'd need to double check that it really works this way

Farmer-Markus: > > > IIRC the IngameWindow::MouseLeftUp won't be called if a control handled it. If that is the case you can add it there.
> > 
> > 
> > I don't think this is necessary. Every window which makes sense to quick-close works with its mouseLeftUp function
> 
> My point was: You can add the close-functionality to IngameWindow::MouseLeftUp if and only if it doesn't get called if a control already handled it. Otherwise the window might be closed when it shouldn't. Hence you'd need to double check that it really works this way

I think I've already done this in my [commit](https://github.com/Return-To-The-Roots/s25client/pull/1804/commits/8d00021691ddeb3de35bcf831e117b2a268ee30f#diff-e8829f77b7d4215f69ca07f04cc1f57f8b90097254156db6437152c9591ad7f5)

libs/s25main/ingameWindows/IngameWindow.cpp
```diff
@@ -249,6 +252,10 @@ void IngameWindow::MouseLeftUp(const MouseCoords& mc)
             }
         }
     }
+
+    // On touch devices dblclick close window
+    if(VIDEODRIVER.IsTouch() && mc.dbl_click && !btnClicked && !IsPinned())
```
Flamefire: This needs at least a check accounting for `closeBehavior_ == CloseBehavior::Custom) // no close button`

libs/s25main/WindowManager.cpp
Flamefire: Almost there. But I wasn't just referring to the title-bar icons but to any "control", i.e. button, list, etc.

In particular in a table control as used in e.g. the save/load windows a double-click means selection.   
Nos I fear 2 things can happen:

1. The user missed the intended row and the window just closes
2. The action could not be performed and the window should stay open but gets closed

However I can't find a way to properly handle that with out current UI system. We have means to detect a "handled" event we don't use it. Not only that an ingamewindow has `Msg_LeftUp`and `MouseLeftUp`. The latter doesn't seem to make sense to me as we have the former for that.

I think this works:
```
            if(!activeWnd.RelayMouseMessage(&Window::Msg_LeftUp, mc))
                activeWnd.Msg_LeftUp(mc);
```

Similar for the other events but that needs some more careful thought.

I'm reasonably sure that for now changing WindowManager.cpp like this would suffice for most cases:
```diff
            // ja, dann Msg_LeftUp aufrufen
            IngameWindow& activeWnd = *windows.back();
-            activeWnd.Msg_LeftUp(mc);
-
-            // und den anderen Fenstern auch Bescheid geben
-            activeWnd.RelayMouseMessage(&Window::Msg_LeftUp, mc);
-
-            // und noch MouseLeftUp vom Fenster aufrufen
-            activeWnd.MouseLeftUp(mc);
+             if(!activeWnd.RelayMouseMessage(&Window::Msg_LeftUp, mc))
+             {
+                 activeWnd.Msg_LeftUp(mc);
+                 activeWnd.MouseLeftUp(mc);
+              }
```

This should at least avoid triggering your close code when the event was already handled. From some searching I'm reasonably sure that this doesn't cause any issues: `MouseLeftUp` only handles the titlebar and so no control could have handled that.

Farmer-Markus: This code breaks not just the touch-dblclick-close but also dblclick with mouse on the title bar which locks left click. (Window gets moved around with cursor)
I've issues understanding this part of the code and I'm not sure how to solve this problem...

libs/s25main/WindowManager.cpp
```diff
@@ -26,6 +26,9 @@
 #include "s25util/MyTime.h"
 #include <algorithm>
 
+// Calc square
+#define SQR(x) ((x) * (x))
```
Flamefire: Please no macros where constants or functions can be used
```suggestion
template<typename T>
constexpr T square(T x)
{
    return x * x;
}
```

libs/s25main/WindowManager.cpp
```diff
@@ -356,13 +359,24 @@ void WindowManager::Msg_LeftUp(MouseCoords mc)
 
     // Ggf. Doppelklick untersuche
     unsigned time_now = VIDEODRIVER.GetTickCount();
-    if(time_now - lastLeftClickTime < DOUBLE_CLICK_INTERVAL && mc.GetPos() == lastLeftClickPos)
+
+    if(!VIDEODRIVER.IsTouch())
```
Flamefire: Please use non-negated conditions when you do if-else for readability.

libs/s25main/WindowManager.cpp
```diff
@@ -356,13 +359,24 @@ void WindowManager::Msg_LeftUp(MouseCoords mc)
...
     {
-        mc.dbl_click = true;
-    } else
+        if(time_now - lastLeftClickTime < DOUBLE_CLICK_INTERVAL && mc.pos == lastLeftClickPos)
+            mc.dbl_click = true;
+
+    } else if(time_now - lastLeftClickTime < TOUCH_DOUBLE_CLICK_INTERVAL)
+    {
+        // Calculate distance between two points
+        unsigned cDistance = SQR(mc.pos.x - lastLeftClickPos.x) + SQR(mc.pos.y - lastLeftClickPos.y);
```
Flamefire: Just a hint for future changes: Use `const auto diff = mc.pos - lastLeftClickPos` to avoid the potential for C&P issues due 2 repeated use of x and y in the same line. The below could then be `if(square(diff.x) + square(diff.y) <= square(TOUCH_MAX_DOUBLE_CLICK_DISTANCE))` which explains why the RHS is squared more directly.

No need to change this, just a suggestion for future changes as this is mostly an opinion

libs/s25main/desktops/dskGameInterface.cpp
```diff
@@ -684,45 +515,251 @@ bool dskGameInterface::Msg_LeftDown(const MouseCoords& mc)
...
-        VIDEODRIVER.SetMousePos(mc.GetPos());
+        VIDEODRIVER.SetMousePos(mc.pos);
+
+        ShowActionWindow(action_tabs, cSel, mc.pos, enable_military_buildings);
+    }
+
+    return true;
+}
+
+void dskGameInterface::Msg_ButtonClick(const unsigned ctrl_id)
```
Flamefire: Did you move those methods around? Any reason for that? The diff is very large and I can't tell if there were changes.

libs/s25main/desktops/dskGameInterface.h
```diff
@@ -117,6 +117,9 @@ class dskGameInterface :
     /// Updatet das Post-Icon mit der Nachrichtenanzahl und der Taube
     void UpdatePostIcon(unsigned postmessages_count, bool showPigeon);
 
+    /// Wird beim Linksklick ausgeführt und überprüft die klick Position auf Gebäude/Straßen
```
Flamefire: English please

extras/videoDrivers/SDL2/VideoSDL2.cpp
Flamefire: @Farmer-Markus Can you do the latter so that we don't loose the error reporting?

libs/s25main/Settings.cpp
```diff
@@ -307,7 +317,16 @@ void Settings::Load()
...
         // {
         interface.autosave_interval = iniInterface->getIntValue("autosave_interval");
-        interface.invertMouse = iniInterface->getValue("invert_mouse", false);
+        try
+        {
+            interface.mapScrollMode = static_cast<MapScrollMode>(iniInterface->getIntValue("map_scroll_mode"));
+        } catch(const std::runtime_error&)
+        {
+            // ScrollSame(old invertMouse) is 0 so invert value
+            interface.mapScrollMode = static_cast<MapScrollMode>(!iniInterface->getBoolValue("invert_mouse"));
```
Flamefire: That cast can't be verified at this point. Just do a ternary here to set the enum value directly. So you don't even need the comment

This will also throw an error if the value does not exist, so please keep the prior one with default value

I don't think you need the warning either as there is nothing to check: The old setting is translated 1:1 to the new one

libs/s25main/TerrainRenderer.cpp
```diff
@@ -44,6 +45,15 @@
  * providing an index and a count into the above arrays.
  */
 
+namespace {
+// gl4es has some problems with GL_MODULATE so we need to change the colors without using a gl function
+#if RTTR_OGL_GL4ES
+constexpr float TEXTURE_COLOR_DIVISOR = 1;
+#else
+constexpr float TEXTURE_COLOR_DIVISOR = 2;
```
Flamefire, line 53: Can you use this constant in `glTexEnvf(GL_TEXTURE_ENV, GL_RGB_SCALE, 2.0f);` so one can tell where the 2 comes from/is used?

libs/s25main/WindowManager.cpp
Flamefire: I'll check in the next couple of days, that was just thinking what could work, haven't tested it.

libs/s25main/desktops/dskGameInterface.cpp
Farmer-Markus: I did not.
I've just created the ContextClick function above. Don't know why the patch was created this weird.

libs/s25main/WindowManager.cpp
Farmer-Markus:
```
// Calculate distance between two points
const auto diff = mc.pos - lastLeftClickPos;
if(square(diff.x) + square(diff.y) <= square(TOUCH_MAX_DOUBLE_CLICK_DISTANCE))
    mc.dbl_click = true;
```

diff would be `Point<int>` but `TOUCH_MAX_DOUBLE_CLICK_DISTANCE` is unsigned which can't be compared.
Do you know a better way than changing the constant to int or using static_cast<float> in the if statement?

Flamefire: Only idea: Make `square` return `std::make_unsigned_t<T>` and do the cast there as the square of an int will always be positive.

extras/videoDrivers/SDL2/VideoSDL2.cpp
Flamefire: > same here, even though that this is not a preprocessor define here.
> 
> @Flamefire  why?

I now remember why I did that: code in runtime ifs ifs checked. We don't really build those configurations locally or on CI so with those instead of PP checks we can at least verify them syntactically

Another reminder to use more comments in such cases 😅

libs/s25client/CMakeLists.txt
Flow86: then this should be documented somewhere, and maybe splitted out completely.

Like we create a static library by default, and have an exe linking the library.
same goes for that approach, so that the jni stuff is explicitly written down

extras/videoDrivers/SDL2/VideoSDL2.cpp
```diff
@@ -235,10 +249,16 @@ bool VideoSDL2::ResizeScreen(const VideoMode& newSize, bool fullscreen)
     return true;
 }
 
+#ifdef __ANDROID__
+void VideoSDL2::PrintError(const std::string&) const
+{ // Not really sure why the nowide::cerr function completely crashes on android
+}
```
Farmer-Markus: I'm not really sure what to do. I could add android logging but this would depend on android/log.h which is only available in the android sdk/ndk. Also the logging would just be visible when debugging the device which most people wouldn't even want do.

Flamefire: > // Not really sure why the nowide::cerr function completely crashes on android

Does that happen with `std::cerr` too? Because `nowide::cerr` on non-Windows is literally an alias of that and I'd be surprised why that would fail

Farmer-Markus: > Does that happen with `std::cerr` too? Because `nowide::cerr` on non-Windows is literally an alias of that and I'd be surprised why that would fail

Yes it does. Just tested it. It seems that std::cout, std::cin... are not supported by android. Android doesn't have a console for the i/o stuff but I'm wondering why std::cerr crashes and the other log implementations don't.

Flamefire: Then not using it here is not a solution: It is naturally used in many more places already

Maybe redirecting stderr to stdout via `nowide::cerr.set_rdbuf(nowide::cout.rdbuf())` works?

Farmer-Markus: > Then not using it here is not a solution: It is naturally used in many more places already
> 
> Maybe redirecting stderr to stdout via `nowide::cerr.set_rdbuf(nowide::cout.rdbuf())` works?

I've tried around and this seems to be a bug with the android std lib. I've fixed this issue by preloading the video driver on the android/java side. The io stuff won't work with separate loaded libraries without preloading.

libs/driver/include/driver/MouseCoords.h
```diff
@@ -14,10 +14,17 @@ struct MouseCoords
...
 };
 
 /// Maximum interval between two clicks to be considered a double-click (in milliseconds)
 constexpr unsigned DOUBLE_CLICK_INTERVAL = 500;
+
+// Max time difference in ms to trigger contextclick
+constexpr unsigned TOUCH_MAX_CLICK_INTERVAL = 250;
+constexpr unsigned TOUCH_DOUBLE_CLICK_INTERVAL = 175;
+// Max distance between the two clicks to trigger doubleclick
+constexpr unsigned TOUCH_MAX_DOUBLE_CLICK_DISTANCE = 30;
```
Flamefire: Missing newline at end of file (causes issues when included in some places)

libs/s25main/WindowManager.cpp
Flamefire: I totally forgot about this and just got around working on this: https://github.com/Return-To-The-Roots/s25client/pull/1850

> This code breaks not just the touch-dblclick-close but also dblclick with mouse on the title bar which locks left click. (Window gets moved around with cursor)

Not exactly sure what you meant by that. Can you check if there is  still an issue with that PR?

Farmer-Markus: > Not exactly sure what you meant by that. Can you check if there is still an issue with that PR?

Works perfectly fine with your changes thanks.
Windows won't close when dbl pressing buttons etc.

libs/driver/include/driver/MouseCoords.h
Farmer-Markus: I'll try to remember it

extras/videoDrivers/SDL2/VideoSDL2.cpp
Farmer-Markus: Should I keep it the way it is in this pr or should I do some additional changes?

libs/s25client/CMakeLists.txt
Farmer-Markus: I'm not really sure what to do now, do you have an idea how/where to document this?

extras/videoDrivers/SDL2/VideoSDL2.cpp
Flamefire: I think I'd go back to constexprs to avoid unexpected compilation errors later on.

IIRC the only issue was that concatenating those ints leads to warnings, e.g. ` if((RTTR_OGL_ES) || (RTTR_OGL_GL4ES))`, wasn't it?

This can be resolved by using bools. So using this instead in the config header should work:
> `constexpr bool RTTR_OGL_ES = @RTTR_OGL_ES@;`

And a comment on top of those that they are used such that syntax will be checked for inactive cases too.

libs/s25client/CMakeLists.txt
Flamefire: Where: `option(RTTR_BUILD_LIB "Executables will be build as shared library." OFF)` (as a comment)

Something like you have written above: "E.g. Android apps require loading shared libraries instead running executables directly. `SDL_main` will call the main function defined in that library"   
Or similar. Maybe with a link to the relevant docs

extras/videoDrivers/SDL2/VideoSDL2.cpp
Farmer-Markus: > > `constexpr bool RTTR_OGL_ES = @RTTR_OGL_ES@;`

I've tried around but some code relies on using pre processor statements.
Eg. 
```
#if RTTR_OGL_GL4ES
#    include <gl4esinit.h>
#endif
```
Or here the GLES2 loader isn't even defined in the header when building for normal GL
```
bool OpenGLRenderer::initOpenGL(OpenGL_Loader_Proc loader)
{
#if RTTR_OGL_ES
    return gladLoadGLES2Loader(loader) != 0;
#else
    return gladLoadGLLoader(loader) != 0;
#endif
}
```

external/glad/openglCfg.hpp.cmake
```diff
@@ -12,5 +12,8 @@
 #define RTTR_OGL_ES @RTTR_OGL_ES@
 // True(thy) if OpenGL compatibility profile should be used
 #define RTTR_OGL_COMPAT @RTTR_OGL_COMPAT@
+// True(thy) if gl4es should be used
+#define RTTR_OGL_GL4ES @RTTR_OGL_GL4ES@
```
Flamefire: Let's avoid the C&P of the var name now:
```suggestion
// 1 if OpenGL ES should be used
#cmakedefine01 RTTR_OGL_ES
// 1 if OpenGL compatibility profile should be used
#cmakedefine01 RTTR_OGL_COMPAT
// 1 if gl4es should be used
#cmakedefine01 RTTR_OGL_GL4ES
```

Edit: seems it doesn't show the unmodified lines in the suggestion.

extras/videoDrivers/SDL2/VideoSDL2.cpp
Flamefire: Ok, looks like it isn't worth it then. Let's use the PP defines then.

external/glad/openglCfg.hpp.cmake
Farmer-Markus: I don't know why but now even the if(RTTR_OGL_ES || RTTR_OGL_GL4es) works without errors, nice

Farmer-Markus: I had to leave the major and minor defines. cmakedefine01 is only able to handle 0 and 1. 
```
#define RTTR_OGL_MAJOR @RTTR_OGL_MAJOR@
#define RTTR_OGL_MINOR @RTTR_OGL_MINOR@
```

Flamefire: Looks like the difference is the parens: https://godbolt.org/z/sMG5KTjMo

> I had to leave the major and minor defines. cmakedefine01 is only able to handle 0 and 1.

That's the definition of that, yes. So it is as good as can be now.

libs/s25main/ingameWindows/IngameWindow.cpp
```diff
@@ -249,6 +252,10 @@ void IngameWindow::MouseLeftUp(const MouseCoords& mc)
             }
         }
     }
+
+    // On touch devices dblclick close window
+    if(VIDEODRIVER.IsTouch() && mc.dbl_click && !btnClicked && !IsPinned() && closeBehavior_ == CloseBehavior::Regular)
```
Flamefire: I guess it should act as-if the close button was pressed:
```suggestion
    if(VIDEODRIVER.IsTouch() && mc.dbl_click && !btnClicked && closeBehavior_ != CloseBehavior::Custom)
```

Flamefire:
https://github.com/Return-To-The-Roots/s25client/pull/1850 will be merged after its CI finishes.

I think the last open issue here was the handling of a double-touch as double-click to close the window. So you can incorporate that (merge or rebase) and check the suggestion I just made. If it works in your tests with those then I'd say this is ready :-)

libs/s25main/ingameWindows/IngameWindow.cpp
Farmer-Markus: I've removed the btnClicked boolean. After your changes, the code just returns after triggering a button.
Works fine with touch & mouse controls

Flow86:
> If it works in your tests with those then I'd say this is ready :-)

maybe a proper rebase & squashing some of the commits to proper commits would be nice ;-)

Farmer-Markus:
> > If it works in your tests with those then I'd say this is ready :-)
> 
> maybe a proper rebase & squashing some of the commits to proper commits would be nice ;-)

I'll do that tomorrow :)

Farmer-Markus:
> maybe a proper rebase & squashing some of the commits to proper commits would be nice ;-)

Is this ok?

Flow86 (approved):

Flamefire:
@Farmer-Markus Do you want to fix the commit messages? otherwise I'd just merge this.

Farmer-Markus:
> @Farmer-Markus Do you want to fix the commit messages? otherwise I'd just merge this.

I'll fix it. I'm just very busy at the moment

Farmer-Markus:
> I'd prefer if every commit builds and is correctly formatted (to avoid unrelated changes in the next commit) and the last 2 commits are not "Updated to last commit [...]" as it isn't clear what that means. To which commit? Also the fix in those could be squashed to the commit introducing the issue.
> 
I've tried to do so the whole day but it kept becoming worse. I can't figure out how to properly use rebase to move diffs to the right commit without breaking everything.

> In case you want to fix the formatting, what I'd do is: Interactive rebase stopping on each commit, run `make clangFormat` and continue. On any conflict select the incoming change (i.e. next commit) and format again.

I've changed the 2 latest commit messages and hope it's not that big of a deal with the other commits.
I'm sorry but I'll remember your tips for the next time :D

Flamefire:
@Farmer-Markus I made a rebase of your branch on current master formatting each commit. 2 other changes:
- Move methods in dskGameInterface to make the diff smaller
- Return `true` when double-tap closes ingame window

Also each commit now builds so I could remove the comments from the last 2 commit titles.

Can you verify that this works for you too? You have to pull and reset your local branch due to the force-push.
I did check the diffs at each step but just want to make sure nothing slipped through, I'd really like to get this in :)

Farmer-Markus:
>     * Return `true` when double-tap closes ingame window
Everything works fine but now windows are closed by dbl clicking on the top bar instead of being minimized.
But I don't know if this was intentional

Flamefire:
> Everything works fine but now windows are closed by dbl clicking on the top bar instead of being minimized.
> But I don't know if this was intentional

It was not and I think this wasn't caused by [my changes](https://github.com/Return-To-The-Roots/s25client/compare/629bc2cbef5fb553c8425aca53eee0aba616266c..1eed84b877a7126bacb212dcff934857e85f3ee6)

Currently minimizing happens only when window-pinning is enabled and double-clicking the title bar:  `if(SETTINGS.interface.enableWindowPinning && mc.dbl_click)`

Your commit 629bc2cbef5fb553c8425aca53eee0aba616266c removed the check if any button was pressed.

Where was/should be the code that minimizes the window on double-tapping the title bar?

Maybe we should:
- always `return true` when any button is hit (`if(IsPointInRect(mc.pos, GetButtonBounds(btn)))`)
- in the `Title` case check `if((SETTINGS.interface.enableWindowPinning || VIDEODRIVER.IsTouch()) && mc.dbl_click)` to enable it for touch

Or (possibly better) use subconditions in the bottom `if(VIDEODRIVER.IsTouch() && mc.dbl_click` to check if inside titlebar and handle either as close or minimize there.   
It might make sense to pull the conditions for modal windows to where they matter instead of continuing or the double-tap check might need to check for the 2 buttons too as they might be clicked but inactive and then we don't want to close, do we?

Farmer-Markus:
> Your commit [629bc2c](https://github.com/Return-To-The-Roots/s25client/commit/629bc2cbef5fb553c8425aca53eee0aba616266c) removed the check if any button was pressed.

Ok now I see why.

> Where was/should be the code that minimizes the window on double-tapping the title bar?

I think I'm just messing up my memories. It should close when dblclick on titlebar and window pinning isn't enabled.

> Maybe we should:
> 
>     * always `return true` when any button is hit (`if(IsPointInRect(mc.pos, GetButtonBounds(btn)))`)
> 
>     * in the `Title` case check `if((SETTINGS.interface.enableWindowPinning || VIDEODRIVER.IsTouch()) && mc.dbl_click)` to enable it for touch
> 
> 
> Or (possibly better) use subconditions in the bottom `if(VIDEODRIVER.IsTouch() && mc.dbl_click` to check if inside titlebar and handle either as close or minimize there. It might make sense to pull the conditions for modal windows to where they matter instead of continuing or the double-tap check might need to check for the 2 buttons too as they might be clicked but inactive and then we don't want to close, do we?

I don't think we need to do additional changes. It works how it's intended :D

>     * always `return true` when any button is hit (`if(IsPointInRect(mc.pos, GetButtonBounds(btn)))`)

We could change this just to keep the window open and not accidentally close the window when trying to drag but I think this shouldn't be a serious problem.

Flamefire:
> I think I'm just messing up my memories. It should close when dblclick on titlebar and window pinning isn't enabled.
>
> We could change this just to keep the window open and not accidentally close the window when trying to drag but I think this shouldn't be a serious problem.

Just to be sure: Double-tap anywhere should close the window except for non-modal windows in these cases:
- When window pinning is enabled double-tap on title will minimize
- double-tap on minimize/pin will minimize/pin (effectively nothing as done twice and hence reverts to previous)

Correct?

Farmer-Markus:
> > I think I'm just messing up my memories. It should close when dblclick on titlebar and window pinning isn't enabled.
> >
> > We could change this just to keep the window open and not accidentally close the window when trying to drag but I think this shouldn't be a serious problem.
> 
> Just to be sure: Double-tap anywhere should close the window except for non-modal windows in these cases:
> - When window pinning is enabled double-tap on title will minimize
> - double-tap on minimize/pin will minimize/pin (effectively nothing as done twice and hence reverts to previous)
> 
> Correct?

Yes Correct

Farmer-Markus:
@Flamefire I've fixed another very annoying bug with disable_mouse when using touch controls.
I hope this didn't broke your rebase or other stuff :/

You would always need to click twice to open a IgWindow if misclicked and opened the wrong one.
Just added the following to `WindowManager::DoShow(`
``` patch
+ if(!VIDEODRIVER.IsTouch())
       disable_mouse = true;
```

Sorry for the many force-pushes my editor is playing games with me and not showing files how the actually are...

Flamefire:
> Sorry for the many force-pushes my editor is playing games with me and not showing files how the actually are...

But why force-push at all? That new change is unrelated to the previous commit so shouldn't be in it.   

I extracted it to a separate commit and queueing this for merge :)

Farmer-Markus:
> I extracted it to a separate commit and queueing this for merge :)

Thank you! I've learned so much during this PR :D

PR #1811, opened 2025-10-25
Convert `MouseCoord` to plain struct

Flamefire:
The only method `GetPos` is no longer required as the `pos` is already a `Point` instance.

I also removed the constructors as `MouseCoords(mc2.GetPos(), false, false, true)` constructs are error prone and hard to read due to the repeated arguments of same type

### DIFF
--- a/libs/driver/include/driver/MouseCoords.h
+++ b/libs/driver/include/driver/MouseCoords.h
@@ -1,4 +1,4 @@
-// Copyright (C) 2005 - 2021 Settlers Freaks (sfteam at siedler25.org)
+// Copyright (C) 2005 - 2025 Settlers Freaks (sfteam at siedler25.org)
 //
 // SPDX-License-Identifier: GPL-2.0-or-later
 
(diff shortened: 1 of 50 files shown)


Flow86:
to be honest, I would have done it the opposite around: only getter functions. 

I agree with the "dangerous" constructors. There I would have used some chainable setters:

```
MouseCoords(...).ldown(true).rdown(true)...
```

so its clear what is getting set. (in our style it would be with `Set...` prefix though)

With public variables you do not have control who and where things read/write to those variables.

Flamefire:
> With public variables you do not have control who and where things read/write to those variables.

Yes but that is already the case.   
And the effect is very limited: This struct is passed around as a const-reference in all(?) cases. So no modification happens.   
It really is just a struct: A collection of strongly related values.

So I wouldn't over-engineer this and just make sure it is const when modification is not intended.

As for the ctor/factory function: We only need that for tests. In all other cases the values are only read by consumers and changed directly in a member of the videodriver where only a single value should be changed/toggled with the rest being constant.

PR #1814, opened 2025-10-29
Fixing build error for gcc-15: Adding include

### LINKED ISSUE #1770: Support GCC 15
GCC 15 has a few more quirks like less implicit includes and more available diagnostics, see #1768

We should add GCC 15 to CI and fix all issues.

sjoblomj:
Fixing build error when using gcc-15. ~~Also adding GitHub Actions Unit test.~~ Fixes #1770.

To observe the build fail to build without this patch, you can save the following Dockerfile and run with `docker build .`. To see the build succeed, you can simulate this patch by uncommenting the line `#RUN sed -i 's/#include </#include <cstdint>\n#include </' ../libs/s25main/notifications/NotificationManager.h`, and then running `docker build .` again.

```
FROM ubuntu:25.10
RUN apt-get update && apt-get install -y \
    libcurl4-gnutls-dev \
    git \
    build-essential \
    cmake \
    libsdl2-dev \
    libsdl2-mixer-dev \
    libboost-all-dev \
    libminiupnpc-dev \
    lua5.3 \
    liblua5.3-dev \
    libbz2-dev \
    gettext

RUN git clone --recursive https://github.com/Return-To-The-Roots/s25client.git

RUN mkdir s25client/build
WORKDIR s25client/build
#RUN sed -i 's/#include </#include <cstdint>\n#include </' ../libs/s25main/notifications/NotificationManager.h
RUN cmake -DRRTR_ENABLE_WERROR=off -DCMAKE_BUILD_TYPE=Release ..
RUN make
```

### DIFF
--- a/libs/s25main/notifications/NotificationManager.h
+++ b/libs/s25main/notifications/NotificationManager.h
@@ -5,6 +5,7 @@
 #pragma once
 
 #include "notifications/Subscription.h"
+#include <cstdint>
 #include <deque>
 #include <functional>
 #include <unordered_map>


Flamefire:
GCC 15 isn't available until Ubuntu 25, I merged a change using a container-based CI job. Just pull your branch to update your local branch

However it seems the tests now start to fail. Do you see that locally too?

Flow86:
until ubuntu 25 is available, I wouldn't add it to the CI. Lets keep this change simple. Add the missing cstdint.h - the ci is already much too complex and big.

Flamefire:
Yes, failures seem to be caused by using containers:
```
ALSA lib conf.c:5205:(_snd_config_evaluate) function snd_func_card_inum returned error: No such file or directory
unknown location(0): fatal error: in "GameClientTests/ClientFollowsConnectProtocol/_0": std::runtime_error: Could not get username
/s25client/tests/s25Main/drivers/testDriverWrapper.cpp(31): fatal error: in "AllAudioDriversAreLoadable": critical check AudioDriverWrapper::inst().LoadDriver(preference) has failed
```

The DeadSingleton-errors shouldn't happen, but anyway: Wouldn't overthink this.

@sjoblomj Can you reset your branch to contain only the fix for the missing include? Would be great to have CI for that but seems to not be worth the effort

sjoblomj:
I force-pushed (hope that's ok) so that now only the include is in the PR. Thank you for looking at it :)

PR #1816, opened 2025-11-02
Move EnableCCache include after submodule check

Flamefire:
It is in libutil which may not exist or be up to date, so check those first


MacThings:
Google said libutil is included in libSystem. Can´t install it through brew.
I´m on the latest macOS (26).

Flow86:
libutil is our own library afaik, so it cannot be in libsystem.

Issue #1806, opened 2025-10-14
Please... Turn off/silence bird chatter


ultimate-underground:
Hej

Can someone tell me how to mute or turn off a specific sound in the game? And that's the birds chirping =D

Flow86:
you can look for `playBirdSounds` in the files. I do not have the right place in my head where the chirp is played though.

ultimate-underground:
I tried to analyze all possible files... the only one that looks promising is sound.scs... I don't know which number is the one that will solve the bird's cry.
And will it work at all by changing the frequency parameter?
Isn't it better to rewrite the code directly in the project?

sound.scs - editing in notepad


```
# Copyright (C) 2005 - 2021 Settlers Freaks <sf-team at siedler25.org>
#
# SPDX-License-Identifier: GPL-2.0-or-later

# Format:
# <itemnr> <source-frequency>
#
# The item at index <itemnr> is converted using <source-frequency> as the frequency
# All others are left untouched
#
051 11025
052 11025
053 11025
054 11025
055 11025
056 11025
057 11025
058 11025
059 11025
# empty
061 11025
062 11025
# empty
064 11025
065 11025
066 11025
067 11025
068 11025
069 11025
070 11025
# empty
072 11025
# empty
074 11025
# empty
076 11025
077 11025
078 6000
# empty
# empty
081 11025
082 9000
# empty
084 11025
085 11025
086 11025
087 11025
088 11025
089 11025
090 11025
091 11025
092 11025
093 11025
094 6000
095 11025
096 11025
# empty
098 11025
099 11025
100 11025
101 11025
102 11025
103 11025
104 06500
105 10025
# empty
107 9000
# empty
109 11025
110 9000
111 11025
112 11025
113 11025
114 11025
```

Flow86:
the sound.scs is only used to convert the sounds to the right pitch (samplerate) - it has nothing to do with the volume itself. The volume is probably set in `playBirdSounds` or in the calling site of it

ultimate-underground:
So manipulating the compiled game won't solve the problem and I'll have to intervene in the code before compiling...

Flow86:
Yes exactly

ultimate-underground:
I found... I solved...

okromario:
Hi @ultimate-underground, I implemented a proper way to disable the little birds with PR #1817. Have fun. 😎
@Flow86 thx for the merge. I guess this issue can be closed then, right?

[...]

PR #1817, opened 2025-11-03
Proposal to turn off/on bird sounds

okromario:
Hi there,

here is a proposal to turn on/off the birds sounds from the game settings menus (in-game & main settings). The new setting is also reflected in the config file as boolean with the name `bird_sounds`. This PR should resolve issue #1806. 💪

I tried to fit in my changes as good as possible in the UI and code as well.

Here are some screenshots of the changed UI windows in 800x600 with 100% UI scaling:

### In-Game Settings:

I placed the new bird toggle button between the effects and the music buttons/sliders. I had to grow the settings window slightly to fit in the new button. I couldn't use the existing free space, because this space is reserved for a future Load button according to the code.
rtts-birds-ingame-settings

### Main Settings:

I reversed the effects and music buttons/sliders and placed the new bird switch buttons between the effects and music buttons.
rtts-birds-main-settings

I tested it also with my full resolution at 3440x1440 with 200% UI scaling and it works fine.

### Issues:

- I don't find a possibility to translate the new UI text into another language, maybe someone can tell me pls, how to do it, if needed.
- If the new switch buttons are used in the main settings, the new state isn't reflected in the in-game settings menu, but it actually works. Other way around, a state change in the in-game settings are reflected in the main settings. I don't know why and how to fix it.

Don't be surprised by my 1st commit, I had to fix the "NotificationManager.h" with a oneliner first, so I was able to build the project successfully.

Pls let me now, if I missed something. 😊

### DIFF
--- a/libs/s25main/ingameWindows/iwOptionsWindow.cpp
+++ b/libs/s25main/ingameWindows/iwOptionsWindow.cpp
@@ -49,7 +49,7 @@ enum
 };
 
 using Offset = DrawPoint;
-constexpr auto windowSize = Extent(300, 545);
+constexpr auto windowSize = Extent(300, 525);
 constexpr auto imageButtonSize = Extent(35, 35);
 constexpr auto optionSizeSmall = Extent(160, 22);
 constexpr auto optionSizeBig = Extent(168, 24);
@@ -112,23 +112,25 @@ iwOptionsWindow::iwOptionsWindow(SoundManager& soundManager)
                    LOADER.GetImageN("io", 114 + !SETTINGS.sound.effectsEnabled)); //-V807
     AddProgress(ID_pgEffectVol, curPos + optionOffset, optionSizeSmall, TextureColor::Green2, 139, 138, 100)
       ->SetPosition((SETTINGS.sound.effectsVolume * 100) / 255);
-    curPos.y += optionSizeBig.y + generalSpacing;
+    curPos.y += imageButtonSize.y;
+    curPos.x += optionOffset.x;
 
-    AddCheckBox(ID_cpBirdSounds, curPos + optionOffset, optionSizeSmall, TextureColor::Green2, _("Bird sounds"),
-                NormalFont, false)
+    AddCheckBox(ID_cpBirdSounds, curPos, optionSizeSmall, TextureColor::Green2, _("Bird sounds"), NormalFont, false)
       ->setChecked(SETTINGS.sound.birdsEnabled);
-    curPos.y += optionSizeSmall.y * 2;
+    curPos.y += optionSizeSmall.y + generalSpacing * 3;
+    curPos.x = leftMargin;
 
     // Music on/off + volume
     AddImageButton(ID_btMusic, curPos, imageButtonSize, TextureColor::Green2,
                    LOADER.GetImageN("io", 116 + !SETTINGS.sound.musicEnabled));
     AddProgress(ID_pgMusicVol, curPos + optionOffset, optionSizeSmall, TextureColor::Green2, 139, 138, 100)
       ->SetPosition((SETTINGS.sound.musicVolume * 100) / 255);
-    curPos.y += optionSizeBig.y + generalSpacing;
+    curPos.y += imageButtonSize.y;
+    curPos.x += optionOffset.x;
 
-    AddTextButton(ID_btMusicPlayer, curPos + optionOffset, optionSizeSmall, TextureColor::Green2, _("Music player"),
-                  NormalFont);
-    curPos.y += optionSizeSmall.y * 2;
+    AddTextButton(ID_btMusicPlayer, curPos, optionSizeSmall, TextureColor::Green2, _("Music player"), NormalFont);
+    curPos.y += optionSizeSmall.y + generalSpacing * 3;
+    curPos.x = leftMargin;
 
     // Buttons at the bottom
     curPos.x = centerPosition - optionSizeBig.x / 2;


libs/s25main/ingameWindows/iwOptionsWindow.cpp
```diff
@@ -84,22 +85,25 @@ iwOptionsWindow::iwOptionsWindow(SoundManager& soundManager)
                    LOADER.GetImageN("io", 114 + !SETTINGS.sound.effectsEnabled)); //-V807
 
     // Musik an/aus
-    AddImageButton(ID_btMusic, DrawPoint(35, 340), Extent(35, 35), TextureColor::Green2,
+    AddImageButton(ID_btMusic, DrawPoint(35, 371), Extent(35, 35), TextureColor::Green2,
```
Flamefire: Can you move this below the bird sound checkbox line please? You did that in the settings screen which IMO is very reasonable to understand the layout better.

libs/s25main/ingameWindows/iwOptionsWindow.cpp
```diff
@@ -84,22 +85,25 @@ iwOptionsWindow::iwOptionsWindow(SoundManager& soundManager)
...
                    LOADER.GetImageN("io", 116 + !SETTINGS.sound.musicEnabled));
 
     // Geräuschlautstärke
     AddProgress(ID_pgEffectVol, DrawPoint(100, 306), Extent(160, 22), TextureColor::Green2, 139, 138, 100)
       ->SetPosition((SETTINGS.sound.effectsVolume * 100) / 255);
 
+    // Vogelgeräusche an/aus
```
Flamefire: Please use English comments only. Feel free to translate neighbouring comments to make it look consistent. But at the very least please don't introduce new German comments.

libs/s25main/ingameWindows/iwOptionsWindow.cpp
```diff
@@ -84,22 +85,25 @@ iwOptionsWindow::iwOptionsWindow(SoundManager& soundManager)
...
+    AddCheckBox(ID_cpBirdSounds, DrawPoint(100, 342), Extent(160, 22), TextureColor::Green2, _("Bird sounds"), NormalFont, false);
+
     // Musiklautstärke
-    AddProgress(ID_pgMusicVol, DrawPoint(100, 346), Extent(160, 22), TextureColor::Green2, 139, 138, 100)
+    AddProgress(ID_pgMusicVol, DrawPoint(100, 377), Extent(160, 22), TextureColor::Green2, 139, 138, 100)
```
Flamefire: It seems the spacing of the progress bars, checkbox and button is even. Can you use a similar pattern to the desktop: `constexpr Extent ctrlSize(160, 22);`, `DrawPoint pos = ...`, and `pos.y += ctrlSize.y + spacingY` and the use `pos` for the position?

The 2 buttons for enable/disable of music&effects would then need to be relative

While the current approach works with more controls added it becomes harder to reason if they are all placed "correctly".

libs/s25main/ingameWindows/iwOptionsWindow.cpp
```diff
@@ -84,22 +85,25 @@ iwOptionsWindow::iwOptionsWindow(SoundManager& soundManager)
...
-    AddImageButton(ID_btMusic, DrawPoint(35, 340), Extent(35, 35), TextureColor::Green2,
+    AddImageButton(ID_btMusic, DrawPoint(35, 371), Extent(35, 35), TextureColor::Green2,
                    LOADER.GetImageN("io", 116 + !SETTINGS.sound.musicEnabled));
 
     // Geräuschlautstärke
     AddProgress(ID_pgEffectVol, DrawPoint(100, 306), Extent(160, 22), TextureColor::Green2, 139, 138, 100)
       ->SetPosition((SETTINGS.sound.effectsVolume * 100) / 255);
 
+    // Vogelgeräusche an/aus
+    AddCheckBox(ID_cpBirdSounds, DrawPoint(100, 342), Extent(160, 22), TextureColor::Green2, _("Bird sounds"), NormalFont, false);
```
Flamefire: This fixes your issue

```suggestion
    AddCheckBox(ID_cpBirdSounds, DrawPoint(100, 342), Extent(160, 22), TextureColor::Green2, _("Bird sounds"), NormalFont, false)->setChecked(SETTINGS.sound.birdsEnabled);
```

Needs formatting and include of "controls/ctrlCheck.h" though

libs/s25main/Settings.cpp
```diff
@@ -258,6 +259,7 @@ void Settings::Load()
         sound.musicVolume = iniSound->getIntValue("musik_volume");
         sound.effectsEnabled = iniSound->getBoolValue("effekte");
         sound.effectsVolume = iniSound->getIntValue("effekte_volume");
+        sound.birdsEnabled = iniSound->getBoolValue("bird_sounds");
```
Flamefire: This clears all existing config settings to their defaults, use a default to make this optional:
```suggestion
        sound.birdsEnabled = iniSound->getValue("bird_sounds", true);
```

Flamefire (changes requested):
Great work thanks! Just a few small remarks.


>    I don't find a possibility to translate the new UI text into another language, maybe someone can tell me pls, how to do it, if needed.

This is done in the https://github.com/Return-To-The-Roots/languages repo but needs updating the language files after the code changes which is a bit involved. I did that update based on the current state of the PR, so you can create a PR in that repo with translations inside the respective files if you want. After the merge of that you need to update the submodule here:
```
cd external/languages
git checkout master
git pull
cd ..
git commit languages
```

>    If the new switch buttons are used in the main settings, the new state isn't reflected in the in-game settings menu, but it actually works. Other way around, a state change in the in-game settings are reflected in the main settings. I don't know why and how to fix it.

You forgot to set the state of that checkbox, see inline comment for fix.

> Don't be surprised by my 1st commit, I had to fix the "NotificationManager.h" with a oneliner first, so I was able to build the project successfully.

Using GCC 15? We merged a fix for that same thing with #1814, so no issue with that. :-)

okromario:
Hi @Flamefire, many thanks for your answers and explanations.

> This is done in the https://github.com/Return-To-The-Roots/languages repo...

aah sry, I didn't checked the other repos for that. Thx for updating the base. I will add the german translation soon. I could add it for more languages, but only translator assisted.

Do I have something to re-generate at the languages repo to reflect my new changes, before I can start with my translation?

> Using GCC 15? We merged a fix for that same thing with #1814, so no issue with that. :-)

yes GCC 15. Ah I see the commit now. My local master branch wasn't up-to-date during branching.

Additionally I found outdated current years at the licence header at the files I touched. Should I correct this as well?
```
                        vvvv
// Copyright (C) 2005 - 2021 Settlers Freaks (sf-team at siedler25.org)
```

Would it be an idea to sort the include lines in alphabetical order, at least at my touched files?

libs/s25main/ingameWindows/iwOptionsWindow.cpp
okromario: With "the desktops" you mean the mechanism at for example my already touched `libs/s25main/desktops/dskOptions.cpp`? This would be really better indeed, due it was way easier for me to add my changes there then in the in-game menu.

I could try this, but I would need some time to implement the same principle to this menu. I'm not very familiar with C++ actually. 😁 I will let you know. 💪

libs/s25main/ingameWindows/iwOptionsWindow.cpp
okromario: sry you are right. I have now translated all German comments into English in all files I have touched. 💪

libs/s25main/ingameWindows/iwOptionsWindow.cpp
```diff
@@ -70,37 +71,37 @@ iwOptionsWindow::iwOptionsWindow(SoundManager& soundManager)
     AddImageButton(ID_btReadme, DrawPoint(35, 160), Extent(35, 35), TextureColor::Green2, LOADER.GetImageN("io", 79));
     AddText(ID_txtReadme, DrawPoint(85, 180), _("Load 'ReadMe' file"), COLOR_YELLOW, FontStyle::BOTTOM, NormalFont);
 
-    // "Spiel laden!"
+    // "Load game!"
```
Flamefire: FWIW: Such comments can actually be removed, it just restates exactly what is below as it appears in the label

libs/s25main/ingameWindows/iwOptionsWindow.cpp
Flamefire: Yes exactly. And you were familiar enough to do this change, so don't worry! :-)

You don't need to do this for this PR, just let me know if you are up to it and I'll wait before merging to avoid introducing this change just to change it again for a "better" solution.

If you need any help feel free to ask here, show intermediate results if that helps you question.

Flamefire (approved):
> yes GCC 15. Ah I see the commit now. My local master branch wasn't up-to-date during branching.

No worries, it was very recent indeed.

> Additionally I found outdated current years at the licence header at the files I touched. Should I correct this as well?
> 
>                         vvvv
> // Copyright (C) 2005 - 2021 Settlers Freaks (sf-team at siedler25.org)

You can do that for the files you changed, yes

> Would it be an idea to sort the include lines in alphabetical order, at least at my touched files?

This is actually already the case and the linter (clang-format based) enforces this. See the failed check:

```diff
 --- libs/s25main/ingameWindows/iwOptionsWindow.cpp	(original)
+++ libs/s25main/ingameWindows/iwOptionsWindow.cpp	(reformatted)
@@ -9,6 +9,7 @@
 #include "Settings.h"
 #include "SoundManager.h"
 #include "WindowManager.h"
+#include "controls/ctrlCheck.h"
 #include "controls/ctrlImageButton.h"
 #include "controls/ctrlProgress.h"
 #include "drivers/AudioDriverWrapper.h"
@@ -20,7 +21,6 @@
 #include "iwTextfile.h"
 #include "ogl/FontStyle.h"
 #include "gameData/const_gui_ids.h"
-#include "controls/ctrlCheck.h"
```

However the order follows a certain pattern: 1. own includes, 2. third-party includes, 3. stdlib includes. And they are sorted within each group.  
You can run the "clangFormat" target (`make clangFormat` if I remember the spelling correctly)  if you have clang-format-10 installed locally to automatically format sources including the sorting.


> Do I have something to re-generate at the languages repo to reflect my new changes, before I can start with my translation?

After adding/removing translatables from the source the template (`*.pot` file) needs to be regenerated using poedit and that reflected in the individual translation files (`*.po`) which I did.   
With that done you can now pull the new branch, add translations and open a PR for that repo similar to this one. When this is merged the submodule update can be added in this repo to include it.    
Only add translations where you are reasonably sure they are correct. Untranslated strings show up in several tools so they are easy to find and English strings are still readable for most. So if in doubt rather leave that.

libs/s25main/ingameWindows/iwOptionsWindow.cpp
okromario: ok thx, I will try, yes pls wait, I let you know 💪

okromario:
Hi @Flamefire, here we go, I'm done. The diff looks not very review friendly, but I can't do anything about it I guess.

The in-game settings menu has now the same placement pattern as the main settings menu (desktops). The placement is only slightly different as before, but better as before in my opinion and the spacing is now standardized.

I gave the (still hidden) Load button now it's own IDs for future use and it's still prepared to be enabled and positioned correctly together with the Save button with ease, if the load function get's implemented. Just comment in the code for the Load button (row 103-105) and remove 2 lines of code marked for this purpose (row 107 & 110) and that's it. Nothing to recalculate or playing with variables. 💪

rttr

In the screenshot you see the 1st dialog was the prev. one, the 2nd the (current checked in) one and the 3rd one with enabled Load button. The whole window size remained unchanged.

Pls let me know, if I need to revise anything or if I have overlooked something.

libs/s25main/ingameWindows/iwOptionsWindow.cpp
```diff
@@ -45,64 +47,100 @@ enum
...
+} // namespace
 
 iwOptionsWindow::iwOptionsWindow(SoundManager& soundManager)
-    : IngameWindow(CGI_OPTIONSWINDOW, IngameWindow::posLastOrCenter, Extent(300, 545), _("Game menu"),
+    : IngameWindow(CGI_OPTIONSWINDOW, IngameWindow::posLastOrCenter, windowSize, _("Game menu"),
                    LOADER.GetImageN("resource", 41)),
       soundManager(soundManager)
 {
+    using Offset = DrawPoint;
+    constexpr Extent buttonImageSize(35, 35);
```
Flamefire: I would put those with the other constants above. Then define the `rowHeight` in terms of the sizes.

From the uses I'd say intended is `rowHeight = buttonImageSize.y + 5` as we want 5px space between the image buttons (btw: `imageButtonSize` might be slightly better as it is `AddImageButton`)

And `rowHeightSmall = optionSizeBig.y + 7`? Or don't define that at all and use calculated values directly. See below

libs/s25main/ingameWindows/iwOptionsWindow.cpp
```diff
@@ -45,64 +47,100 @@ enum
...
+    constexpr Extent optionSizeSmall(160, 22);
+    constexpr Extent optionSizeBig(168, 24);
+    constexpr Offset textOffset(50, 24);
+    constexpr Offset optionOffset(65, 6);
+
+    DrawPoint curPos = headerStartPosition;
```
Flamefire: I'd say use the value directly. No need for a constant used only once if it is self-explanatory

libs/s25main/ingameWindows/iwOptionsWindow.cpp
```diff
@@ -41,65 +44,103 @@ enum
...
+
+    AddText(ID_txtVersion, curPos, rttr::version::GetReadableVersion(), COLOR_YELLOW, FontStyle::CENTER, NormalFont);
+    curPos.y += textSpacing;
+
+    AddFormattedText(ID_txtCopyright, curPos,
                      "\xC2\xA9"
                      "2005 - %s Settlers Freaks",
                      COLOR_YELLOW, FontStyle::CENTER, NormalFont)
       % rttr::version::GetYear();
+    curPos.y += rowHeightSmall;
```
Flamefire: Maybe:

```suggestion
    curPos.y += textSpacing * 2;
```
(3px bigger but easier to reason about)

libs/s25main/ingameWindows/iwOptionsWindow.cpp
```diff
@@ -41,65 +44,103 @@ enum
...
+    curPos.x = textPadding;
```
Flamefire: Name seems off: It isn't used for text

```suggestion
    curPos.x = leftMargin;
```

libs/s25main/ingameWindows/iwOptionsWindow.cpp
```diff
@@ -41,65 +44,103 @@ enum
...
-    AddTextButton(ID_btAdvanced, DrawPoint(67, 412), Extent(168, 24), TextureColor::Green2, _("Advanced"), NormalFont);
-    AddTextButton(ID_btSurrender, DrawPoint(67, 443), Extent(168, 24), TextureColor::Red1, _("Surrender"), NormalFont);
-    AddTextButton(ID_btEndGame, DrawPoint(67, 474), Extent(168, 24), TextureColor::Red1, _("End game"), NormalFont);
+    curPos.y += rowHeight;
+
+    // Buttons at the bottom
+    curPos.x = centerPosition;
+    constexpr Offset btOffset(-(optionSizeBig.x / 2), 0);
+    AddTextButton(ID_btAdvanced, curPos + btOffset, optionSizeBig, TextureColor::Green2, _("Advanced"), NormalFont);
+    curPos.y += rowHeightSmall;
```
Flamefire:
```suggestion
    curPos.y += optionSizeBig.y + 5;
```

(instead of `+ 7` to match the image button spacing?)

Same below

libs/s25main/ingameWindows/iwOptionsWindow.cpp
```diff
@@ -41,65 +44,103 @@ enum
...
-
-    // Musik an/aus
-    AddImageButton(ID_btMusic, DrawPoint(35, 340), Extent(35, 35), TextureColor::Green2,
-                   LOADER.GetImageN("io", 116 + !SETTINGS.sound.musicEnabled));
-
-    // Geräuschlautstärke
-    AddProgress(ID_pgEffectVol, DrawPoint(100, 306), Extent(160, 22), TextureColor::Green2, 139, 138, 100)
+    AddProgress(ID_pgEffectVol, curPos + optionOffset, optionSizeSmall, TextureColor::Green2, 139, 138, 100)
       ->SetPosition((SETTINGS.sound.effectsVolume * 100) / 255);
+    curPos.y += rowHeight;
```
Flamefire: I'd move the bird sounds box closer to the bar so make it clearer they belong together. E.g.:

```suggestion
    curPos.y += optionSizeSmall.y + 5;
```

(5px to keep the 5px spacing in other parts)

Same for the music player below

libs/s25main/ingameWindows/iwOptionsWindow.cpp
```diff
@@ -160,3 +201,15 @@ void iwOptionsWindow::Msg_ProgressChange(const unsigned ctrl_id, const unsigned
...
+void iwOptionsWindow::Msg_CheckboxChange(const unsigned ctrl_id, const bool checked)
+{
+    switch(ctrl_id)
+    {
+        case ID_cpBirdSounds:
+        {
+            SETTINGS.sound.birdsEnabled = checked;
+            break;
+        }
+    }
```
Flamefire: Maybe just: 

```suggestion
    RTTR_Assert(ctrl_id == ID_cpBirdSounds);
    SETTINGS.sound.birdsEnabled = checked;
```

This way it is clear that there are intentionally no other cases.

Flamefire (changes requested):
Awesome, thanks!

I only have 2 suggestions for naming changes and I'd define offsets in terms of sizes which is a major strength of this way of layouting: If the size changes the controls will move accordingly.

And a small layout change that wasn't easily possible before: The bird sounds should be closer to the effects volume than to the next control to make it visually clear that they belong together: The volume of the birds are also affected.   
Similar for the music player.   
What do you think about that?

Also thanks for attaching screenshots. That makes it much easier to review!

okromario:
Hi @Flamefire, ah you are right, I don't considered the sizes of the controls 🫣. Thx for your suggestions, I have implemented your changes, removed now unused spacing constants, but added 1 for the general spacing of the 5px (because it's used on many places and I guess it's convenient to have a general spacing one).

rttr2

In the screenshot above your see now the 1st dialog as the prev. one again, the 2nd one with my new changes and the 3rd one same but with enabled load button.

I hope I did it well. Pls let me know again, if there is something more. 💪

libs/s25main/notifications/NotificationManager.h
```diff
@@ -5,9 +5,9 @@
 #pragma once
 
 #include "notifications/Subscription.h"
-#include <cstdint>
 #include <deque>
 #include <functional>
+#include <cstdint>
```
Flamefire: This causes the issue. If you merge master you can remove this

libs/s25main/ingameWindows/iwOptionsWindow.cpp
```diff
@@ -41,65 +44,100 @@ enum
...
-
-    // Musik an/aus
-    AddImageButton(ID_btMusic, DrawPoint(35, 340), Extent(35, 35), TextureColor::Green2,
-                   LOADER.GetImageN("io", 116 + !SETTINGS.sound.musicEnabled));
-
-    // Geräuschlautstärke
-    AddProgress(ID_pgEffectVol, DrawPoint(100, 306), Extent(160, 22), TextureColor::Green2, 139, 138, 100)
+    AddProgress(ID_pgEffectVol, curPos + optionOffset, optionSizeSmall, TextureColor::Green2, 139, 138, 100)
       ->SetPosition((SETTINGS.sound.effectsVolume * 100) / 255);
+    curPos.y += imageButtonSize.y;
```
Flamefire:
```suggestion
    curPos.y += optionSizeSmall.y + generalSpacing;
```

libs/s25main/ingameWindows/iwOptionsWindow.cpp
```diff
@@ -41,65 +44,100 @@ enum
...
+    curPos.y += optionSizeSmall.y * 2;
 
-    // Musiklautstärke
-    AddProgress(ID_pgMusicVol, DrawPoint(100, 346), Extent(160, 22), TextureColor::Green2, 139, 138, 100)
+    // Music on/off + volume
+    AddImageButton(ID_btMusic, curPos, imageButtonSize, TextureColor::Green2,
+                   LOADER.GetImageN("io", 116 + !SETTINGS.sound.musicEnabled));
+    AddProgress(ID_pgMusicVol, curPos + optionOffset, optionSizeSmall, TextureColor::Green2, 139, 138, 100)
       ->SetPosition((SETTINGS.sound.musicVolume * 100) / 255);
+    curPos.y += imageButtonSize.y;
```
Flamefire:
```suggestion
    curPos.y += optionSizeSmall.y + generalSpacing;
```

libs/s25main/ingameWindows/iwOptionsWindow.cpp
```diff
@@ -41,65 +44,100 @@ enum
...
+    AddTextButton(ID_btMusicPlayer, curPos + optionOffset, optionSizeSmall, TextureColor::Green2, _("Music player"),
                   NormalFont);
-    AddTextButton(ID_btAdvanced, DrawPoint(67, 412), Extent(168, 24), TextureColor::Green2, _("Advanced"), NormalFont);
-    AddTextButton(ID_btSurrender, DrawPoint(67, 443), Extent(168, 24), TextureColor::Red1, _("Surrender"), NormalFont);
-    AddTextButton(ID_btEndGame, DrawPoint(67, 474), Extent(168, 24), TextureColor::Red1, _("End game"), NormalFont);
+    curPos.y += optionSizeSmall.y * 2;
+
+    // Buttons at the bottom
+    curPos.x = centerPosition;
+    constexpr Offset btOffset(-(optionSizeBig.x / 2), 0);
```
Flamefire: `-` on unsigned then to int doesn't (always) work.

Try:

```suggestion
    curPos.x = centerPosition - optionSizeBig.x / 2;
```

and remove `btOffset`

Flamefire (changes requested):
There is a duplicate include now which triggers the linter. Merge master and run clang-format to get rid of it.

>  but added 1 for the general spacing of the 5px (because it's used on many places and I guess it's convenient to have a general spacing one)

Good idea! 👍 

I would use that for the checkbox&button too. Can you see how that looks and if you like it apply it? My idea is that this a) keeps it even closer and b) avoids using another spacing (effectively now: `imgButtonSize - optionSizeSmall`)

okromario:
Hi @Flamefire, ah sry this include was my own fix from my 1st commit, that is of course now redundant. I reverted this commit now.

I implemented your suggestions. It looks like following:

rttr3

The 1st dialog is the prev. one, the 2nd your suggestion with "optionSizeSmall" and the 3rd one with "optionSizeBig".

The bird/music player button looks now to cramped to me, "overlapping" the image button on the left side, so I did the (3rd dialog) one with "optionSizeBig.y" that is 2px larger and then there would be no "overlapping" anymore, but the spacing between these buttons and the buttons at the bottom isn't the same anymore.

Here is a screenshot in detail:

rttr3_1

What spacing do you prefer? I like the "optionSizeBig" a bit more, but I will do as you like. I committed the big one currently.

okromario:
ah because there is now more space at the botton, I could reduce the window size to the original value again, or?

Flamefire:
> What spacing do you prefer? I like the "optionSizeBig" a bit more, but I will do as you like. I committed the big one currently.

I'm fine with either one. Having the top and bottom align is also nice.   
However the position calculation looks odd: I'd expect `curPos.y += imageButtonSize.y` not `optionSizeBig` which is only used for the buttons at the bottom (maybe should be renamed accordingly?)   
I guess ` imageButtonSize.y + generalSpacing + optionOffset.y` only happens to yield the same result.   
So I guess the easiest way would be to revert to  `curPos.y += imageButtonSize.y`  and use `curPos + Offset(optionOffset.x, 0)` for the placement. Is that correct?

> ah because there is now more space at the bottom, I could reduce the window size to the original value again, or?

Sure why not. Maybe such that the space between the last button and the bottom is the same as between the buttons? Or twice that? Just check what looks good to you and use that. It isn't worth to make that too exact, just a little less than after the current moves.

okromario:
Hi @Flamefire, ahaaa now I know why imageButtonSize.y led to a too large gap that prevented me to do so with the imageButtonSize.y alone. Many thanks for pointing that out. I hope this is okay now.

I corrected the calculation and reduced the vertical size of the dialog. The spacing is only very slighly different.

rttr4

The screenshot shows the same again, prev., current and current with enabled save button.

libs/s25main/ingameWindows/iwOptionsWindow.cpp
```diff
@@ -112,23 +112,25 @@ iwOptionsWindow::iwOptionsWindow(SoundManager& soundManager)
                    LOADER.GetImageN("io", 114 + !SETTINGS.sound.effectsEnabled)); //-V807
     AddProgress(ID_pgEffectVol, curPos + optionOffset, optionSizeSmall, TextureColor::Green2, 139, 138, 100)
       ->SetPosition((SETTINGS.sound.effectsVolume * 100) / 255);
-    curPos.y += optionSizeBig.y + generalSpacing;
+    curPos.y += imageButtonSize.y;
+    curPos.x += optionOffset.x;
 
-    AddCheckBox(ID_cpBirdSounds, curPos + optionOffset, optionSizeSmall, TextureColor::Green2, _("Bird sounds"),
-                NormalFont, false)
+    AddCheckBox(ID_cpBirdSounds, curPos, optionSizeSmall, TextureColor::Green2, _("Bird sounds"), NormalFont, false)
```
okromario, line 118: Not sure why the clang-format did this now and not before, but anyways 😁.

libs/s25main/ingameWindows/iwOptionsWindow.cpp
```diff
@@ -112,23 +112,25 @@ iwOptionsWindow::iwOptionsWindow(SoundManager& soundManager)
...
     AddProgress(ID_pgMusicVol, curPos + optionOffset, optionSizeSmall, TextureColor::Green2, 139, 138, 100)
       ->SetPosition((SETTINGS.sound.musicVolume * 100) / 255);
-    curPos.y += optionSizeBig.y + generalSpacing;
+    curPos.y += imageButtonSize.y;
+    curPos.x += optionOffset.x;
 
-    AddTextButton(ID_btMusicPlayer, curPos + optionOffset, optionSizeSmall, TextureColor::Green2, _("Music player"),
-                  NormalFont);
-    curPos.y += optionSizeSmall.y * 2;
+    AddTextButton(ID_btMusicPlayer, curPos, optionSizeSmall, TextureColor::Green2, _("Music player"), NormalFont);
```
okromario, line 131: Here too, this was actually clang-format. Of course the old row 131 was my change.

PR #1820, opened 2025-11-08
Incrase max lobby login name length to 30

Flamefire:
Don't limit to 15 chars.

Also refactor for using relative positions and English comments.

### DIFF
--- a/libs/s25main/ingameWindows/iwLobbyConnect.cpp
+++ b/libs/s25main/ingameWindows/iwLobbyConnect.cpp
@@ -57,7 +57,7 @@ iwLobbyConnect::iwLobbyConnect()
 
     AddText(ID_txtUser, curLblPos, _("Username:"), COLOR_YELLOW, FontStyle{}, NormalFont);
     ctrlEdit* user =
-      AddEdit(ID_edtUser, DrawPoint(ctrlStartPos, curLblPos.y), edtSize, TextureColor::Green2, NormalFont);
+      AddEdit(ID_edtUser, DrawPoint(ctrlStartPos, curLblPos.y), edtSize, TextureColor::Green2, NormalFont, 30);
     user->SetFocus();
     user->SetText(SETTINGS.lobby.name); //-V807
     curLblPos.y += 30;


libs/s25main/ingameWindows/iwLobbyConnect.cpp
```diff
@@ -49,36 +49,49 @@ iwLobbyConnect::iwLobbyConnect()
...
 {
-    AddText(ID_txtUser, DrawPoint(20, 40), _("Username:"), COLOR_YELLOW, FontStyle{}, NormalFont);
-    ctrlEdit* user = AddEdit(ID_edtUser, DrawPoint(260, 40), Extent(220, 22), TextureColor::Green2, NormalFont, 15);
+    DrawPoint curLblPos(20, 40);
+    constexpr Extent edtSize(220, 22);
+    constexpr auto ctrlStartPos = 260;
+    constexpr Extent btSize(edtSize.x / 2 - 5, edtSize.y);
+    constexpr auto secCtrlStartPos = ctrlStartPos + btSize.x + 10;
+
+    AddText(ID_txtUser, curLblPos, _("Username:"), COLOR_YELLOW, FontStyle{}, NormalFont);
```
Flow86, line 58: change the limit to 30. the registration now also have a proper limit of 15, but I can extend it to 30 then

libs/s25main/ingameWindows/iwLobbyConnect.cpp
```diff
@@ -164,31 +170,23 @@ void iwLobbyConnect::Msg_ButtonClick(const unsigned ctrl_id)
...
     switch(ctrl_id)
     {
-        case ID_optSavePw: // Passwort speichern Ja/Nein
-        {
-            SETTINGS.lobby.save_password = (selection == 1);
-        }
-        break;
-        case ID_optProtocol: // IPv6 Ja/Nein
-        {
+        case ID_optSavePw: SETTINGS.lobby.save_password = (selection == 1); break;
```
Flow86, line 186: inconsistent, put the Settings line to a new line

libs/s25main/ingameWindows/iwLobbyConnect.cpp
```diff
@@ -211,47 +209,30 @@ bool iwLobbyConnect::Msg_KeyDown(const KeyEvent& ev)
...
-    auto* b2 = GetCtrl<ctrlButton>(ID_btRegister);
-
-    // Text setzen
     t->SetTextColor(color);
     t->SetText(text);
 
-    // Button (de)aktivieren
-    b->SetEnabled(button);
-    if(b2)
-        b2->SetEnabled(button);
```
Flow86, line 231: why did we check if the button exist?

libs/s25main/ingameWindows/iwLobbyConnect.cpp
```diff
@@ -211,47 +209,30 @@ bool iwLobbyConnect::Msg_KeyDown(const KeyEvent& ev)
...
- */
 void iwLobbyConnect::LC_Status_Waiting()
 {
     SetText(_("Waiting for Reply..."), COLOR_YELLOW, false);
 }
 
-/**
- *  Status: Benutzerdefinierter Fehler (inkl Conn-Reset u.ä)
- */
+// User defined error (e.g. connection reset)
```
Flow86: missing the third doxygen `/`

libs/s25main/ingameWindows/iwLobbyConnect.cpp
Flamefire: No idea, historical... Maybe the button was added on a specific condition only.

libs/s25main/ingameWindows/iwLobbyConnect.cpp
Flamefire: That's clang-formats decision.
Reason why the other isn't on the same line is the comment. I kept the comment because the setting is actually IPv4/IPv6 which we only translate to IPv6 y/n

libs/s25main/ingameWindows/iwLobbyConnect.cpp
Flamefire: done

libs/s25main/ingameWindows/iwLobbyConnect.cpp
Flow86: add an empty comment behind the case:

```
case ID_optSavePw: //
```

then it will make it consistent

Flamefire: I don't think this is worth it. Especially I actually like it on one line: It reads quite natural like an English sentence

Issue #1822, opened 2025-11-15
scrolling in game with the mousewheel is slightly off


stefson:
hey everyone, 

I just noticed that you can scroll into the game from 100% to 103%, and if you scroll back out it will go to 99%, and if you scroll in again it will be 102%. 

you can go back to 100% by saving and reloading the hame, but thats just a workaround.

PR #1823, opened 2025-11-15
Make sure zooming via mouse goes back to 1.0 

Flamefire:
Fix for #1822 

When the zoom factor is closer to 1.0 than the mouse-increment set it to 1

Also make sure that using the keys to zoom behave consistently: If we set a "close" zoom level then either "z" or "Z" would jump to 1 level off the expected.

Example: Zoom in to 1.03, between level 4 and 5. If we set the level to "4" than zooming out will jump to 3 not back to 4. Similar for setting it to 5 and zooming in

### DIFF
--- a/libs/s25main/desktops/dskGameInterface.cpp
+++ b/libs/s25main/desktops/dskGameInterface.cpp
@@ -97,7 +97,23 @@ enum
     ID_btPost,
     ID_txtNumMsg
 };
+
+float getNextZoomLevel(const float currentZoom)
+{
+    // Get first level bigger than current zoom
+    // NOLINTNEXTLINE(readability-qualified-auto)
+    auto it = std::upper_bound(ZOOM_FACTORS.begin(), ZOOM_FACTORS.end(), currentZoom);
+    return (it == ZOOM_FACTORS.end()) ? ZOOM_FACTORS.front() : *it;
 }
+
+float getPreviousZoomLevel(const float currentZoom)
+{
+    // Get last level bigger or equal than current zoom
+    // NOLINTNEXTLINE(readability-qualified-auto)
+    auto it = std::lower_bound(ZOOM_FACTORS.begin(), ZOOM_FACTORS.end(), currentZoom);
+    return (it == ZOOM_FACTORS.begin()) ? ZOOM_FACTORS.back() : *(--it);
+}
+} // namespace
 
 dskGameInterface::dskGameInterface(std::shared_ptr<Game> game, std::shared_ptr<const NWFInfo> nwfInfo,
                                    unsigned playerIdx, bool initOGL)
@@ -105,7 +121,7 @@ dskGameInterface::dskGameInterface(std::shared_ptr<Game> game, std::shared_ptr<c
       worldViewer(playerIdx, const_cast<Game&>(*game_).world_),
       gwv(worldViewer, Position(0, 0), VIDEODRIVER.GetRenderSize()), cbb(*LOADER.GetPaletteN("pal5")),
       actionwindow(nullptr), roadwindow(nullptr), minimap(worldViewer), isScrolling(false),
-      zoomLvl_(ZOOM_DEFAULT_INDEX), cheats_(const_cast<Game&>(*game_).world_), cheatCommandTracker_(cheats_)
+      cheats_(const_cast<Game&>(*game_).world_), cheatCommandTracker_(cheats_)
 {
     road.mode = RoadBuildMode::Disabled;
     road.point = MapPoint(0, 0);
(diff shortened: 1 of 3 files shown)


extras/videoDrivers/WinAPI/WinAPI.cpp
```diff
@@ -539,8 +541,10 @@ void VideoWinAPI::OnWMChar(unsigned c, bool disablepaste, LPARAM lParam)
  */
 void VideoWinAPI::OnWMKeyDown(unsigned c, LPARAM lParam)
 {
-    KeyEvent ke = {KeyType::Invalid, 0, (GetKeyState(VK_CONTROL) & 0x8000) != 0, (GetKeyState(VK_SHIFT) & 0x8000) != 0,
-                   (lParam & KF_ALTDOWN) != 0};
+    KeyEvent ke;
+    ke.ctrl = (GetKeyState(VK_CONTROL) & 0x8000) != 0;
+    ke.shift = (GetKeyState(VK_SHIFT) & 0x8000) != 0;
+    ke.alt = (lParam & KF_ALTDOWN) != 0;
```
Flow86: thats now 3 times in this file alone? maybe also a `setSpecialKeys`?

libs/s25main/CheatCommandTracker.cpp
```diff
@@ -38,25 +41,26 @@ void CheatCommandTracker::onChatCommand(const std::string& cmd)
...
-
-    return true;
 }
 
 void CheatCommandTracker::onCharKeyEvent(const KeyEvent& ke)
 {
-    lastChars_.push_back(ke.c);
+    // Handle only ASCII chars
+    if(ke.c > 0x7F)
+        lastChars_.clear();
```
Flow86, line 58: do we do this? I thought we allow utf-8 input?

extras/videoDrivers/WinAPI/WinAPI.cpp
```diff
@@ -710,8 +714,10 @@ LRESULT CALLBACK VideoWinAPI::WindowProc(HWND window, UINT msg, WPARAM wParam, L
  */
 KeyEvent VideoWinAPI::GetModKeyState() const
 {
-    const KeyEvent ke = {KeyType::Invalid, 0, (GetKeyState(VK_CONTROL) & 0x8000) != 0,
-                         (GetKeyState(VK_SHIFT) & 0x8000) != 0, (GetKeyState(VK_MENU) & 0x8000) != 0};
+    KeyEvent ke;
+    ke.ctrl = (GetKeyState(VK_CONTROL) & 0x8000) != 0;
+    ke.shift = (GetKeyState(VK_SHIFT) & 0x8000) != 0;
+    ke.alt = (GetKeyState(VK_MENU) & 0x8000) != 0;
```
Flow86, line 727: oh this time the `alt` key is different? insteresting?

libs/s25main/desktops/dskGameInterface.cpp
```diff
@@ -879,20 +879,38 @@ bool dskGameInterface::Msg_KeyDown(const KeyEvent& ke)
...
-
-            gwv.SetZoomFactor(ZOOM_FACTORS[zoomLvl]);
+            {
+                // Get last level bigger or equal than current zoom
+                // NOLINTNEXTLINE(readability-qualified-auto)
+                auto it = std::lower_bound(ZOOM_FACTORS.begin(), ZOOM_FACTORS.end(), gwv.GetCurrentTargetZoomFactor());
+                zoomLvl_ = (it == ZOOM_FACTORS.begin()) ? 0 : std::distance(ZOOM_FACTORS.begin(), it) - 1;
+            }
+            gwv.SetZoomFactor(ZOOM_FACTORS[*zoomLvl_]);
             return true;
```
Flow86, line 906: can we maybe not do this in the switch case? its so hard to read. Maybe add a class handling the zoom factors and everything zoom related?

extras/videoDrivers/WinAPI/WinAPI.cpp
Flamefire: Once it is `GetKeyState`, the other uses `lParam`, those are different, yes

libs/s25main/CheatCommandTracker.cpp
Flamefire: We do but not here. Cheats are ASCII only

stefson:
works as intended, thanks! :)

libs/s25main/desktops/dskGameInterface.cpp
Flamefire: Factored out into 2 functions and noticed we don't need the class variable for the level at all as the same code works when we are at any of the fixed zoom levels

Flamefire:
@Flow86 Jenkins fails with
> Error response from daemon: client version 1.52 is too new. Maximum supported API version is 1.41

extras/videoDrivers/WinAPI/WinAPI.cpp
Flamefire: Done. Fixed a bug too: KF_ALTDOWN needs to check the HIWORD

Flamefire: See also #1831 which includes this commit and 2 more from this PR

Flow86: rebase/conflict resolving necessary

Flow86:
> @Flow86 Jenkins fails with
> 
> > Error response from daemon: client version 1.52 is too new. Maximum supported API version is 1.41

yes, already fixed, but I forgot to make it permanent. now it should work

extras/videoDrivers/WinAPI/WinAPI.cpp
Flamefire: Rebased. This change is now gone (from this PR)

Flow86 (approved):

PR #1829, opened 2025-11-16
Make all game speeds available to "v" key

### LINKED ISSUE #1828: speed settings (bug?)
I've noticed that the game speed has changed across different versions. Overall, it's became faster (was it intended or is it a bug?). That's not a big deal in itself, but I can no longer find the setting I preferred.

I seem to recall that some earlier nightly builds had more options available. Perhaps I'm mistaken, but the available speeds aren't what they used to be, so I can no longer achieve "my" speed (minus 2) as easily as before.

One could restore the "normal" speed, as it was in the original S2, or could provide more options. As far as I understand, there are now 7 (or 8) levels.

There are a few options I would suggest:

1) Make all 7 (or 8) speed levels accessible in the options menu (currently there are only 5 to choose).
2) Replace the slowest speed setting in the options menu (Minus 3 is currently enabled in version s25rttr_20250618, which, oddly enough, isn't even accessible via the V hotkey).
3) Make the speed levels variable and adjustable during gameplay. My idea is that each player sets their own speed, and the computer uses the lowest speed setting of any player as a consensus and adopts that setting. For example: If the slowest player increases his speed, the computer will increase it until the next slowest speed setting of another player.

Flamefire:
Introduce constants for the speed and speed steps, assert the min and max are reachable
and adapt conditions for step size and minimal/maximal step to debug speeds.

Closes #1828

### DIFF
--- a/tests/s25Main/lua/testLuaGUI.cpp
+++ b/tests/s25Main/lua/testLuaGUI.cpp
@@ -38,22 +38,22 @@ BOOST_AUTO_TEST_CASE(MissionStatement)
     executeLua("rttr:MissionStatement(1, 'Title', 'Text')");
     const auto* wnd = dynamic_cast<iwMissionStatement*>(WINDOWMANAGER.GetTopMostWindow());
     BOOST_TEST_REQUIRE(wnd);
-    BOOST_TEST_REQUIRE(wnd->IsActive());
-    BOOST_TEST_REQUIRE(wnd->GetTitle() == _("Title"));
+    BOOST_TEST(wnd->IsActive());
+    BOOST_TEST(wnd->GetTitle() == _("Title"));
 
     // double windows stack
     executeLua("rttr:MissionStatement(1, 'Title2', 'Text')");
     const auto* wnd2 = dynamic_cast<iwMissionStatement*>(WINDOWMANAGER.GetTopMostWindow());
     BOOST_TEST_REQUIRE(wnd2);
     // Other window still on top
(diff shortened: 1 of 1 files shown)


libs/s25main/gameData/GameConsts.h
```diff
@@ -11,9 +11,16 @@
...
+/// Length of GameFrames for each speed level
 constexpr helpers::EnumArray<std::chrono::duration<unsigned, std::milli>, GameSpeed> SUPPRESS_UNUSED
   SPEED_GF_LENGTHS = {{80ms, 60ms, 50ms, 40ms, 30ms}};
+constexpr auto MAX_SPEED = 10ms;
+constexpr auto MIN_SPEED = SPEED_GF_LENGTHS[GameSpeed::VerySlow];
+/// Max/Max speed for debug mode (includes replays)
+constexpr auto MIN_SPEED_DEBUG = 1s;
+constexpr auto MAX_SPEED_DEBUG = 1ms;
+/// Difference in which to change speed levels in-game.
+constexpr auto SPEED_STEP = 10ms;
```
Flow86, line 23: can we maybe make a small wrapper class handling all the speed steps instead?

Now (and in the past) the handling is shattered around the code.

Flamefire: There is exactly one place that uses the speed steps: The functions `IncreaseSpeed` and `DecreaseSpeed`

However I factored out the handling of reference speed conversion to 2 functions

Flamefire:
Could you fix Jenkins and rerun the builds please?

PR #1841, opened 2025-11-27
Fix typo in allowPortraitChange variable name and update LUA feature level

Flamefire:
When adding functions we need to update the feature level of the LUA interface so scripts can check that.

### DIFF
--- a/data/RTTR/campaigns/roman/MISS200.lua
+++ b/data/RTTR/campaigns/roman/MISS200.lua
@@ -21,7 +21,7 @@ function isMapPreviewEnabled()
     return false
 end
 
-local requiredFeature = 4
+local requiredFeature = 6
 function checkVersion()
     local featureLevel = rttr:GetFeatureLevel()
     if(featureLevel < requiredFeature) then
--- a/data/RTTR/campaigns/roman/MISS201.lua
+++ b/data/RTTR/campaigns/roman/MISS201.lua
@@ -22,7 +22,7 @@ function isMapPreviewEnabled()
     return false
 end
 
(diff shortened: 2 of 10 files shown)


doc/lua/functions.md
```diff
@@ -242,7 +242,8 @@ Add an AI or change its difficulty.
 Change the player's name.
 
 **SetPortrait(portraitIndex)**  
-Change the player's portrait.
+Change the player's portrait.  
+-- Added in LUA version 1.6
```
Flow86, line 246: shouldn't it be simply `6`?

doc/lua/functions.md
```diff
@@ -383,7 +384,8 @@ rttr:GetWorld():AddAnimal(41, 44, SPEC_DUCK)
 [Back](#Lua-objects-and-their-methods)  
 
 **SetComputerBarrier(radius, x, y)**
-Defines an area with a specified (inclusive) radius around map point {x, y} in which the AI cannot build military buildings.
+Defines an area with a specified (inclusive) radius around map point {x, y} in which the AI cannot build military buildings.  
+-- Added in LUA version 1.5
```
Flow86, line 388: shouldn't it be simply `5`?

doc/lua/functions.md
Flamefire: We have a major and minor version (breaking changes and feature level), hence I put in the full version here.

With the LUA code changes I believe we should have `getRequiredLuaVersion` allow to return the feature level too, not only the major version. But I remember some issues trying to return "1.2". Got to double check

Issue #1844, opened 2025-11-29
playlist automatically restarts if the sound menue is closed


stefson:
I just noticed that the playlist automatically restarts, once you close the sound menue. It doesn't seem to matter if you just open the sound menue, or if you activly change the playing song, or if you manipulate the playlist.

PR #1845, opened 2025-11-30
Fix detection of selected playlist on window open

Flamefire:
The path to the playlist needs to be resolved before comparison or it
will always fail.
That in turn leads to selecting the first playlist as-if the current one
didn't exist, and restarting the playlist.

Also ensure that in this case the save/delete buttons are correctly disabled.

Fixes https://github.com/Return-To-The-Roots/s25client/issues/1844


Flow86 (approved):

stefson:
I patched my local source with the patch from this pr, it works as intended! :-)

PR #1849, opened 2025-12-26
Wrap around from highest to lowest speed with "V" key

### LINKED ISSUE #1828: speed settings (bug?)

Flamefire:
Fixes #1828

Regression in #1829 which assumed that `IncreaseSpeed` only increases speed but with "V" we want to wrap around. But only there, so add a param for wraparound

### DIFF
--- a/libs/s25main/desktops/dskGameInterface.cpp
+++ b/libs/s25main/desktops/dskGameInterface.cpp
@@ -830,7 +830,7 @@ bool dskGameInterface::Msg_KeyDown(const KeyEvent& ke)
                 RTTR_Assert(worldViewer.GetPlayerId() == oldPlayerId || worldViewer.GetPlayerId() == playerIdx);
             } else if(playerIdx < worldViewer.GetWorld().GetNumPlayers())
             {
-                // On multi player this currently asyncs, but as this is a debug feature anyway just disable it.
+                // On multiplayer this currently asyncs, but as this is a debug feature anyway just disable it.
                 // If this should be enabled again, look into the handling/clearing of accumulated GCs
                 if(game_->world_.IsSinglePlayer())
                 {
@@ -874,7 +874,7 @@ bool dskGameInterface::Msg_KeyDown(const KeyEvent& ke)
         case 'm': // Show main menu
             WINDOWMANAGER.ToggleWindow(std::make_unique<iwMainMenu>(gwv, GAMECLIENT));
             return true;
-        case 'n': // Show post window
+        case 'n': // Show message window
             WINDOWMANAGER.ToggleWindow(std::make_unique<iwPostWindow>(gwv, GetPostBox()));
             UpdatePostIcon(GetPostBox().GetNumMsgs(), false);
             return true;
@@ -891,10 +891,10 @@ bool dskGameInterface::Msg_KeyDown(const KeyEvent& ke)
         case 26: // ctrl+z
             gwv.SetZoomFactor(ZOOM_FACTORS[ZOOM_DEFAULT_INDEX]);
             return true;
-        case 'z':       // zoom in
+        case 'z':
             if(ke.ctrl) // Reset zoom
                 gwv.SetZoomFactor(ZOOM_FACTORS[ZOOM_DEFAULT_INDEX]);
-            else
+            else // zoom in
                 gwv.SetZoomFactor(getNextZoomLevel(gwv.GetCurrentTargetZoomFactor()));
             return true;
         case 'Z': // shift-z, zoom out


libs/s25main/desktops/dskGameInterface.cpp
```diff
@@ -832,7 +830,7 @@ bool dskGameInterface::Msg_KeyDown(const KeyEvent& ke)
                 RTTR_Assert(worldViewer.GetPlayerId() == oldPlayerId || worldViewer.GetPlayerId() == playerIdx);
             } else if(playerIdx < worldViewer.GetWorld().GetNumPlayers())
             {
-                // On mutiplayer this currently asyncs, but as this is a debug feature anyway just disable it there.
+                // On multi player this currently asyncs, but as this is a debug feature anyway just disable it.
```
Spikeone: heißt es nicht multiplayer?

libs/s25main/desktops/dskGameInterface.cpp
```diff
@@ -844,64 +842,62 @@ bool dskGameInterface::Msg_KeyDown(const KeyEvent& ke)
             return true;
         }
 
-        case 'b': // Zur lezten Position zurückspringen
-            gwv.MoveToLastPosition();
-            return true;
+        case 'b': gwv.MoveToLastPosition(); return true;
```
Spikeone, line 845: warum hier der single line style, in z.b. L851 wird das nicht gemacht

libs/s25main/desktops/dskGameInterface.cpp
```diff
@@ -844,64 +842,62 @@ bool dskGameInterface::Msg_KeyDown(const KeyEvent& ke)
...
-        case 's': // Produktivität anzeigen
+        case 's': // Show/hide productivity overlay
             gwv.ToggleShowProductivity();
             return true;
         case 26: // ctrl+z
             gwv.SetZoomFactor(ZOOM_FACTORS[ZOOM_DEFAULT_INDEX]);
             return true;
-        case 'z': // zoom
-            if(ke.ctrl)
+        case 'z':       // zoom in
```
Spikeone: wozu spaces for dem comment?

libs/s25main/desktops/dskGameInterface.cpp
```diff
@@ -844,64 +842,62 @@ bool dskGameInterface::Msg_KeyDown(const KeyEvent& ke)
...
-        case 'l': // Minimap anzeigen
+        case 'l': // Show minimap
             WINDOWMANAGER.ToggleWindow(std::make_unique<iwMinimap>(minimap, gwv));
             return true;
-        case 'm': // Hauptauswahl
+        case 'm': // Show main menu
             WINDOWMANAGER.ToggleWindow(std::make_unique<iwMainMenu>(gwv, GAMECLIENT));
             return true;
-        case 'n': // Show Post window
+        case 'n': // Show post window
```
Spikeone: I glaube heißt im englische message window

Spikeone (dismissed):

libs/s25main/desktops/dskGameInterface.cpp
Flamefire: clang-format aligns it with below comment which makes it easier to read in general. Moved this comment to the branch though similar to the other

libs/s25main/desktops/dskGameInterface.cpp
Flamefire: clang-format: Hier ist kein comment im Weg und es passt

libs/s25main/network/GameClient.cpp
```diff
@@ -1172,14 +1172,17 @@ void GameClient::IncreaseSpeed()
...
     if(framesinfo.gfLengthReq == MIN_SPEED_DEBUG)
-        framesinfo.gfLengthReq = MIN_SPEED;
+        framesinfo.gfLengthReq = MIN_SPEED; // NOLINT(bugprone-branch-clone)
     else if(framesinfo.gfLengthReq >= MAX_SPEED + SPEED_STEP)
         framesinfo.gfLengthReq -= SPEED_STEP;
-    else if((replayMode || debugMode))
+    else if((replayMode || debugMode) && framesinfo.gfLengthReq > MAX_SPEED_DEBUG) // 1 more step in debug/replay mode
         framesinfo.gfLengthReq = MAX_SPEED_DEBUG;
+    else if(wraparound) // Highest speed, wrap around to slowest if requested
+        framesinfo.gfLengthReq = MIN_SPEED;
```
Flow86, line 1185: why are we not always doing this?

Flamefire: Because the name of the function is "IncreaseSpeed` so it might be unexpected when it slows it down.
The parameter makes this clearer even though we currently don't use other values

Flow86 (approved):

Flamefire:
@Flow86 Can you enable the squash option for PRs? Here I would have squashed the 2 commits into one instead of a full merge.   
Or what do you think?

Flow86:
I think thats fine, its okay as long as the commits are different enough

Issue #1843, opened 2025-11-28
unknown crash


stefson:
sadly there is no backtrace, as I wasn't running the game through gdb - it hasn't crashed for me in a very long time. 

so, here is the replay: [2025-11-28_21-13-08.rpl.gz](https://github.com/user-attachments/files/23829751/2025-11-28_21-13-08.rpl.gz)

feel free to close this issue, if you need a gdb backtrace

Spikeone:
@stefson does the replay crash as well when you play it until the end and press P? What version were you using?

stefson:
my version was the current head of git on the date I posted the issue, I always recompile my local binary, once a commit is posted. 

its been a very long time since I played a replay myself - wasn't there a framecounter somewhere in the UI? I'm able to load my own replay from above, but it doesn't show the number of frame when loaded. So I can't jump to that point right now, and press P to continue.

Spikeone:
You need to enable the debug overlay in the settings:

<img width="1922" height="1112" alt="Image" src="https://github.com/user-attachments/assets/bfff703e-90e6-4e4b-9ca4-7c467b23f2f1" />

stefson:
ok nice, I must have forgotten about that. 

anyway, the replay does crash immediatly when you jump to the end and press P, here is the backtrace from this: 

```
Thread 1 "s25client" received signal SIGTRAP, Trace/breakpoint trap.
0x00005555555e073b in nobHarborBuilding::AddWare (this=0x55555ddfc5a0, ware=...)
    at /var/tmp/portage/games-strategy/s25rttr-9999-r6/work/s25rttr-9999/libs/s25main/buildings/nobHarborBuilding.cpp:654
warning: 654	/var/tmp/portage/games-strategy/s25rttr-9999-r6/work/s25rttr-9999/libs/s25main/buildings/nobHarborBuilding.cpp: Datei oder Verzeichnis nicht gefunden
(gdb) bt
#0  0x00005555555e073b in nobHarborBuilding::AddWare (this=0x55555ddfc5a0, ware=...)
    at /var/tmp/portage/games-strategy/s25rttr-9999-r6/work/s25rttr-9999/libs/s25main/buildings/nobHarborBuilding.cpp:654
[...]

PR #1852, opened 2026-01-04
Correctly handle wares carried into harbors with no path to goal

Flamefire:
When the goal for a ware becomes unreachable while it is carried into a harbor
`RecalcRoute` will call `FindRouteToWarehouse`.
Usually the selected warehouse will be the harbor as it is trivially the closest.
However when the harbor does not accept this ware type another one might be chosen
and the ware needs to be moved out instead of stored.

But `FindRouteToWarehouse` does NOT set `next_dir` while the ware is carried even when
called from `RecalcRoute` where the caller assumes it is set.
Changing this causes replays to go async, so work around it by calling `RecalcRoute` again
which now succeeds as it reached the goal in the previous call.

This bug also affects `WantInBuilding` which also calls `RecalcRoute` assuming it does so in all cases.

Fixes https://github.com/Return-To-The-Roots/s25client/issues/1843

The test case was a bit tricky to write and I added the case where no goal is possible in which case the ware should be accepted in the ware house anyway.

### DIFF
--- a/libs/s25main/BuildingRegister.cpp
+++ b/libs/s25main/BuildingRegister.cpp
@@ -153,7 +153,7 @@ helpers::EnumArray<uint16_t, BuildingType> BuildingRegister::CalcProductivities(
 
 unsigned BuildingRegister::CalcAverageProductivity(BuildingType bldType) const
 {
-    if(!BLD_WORK_DESC[bldType].producedWare)
+    if(holds_alternative<boost::none_t>(BLD_WORK_DESC[bldType].producedWare))
         return 0;
     unsigned productivity = 0;
     const auto& buildings = GetBuildings(bldType);
@@ -173,7 +173,7 @@ unsigned short BuildingRegister::CalcAverageProductivity() const
     unsigned numBlds = 0;
     for(const auto bldType : helpers::enumRange<BuildingType>())
     {
-        if(!BLD_WORK_DESC[bldType].producedWare)
+        if(holds_alternative<boost::none_t>(BLD_WORK_DESC[bldType].producedWare))
             continue;
 
         const auto& buildings = GetBuildings(bldType);
--- a/libs/s25main/gameData/BuildingConsts.h
+++ b/libs/s25main/gameData/BuildingConsts.h
@@ -70,9 +70,8 @@ const helpers::EnumArray<BldWorkDescription, BuildingType> SUPPRESS_UNUSED BLD_W
   {Job::Helper, GoodType::Water},
   {Job::Shipwright, GoodType::Boat, WaresNeeded(GoodType::Boards)},
   {Job::Farmer, GoodType::Grain},
-  {Job::DonkeyBreeder, GoodType::Nothing,
-   WaresNeeded(GoodType::Grain, GoodType::Water)}, // Produces a job. TODO: Better way
-  {},                                              // Harbour
+  {Job::DonkeyBreeder, Job::PackDonkey, WaresNeeded(GoodType::Grain, GoodType::Water)},
+  {}, // Harbour
 }};
 
 /// Smoke consts for all buildings and nations
--- a/libs/s25main/gameTypes/BuildingTypes.h
+++ b/libs/s25main/gameTypes/BuildingTypes.h
@@ -9,6 +9,7 @@
 #include "JobTypes.h"
 #include "Point.h"
 #include "helpers/OptionalEnum.h"
+#include "variant.h"
 #include <cassert>
 
 struct BuildingCost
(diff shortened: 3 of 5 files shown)


stefson:
it seems as if I drove the logistics into an interesting corner case? :-)

libs/s25main/figures/nofWarehouseWorker.cpp
```diff
@@ -62,84 +60,78 @@ void nofWarehouseWorker::Draw(DrawPoint drawPt)
...
     if(!shouldBringWareIn)
     {
-        // Ware an der Fahne ablegen ( wenn noch genug Platz ist, 8 max pro Flagge!)
-        // außerdem ggf. Waren wieder mit reinnehmen, deren Zi­el zerstört wurde
-        // ( dann ist goal = location )
-        if(world->GetSpecObj<noFlag>(pos)->HasSpaceForWare() && carried_ware->GetGoal() != carried_ware->GetLocation()
-           && carried_ware->GetGoal() != wh)
+        // Put ware down at flag if enough space.
+        // Might need to take it back in if goal was destroyed or changed to the warehouse
+        if(flag->HasSpaceForWare() && carried_ware->GetGoal() && carried_ware->GetGoal() != wh)
```
Flow86, line 70: can there be no flag? (e.g `flag` being null?)
we should at least assert it

libs/s25main/figures/nofWarehouseWorker.cpp
```diff
@@ -62,84 +60,78 @@ void nofWarehouseWorker::Draw(DrawPoint drawPt)
...
         {
-            carried_ware->WaitAtFlag(world->GetSpecObj<noFlag>(pos));
+            // TODO: Remove assert. Was added to verify prior condition
```
Flow86, line 72: how do we resolve the TODO?

libs/s25main/Ware.cpp
```diff
@@ -112,6 +112,7 @@ void Ware::RecalcRoute()
             static_cast<nobHarborBuilding*>(goal)->WareDontWantToTravelByShip(this);
         } else
         {
+            // TODO(Replay) This should calculate the next dir even when carried
```
Flow86, line 115: theres another `TODO`?

libs/s25main/figures/nofWarehouseWorker.cpp
```diff
@@ -62,84 +60,78 @@ void nofWarehouseWorker::Draw(DrawPoint drawPt)
...
+            // Add to warehouse inventory (if it still exists and was not burnt down)
             if(world->GetNO(pos)->GetType() == NodalObjectType::Building)
             {
                 auto* wh = world->GetSpecObj<nobBaseWarehouse>(pos);
-                if(carried_ware->GetGoal() == carried_ware->GetLocation() || carried_ware->GetGoal() == wh)
+                // Store the ware if its goal is this warehouse or it has no goal (anymore)
+                // Else it wants to go somewhere else, so add to waiting wares
+                if(!carried_ware->GetGoal() || carried_ware->GetGoal() == wh)
+                {
+                    // TODO: Remove assert. Was added to verify prior condition
```
Flow86, line 113: another TODO?

Flow86 (changes requested):

libs/s25main/figures/nofWarehouseWorker.cpp
Flamefire: If there was no flag, there is no warehouse and the worked would have been notified to abort the work. So yes there is always a flag

libs/s25main/figures/nofWarehouseWorker.cpp
Flamefire: Keep it for a short while, then remove. Possible to remove now as the auto-play tests succeeded but might be worth collecting a bit more evidence although I'm very sure this is correct.

libs/s25main/figures/nofWarehouseWorker.cpp
Flamefire: Same as above

libs/s25main/Ware.cpp
Flamefire: Yes: "TODO(Replay)" marks places that should be addressed when the replay version is increased, i.e. breaking old replays. I added this because it would be correct to do but can't be done as replays will go async

libs/s25main/figures/nofWarehouseWorker.cpp
Flow86: well, then we should assert it. I don't like non checked pointers

Flamefire: Ok done. Also using `wh` where it makes sense to avoid fetching it multiple times and `wh->GetFlag()` is cleaner

Flow86 (approved):

Flamefire:
> it seems as if I drove the logistics into an interesting corner case? :-)

Yeah, it's a timing issue and hence difficult to reproduce in "normal" circumstances so I cheated a bit for the test case.

PR #1854, opened 2026-01-06
Order nations in game lobby as in original S2

### LINKED ISSUE #1853: Nations wrongly ordered
Currently in RttR we have:
Romans
Vikings
Babylonians
Africans
Japanese

Just as in the original it should be:
Romans
Vikings
Japanese
Nubian
new Babylonian

Funny thing is, that the portraits are oredered correctly.

Flamefire:
Use an array for the order to avoid changing the enum which breaks savegames and replays.

Closes #1853

### DIFF
--- a/libs/s25main/desktops/dskGameLobby.cpp
+++ b/libs/s25main/desktops/dskGameLobby.cpp
@@ -1,4 +1,4 @@
-// Copyright (C) 2005 - 2021 Settlers Freaks (sf-team at siedler25.org)
+// Copyright (C) 2005 - 2025 Settlers Freaks (sf-team at siedler25.org)
 //
 // SPDX-License-Identifier: GPL-2.0-or-later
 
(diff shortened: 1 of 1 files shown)


libs/s25main/desktops/dskGameLobby.cpp
```diff
@@ -106,6 +108,21 @@ constexpr T nextEnumValue(T value)
...
+
+std::array NATION_ORDER = {
+  Nation::Romans, Nation::Vikings, Nation::Japanese, Nation::Africans, Nation::Babylonians,
+};
+static_assert(NATION_ORDER.size() == helpers::NumEnumValues_v<Nation>);
+
+Nation nextNation(const Nation value)
+{
+    // NOLINTNEXTLINE(readability-qualified-auto)
+    auto it = helpers::find(NATION_ORDER, value);
```
Flow86, line 120: why not add the `*` to the `auto` to avoid the `NOLINT`?
or better, if thats an iterator, why is the NOLINT even necessary?

Flamefire: Because it is an iterator that only happens to be a pointer. Clang-tidy 21 allows to suppress this: https://github.com/llvm/llvm-project/issues/63461

For the purpose of this code this is an iterator, so I didn't want to declare it a pointer.

PR #1859, opened 2026-01-12
Add missing Hungarian accented glyphs to font

anon569:
This PR adds a small number of missing Hungarian accented characters to the font.

[fonts.zip](https://github.com/user-attachments/files/24586921/fonts.zip)
Képernyőfotó 2026-01-13 - 11 08 17
Képernyőfotó 2026-01-13 - 11 08 42
Képernyőfotó 2026-01-13 - 11 08 51


Flamefire:
I assume you tested the updated archive locally and it looks fine for you?

Could you show/attach the added glyphs here in the PR? Just to make it easier to see the change/addition, maybe @Spikeone wants to have a look to

anon569:
Hi @Flamefire,

I've been using this locally for a few months without any issues, and I also shared it with a few others who haven’t given any negative feedback.  

I’ve now uploaded screenshots of the new glyphs and attached them as a ZIP.

Flamefire:
LGTM but I'd like to hear the opinion of @Spikeone who knows more about the graphics

Only thing that I notice is that it looks like some can be cropped to the minimal size (without the empty space at the bottom) by using `ny`. But IIRC the RLE format used to store them makes that go away on disk. So not important, just noticed this

Spikeone:
As it's not my native language, I have no idea if they are looking good or not:
grafik

but looking fine for me

Flamefire (approved):

PR #1862, opened 2026-01-16
Add option to jump by specific number of GFs to jump window

### LINKED ISSUE #1861: Enhance Jump-Screen
I'm using the Jump-Screen quite often, and it'd be nice to have some options to jump fixed frames. So one doesn't have to enter values manually.
e.g.: +100, +1000, +5000, +10000, and maybe even a + button with an input next to it, so if I always want to jump 2500, I could just enter that value (which maybe makes some of the previous values obsolete).

Flamefire:
Closes #1861

{ACC24F65-449F-45B1-9E40-55871498D153}

@Spikeone Like this?

### DIFF
--- a/libs/s25main/ingameWindows/iwSkipGFs.cpp
+++ b/libs/s25main/ingameWindows/iwSkipGFs.cpp
@@ -64,7 +64,7 @@ iwSkipGFs::iwSkipGFs(GameWorldView& gwv)
     btSize.x = (availWidth - spacing * (jumpPresets.size() - 1)) / jumpPresets.size();
     for(auto i : jumpPresets)
     {
-        AddTextButton(ID_btJumpPresetStart + i, curPos, btSize, TextureColor::Green1, "+" + std::to_string(i),
+        AddTextButton(ID_btJumpPresetStart + i, curPos, btSize, TextureColor::Green2, "+" + std::to_string(i),
                       NormalFont);
         curPos.x += btSize.x + spacing;
     }


libs/s25main/ingameWindows/iwSkipGFs.cpp
```diff
@@ -5,39 +5,91 @@
...
 
-    // OK-Button
-    AddTextButton(2, DrawPoint(110, 65), Extent(80, 22), TextureColor::Green2, _("OK"), NormalFont);
+    curPos.x = contentOffset.x + spacing;
+    curPos.y += edtSize.y + 5;
+    const auto availWidth = GetIwSize().x - spacing * 2;
+    btSize.x = (availWidth - spacing * (jumpPresets.size() - 1)) / jumpPresets.size();
+    for(auto i : helpers::range(jumpPresets.size()))
+    {
+        AddTextButton(ID_btJumpPresetStart + i, curPos, btSize, TextureColor::Green1,
```
Flow86: why not simply `+ jumpPresets[i]`? (e.g iterate over the values)

it doesnt matter how big those ids are or if there are gaps.

Flamefire: Makes sense, yes.

Spikeone:
@Flamefire I'm loving it - except the button texture which is wrong for the 4 lower buttons ;)

Flamefire:
I used a different texture for the buttons at the bottom to make them stick out a bit. Can revert it or use another one. Which would you like?

Spikeone:
The same texture as the two upper right buttons - I don't know why but these look like the only correct ones (gray, red and green)

Flamefire:
Ok done

Spikeone (approved):

PR #1864, opened 2026-01-17
Add Korean glyphs

### LINKED ISSUE #1860: ??? shows on text
Recent 5452ee91d43f5f2d7dd539c16e82769c8c509787 nightly build, it may select default system language file correctly but not shown properly.   
My local language is Korean.  
It may font is missing on game resources.  
  
Image

Text shows ??? ?? on screens.  
When change to English or chinese, it has no problem.

Flamefire:
Contributed by @rasberryrabbit
Closes #1860


Flow86:
so basically this blows up the file from 2kb to 11mb? how much time&memory does reading the font cost? maybe we need to either publish this extra (e.g for korean support install this file ...)

Flamefire:
> so basically this blows up the file from 2kb to 11mb?

Where did you get those numbers from? File size goes from 10.7MB to 11.8MB. So 10%, not too bad for that amount of glyphs

rasberryrabbit:
It also include japanese unicode characters.

Flow86:
> > so basically this blows up the file from 2kb to 11mb?
> 
> Where did you get those numbers from? File size goes from 10.7MB to 11.8MB. So 10%, not too bad for that amount of glyphs

oh okay, somehow github showed me weird numbers :-)

okay. Can we have a diff of the glyphs contained before/after? only to check that everything is right? (Maybe we should add some reporting to the github actions for changes like this?

Flamefire:
> okay. Can we have a diff of the glyphs contained before/after? only to check that everything is right? (Maybe we should add some reporting to the github actions for changes like this?

I manually verified that only new images were added. If I didn't make a mistake this should be safe

rasberryrabbit:
Korean unicode list -> 2350 glyphs  
[KS1001.txt](https://github.com/user-attachments/files/24699777/KS1001.txt)
  
0x3040-0x309F -> Japanese Hiragana
0x30A0-0x30FF ->  Japanese Katakana
  
It contains in new fonts.lst.  

There is individual glyphs archive file at #1860 .

PR #1865, opened 2026-01-17
Fix waiting attackers becoming stuck when target flag becomes unreachable.

### LINKED ISSUE #1863: Crash on save on Windows after long playtime, pure virtual method called
### Description
RTTR crashes on **Windows** during a save operation with the following runtime error:

```
pure virtual method called
terminate called without an active exception
```

The same savegame behaves differently on **Linux**: it loads and saves correctly.  
After re-saving the game on Linux and copying the savefile back to Windows, the crash no longer occurs (at least so far).

This suggests **platform-dependent behavior**, possibly related to savegame serialization or object lifetime handling.

---

#### Windows (crash occurs)
- OS: Windows 11 Pro (also reproduced on Windows 10)
- RTTR version: `v20260114-5452ee9`
- Architectures tested: x86_64 and i386
- CPU: Intel Core i11 (11th gen)
- GPU: NVIDIA GeForce RTX 4070
- Graphics backend: SDL2 OpenGL (OpenGL 4.6 supported)
- Audio backend: SDL2_mixer
- Install path: `E:\Settlers2`
- User data path:  
  `C:\Users\Bell®\Saved Games\Return To The Roots`

#### Linux (no crash observed)
- OS: Ubuntu 24.04.3
- RTTR version: `v20260114-5452ee9`
- Runtime: VMware® Workstation 17 Pro  
  Version 17.6.4 (build-24832109)

---

### Steps to reproduce (Windows)
1. Start RTTR on Windows
2. Load the attached savegame
3. Play the game for a longer period of time
4. Attempt to save the game manually
5. Game crashes with:
```
pure virtual method called
terminate called without an active exception
```

---

### Steps to reproduce (Linux)
1. Start RTTR on Linux
2. Load the same savegame
3. Play for some time
4. Save the game

**Result:** save succeeds, no crash.

---

### Additional timing details
- The crash occurred after approximately **20 hours and 31 minutes of gameplay on the same map**
- Before the crash:
- Autosave was enabled with a **10-minute interval**
- Autosave worked reliably for the entire session (≈120 autosaves)
- The crash happened during a **manual save attempt**

This points to a **long-running session / accumulated state issue** rather than a general save failure.

---

### Workaround / Additional findings
- Saving the game on Linux and copying the newly created savefile back to Windows prevents the crash (so far).
- Linux appears to re-serialize the game state in a way that avoids triggering the issue on Windows.

---

### Expected behavior
Saving the game should not crash on any platform.

### Actual behavior
- Windows: crash during save
- Linux: save succeeds

---

### Notes / Hypothesis
- The error looks like a C++ runtime issue (`pure virtual method called`), possibly caused by invalid object lifetime or use-after-free.
- The issue appears platform-specific and may be related to:
- savegame serialization/deserialization
- long-running sessions with accumulated game state
- undefined behavior manifesting differently on Windows vs Linux (compiler / STL / memory layout differences)

---

### Attachments
- Original savegame file (before crash)
- Savegame file re-saved on Linux
- Full Windows console log

[LOGS.zip](https://github.com/user-attachments/files/24683736/LOGS.zip)

Flamefire:
When the defender kills the current attacker (at the flag) `FindAttackerNearBuilding` is used to get the next one for the defender to fight. This takes the closest attacker that is ready, i.e. waiting around the flag and calls `AttackDefenderAtFlag` which usually causes the attacker to start walking to the flag. However if there is no path to the flag (anymore, e.g. after destroying a road) the attacker won't do anything and the function returns `false`.

As this isn't checked the defender will keep waiting for it.

This can lead to use-after-free if the attacker is destroyed (defeated in free fight, went back home, ...) as the defender will then have a dangling pointer to it.

Fixes #1863

### DIFF
--- a/libs/s25main/buildings/nobBaseMilitary.cpp
+++ b/libs/s25main/buildings/nobBaseMilitary.cpp
@@ -188,8 +188,9 @@ MapPoint nobBaseMilitary::FindAnAttackerPlace(unsigned short& ret_radius, const 
     // Check all points around the flag and take shortest
     unsigned min_length = std::numeric_limits<unsigned>::max();
     MapPoint minPt = MapPoint::Invalid();
-    ret_radius = 100;
-    for(const auto& node : world->GetPointsInRadius(flagPos, 3, ReturnMapPointWithRadius{}))
+    constexpr auto MAX_RADIUS = 3;
+    ret_radius = MAX_RADIUS;
+    for(const auto& node : world->GetPointsInRadius(flagPos, MAX_RADIUS, ReturnMapPointWithRadius{}))
     {
         // We found a point with a better radius
         if(node.second > ret_radius)


libs/s25main/buildings/nobBaseMilitary.cpp
```diff
@@ -165,29 +165,31 @@ nofAttacker* nobBaseMilitary::FindAggressor(nofAggressiveDefender& defender)
...
-    const MapPoint soldierPos = soldier.GetPos();
-    // Get points AROUND the flag. Never AT the flag
-    const auto nodes = world->GetPointsInRadius(flagPos, 3, ReturnMapPointWithRadius{});
-
-    // Weg zu allen möglichen Punkten berechnen und den mit den kürzesten Weg nehmen
-    // Die bisher kürzeste gefundene Länge
+    // Check all points around the flag and take shortest
     unsigned min_length = std::numeric_limits<unsigned>::max();
     MapPoint minPt = MapPoint::Invalid();
     ret_radius = 100;
```
Flow86: should this then start at MAX_ATTACKING_RUN_DISTANCE  too?

Flamefire: Actually this assignment could be removed. It is never used because the value will be overwritten when `minPt` gets set and if it isn't then `ret_radius` has no meaning

Flow86: can we explicitly make it `std::optional` then? maybe its even eases up the do-while?

Flamefire: Wait, this is used as the max value in the loop.
So it corresponds to `GetPointsInRadius`.

Added a constant to clarify

Issue #1873, opened 2026-01-24
spacebar for showing construcion aid is broken (on X)


stefson:
hey, I just reemerged current head of git and was met with a non-working hotkey of spacebar --> show me construction aid icons

I wonder if it has been broken here: https://github.com/Return-To-The-Roots/s25client/pull/1872

I'm on xorg

stefson:
downgrade to https://github.com/Return-To-The-Roots/s25client/commit/12da8bf447c6226ed4c126f5aaedbcf8fd34df5c fixed it for me.

Flow86:
Hmm, does it maybe trigger "both" - e.g two events, one with Spacebar, and one with text " "?

pnlabs:
Yes, space do not work anymore on Ubuntu Linux / Gnome

JanSimek:
I was afraid of that happening.

> Here's a fix that works for me, but I am not sure how it would behave on other platforms - if it would toggle the overlay twice, for example.

I think the proper fix would be to call `SDL_StartTextInput()` only when accepting text input, followed by `SDL_StopTextInput()` when done, but, as mentioned by @Flamefire, it would be a significant change. 

Here is a PR to revert the change: https://github.com/Return-To-The-Roots/s25client/pull/1875

Flamefire:
I forgot: we should remove the space bar  hockey and only keep the character one you added

JanSimek:
OK, I created another PR to remove the key press event handling for spacebar: https://github.com/Return-To-The-Roots/s25client/pull/1876

Btw, do you also use Wayland, @pnlabs ?

rEVolution-18650:
I just discovered RttR and what a flashback to my childhood this is! A **huge** thanks to everybody involved! Looking forward to play including the fun additions such as wine/temple or looking at soldier health. Widescreen support and smooth zoom would have blown my mind back then for sure ;-)

Running it on OS X M1. Everything is working so far, except for this issue described here. No construction aid for me, also not when using X as a hotkey as described by @stefson in another thread.

How can I "downgrade to 12da8bf" as suggested? I just downloaded the 0.9.5 app bundle and I'm more the hardware guy when it comes to computers ;-)

stefson:
I downgraded it "myself" by compiling the code into a binary. The fix should land in the near future, so you can either wait a bit for an updated nightly, or try to download an older nightly binary if thats possible. 

[...]

PR #1876, opened 2026-01-25
Capture spacebar as a character only (#1873)

JanSimek:


stefson:
this works on top of current head of git, construction aid is back on X

PR #1720, opened 2024-12-01
Add leather addon

ottml:
Fixes #1711

PR #1882, opened 2026-02-01
Fix collection of jobs to warehouses

Flamefire:
When the "Collect" option in a warehouse for a job is set the actual job subclass will be created which expects its workplace as the goal, i.e. a `nobUsual` for most.
In this situtation the goal is the warehouse to which the figure should go which triggers an assertion.

So check that in `OrderJob` and create a `nofPassiveWorker` instead similar to the  `Send` option handler.

Contains a bit of #1720 to reduce conflicts

A major role in the misunderstanding was likely the `nofBuildingWorker(const Job job, const MapPoint pos, const unsigned char player, nobBaseWarehouse* goalWh)`  overload (warehouse last) which was never used but looks intended for this case. `nofPassiveWorker` handles this case already without having to maintain a ctor through the whole hierarchy

### DIFF
--- a/libs/s25main/GamePlayer.cpp
+++ b/libs/s25main/GamePlayer.cpp
@@ -620,35 +620,35 @@ void GamePlayer::RoadDestroyed()
     }
 }
 
-bool GamePlayer::FindCarrierForRoad(RoadSegment* rs) const
+bool GamePlayer::FindCarrierForRoad(RoadSegment& rs) const
 {
-    RTTR_Assert(rs->GetF1() != nullptr && rs->GetF2() != nullptr);
+    RTTR_Assert(rs.GetF1() != nullptr && rs.GetF2() != nullptr);
     std::array<unsigned, 2> length;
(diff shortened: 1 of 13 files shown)


libs/common/include/commonDefines.h
```diff
@@ -32,6 +32,13 @@ inline T checkedCast(T_Src* src)
     return static_cast<T>(src);
 }
 
+template<typename T, typename T_Src>
+inline T& checkedCast(T_Src& src)
+{
+    RTTR_Assert(dynamic_cast<T*>(&src));
+    return static_cast<T&>(src);
```
Flow86, line 45: static_cast here is wrong. you also use dynamic_cast with references.

using a dynamic_cast, you normally get a `bad_cast` exception if the dynamic_cast fails

Flamefire: The cast can never fail, this is the whole purpose of this function, see the other overload too. So no need for a release-enabled `dynamic_cast`.
I copied and enhanced the doc from the other overload to clarify.

Issue #183, opened 2015-08-09
[Addon] Destroy inactive enemy ships


Spikeone:
It would be great if there was an option which just destroys enemy ships if they are within your borders and inactive.
This could happen if you overtake an enemy island or just destroy most enemy harbors so some ships just idle.

Flow86:
well, they could simply leave and go to the next harbor perhaps, if they're inside enemy territory - if none exist, they should go to nearest harbor point which is closest to a friendly territory

Spikeone:
Maybe both - an other dropdown menu. I personally like the idea of actually being able to destroy enemy ships. Thats also kinda important when you defeat an enemy and there are only ships left all over the coast(s). Currently ships are the only thing a player can build which could never be destroyed nor decayed nor anything else. Therefore I liked that idea of losing more than just a harbor building but also ships.

Flow86:
well there could be the catapult addon come in - if the ships empty and i.e hit 3 times, the ship will be destroyed and sink (perhaps replace the position with a "sunken ship" if its near the land

Spikeone:
Well, if the catapults hit ships addon had an (referring my explanation) T0 state: only attack idle ships, I could go with that.
But please DO NOT replace those ships with sunken ships, that way harbor spots could be blocked really easy (currently I have that problem when editing maps, the new editor doesn't show those ships but in RttR they can block harbors even if they are 2 nodes away), a fire would be enough in my opinion. Or maybe if the distance is +10 to any harbor, but this could also block some rivers thats why I think it could be bad to add such a function.

bluss:
Half-built ships also block harbors spots (and you can't remove them, except if the original player completes the ship).

Flamefire:
What about: If the player has no more harbours at that sea, all ships are destroyed. Same as people wander around and die when they don't find any warehouse.

Spikeone:
Thats a bad idea, sometimes you only have 1 harbour and they might be destroyed very fast. Also you may build shipts before you build a harbour

Flamefire:
Then something similar. E.g. no harbour and no shipbuilder or so... At least if player is destroyed then ships are also destroyed (if this is not yet the case)

Spikeone:
Still - destroying them if you overtake the land where they stand would be great as well. Sometimes players are not dead yet.
Would be quite easy compared to other solutions.
Also maybe we should just add a prohibition for shipwrights to build ships near harbour spots and blocking them that way.

OR:
[...]

PR #1883, opened 2026-02-02
 Destroy ships when player is defeated 

Flamefire:
As suggested in https://github.com/Return-To-The-Roots/s25client/issues/183 ships will now be destroyed when a player is defeated or surrenders.

Requires:
- [x] #1882

### DIFF
--- a/libs/s25main/GamePlayer.cpp
+++ b/libs/s25main/GamePlayer.cpp
@@ -1512,6 +1512,10 @@ void GamePlayer::Surrender()
     if(isDefeated)
         return;
 
+    const auto shipsCopy = ships; // copy to avoid modification during iteration
+    for(auto* ship : shipsCopy)
+        ship->Sink();
+
     isDefeated = true;
 
     // GUI Bescheid sagen
--- a/libs/s25main/GamePlayer.cpp
+++ b/libs/s25main/GamePlayer.cpp
@@ -1981,12 +1981,10 @@ bool GamePlayer::OrderShip(nobHarborBuilding& hb)
     return false;
 }
 
-/// Meldet das Schiff wieder ab
-void GamePlayer::RemoveShip(noShip* ship)
+void GamePlayer::RemoveShip(noShip& ship)
 {
-    auto it = helpers::find(ships, ship);
-    RTTR_Assert(it != ships.end());
-    ships.erase(it);
+    RTTR_Assert(helpers::contains(ships, &ship));
+    helpers::erase(ships, &ship);
 }
 
 /// Versucht, für ein untätiges Schiff eine Arbeit zu suchen
--- a/libs/s25main/nodeObjs/noShip.h
+++ b/libs/s25main/nodeObjs/noShip.h
@@ -1,4 +1,4 @@
-// Copyright (C) 2005 - 2021 Settlers Freaks (sf-team at siedler25.org)
+// Copyright (C) 2005 - 2026 Settlers Freaks (sf-team at siedler25.org)
 //
 // SPDX-License-Identifier: GPL-2.0-or-later
 
@@ -224,4 +224,7 @@ public:
     void HarborDestroyed(nobHarborBuilding* hb);
     /// Sagt dem Schiff, dass ein neuer Hafen erbaut wurde
     void NewHarborBuilt(nobHarborBuilding* hb);
+
+    /// Destroy the ship immediately
+    void Sink();
 };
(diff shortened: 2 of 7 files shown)


Spikeone:
Juast asking - this is not an addon, is it? I'm actually not sure if it should or not. Also if it should depend on if a player is defeated with destroying buildings or without. Guess this is just always which should be fine I guess, although this might stop ongoing sea attacks opposed to a normal attack which would still be going on.

Flamefire:
This is not an addon. When a player is defeated (no more military, warehouses/HQs) or he surrenders his ships are destroyed/removed.

I don't see any issue with that and it's what I got out of #183

> Thats also kinda important when you defeat an enemy and there are only ships left all over the coast(s).

For people etc. they will wander around and eventually die, so why not do the same for ships as everything else will be gone too.

Flow86:
Rebase necessary

Flamefire:
@Flow86 *ping

libs/s25main/GamePlayer.cpp
```diff
@@ -1512,6 +1512,10 @@ void GamePlayer::Surrender()
     if(isDefeated)
         return;
 
+    const auto shipsCopy = ships; // copy to avoid modification during iteration
+    for(auto* ship : shipsCopy)
+        ship->Sink();
+
```
Flow86, line 1518: shouldn't this be an addon? in the original RttR, defeated ships were not destroyed

libs/s25main/nodeObjs/noShip.h
```diff
@@ -224,4 +224,7 @@ class noShip : public noMovable
     void HarborDestroyed(nobHarborBuilding* hb);
     /// Sagt dem Schiff, dass ein neuer Hafen erbaut wurde
     void NewHarborBuilt(nobHarborBuilding* hb);
+
+    /// Destroy the ship immediately
+    void Sink();
```
Flow86: more likely 

```suggestion
    /// Remove the ship and its contents
    void Sink();
```

Flow86:
> I guess, although this might stop ongoing sea attacks opposed to a normal attack which would still be going on.

Well, this is maybe not wanted by default. If you loose your HQ, you still can reconquer stuff right? but I assume then you not count as "defeated" yet, right?

libs/s25main/GamePlayer.cpp
Spikeone: Depends on - in the original only the player ever had ships, so it's a quite RttRish problem. Although I think just removing them is somewhat funny, since there is no burning/sinking animation. Maybe we should add some kind of animation (even the fire used for buildings does make sense I think).

Flamefire:
IIRC attacking soldiers get "cancelled" when their home building is destroyed, at least up to a certain point in their conquering.

"Defeated" means you have no more military buildings and no more storehouses. 
But in any case: Even if they conquer anything the player cannot do anything else as he will never again have carriers or wares.

libs/s25main/GamePlayer.cpp
Flow86: yep some burning/oil spill/whatever would be cool

Flamefire: Using the small building fire as a vanishing, non-blocking object would work and not cause much trouble. Or do you want to make a new animation?

Flow86: if that works, I think thats a good thing.

Spikeone: I think we don't need a new animation and would like the small building fire as you described it @Flamefire

Spikeone:
> "Defeated" means you have no more military buildings and no more storehouses.
> But in any case: Even if they conquer anything the player cannot do anything else as he will never again have carriers or wares.

Well yeah it's hard to know what defeated in most cases means. For me a player counts as defeated, when he lost control. This can be either giving up or by game goal. Players who have no storehouse left, can still attack and are maybe in terms of winnig often defeated, but they can still attack and annoy you. So only if the player is really defeated (lost control) then they should lose their ships I think.

Flamefire:
>> "Defeated" means you have no more military buildings and no more storehouses.
> Players who have no storehouse left, can still attack and are maybe in terms of winnig often defeated, but they can still attack and annoy you.

That's our definition of defeated (in the code) so those players cannot attack you anymore. Hence it's safe to remove their ships, isn't it?

Spikeone:
> That's our definition of defeated (in the code) so those players cannot attack you anymore. Hence it's safe to remove their ships, isn't it?

yes

Issue #1884, opened 2026-02-02
Player Colors set by LUA not working correctly


matony1989:
I am trying to assign colors to characters (portraits) in custom campaign. It would be more memorable to know when you are fighting “red” opponent, that it is always the same “character”. What I noticed is that colors set in LUAs are acting strange.

Tested on versions 20260124 and 20250626

luas and maps can be downloaded here:
[Eric Ver1.1.zip](https://github.com/user-attachments/files/25011614/Eric.Ver1.1.zip)

<img width="1516" height="359" alt="Image" src="https://github.com/user-attachments/assets/ae2eb51b-c82b-46cf-b1b6-89010b9ffeb2" />

Color index : 
0    Blue
1    Yellow
2    Red
3    Pink
4    Grey
5    Green
6    Orange
7    Turquoise
8    White
9    Brown
10    Purple

Mission 2 example:
Player 1 = white (8)
Player 2 = red (2)
Player 3 = turquoise (7)

8(white) 2(red) 7(turquoise) results in 8(white) 3(pink) 7(turquoise) in game menu

I tried :
8 7 2 => 8 7 2 correct – Just changed order of colors (cannot be used in every mission)
8 0 7 => 8 0 7 correct
8 1 7 => 8 1 7 correct
8 3 7 => 8 3 7 correct ( same result as 8 2 7)
8 4 7 => 8 4 7 correct (Initially I thought grey is causing issues, you will see why later)
5, 6, 9 , 10 correct

All colors are working, except 2-red for map 2 on player position 2.
Tried also 8 7 9 – result 8 7 9 correct

Map 3:
8 0 2 => 8 0 2 Correct
Map 4:
8 5 2 7 => 8 5 2 7 Correct
Map 5 same setup as map 2, same result
Map 6
8 1 2 6 => 8 1 2 6 Correct
Map 7
8 10 2 =>  8 10 2 Correct
Tried 8 2 10 => 8 3 10 

Map 8
8 0 2 7 5 => correct

Map 9
8 1 6 2 =>  correct
Map 10
8 10 5 4 2 => 8 10 5 6 2 – This is where I first noticed it. Instead of Grey 4 I got first available color from 4 up = 6
Map 11
8 10 1 2 7 - correct
Map 12
8 3 9 2 7 => 8 5 9 2 7
Map 13
8 5 2 4 0 => 8 5 2 6 0
Map 14
8 1 2 3 6 7 4 => 8 1 2 3 7 9 4
[...]

PR #1885, opened 2026-02-03
 Allow changing color of unused player slots 

Flamefire:
This is useful for use by LUA to set player colors before filling the
slots avoiding temporary conflicts (same colors).

See https://github.com/Return-To-The-Roots/s25client/issues/1884

Mention this in the docs which were wrong in this regard

Edit: doesn't work as in single player games all slots are filled by AI

### DIFF
--- a/doc/lua/functions.md
+++ b/doc/lua/functions.md
@@ -230,7 +230,10 @@ Change the players team.
 
 **SetColor(color or colorIdx)**  
 Sets the players color by index or directly if its alpha value is not zero.
-Duplicate color values are possible, so you have to ensure unique colors if you want them!
+When the color is currently used by an active player/AI, a new one will be chosen out of the standard player colors.  
+You can change the colors freely for slots not yet taken, so if the final distribution of colors is unique,
+they will be kept after using e.g. `SetAI`.  
+In any case, using `SetAI` will ensure a unique color.
 
 **Close()**  
 Closes a spot kicking any player or AI there.


Flow86:
can we maybe change/extend the lua code so you have to specify all colors at once, e.g to set all colors at once?

Flamefire:
That would still require adapting the protocol as it is the server that sets the colors.
So my best idea is to have a "force" parameter for the protocol that changes the *other* conflicting color instead of the current/new one.

This way you get exactly what you want when you do `GetPlayer:SetColor` for this player, but if you don't set the other player you'll get a different one there.

Or we remove that part checking colors on set: We have a readiness check that verifies unique colors: https://github.com/Flamefire/s25client/blob/fb01a4f33a076b7fd63b174d0ac7bde0bb176544/libs/s25main/network/GameServer.cpp#L1446-L1458

With that we could check the colors only before starting. Given the original documentation of the Lua function we might have even intended to allow 2 players with the same color, at least for scripted maps

Spikeone:
> With that we could check the colors only before starting. Given the original documentation of the Lua function we might have even intended to allow 2 players with the same color, at least for scripted maps

Well, this does sound great, since you then could place 2 HQs somewhere and make it seem as if they are the same player, or just make alle enemies red.

On the other hand, this might lead to unexpected results when inexperienced users try to write LUA, but I think with proper documentation, this should be fine.

Flamefire:
> On the other hand, this might lead to unexpected results when inexperienced users try to write LUA, but I think with proper documentation, this should be fine.

I mean that is what is documented right now so making sure it works would be a bugfix.

When using the color toggle manually it picks a free color. There is only a small chance when 2 people change their color at the "same" time to have duplicates.   
So shall we check at all for duplicates, only when setting and/or allow when set by Lua?
I would go the easy route and accept any color requested by a client/player but make the UI select the next free color when clicked, which might be duplicate if another player is just changing to this color too

Flamefire:
I went with that solution: https://github.com/Return-To-The-Roots/s25client/pull/1888

PR #1886, opened 2026-02-03
Allow relative path for campaign images

Flamefire:
Custom campaigns are not in `<RTTR_RTTR>` but in the user-data folder.
Hence it would be impossible to refer to custom images as shown in the example.

Allow a single subfolder level for images and adapt the docs

### DIFF
--- a/doc/AddingCustomCampaign.md
+++ b/doc/AddingCustomCampaign.md
@@ -147,3 +147,6 @@ RTTR/campaigns/garden/mapscreen/map_mask.bmp
 RTTR/campaigns/garden/mapscreen/marker.bmp
 RTTR/campaigns/garden/mapscreen/conquered.bmp
 ```
+
+Note that for custom campaigns, placed in `<RTTR_USERDATA>/campaigns/` the top-level `RTTR` folder won't exist.
+Hence, `<RTTR_RTTR>/campaigns/` should not be used in the campaign description file to refer to campaign files.
--- a/doc/AddingCustomCampaign.md
+++ b/doc/AddingCustomCampaign.md
@@ -6,7 +6,7 @@ SPDX-License-Identifier: GPL-2.0-or-later
 
 # Add a custom campaign
 
-You can have a look into the already existing original Settlers 2 campaigns in the subfolders `RTTR/campaigns/roman` and `RTTR/campaigns/continent` for an example.
+You can take a look into the already existing original Settlers 2 campaigns in the subfolders `RTTR/campaigns/roman` and `RTTR/campaigns/continent` for an example.
 We will now create an example campaign `garden`.
 
 ## Location for adding a new campaign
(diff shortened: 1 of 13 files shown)


Spikeone:
Thats pretty cool, so one could create his own world map?

libs/libGamedata/gameData/CampaignDescription.cpp
```diff
@@ -1,25 +1,55 @@
...
+            // If it is only a file/folder name or empty use path relative to campaign folder
+            if(!tmpPath.has_parent_path())
+                return (campaignPath / tmpPath).string();
+            if(!isFolder)
+            {
+                // For files only allow a single sub folder
+                const auto parentPath = tmpPath.parent_path();
+                if(!parentPath.parent_path().has_parent_path())
+                {
+                    // Only alpha-numeric folder names are allowed
```
Spikeone, line 28: so my-file.bmp is forbidden? And why is that at all?

doc/AddingCustomCampaign.md
```diff
@@ -41,18 +41,20 @@ campaign = {
     name = _"name",
     shortDescription = _"shortDescription",
     longDescription = _"longDescription",
-    image = "<RTTR_RTTR>/campaigns/garden/garden.bmp",
+    image = "<RTTR_RTTR>/campaigns/garden/garden.bmp", -- Same as: image = "garden.bmp",
```
Spikeone, line 44: what formats are supported? LBM, BMP?

libs/libGamedata/gameData/CampaignDescription.cpp
Flow86: maybe we can allow - and _ ?

Flamefire: No, only the folder name is checked. I don't see a need for `campaign-images` or similar inside a campaign folder

doc/AddingCustomCampaign.md
Flamefire: Pretty much anything that has a single image. So yes, LBM & BMP

Flamefire: This is now documented.
I'm wondering if this should support indexed archives similar to the selection map. But probably overkill

Spikeone: well, I was about to ask and hope if we could support loading custom objects as well, as I'm still hoping for that addition

Flamefire: What do you mean by  "loading custom objects"?

Do you have any concrete use case for that as `image`?

Spikeone: I do, some examples do include but are not limited:
- I want to add 4 gates, only the last one reached will activate, all others I'd like to replace with a broken gate (as if they crumble after activation)
- I want to add some custom objects for a campaign where you search for armor pieces
- I would like to add some kind of either treasure chest/dig site where you find something
- I would like to add custom barrier objects
- I would possibly like to add some ruin variations of different buildings
- Also (but that would require loading it into IO and not as a missbob) would like to have custom graphics for the messageboxEx

And now that I'm wiring this, I realize, that his would not work in multiplayer at all, or rather would require further implementations. Maybe if we put it into a .lst file we'd distribute then.

Flamefire: Ok, so that is unrelated to the campaign description and could be done later.

As for placing objects: `AddStaticObject` can use many graphics already. We could make it accept a resource ID instead of a file number to refer to other archives.

As for custom graphics we would indeed be limited to campaigns as we don't transfer archives to clients. Even then we'd need to somehow pass the campaign folder/subfolder to the loader but it supports such overwriting already and we use it for e.g. addons already

PR #1888, opened 2026-02-07
Allow duplicate colors when map has LUA script

Flamefire:
For e.g. campaigns the creator might intend duplicate colors so allow that.
This also resolves the issue where color set by LUA is changed as it is
still taken even though the conflicting players color might get changed
right afterwards so colors would be unique when the script is done.
That makes setting the colors dependent on the order which is hard.

This also makes the behavior consistent with the documentation.

Fixes https://github.com/Return-To-The-Roots/s25client/issues/1884

### DIFF
--- a/extras/videoDrivers/SDL2/VideoSDL2.cpp
+++ b/extras/videoDrivers/SDL2/VideoSDL2.cpp
@@ -392,29 +392,43 @@ bool VideoSDL2::MessageLoop()
             case SDL_MOUSEBUTTONDOWN:
                 mouse_xy.pos = getGuiScale().screenToView(Position(ev.button.x, ev.button.y));
 
-                if(/*!mouse_xy.ldown && */ ev.button.button == SDL_BUTTON_LEFT)
+                switch(ev.button.button)
                 {
-                    mouse_xy.ldown = true;
-                    CallBack->Msg_LeftDown(mouse_xy);
-                }
-                if(/*!mouse_xy.rdown &&*/ ev.button.button == SDL_BUTTON_RIGHT)
-                {
-                    mouse_xy.rdown = true;
-                    CallBack->Msg_RightDown(mouse_xy);
+                    case SDL_BUTTON_LEFT:
+                        mouse_xy.ldown = true;
+                        CallBack->Msg_LeftDown(mouse_xy);
+                        break;
+
+                    case SDL_BUTTON_RIGHT:
+                        mouse_xy.rdown = true;
+                        CallBack->Msg_RightDown(mouse_xy);
+                        break;
+
+                    case SDL_BUTTON_MIDDLE:
+                        mouse_xy.mdown = true;
+                        CallBack->Msg_MiddleDown(mouse_xy);
+                        break;
                 }
                 break;
             case SDL_MOUSEBUTTONUP:
                 mouse_xy.pos = getGuiScale().screenToView(Position(ev.button.x, ev.button.y));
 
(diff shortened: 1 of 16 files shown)


Spikeone:
@Flamefire I just had a look at the docu (probably for that one the first time):
```
SetColor(color or colorIdx)
Sets the players color by index or directly if its alpha value is not zero. Duplicate color values are possible, so you have to ensure unique colors if you want them!
```

So we can add any color? And can we maybe document the colorIdxs (guess there is an array somewhere). Also what does happen if we use an invalid index (guess just nothing)

libs/s25main/network/GameServer.cpp
```diff
@@ -177,7 +177,11 @@ bool GameServer::Start(const CreateServerInfo& csi, const MapDescription& map, c
         LOG.write("Map %1% has no players!\n") % mapinfo.filepath;
         return false;
     }
-
+    if(playerInfos.size() > MAX_PLAYERS)
```
Spikeone, line 180: Is MAX_PLAYERS == 8 or 7? There are some maps out there which have 8 players already

libs/s25main/network/GameServer.cpp
```diff
@@ -1584,32 +1591,36 @@ void GameServer::SendAsyncLog(const bfs::path& asyncLogFilePath)
     }
 }
 
-void GameServer::CheckAndSetColor(unsigned playerIdx, unsigned newColor)
+void GameServer::CheckAndSetColor(unsigned playerIdx, unsigned newColor, bool ensureUnique)
 {
     RTTR_Assert(playerIdx < playerInfos.size());
-    RTTR_Assert(playerInfos.size() <= PLAYER_COLORS.size()); // Else we may not find a valid color!
+    // If either of those fails we may not find a valid color!
+    static_assert(PLAYER_COLORS.size() >= MAX_PLAYERS, "Not enough player colors defined!");
```
Spikeone, line 1598: I fail to wrap my head around this - shouldn't player clolors always be more than MAX_PLAYERS? 

Expected this rather to be `PLAYER_COLORS.size() < MAX_PLAYERS`

libs/s25main/network/GameServer.cpp
```diff
@@ -1584,32 +1591,36 @@ void GameServer::SendAsyncLog(const bfs::path& asyncLogFilePath)
...
+    RTTR_Assert(playerInfos.size() <= PLAYER_COLORS.size());
```
Spikeone, line 1599: Also this one, shouldn't the same size work fine?

libs/s25main/network/GameServer.cpp
Flamefire: That's exactly what the assertion says: Assert at least as many colors as max players

libs/s25main/network/GameServer.cpp
Flamefire: "assert" (roughly) means "make sure this is true" and `<=` includes "equal"

Flamefire:
> So we can add any color? And can we maybe document the colorIdxs (guess there is an array somewhere). Also what does happen if we use an invalid index (guess just nothing)

Yes any color. I enhanced the documentation to refer to the array and stated that custom colors are ARGB.
An invalid index will cause the function to fail/abort. I didn't document this as it is clear this is an error and we don't document such things in similar places, e.g. for `GetPlayer(idx)`

Flamefire:
@Flow86 *ping

Issue #1785, opened 2025-07-06
Extreme lag on large maps transportation network


Bark-777:
Main-version: 20250506
Sub-version: 0d2a837

Maybe I'm playing the game wrong/too long, so don't know if this is expected or not on large maps.
Creating issue with attached save game in case anyone wants to check it out and see if there is any obvious issues or improvements that can be made.
To reproduce just connect a new road to the existing road-network.

[Savegame](https://github.com/user-attachments/files/21091455/Large-map-with-lag.zip)

I'm playing a very large map 512x512 autogenerated by https://settlers2.net/map-generator/
I play mostly just to explore and settle new lands and to connect everything with roads while listening on podcasts or youtube,
while slowly pushing into computer territory.

Now when controlling about 75% of the map the problem is that when connecting a new roads to the road network the games freezes
for 5-10 seconds. During this time the CPU % goes from 2-3% to 25-30% in the task manager. The computer is a i5 9500t with an
d 48GB RAM.
Placing new flag-post in between on existing road does not have any impact.
Removing roads connected to the network does sometimes freeze the game but not always.

Flamefire:
Such large maps are indeed not well supported as especially large road networks cause a long time for the routing of wares and peoples in that.

There have been improvements to that but further improvements likely require fundamental changes that are unlikely to happen in the foreseeable future.

So I'd suggest to keep that map size at 256x256 or below

Noseey:
After having a more detailed look at this, it appears to be related to the copious amount of mine construction sites set on this map in combination with the large path grid. 

I could see the freeze in two occassions: 
- Placing down a new road (like described above), letting `FindWarehouseForAllJobs` go through the 455 items.
- Having a plank brought into a storage area, letting `FindMaterialForBuildingSites` go through the 455 items.

Those mine sites were not connected to the grid, so all the path calculation was in vain. 

Find attached the cleaned-up version of the map removing the sites. This reduces the freeze to a mere microstutter (if anything).
[Large-map-with-lag_cleanedup-mine-constructions.zip](https://github.com/user-attachments/files/22626836/Large-map-with-lag_cleanedup-mine-constructions.zip)

> Maybe I'm playing the game wrong

Well, some people might call you a madman for that indeed... 😉 However, I can understand this approach using mine sites as a placeholder to be build later. Maybe this can be supported somehow, such that the list these two functions iterate through eliminates sites that were never connected to any roads?

Flamefire:
[...]

Issue #164, opened 2015-08-06
Road network identifiers


MarcusSt:
Every road node (noRoadNode) gets an id of the unique network (as in connected nodes) it is attached to.

**Advantages**

1. Eliminates the worst case for pathfinding, which would visit all nodes, putting them into a tree structure etc. and happens a lot while fighting or for new buildings that are not yet connected. Also the well-known strategy of disconnecting military buildings from the network is problematic.
2. Possibility to implement a "reachable" check to replace pathfinding in many cases. May be used as heuristics together with CalcDistance() when there's no exact walking distance needed.

**Implementation**

1. New road between n1 and n2 gets built.
  - Different id? (happens way less often than worst case pathfinding, mostly only by user interaction)
    - Flood-fill (non-recursive!) part with lower id with higher id.
    - Re-calculate transport (buildings in net 1 may need wares from storehouses in 2 etc.)
2. Road between two nodes n1 and n2 gets destroyed
  - No other path from n1 to n2? (should also not happen that often)
    - flood-fill n2 with new id
3. Seafaring
  - Ids stop at harbors
  - set of pairs (low, high) to look up connections
4. Pathfinding
  - Different id? Check harbors, otherwise no path possible.

Flow86:
sounds good, I think the old s2 has done something similar, since they had unique ids for the "seas"

Flow86:
https://github.com/Return-To-The-Roots/s25client/compare/master...road-network-id

PR #1889, opened 2026-02-11
Pathfinding: reject unconnected buildings / construction sites in specific scenarios

Noseey:
Related Issue:
Closes https://github.com/Return-To-The-Roots/s25client/issues/1785

**Changes Summary**
This will not implement the algorithm mentioned in https://github.com/Return-To-The-Roots/s25client/issues/164, but is an attempt to fix the immediate issues in https://github.com/Return-To-The-Roots/s25client/issues/1785 by rejecting buildings that are not connected to anything / are isolated. This is implemented for the following cases:

- OrderConstructionMaterial
- FindWarehouseForJob
- FindClientForWare
- FindClientForCoin

### DIFF
--- a/libs/s25main/buildings/noBuildingSite.cpp
+++ b/libs/s25main/buildings/noBuildingSite.cpp
@@ -149,7 +149,7 @@ void noBuildingSite::OrderConstructionMaterial()
         return;
 
     // Optimization: return early if building is isolated
-    if(!this->IsConnected())
+    if(!IsConnected())
         return;
 
     // Bretter


Noseey:
~Nevermind, please ignore it for now, I opened it too soon as it seems to not yet correctly work. I will have a further look into it.~

I might have found a less invasive solution that seems promising. Some further testing is ofc needed.

libs/s25main/buildings/noBaseBuilding.cpp
```diff
@@ -207,6 +207,35 @@ MapPoint noBaseBuilding::GetFlagPos() const
     return world->GetNeighbour(pos, Direction::SouthEast);
 }
 
+bool noBaseBuilding::IsConnectedToRoadSystem(const noFlag* flag) const
```
Flamefire: Note on style (at least for new code): If the pointer is always non-NULL make it a reference. You can always get back the pointer if required and it makes it clear for the reader where existing objects must be passed.

libs/s25main/GamePlayer.cpp
```diff
@@ -726,7 +726,14 @@ void GamePlayer::FindCarrierForAllRoads()
...
 {
     for(noBuildingSite* bldSite : buildings.GetBuildingSites())
-        bldSite->OrderConstructionMaterial();
+    {
+        // is it connected?
+        if (!bldSite->IsConnectedToRoadSystem(bldSite->GetFlag()))
+        {
+            continue;
+        }
+        bldSite->OrderConstructionMaterial();
```
Flamefire: Note on style: This could be rather written as positive: 
```suggestion
        if (bldSite->IsConnectedToRoadSystem(bldSite->GetFlag()))
            bldSite->OrderConstructionMaterial();       
```

Less code to read and clearer on intention.

Flamefire:
Thanks for the attempt anyway.  
I added 2 general notes on style in case you pursue this further. :)

Noseey:
> Thanks for the attempt anyway. I added 2 general notes on style in case you pursue this further. :)

I'm not giving up on the probem just yet. 😉 But the current solution still seems a bit too easy for my taste, let me know if I missed something substantial here. At least the large map from the issue seems to run fine now, only very minor stuttering upon posting a new flag.

Flamefire:
2 issues I see:
- It breaks replays as the order and amount of checks, and hence RNG-calls change. See the failing CI
- Pathfinding is not symmetric in general. E.g. for wares the amount of waiting wares at a flag is taken into account which will be wrong if you swap the directions

An idea: I guess "unconnected" buildings/construction sites usually have simply no road from the building flag to anywhere. We could check for that (only required at the target) before running the full pathfinding.
There is also the idea (see TODO) to search only up to the buildings flag to shave off a (little) bit.

In any case this will break replays so it is hard to measure how much benefit this provides. Only empirically by running many longer replays on a large map ignoring the async, which provides at least an indicator.

Noseey:
> An idea: I guess "unconnected" buildings/construction sites usually have simply no road from the building flag to anywhere. We could check for that (only required at the target) before running the full pathfinding.

Yes, this goes in the direction of my first attempt visible in the first two commits. It resolved the issue with the map in the ticket, but I found it too specific as it fails to work as soon as some roads are attached to the construction sites. But it could be a last resort. So you would say that even with that solution of checking just the target flag, the replay compatibility is endangered? Could you point me to an example in the code where randomness is factored into the pathfinding in these scenarios?

Flamefire:
> So you would say that even with that solution of checking just the target flag, the replay compatibility is endangered? Could you point me to an example in the code where randomness is factored into the pathfinding in these scenarios?

Looks like i misremembered that. Only place I found is unrelated to road pathfinding: https://github.com/Return-To-The-Roots/s25client/blob/e6489dca6124c27fd5da763cdb773cb1714d9ca7/libs/s25main/pathfinding/FreePathFinder.cpp#L103-L105

Given that I guess that checking the flag connections should not impact the replays.

> it fails to work as soon as some roads are attached to the construction sites. 

What situation would that be? Usually either the building is fully connected or not connected at all, isn't it?
And swapping start & goal just reverses the issue. Does that really help? And as mentioned for wares it can cause wrong results for the "time" estimates and then choose different warehouses to order wares where the roads are already full.

Noseey:
> And as mentioned for wares it can cause wrong results for the "time" estimates and then choose different warehouses to order wares where the roads are already full.

I think I know what you mean now, it is related to the `AdditonalCosts::Carrier()` in `FindPath` ~introducing a time-dependency to the pathfinding calculation, as the situation upon arrival on high cost road segments is dynamic and thus can be different when swapping start and goal?~ because the `GetPunishmentPoints` depends on the direction?

If that is the case, I guess the current solution is then out of question. There would be an intermediate solution by additionally calculating if a path exists from the goal to nearby warehouses (see the third and forth commit) before, which also resolves the problem with isolated buildings aswell (at least in my tests) but adds additional pathfinding overall, which is not really great.

> What situation would that be? Usually either the building is fully connected or not connected at all, isn't it? 

Yes, I thought it could be somehow integrated into the pathfinding itself without additional checks and have the benefit of covering also "non-connected networks", but given the restrictions it might really be best to just check the building flag itself, as it would still cover most user scenarios. I will update the PR.

~Edit: I think also by only rejecting isolated buildings, there still is the timing dependency to the dynamic situation of high cost road segments, so I guess it could really break replay compatibility over longer periods of time when a high numbers of isolated buildings are involved. I would say your rightfully claimed it as an issue that takes places, no matter how this is fixed.~ Ignore this, since the call is blocking it shouldn't do anything.

Flamefire:
> I think I know what you mean now, it is related to the `AdditonalCosts::Carrier()` in `FindPath` because the `GetPunishmentPoints` depends on the direction?

Right: That is for making sure wares can go around a flag that has currently so many wares in the same direction that the other way is faster even if longer.

> There would be an intermediate solution by additionally calculating if a path exists from the goal to nearby warehouses

We could only do that for the `PathExists` which doesn't care about costs/distances. But if we need costs or directions then there is no alternative.

> > What situation would that be? Usually either the building is fully connected or not connected at all, isn't it?
> 
> Yes, I thought it could be somehow integrated into the pathfinding itself without additional checks and have the benefit of covering also "non-connected networks",

When there is no path from A to B then there is also no path from B to A. So I don't really see a benefit of always swapping the direction. It only helps if the network of B has less nodes than that of A but you don't know that before, do you? So swapping might even make it worse.
That was my question: In which situations is the network from the warehouse smaller than the target buildings network? Is that much more common than the other way round?

> best to just check the building flag itself, as it would still cover most user scenarios. 

My intention is that unconnected buildings are usually e.g. military buildings where no gold or soldiers should go in/out and hence have no connection at all. And this is common enough to do additional checks before running the full pathfinding.

libs/s25main/GamePlayer.cpp
```diff
@@ -821,7 +827,9 @@ void GamePlayer::FindWarehouseForAllJobs()
 {
     for(auto it = jobs_wanted.begin(); it != jobs_wanted.end();)
     {
-        if(FindWarehouseForJob(it->job, *it->workplace))
+        noBaseBuilding* bld = static_cast<noBaseBuilding*>(it->workplace);
```
Flamefire: Not all road nodes are buildings, so you need a check before the cast.
Why not do the check at the lowest place where it makes visually most sense. I.e. in `FindWarehouseForJob` with a comment like: "Optimization: Return early if no path possible"

That covers all current and future uses in a single place

libs/s25main/GamePlayer.cpp
```diff
@@ -726,7 +726,14 @@ void GamePlayer::FindCarrierForAllRoads()
 void GamePlayer::FindMaterialForBuildingSites()
 {
     for(noBuildingSite* bldSite : buildings.GetBuildingSites())
+    {
+        // isolated building site?
+        if(!bldSite->IsConnected())
```
Flamefire: See below: Do the check in `OrderConstructionMaterial` to cover all cases

libs/s25main/buildings/noBaseBuilding.cpp
```diff
@@ -207,6 +207,21 @@ MapPoint noBaseBuilding::GetFlagPos() const
...
 
+bool noBaseBuilding::IsConnected() const
+{
+    const helpers::EnumArray<RoadSegment*, Direction> routes = this->GetFlag()->getRoutes();
+
+    // Check paths in all directions
+    for(const auto dir : helpers::EnumRange<Direction>{})
+    {
+        const auto* route = routes[dir];
+        if(route && dir != Direction::NorthWest)
```
Flamefire: No need for the variable and checking the dir first makes it more likely for the compiler to optimize away
```suggestion
        if(dir != Direction::NorthWest && routes[dir])
```

libs/s25main/buildings/noBaseBuilding.cpp
```diff
@@ -207,6 +207,21 @@ MapPoint noBaseBuilding::GetFlagPos() const
     return world->GetNeighbour(pos, Direction::SouthEast);
 }
 
+bool noBaseBuilding::IsConnected() const
+{
+    const helpers::EnumArray<RoadSegment*, Direction> routes = this->GetFlag()->getRoutes();
+
+    // Check paths in all directions
```
Flamefire:
```suggestion
    // Check paths in all directions except back to building
```

libs/s25main/GamePlayer.cpp
```diff
@@ -1142,6 +1155,10 @@ nobBaseMilitary* GamePlayer::FindClientForCoin(const Ware& ware) const
     {
         unsigned way_points;
 
+        // isolated military bulding?
```
Flamefire: Move to top of loop, or maybe move the `way_points` before `FindPathForWareOnRoads` below

libs/s25main/GamePlayer.cpp
```diff
@@ -1079,6 +1089,9 @@ noBaseBuilding* GamePlayer::FindClientForWare(const Ware& ware)
         if(possibleClient.bld == lastBld)
             continue;
 
+        if(!possibleClient.bld->IsConnected())
```
Flamefire: This can be done before adding a building to `possibleClients` to not waste allocations and time for sorting

libs/s25main/buildings/noBaseBuilding.cpp
```diff
@@ -207,6 +207,21 @@ MapPoint noBaseBuilding::GetFlagPos() const
...
+    for(const auto dir : helpers::EnumRange<Direction>{})
+    {
+        const auto* route = routes[dir];
+        if(route && dir != Direction::NorthWest)
+            return true;
+    }
+
+    return false;
```
Flamefire, line 224: I think we need to check for finished harbors here: They can be reached from sea, not only land so with this they won't get any wares, will they?

Noseey: I did add a check for now, but will need to test this a bit further.

Flamefire: The check needs to be outside the loop and I geuss GetBuildingType is enough as we already know it is a building

Noseey: I moved the check out of the loop and you were right, it should only be checked against `GetBuildingType`. I did a test run, building a harbor on an island without any further connections and then requested to store planks, so it worked no matter if the check was commented out or not. I will leave it in for robustness sake anyways.

libs/s25main/buildings/noBaseBuilding.cpp
```diff
@@ -207,6 +207,24 @@ MapPoint noBaseBuilding::GetFlagPos() const
...
 }
 
+bool noBaseBuilding::IsConnected() const
+{
+    const helpers::EnumArray<RoadSegment*, Direction> routes = this->GetFlag()->getRoutes();
+
+    // Harbors are always considered connected via sea
+    if(this->GetBuildingType() == BuildingType::HarborBuilding)
+        return true;
+
```
Flamefire: No need for the double-pointer access to routes for harbors, and reference to avoid copy of 6 pointers
```suggestion
    // Harbors are always considered connected via sea
    if(GetBuildingType() == BuildingType::HarborBuilding)
        return true;

    const helpers::EnumArray<RoadSegment*, Direction>& routes = GetFlag()->getRoutes();
```

BTW: Any reason you used `this->` explicitely? Looks at least inconsistent with the rest of the code in this file

libs/s25main/GamePlayer.cpp
```diff
@@ -804,6 +804,14 @@ void GamePlayer::ToolOrderProcessed(Tool tool)
 
 bool GamePlayer::FindWarehouseForJob(const Job job, noRoadNode& goal) const
 {
+    // Optimization: return early if building is isolated
+    if(goal.GetType() == NodalObjectType::Building || goal.GetType() == NodalObjectType::Buildingsite)
+    {
+        auto* bld = static_cast<noBaseBuilding*>(&goal);
+        if(!bld->IsConnected())
```
Flamefire: You can cast to a reference directly:

```suggestion
        if(!static_cast<noBaseBuilding&>(goal).IsConnected())
```

libs/s25main/GamePlayer.cpp
```diff
@@ -1026,6 +1034,10 @@ noBaseBuilding* GamePlayer::FindClientForWare(const Ware& ware)
             // Bei Baustellen die Extraliste abfragen
             for(noBuildingSite* bldSite : buildings.GetBuildingSites())
             {
+                // Optimization: return early if building is isolated
```
Flamefire: C&P mistake:

```suggestion
                // Optimization: Ignore if unconnected
```

libs/s25main/GamePlayer.cpp
```diff
@@ -1039,6 +1051,10 @@ noBaseBuilding* GamePlayer::FindClientForWare(const Ware& ware)
             // Für übrige Gebäude
             for(nobUsual* bld : buildings.GetBuildings(bldType))
             {
+                // Optimization: return early if building is isolated
```
Flamefire: Same:
```suggestion
                // Optimization: Ignore if unconnected
```

libs/s25main/GamePlayer.cpp
```diff
@@ -1140,12 +1156,16 @@ nobBaseMilitary* GamePlayer::FindClientForCoin(const Ware& ware) const
     // Militärgebäude durchgehen
     for(nobMilitary* milBld : buildings.GetMilitaryBuildings())
     {
-        unsigned way_points;
+        // Optimization: return early if building is isolated
```
Flamefire:
```suggestion
        // Optimization: Ignore if unconnected
```

libs/s25main/buildings/noBuildingSite.cpp
```diff
@@ -148,6 +148,10 @@ void noBuildingSite::OrderConstructionMaterial()
     if(state == BuildingSiteState::Planing)
         return;
 
+    // Optimization: return early if building is isolated
+    if(!this->IsConnected())
```
Flamefire: One more
```suggestion
    if(!IsConnected())
```

Noseey:
> If you are happy with that mark the PR as ready so we can merge it. Have you checked how much this change gains?

Yes, I did check it against the large example map posted in the issue ticket to see if it resolves the seconds long freezes in the following scenarios:

1. when building material is stored in warehouses
2. when building material is created and dropped infront of the building
3. when a new road is connected

For 1) and 2) it resolved the freezes completely. For 3) it was decreased down to a fraction of a second, due to the still large road network.

PR #1890, opened 2026-02-15
Add support for borderless Windows

### LINKED ISSUE #1512: Lock window size

Flamefire:
Continuation of #1602 to add a 3rd mode:
- Fullscreen
- Windowed
- **Borderless window**

Additionally adds an option to disable window resizing

The window size controls will only be visible when windowed mode is selected.
 Similar for fullscreen resolution.
The window size and resize lock change is directly applied if windowed mode is active

This is how it looks like now

windowsize

Closes #1602 Fixes #1512

In the options screen I added a similar drop-down.

As an extension I would change "Fullscreen resolution" to "Window size" when selecting a windowed option in both settings. There all resolution options and the current window size will be added. Shall I? @Spikeone

### DIFF
--- a/data/RTTR/texte/de/keyboardlayout.txt
+++ b/data/RTTR/texte/de/keyboardlayout.txt
@@ -28,6 +28,7 @@ F1:................... (Spiel laden)
 F2:................... Spiel speichern
 F8:................... Tastaturbelegung anzeigen
 F9:................... ReadMe-Datei anzeigen
+F10:.................. Einstellungen (UI, ...)
 F11:.................. Musik-Spieler
 F12:.................. Kontrollfenster
 
@@ -47,6 +48,6 @@ P:.................... Pause/Fortsetzen (Auch nach GF Sprung)
 1-7:.................. Spielerwechsel
 
 ------------------------------------------------------------------------
-https://www.siedler25.org        Copyright (C) 2005-2022 Settlers Freaks
+https://www.siedler25.org        Copyright (C) 2005-2026 Settlers Freaks
 ------------------------------------------------------------------------
 
--- a/extras/videoDrivers/SDL2/VideoSDL2.cpp
+++ b/extras/videoDrivers/SDL2/VideoSDL2.cpp
@@ -32,11 +32,17 @@
 #    include <gl4esinit.h>
 #endif
 
+namespace {
+
 /// Check that the (SDL) call returns success or print the error
 /// Can be used in conditions: if(CHECK_SDL(SDL_Foo()))
-#define CHECK_SDL(call) ((call) >= 0 || (PrintError(), false))
-
-namespace {
+bool CHECK_SDL(int sdlResult)
+{
+    if(sdlResult >= 0)
+        return true;
+    VideoSDL2::PrintError();
+    return false;
+}
 
 template<typename T>
 struct SDLMemoryDeleter
--- a/extras/videoDrivers/SDL2/VideoSDL2.cpp
+++ b/extras/videoDrivers/SDL2/VideoSDL2.cpp
@@ -540,11 +540,12 @@ unsigned long VideoSDL2::GetTickCount() const
     return SDL_GetTicks();
 }
 
-void VideoSDL2::ListVideoModes(std::vector<VideoMode>& video_modes) const
+std::vector<VideoMode> VideoSDL2::ListVideoModes() const
 {
     int display = SDL_GetWindowDisplayIndex(window);
     if(display < 0)
         display = 0;
+    std::vector<VideoMode> video_modes;
     for(int i = SDL_GetNumDisplayModes(display) - 1; i >= 0; --i)
     {
         SDL_DisplayMode mode;
@@ -557,6 +558,7 @@ void VideoSDL2::ListVideoModes(std::vector<VideoMode>& video_modes) const
                 video_modes.push_back(vm);
         }
     }
+    return video_modes;
 }
 
 OpenGL_Loader_Proc VideoSDL2::GetLoaderFunction() const
--- a/extras/videoDrivers/SDL2/VideoSDL2.cpp
+++ b/extras/videoDrivers/SDL2/VideoSDL2.cpp
@@ -142,7 +142,7 @@ static VideoMode getDesktopSize(VideoMode fallback)
     return fallback;
 }
 
-bool VideoSDL2::CreateScreen(const std::string& title, const VideoMode& size, DisplayMode displayMode)
+bool VideoSDL2::CreateScreen(const std::string& title, const VideoMode size, DisplayMode displayMode)
 {
     if(!initialized)
         return false;
@@ -228,7 +228,7 @@ bool VideoSDL2::CreateScreen(const std::string& title, const VideoMode& size, Di
     return true;
 }
 
-bool VideoSDL2::ResizeScreen(const VideoMode& reqSize, DisplayMode displayMode)
+bool VideoSDL2::ResizeScreen(VideoMode reqSize, DisplayMode displayMode)
 {
     if(!initialized)
         return false;
--- a/extras/videoDrivers/SDL2/VideoSDL2.cpp
+++ b/extras/videoDrivers/SDL2/VideoSDL2.cpp
@@ -1,8 +1,9 @@
-// Copyright (C) 2005 - 2025 Settlers Freaks (sf-team at siedler25.org)
+// Copyright (C) 2005 - 2026 Settlers Freaks (sf-team at siedler25.org)
 //
 // SPDX-License-Identifier: GPL-2.0-or-later
 
 #include "VideoSDL2.h"
+#include "RTTR_Assert.h"
 #include "driver/Interface.h"
 #include "driver/VideoDriverLoaderInterface.h"
 #include "driver/VideoInterface.h"
@@ -31,17 +32,12 @@
 #    include <gl4esinit.h>
 #endif
 
-#define CHECK_SDL(call)                 \
-    ([&]() -> bool {                    \
-        if((call) < 0)                  \
-        {                               \
-            PrintError(SDL_GetError()); \
-            return false;               \
-        }                               \
-        return true;                    \
-    })()
+/// Check that the (SDL) call returns success or print the error
+/// Can be used in conditions: if(CHECK_SDL(SDL_Foo()))
+#define CHECK_SDL(call) ((call) >= 0 || (PrintError(), false))
 
 namespace {
+
 template<typename T>
 struct SDLMemoryDeleter
 {
@@ -102,7 +98,7 @@ bool VideoSDL2::Initialize()
     SDL_SetHint(SDL_HINT_TOUCH_MOUSE_EVENTS, "0");
     if(SDL_InitSubSystem(SDL_INIT_VIDEO) < 0)
     {
-        PrintError(SDL_GetError());
+        PrintError();
         return false;
     }
 
@@ -156,37 +152,53 @@ bool VideoSDL2::CreateScreen(const std::string& title, const VideoMode& size, Di
     CHECK_SDL(SDL_GL_SetAttribute(SDL_GL_DOUBLEBUFFER, 1));
 
     const int wndPos = SDL_WINDOWPOS_CENTERED;
-    const auto fullscreen = (displayMode & DisplayMode::Fullscreen) != DisplayMode::None;
-    const auto resizable = (displayMode & DisplayMode::Resizable) != DisplayMode::None;
+    const auto fullscreen = displayMode == DisplayMode::Fullscreen;
     const auto requestedSize = fullscreen ? FindClosestVideoMode(size) : size;
-    const unsigned commonFlags = SDL_WINDOW_OPENGL | (resizable ? SDL_WINDOW_RESIZABLE : 0);
-    const unsigned fullscreenFlag = (fullscreen ? SDL_WINDOW_FULLSCREEN : 0);
+    const unsigned commonFlags = SDL_WINDOW_OPENGL;
     // TODO: Fix GUI scaling with High DPI support enabled.
     // See https://github.com/Return-To-The-Roots/s25client/issues/1621
     // commonFlags |= SDL_WINDOW_ALLOW_HIGHDPI;
 
-    window = SDL_CreateWindow(title.c_str(), wndPos, wndPos, requestedSize.width, requestedSize.height,
-                              commonFlags | fullscreenFlag);
+    const unsigned windowTypeFlag =
+      fullscreen ? SDL_WINDOW_FULLSCREEN :
+                   (displayMode == DisplayMode::BorderlessWindow ? SDL_WINDOW_BORDERLESS : SDL_WINDOW_RESIZABLE);
 
-    // Fallback to non-fullscreen
-    if(!window && fullscreen)
+    window = SDL_CreateWindow(title.c_str(), wndPos, wndPos, requestedSize.width, requestedSize.height,
+                              commonFlags | windowTypeFlag);
+
+    if(!window)
     {
-        window = SDL_CreateWindow(title.c_str(), wndPos, wndPos, requestedSize.width, requestedSize.height,
-                                  commonFlags | SDL_WINDOW_RESIZABLE);
+        PrintError();
+        // Fallback to borderless fullscreen
+        if(fullscreen)
+        {
+            SDL_DisplayMode dskSize;
+            if(!CHECK_SDL(SDL_GetDesktopDisplayMode(0, &dskSize)))
+            {
+                dskSize.w = size.width;
+                dskSize.h = size.height;
+            }
+            window = SDL_CreateWindow(title.c_str(), wndPos, wndPos, dskSize.w, dskSize.h,
+                                      commonFlags | SDL_WINDOW_BORDERLESS);
+        }
+        // No borderless -> Resizable
+        if(!window)
+        {
+            window = SDL_CreateWindow(title.c_str(), wndPos, wndPos, size.width, size.height,
+                                      commonFlags | SDL_WINDOW_RESIZABLE);
+        }
     }
 
     if(!window)
     {
-        PrintError(SDL_GetError());
+        PrintError();
         return false;
     }
 
-    const auto flags = SDL_GetWindowFlags(window);
-    SetFullscreenFlag((flags & SDL_WINDOW_FULLSCREEN) != 0);
-    SetResizableFlag((flags & SDL_WINDOW_RESIZABLE) != 0);
+    UpdateCurrentDisplayMode();
     UpdateCurrentSizes();
 
-    if(!IsFullscreen())
+    if(displayMode_ != DisplayMode::Fullscreen)
         MoveWindowToCenter();
 
     SDL_Surface* iconSurf =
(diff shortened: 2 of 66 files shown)


Spikeone:
Sounds good to me - just some question, I never thought about borderless window (Although this is probably what I'd like to have anyway) but wouldn't this be somewhat confusing for people since most would think its fullscreen windowed (as most modern game name it) but we don't have such option at all?

Flamefire:
My concern is rather: Without a title bar you cannot move it and it will always be centered. Not sure this suits everyone but you can't have that anyway.

> fullscreen windowed (as most modern game name it) but we don't have such option at all?

The closest would be to set the size accordingly.
Additionally, if fullscreen isn't possible I implemented a fallback to a fullscreen, borderless window

That being said we can add a 4th option "Fullscreen windowed" (or maybe "windowed fullscreen" or "fullscreen window"?) to automatically set the size to the desktop size. IIRC SDL has even a flag for that, something called `FULLSCREEN_DESKTOP` or so.

Spikeone:
Oh, so the borderless window is the one that is not resizable? I honestly thought we'd just ignore the resize event and add an option to ignore that event

Flamefire:
Having no border makes it clear that the window cannot be resized as there is nothing to do that.
For the normal window we actually set a flag to make it resizable. The original PR #1602 was using a bitset so we can represent that in our settings too. Downside of course are some combinations like "resizable fullscreen" make no sense. I'd ignore that though.

I see it rather as an UI issue: Now we have 3 clear display modes. How/Where would be disable window resizing? Or shall we have a "fixed size window" option there?

Spikeone:
Well, what I was thinking about in general:
- Fullscreen
- Fullscreen Borderless Windowed
- Windowed

Those are the 3 modes I'd expect. Then I'd just love to have an additional option to:
- Enable / Disable Resize window

Which would only apply to windowed mode. I still want to move the window around, but I don't want to resize it by accident. That was my initial intention when I asked @falbrechtskirchinger for this :)

Flamefire:
Hm, back to a bitset for the option then:
- Fullscreen
- Windowed
- Resizeable

`Fullscreen | Windowed ` would then be a borderless window and `Resizeable` only specifically handled for non-Fullscreen, ignoring it for fullscreen option

Spikeone:
> Hm, back to a bitset for the option then:
> 
>     * Fullscreen
> 
>     * Windowed
> 
>     * Resizeable
> 
> 
> `Fullscreen | Windowed ` would then be a borderless window and `Resizeable` only specifically handled for non-Fullscreen, ignoring it for fullscreen option

If I don't missunderstand this - yes. (Honestly, as long as I can play windowed, move the window and prevent it from resizing by accident, I'm happy)

Flamefire:
This should now be ready. I also included a couple of bug fixes related to resizing that I noticed during the tests.
Doing all the work twice for WinAPI and SDL2 was annoying...

@Spikeone Added your extra button. The window has still a border but it can't be dragged.
"Borderless window" is now what you know from other games: A window in full desktop resolution

Spikeone:
> This should now be ready. I also included a couple of bug fixes related to resizing that I noticed during the tests. Doing all the work twice for WinAPI and SDL2 was annoying...
> 
> @Spikeone Added your extra button. The window has still a border but it can't be dragged. "Borderless window" is now what you know from other games: A window in full desktop resolution

Sounds great - just to make sure - not dragged, does mean can't resize and not can't move it right?

Flamefire:
> Sounds great - just to make sure - not dragged, does mean can't resize and not can't move it right?

Right. When the mouse is at the border the cursor won't change to the resize cursor and you can't drag the border to resize it. You can move the window dragging the title bar as before

data/RTTR/texte/de/keyboardlayout.txt
Spikeone: `https://www.siedler25.org        Copyright (C) 2005-2026 Settlers Freaks` sollte hier wohl auch geändert werden

Spikeone (dismissed):
Den restlichen Code kann ich nicht überschauen 😅

extras/videoDrivers/SDL2/VideoSDL2.cpp
```diff
@@ -102,7 +104,7 @@ bool VideoSDL2::Initialize()
     SDL_SetHint(SDL_HINT_TOUCH_MOUSE_EVENTS, "0");
     if(SDL_InitSubSystem(SDL_INIT_VIDEO) < 0)
```
Flow86: Can't we use the macro here too?

extras/videoDrivers/WinAPI/WinAPI.cpp
```diff
@@ -173,97 +105,111 @@ bool VideoWinAPI::CreateScreen(const std::string& title, const VideoMode& newSiz
...
 
-    ShowWindow(screen, SW_HIDE);
-
-    VideoMode windowSize = fullscreen ? FindClosestVideoMode(newSize) : newSize;
+    VideoMode windowSize = (displayMode == DisplayMode::Fullscreen) ? FindClosestVideoMode(newSize) : newSize;
     // Try to switch full screen first
-    if(isFullscreen_ && !fullscreen)
+    if(displayMode_ == DisplayMode::Fullscreen && displayMode != DisplayMode::Fullscreen)
     {
+        ShowWindow(screen, SW_HIDE);
```
Flow86, line 120: Isn't this a bit dangerous if we fail to switch it's not restored back

extras/videoDrivers/WinAPI/WinAPI.cpp
```diff
@@ -173,97 +105,111 @@ bool VideoWinAPI::CreateScreen(const std::string& title, const VideoMode& newSiz
...
         if(ChangeDisplaySettings(nullptr, 0) != DISP_CHANGE_SUCCESSFUL)
             return false;
-    } else if(isFullscreen_ || fullscreen)
+    } else if(displayMode == DisplayMode::Fullscreen)
     {
+        ShowWindow(screen, SW_HIDE);
```
Flow86, line 128: Same

extras/videoDrivers/WinAPI/WinAPI.cpp
```diff
@@ -173,97 +105,111 @@ bool VideoWinAPI::CreateScreen(const std::string& title, const VideoMode& newSiz
...
     }
-    wRect.right = wRect.left + size.width;
-    wRect.bottom = wRect.top + size.height;
-
-    std::pair<DWORD, DWORD> style = GetStyleFlags(fullscreen);
-    // Calculate real right/bottom based on the window style
-    AdjustWindowRectEx(&wRect, style.first, false, style.second);
 
-    return wRect;
+    BOOST_UNREACHABLE_RETURN(std::make_pair(RECT{0, 0, requestedSize.width, requestedSize.height}, requestedSize));
```
Flow86, line 215: Hu, why unreachable?

libs/s25main/controls/ctrlList.cpp
```diff
@@ -162,42 +157,26 @@ void ctrlList::Draw_()
...
 }
 
-/**
- *  Verändert einen String
- */
-void ctrlList::SetString(const std::string& text, const unsigned id)
+void ctrlList::SetItemText(const unsigned id, const std::string& text)
 {
-    lines[id] = text;
+    lines.at(id) = text;
```
Flow86, line 170: So we do want it to exist?

libs/s25main/desktops/dskGameInterface.cpp
```diff
@@ -300,8 +301,10 @@ void dskGameInterface::Resize(const Extent& newSize)
     cbb.buildBorder(newSize, borders);
 
     // move buttons
-    DrawPoint barPos((newSize.x - LOADER.GetImageN("resource", 29)->getWidth()) / 2 + 44,
-                     newSize.y - LOADER.GetImageN("resource", 29)->getHeight() + 4);
+    // Get real renderer size as newSize may get capped but we want to keep the manually drawn borders intact
+    const Extent realNewSize = VIDEODRIVER.GetRenderSize();
+    DrawPoint barPos = DrawPoint((realNewSize.x - LOADER.GetImageN("resource", 29)->getWidth()) / 2 + 44,
+                                 realNewSize.y - LOADER.GetImageN("resource", 29)->getHeight() + 4);
```
Flow86: I hate that we read the resource twice, even though its not really a problem,  but its ugly.

Can we somehow also add some proper "anti-magic" constants?

Maybe we should  think of "layout"-intermediate classes which would make (re)scaling even simpler (not for this MR of course ...)

libs/s25main/desktops/dskGameInterface.cpp
```diff
@@ -826,6 +829,7 @@ bool dskGameInterface::Msg_KeyDown(const KeyEvent& ke)
         case KeyType::F9:
             WINDOWMANAGER.ToggleWindow(std::make_unique<iwTextfile>("readme.txt", _("Readme!")));
             return true;
+        case KeyType::F10: WINDOWMANAGER.ToggleWindow(std::make_unique<iwSettings>()); return true;
```
Flow86, line 840: weird formatting. since its doing two things, make it seperate lines like the rest

libs/s25main/desktops/dskMainMenu.cpp
```diff
@@ -73,9 +73,9 @@ void dskMainMenu::Msg_MsgBoxResult(const unsigned msgbox_id, const MsgboxResult
     if(msgbox_id == 100)
     {
         if(mbr == MsgboxResult::Yes)
```
Flow86: we should also add some constants for "0, 1 and 2" so its clear what they man

libs/s25main/drivers/VideoDriverWrapper.cpp
```diff
@@ -456,25 +428,34 @@ void VideoDriverWrapper::SetMousePos(const Position& newPos)
...
+                      [](const auto& m) { return m.width < MinWindowSize.width && m.height < MinWindowSize.height; });
+    return videoModes;
+}
 
-    videodriver->ListVideoModes(video_modes);
+std::vector<VideoMode> VideoDriverWrapper::GetDefaultWindowSizes() const
+{
+    std::vector<VideoMode> defaultWindowSizes = {
+      VideoMode(800, 600),  VideoMode(1024, 768), VideoMode(1152, 648), VideoMode(1280, 720),  VideoMode(1280, 800),
+      VideoMode(1366, 768), VideoMode(1440, 810), VideoMode(1600, 900), VideoMode(1680, 1050), VideoMode(1920, 1080)};
```
Flow86: can you add some comments so its formatted one per line, its really hard to read (or later edit)

libs/s25main/drivers/VideoDriverWrapper.cpp
```diff
@@ -456,25 +428,34 @@ void VideoDriverWrapper::SetMousePos(const Position& newPos)
...
+    std::vector<VideoMode> windowSizes = ListVideoModes();
+    windowSizes.insert(windowSizes.end(), defaultWindowSizes.begin(), defaultWindowSizes.end());
+    helpers::makeUnique(windowSizes);
```
Flow86: can't we simply use a  `std::set` instead?

libs/s25main/Settings.cpp
```diff
@@ -239,10 +244,18 @@ void Settings::Load()
...
-        video.fullscreen = iniVideo->getBoolValue("fullscreen");
+        const auto displayMode = iniVideo->getValue("displayMode", -1);
+        if(displayMode >= 0 && static_cast<unsigned>(displayMode) <= helpers::MaxEnumValue_v<DisplayMode::Type>)
+            video.displayMode = DisplayMode(displayMode);
+        else
+        {
+            video.displayMode =
+              iniVideo->getValue("fullscreen", false) ? DisplayMode::Fullscreen : DisplayMode::Windowed;
+        }
+        video.displayMode.resizeable = !iniVideo->getValue("lock_window_size", false);
```
Flow86, line 259: can't we do some explicit "migration path" before loading instead?

extras/videoDrivers/WinAPI/WinAPI.cpp
Flamefire: Never has been an issue, but sure.

extras/videoDrivers/WinAPI/WinAPI.cpp
Flamefire: done

extras/videoDrivers/WinAPI/WinAPI.cpp
Flamefire: switch-case above is exhaustive, we verify that via warnings

libs/s25main/controls/ctrlList.cpp
Flamefire: Yes. We could do this in all setters & getters but this is the one I touched and most have an assert already, can't remember why I use `at`  only here, maybe I ran into this, it didn't have the assert or copied from somewhere else

libs/s25main/desktops/dskGameInterface.cpp
Flamefire: They are used in multiple places, yes. Extracted those constants

libs/s25main/desktops/dskGameInterface.cpp
Flamefire: line breaks here are controlled by clang-format via line length which we have to not needing to discuss formatting.
I don't think it matters here anyway as it is only a `return` and there is no other logic or `break`

libs/s25main/drivers/VideoDriverWrapper.cpp
Flamefire: VideoMode has no meaningfull less-than, so set doesn't work.

libs/s25main/drivers/VideoDriverWrapper.cpp
Flamefire: done

libs/s25main/Settings.cpp
Flamefire: What do you mean? You are looking at that: Convert the prior values losslessly to the new ones

extras/videoDrivers/WinAPI/WinAPI.cpp
Flow86: if its exchaustive, we should not need a macro here

libs/s25main/desktops/dskGameInterface.cpp
Flow86: Ah ok. still hate it how it looks, but I never looked into the formatter config anymore to adapt it.

libs/s25main/drivers/VideoDriverWrapper.cpp
Flow86: hmm, we could add one, right?

Something like
`a.width < b.width || (a.width == b.width && a.height < b.height)`

since 800x600 < 1920x600

or what AI suggested:

```
lexicographic by area → width → height
Useful when two resolutions have the same pixel count but different shapes.

Example:
1280×720 (16:9) vs 1024×900 (roughly 1.14:1)
Both are ~921k pixels, but one is wider, one is taller.

Comparison rule:
- Compare total pixels
- If equal, compare width
- If still equal, compare height
```

libs/s25main/Settings.cpp
Flow86: didnt we had some kind of "version" number in the config?
I mean I don't want to have that kind of conversion code shattered around the loading place (where you don't really remember which one is "old" or when was it introduced.

I mean a separate conversion method run before the main load function (or as first option)

e.g

```
if version <= 5) { // we replaced fullscreen by displaymode
   write('displaymode') = read('fullscreen') ? DM::FullScreen : DM::Windowed  ;
   erase('fullscreen')
}
```

and you only read displaymode at the real loading place.

extras/videoDrivers/WinAPI/WinAPI.cpp
Flamefire: We need it because technically `mode.type = DisplayMode(1234567)` is possible.
The macro basically tells the compiler we didn't do that.

libs/s25main/desktops/dskMainMenu.cpp
Flamefire: Done

libs/s25main/Settings.cpp
Flamefire: We don't have a version for the settings, just:
- revision changed -> show warning
- migrate on-the-fly where possibly
- any hard error -> load defaults


We don't need anything more complicated than that: We do the loading only here at a single place. Adding such conversion code would decouple it so much you'd need comments explaining what you'll do with the values.
`video.displayMode = iniVideo->getValue("fullscreen", false) ? DisplayMode::Fullscreen : DisplayMode::Windowed;` is quite clear.

> where you don't really remember which one is "old" or when was it introduced.

Why would you care about the "when" when reading the code?
About what is "old": Both branches set displayMode. One directly other through a differently named option.  I'd say it is easy enough to know what the new/correct one is. But we could add a comment that this is compatibility code if you want.

libs/s25main/drivers/VideoDriverWrapper.cpp
Flamefire: Haven't found it while searching earlier: We had `cmpVideoModes` doing what you suggested.
I also like the other approach better although it could be confusing in the selection list: 
We now have:
1440x480
1128x634
1152x648
1024x768
1440x576
1280x720
1152x870

So using the one we had before:
1024x768
1128x634
1152x648
1152x870
1280x720
1440x480
1440x576

libs/s25main/Settings.cpp
Flow86: yes a comment would be nice, so its 1000% clear which one is the old part

libs/s25main/drivers/VideoDriverWrapper.cpp
Flow86: the 2nd one is better, the first one  is simply weird.

Normally you go for width first, right?

extras/videoDrivers/WinAPI/WinAPI.cpp
Flow86: so maybe simply do a RTTR_ASSERT and a `throw "unreachable"`?

libs/s25main/Settings.cpp
```diff
@@ -239,10 +247,18 @@ void Settings::Load()
         video.windowedSize.height = iniVideo->getIntValue("windowed_height");
         video.fullscreenSize.width = iniVideo->getIntValue("fullscreen_width");
         video.fullscreenSize.height = iniVideo->getIntValue("fullscreen_height");
-        video.fullscreen = iniVideo->getBoolValue("fullscreen");
+        const auto displayMode = iniVideo->getValue("displayMode", -1);
+        if(displayMode >= 0 && static_cast<unsigned>(displayMode) <= helpers::MaxEnumValue_v<DisplayMode::Type>)
+            video.displayMode = DisplayMode(displayMode);
+        else
+        {
+            video.displayMode =
```
Flamefire, line 256:
```suggestion
            // Compatibility with prior settings format
            video.displayMode =
```

extras/videoDrivers/WinAPI/WinAPI.cpp
Flamefire: No, we really mean "unreachable" here. The Boost macro expands to the corresponding compiler intrinsic for that allowing for better codegen. The "return" part is only required when the compiler doesn't support such thing and hence would require a return for a non-void function even though it isn't reached

libs/s25main/drivers/VideoDriverWrapper.cpp
Flamefire: I'd say so, yes.

Flow86 (approved):

PR #1893, opened 2026-02-16
Allow dragging ingameWindows with middle mouse button

Farmer-Markus:
Just a small idea to let users drag ingame windows with the middle mouse button.
Very useful when playing on a laptop

### DIFF
--- a/extras/videoDrivers/SDL2/VideoSDL2.cpp
+++ b/extras/videoDrivers/SDL2/VideoSDL2.cpp
@@ -392,39 +392,43 @@ bool VideoSDL2::MessageLoop()
             case SDL_MOUSEBUTTONDOWN:
                 mouse_xy.pos = getGuiScale().screenToView(Position(ev.button.x, ev.button.y));
 
-                if(/*!mouse_xy.ldown && */ ev.button.button == SDL_BUTTON_LEFT)
+                switch(ev.button.button)
                 {
-                    mouse_xy.ldown = true;
-                    CallBack->Msg_LeftDown(mouse_xy);
-                }
-                if(/*!mouse_xy.rdown &&*/ ev.button.button == SDL_BUTTON_RIGHT)
-                {
-                    mouse_xy.rdown = true;
-                    CallBack->Msg_RightDown(mouse_xy);
-                }
-                if(ev.button.button == SDL_BUTTON_MIDDLE)
-                {
-                    mouse_xy.mdown = true;
-                    CallBack->Msg_MiddleDown(mouse_xy);
+                    case SDL_BUTTON_LEFT:
+                        mouse_xy.ldown = true;
+                        CallBack->Msg_LeftDown(mouse_xy);
+                        break;
+
+                    case SDL_BUTTON_RIGHT:
+                        mouse_xy.rdown = true;
+                        CallBack->Msg_RightDown(mouse_xy);
+                        break;
+
+                    case SDL_BUTTON_MIDDLE:
+                        mouse_xy.mdown = true;
+                        CallBack->Msg_MiddleDown(mouse_xy);
+                        break;
                 }
                 break;
             case SDL_MOUSEBUTTONUP:
                 mouse_xy.pos = getGuiScale().screenToView(Position(ev.button.x, ev.button.y));
 
-                if(/*mouse_xy.ldown &&*/ ev.button.button == SDL_BUTTON_LEFT)
+                switch(ev.button.button)
                 {
-                    mouse_xy.ldown = false;
-                    CallBack->Msg_LeftUp(mouse_xy);
-                }
-                if(/*mouse_xy.rdown &&*/ ev.button.button == SDL_BUTTON_RIGHT)
-                {
-                    mouse_xy.rdown = false;
-                    CallBack->Msg_RightUp(mouse_xy);
-                }
-                if(ev.button.button == SDL_BUTTON_MIDDLE)
-                {
-                    mouse_xy.mdown = false;
-                    CallBack->Msg_MiddleUp(mouse_xy);
+                    case SDL_BUTTON_LEFT:
+                        mouse_xy.ldown = false;
+                        CallBack->Msg_LeftUp(mouse_xy);
+                        break;
+
+                    case SDL_BUTTON_RIGHT:
+                        mouse_xy.rdown = false;
+                        CallBack->Msg_RightUp(mouse_xy);
+                        break;
+
+                    case SDL_BUTTON_MIDDLE:
+                        mouse_xy.mdown = false;
+                        CallBack->Msg_MiddleUp(mouse_xy);
+                        break;
                 }
                 break;
             case SDL_MOUSEWHEEL:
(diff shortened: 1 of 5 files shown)


Flow86:
Useful extension, especially since you do not need to find the (relatively) small title bar to move the window around.

Approved

extras/videoDrivers/WinAPI/WinAPI.h
```diff
@@ -84,8 +84,9 @@ class VideoWinAPI final : public VideoDriver
     static LRESULT CALLBACK WindowProc(HWND window, UINT msg, WPARAM wParam, LPARAM lParam);
 
 private:
-    bool mouse_l;    /// Status der Linken Maustaste.
-    bool mouse_r;    /// Status der Rechten Maustaste.
+    bool mouse_l;    /// State of left mouse button
```
Flamefire: I think those are left overs, SDL doesn't have them. Or is there a reason to have 
`mouse_m = mouse_xy.mdown = false;`? If not I'd just remove `mouse_l/mouse_r` instead of adding another copied value.

libs/s25main/ingameWindows/IngameWindow.cpp
```diff
@@ -263,6 +263,20 @@ bool IngameWindow::Msg_LeftUp(const MouseCoords& mc)
     return false;
 }
 
+bool IngameWindow::Msg_MiddleDown(const MouseCoords& mc)
+{
+    isMoving = true;
```
Flamefire: This is copied from clicking the title bar. So I'd put it into a method, e.g. `StartDragging(pos)`, especially with the handling of window snapping that might change and need to be kept in sync

libs/s25main/ingameWindows/IngameWindow.cpp
```diff
@@ -263,6 +263,20 @@ bool IngameWindow::Msg_LeftUp(const MouseCoords& mc)
...
+    snapOffset_ = SnapOffset::all(0);
+    lastMousePos = mc.pos;
+    return true;
+}
+
+bool IngameWindow::Msg_MiddleUp(const MouseCoords&)
+{
+    isMoving = false;
+    return false;
```
Flamefire: IIRC the return value was meant to be "handled the event", so I guess returning true here, at least if `isMoving` was true

Spikeone:
I still think this should be an option you can disable - at least that way I hope that people who, for whatever reason don't want this, aren't bothered and maybe it's a bit more prominent to find and actually know that it is possible at all.

Flamefire:
IMO it is so unintrusive we don't need an option: Zoom is the only other use for the mousewheel we have so I doubt people will unintentionally use this
Not sure if we have documentation for the zoom function, but if we could add this there. If not we should likely add both

extras/videoDrivers/WinAPI/WinAPI.h
Farmer-Markus: > Or is there a reason to have
> `mouse_m = mouse_xy.mdown = false;`?

There isn't. I've just done this because of the other mouse_l/mouse_r values.
I'll remove them

libs/s25main/ingameWindows/IngameWindow.cpp
Farmer-Markus: Wouldn't the same issue be in Msg_LeftUp?

```c++
bool IngameWindow::Msg_LeftUp(const MouseCoords& mc)
{
    isMoving = false;

    [..]

    return false;
}
```

Flamefire: Yes. I'm not sure where and how the return value is used here, but I'd guess we should return true when we handled it by stopping dragging

extras/videoDrivers/WinAPI/WinAPI.cpp
```diff
@@ -665,25 +665,27 @@ LRESULT CALLBACK VideoWinAPI::WindowProc(HWND window, UINT msg, WPARAM wParam, L
...
             break;
         case WM_RBUTTONUP:
-            pVideoWinAPI->mouse_r = false;
             pVideoWinAPI->mouse_xy.rdown = false;
             pVideoWinAPI->CallBack->Msg_RightUp(pVideoWinAPI->mouse_xy);
             break;
+        case WM_MBUTTONDOWN:
+            pVideoWinAPI->mouse_xy.mdown = true;
+            pVideoWinAPI->CallBack->Msg_MiddleDown(pVideoWinAPI->mouse_xy);
+        case WM_MBUTTONUP:
```
Flamefire, line 687: Missing break

```suggestion
        break;
        case WM_MBUTTONUP:
```

extras/videoDrivers/SDL2/VideoSDL2.cpp
```diff
@@ -402,6 +402,11 @@ bool VideoSDL2::MessageLoop()
                     mouse_xy.rdown = true;
                     CallBack->Msg_RightDown(mouse_xy);
                 }
+                if(ev.button.button == SDL_BUTTON_MIDDLE)
```
Flamefire: Maybe make a switch-case (here and below) on `ev.button.button` as we now have 3 ifs where only 1 can be activated which isn't fully clear.

libs/s25main/ingameWindows/IngameWindow.cpp
```diff
@@ -200,9 +201,7 @@ bool IngameWindow::Msg_LeftDown(const MouseCoords& mc)
             if(btn == IwButton::Title)
             {
                 // start moving
```
Flamefire:
```suggestion
```

libs/s25main/ingameWindows/IngameWindow.cpp
```diff
@@ -212,7 +211,8 @@ bool IngameWindow::Msg_LeftDown(const MouseCoords& mc)
 
 bool IngameWindow::Msg_LeftUp(const MouseCoords& mc)
 {
-    isMoving = false;
+    // If we handled the msg event
+    bool handled = StopDragging();
```
Flamefire: If we were dragging no other button could be activated, can it? Not sure, but don't see how.
In that case:

```suggestion
    if(StopDragging())
        return true;
```

libs/s25main/ingameWindows/IngameWindow.cpp
```diff
@@ -493,3 +504,20 @@ Rect IngameWindow::GetButtonBounds(IwButton btn) const
...
+}
+
+bool IngameWindow::StopDragging()
+{
+    if(isMoving)
+    {
+        isMoving = false;
+        return true;
+    }
+    return false;
```
Flamefire, line 521: IIRC this is what `std::exchange` does:

```suggestion
    return std::exchange(isMoving, false);
```

I'm undecided though if this is clearer, but wanted to mention it. See what you like better

Farmer-Markus: > I'm undecided though if this is clearer, but wanted to mention it. See what you like better

I would keep it. You instantly know what the function does without knowing all the std functions.
And there doesn't seem to be a  big performance difference either

libs/s25main/ingameWindows/IngameWindow.cpp
Farmer-Markus: I think this would stop closing the window with dblclick(on top bar) on touch devices right?
The fingerdown could trigger the movement and the fingerup will stop the moving and prevent the window close code from being executed

Flamefire: Good point! It only matters when the first tap is on the title bar and window pinning is disabled.
Maybe add this in the comment. And:

```suggestion
    const bool handled = StopDragging();
```

Flamefire (approved):
LGTM, thanks a lot!

Issue #1784, opened 2025-07-06
Ships get stuck on landing spot behind their first landing spot, then building harbor on first spot the game crashes


Bark-777:
Main-version: 20250626
Sub-version: f7859b8

Expedition lands on harbor building site, once arrived I get option to continue expedition to another spot, which turns out is the spot just behind the first harbor site, and here all ships being sent get stuck and cannot continue or cancel expedition.
Send once again another expedition and now builds harbor on first spot, and game crashes when harbor is finished.

Harbor causing crash is upper right part of the map.
[auto_154x192 (Auto-Save).zip](https://github.com/user-attachments/files/21091348/auto_154x192.Auto-Save.zip)

Noseey:
I had a look at this and know why it crashes: the "stranded" ships - Nipia, Otacilia, Maelia, Secundinia, Iuventia - next to the about to be constructed harbor will get a trigger to `StartDrivingToHarborPlace` but fail to find a way to this newly builded harbor since they try to cross from land terrain to land terrain. They then fail the `if(!routeFound)` check, even if they are close by. All of the affected ships have their `lost` attribute set and are in status `ExpeditionDriving`.

~A fix for the crash might be to not call an `RTTR_Assert(false);` in such a case but to flag / keep the ship as lost, so it remains stuck but the game continues.~ Probably related to the debug build, need to check it further.

However, what I was not able to reproduce is how the ships got stuck there in the first place. I tried destroying both harbors and started expeditions to the nothern spot, then southern and vice versa.. but they wouldn't get stuck. Even deleting the stranded ships didn't make subsequent expeditions stuck. I know it has been a while, but do you happen to recall more details or have a save before they got stuck?

Bark-777:
I didn't have any previous saves, but was able to reproduce getting a ship stuck.
Once the expedition lands, click the expeditions continue-arrow, the ship will then back out and re-enter the same spot.
Do this 2 or 3 more times and it will sail into one of the "stuck spots" behind the first landing spot.

Attached savegame with ship Antia mounting expedition to the problematic landing spot.
[antia-getting-stuck.zip](https://github.com/user-attachments/files/25328542/antia-getting-stuck.zip)

Flamefire:
> Attached savegame with ship Antia mounting expedition to the problematic landing spot.
> [antia-getting-stuck.zip](https://github.com/user-attachments/files/25328542/antia-getting-stuck.zip)

[...]

PR #1895, opened 2026-02-17
Ships: Fix crash when doing expedition between adjacent harbor places

Noseey:
Related Issue:
Closes https://github.com/Return-To-The-Roots/s25client/issues/1784

**Changes Summary**
This fixes a crash when expeditions are done between harbor places that are directly adjacent to another (e. g. on autogenerated maps), so the ship doesn't need to move at all for reaching them and thus its `route_` is empty. For more details, see related issue.

Note:
@Flamefire: Like discussed, could you add also a test for this if possible?

### DIFF
--- a/libs/s25main/world/GameWorldBase.cpp
+++ b/libs/s25main/world/GameWorldBase.cpp
@@ -339,25 +339,23 @@ HarborId GameWorldBase::GetHarborInDir(const MapPoint pt, const HarborId originH
 {
     RTTR_Assert(originHarborId);
 
-    // Herausfinden, in welcher Richtung sich dieser Punkt vom Ausgangspunkt unterscheidet
-    helpers::OptionalEnum<Direction> coastal_point_dir;
     const MapPoint hbPt = GetHarborPoint(originHarborId);
 
-    for(const auto dir : helpers::EnumRange<Direction>{})
+    // Find sea of harbor point
+    SeaId seaId;
+    for(const auto d : helpers::EnumRange<Direction>{})
     {
-        if(GetNeighbour(hbPt, dir) == pt)
+        if(GetNeighbour(hbPt, d) == pt)
         {
-            coastal_point_dir = dir;
+            seaId = GetSeaId(originHarborId, d);
             break;
         }
     }
+    RTTR_Assert(seaId);
 
-    RTTR_Assert(coastal_point_dir);
-
-    SeaId seaId = GetSeaId(originHarborId, *coastal_point_dir);
     const std::vector<HarborPos::Neighbor>& neighbors = GetHarborNeighbors(originHarborId, dir);
 
-    for(auto neighbor : neighbors)
+    for(const auto& neighbor : neighbors)
     {
         if(IsHarborAtSea(neighbor.id, seaId) && isHarborOk(neighbor.id))
             return neighbor.id;


libs/s25main/nodeObjs/noShip.cpp
```diff
@@ -517,7 +517,12 @@ noShip::Result noShip::DriveToHarbourPlace()
...
     }
 
-    StartDriving(route_[curRouteIdx++]);
+    // Already arrived? (e. g. adjacent harbor)
+    if(curRouteIdx == route_.size())
+        return Result::GoalReached;
+    else
+        StartDriving(route_[curRouteIdx++]);
+
     return Result::Driving;
```
Flamefire: This check exists already at the top of the function, which got me confused.

I think the only situation would be a route recalculation by one of the 2 `FindShipPathToHarbor`
So either put a `route_.empty()` check there after `curRouteIdx = 0;` or something like:


```suggestion
    // FindShipPathToHarbor above found we arrived already?
    if(route_.empty())
        return Result::GoalReached;

    RTTR_Assert(curRouteIdx < route_.size();
    StartDriving(route_[curRouteIdx++]);
    return Result::Driving;
```

(the `else` was also misleading as it can be omitted after `return` and looked like `return Result::Driving;` would always be executed)

libs/s25main/nodeObjs/noShip.cpp
```diff
@@ -557,7 +562,14 @@ void noShip::ContinueExpedition(const ShipDirection dir)
...
+
+    // Already arrived? (e. g. adjacent harbor)
+    if(curRouteIdx == route_.size())
+    {
+        state = State::ExpeditionWaiting;
+        return;
+    } else
+        state = State::ExpeditionDriving;
 
     StartDriving(route_[curRouteIdx++]);
```
Flamefire: Misleading indentation, and check can be against "empty"
And we don't need to set the state (see check at top)

```suggestion
    if(!route_.empty())
    {
        state = State::ExpeditionDriving;
        StartDriving(route_[curRouteIdx++]);
    }
```

Flamefire: I also think we need to handle the "arrived" as an event to e.g. notify the player. Maybe simply:

```suggestion
    state = State::ExpeditionDriving;
    HandleState_ExpeditionDriving();
```

libs/s25main/pathfinding/FreePathFinderImpl.h
```diff
@@ -171,6 +171,10 @@ bool FreePathFinder::CheckRoute(const MapPoint start, const std::vector<Directio
 {
     RTTR_Assert(pos < route.size());
 
+    // Empty routes are not considered passable
+    if(route.empty())
```
Flamefire: Not sure where else this is used but I'd say an empty route is valid: It just doesn't change the current point, i.e. `*dest = pos`

Or would that cause issues anywhere?

libs/s25main/nodeObjs/noShip.cpp
Noseey: > This check exists already at the top of the function, which got me confused.

Yes, in the end it is just another "are we there yet?" check , just after the route recalculations. I found this to be more in line with the rest of the code and since `curRouteIdx` is set to 0 in these cases, it would also work for empty routes in my opinion. But I can change it back to an check against it being empty.

About the else, I agree, that was not necessary.

Flamefire: My issue as I was reading the function was: This check already exists, why do it again? And what is it about `adjacent harbor`? `DriveToHarbourPlace` isn't only called from a harbor so without the background knowledge of this issue it might be confusing.

Only after checking all uses of `route_` I remembered that `FindShipPathToHarbor` changes it's parameter and wanted to keep that knowledge.

Maybe:
```suggestion
    // Already arrived after possibly recalculating route (FindShipPathToHarbor)
    if(curRouteIdx == route_.size())
        return Result::GoalReached;

    RTTR_Assert(curRouteIdx < route_.size();
    StartDriving(route_[curRouteIdx++]);
    return Result::Driving;
```

The other idea was to have the check local to where it matters: Above
```diff
-curRouteIdx = 0;
+curRouteIdx = 0;
+if(route_.empty())
+    return Result::GoalReached;
```

And then have only `RTTR_Assert(curRouteIdx < route_.size();` here

Noseey: I probably haven't taken into account the history of the example map when adding this additional check. For the already stuck ships (due to them driving regardless the access to routes_ was OOB), the `curRouteIdx` was 1. 

With a proper check in place in `noShip::ContinueExpedition`, this should no longer be possible. I would say, just to be save, we change the check on the top of this function to >= `if(curRouteIdx >= route_.size())` and remove the additional check at the bottom. What do you say?

libs/s25main/pathfinding/FreePathFinderImpl.h
Noseey: Apart from the `DriveToHarbourPlace`, it seems to be also used in two functions for checking the Trade Route, being `TradeRoute::GetNextDir` (which checks the path route size before calling the function) and `TradePathCache::pathExists`, which I need check further.

libs/s25main/nodeObjs/noShip.cpp
Flamefire: > just to be save

When `curRouteIdx > route_.size()` we already had an OOB and things are likely damaged already. So that situation cannot happen in a state we can recover from anyways.
It most I'd add an `RTTR_Assert(curRouteIdx < route_size())` when `==` failed, i.e. not returned

So while I appreciate being careful, accounting for a situation that already caused fatal issues is a bit to much for me.

libs/s25main/pathfinding/FreePathFinderImpl.h
Flamefire: Just checked loosely and found: https://github.com/Return-To-The-Roots/s25client/blob/6e122731ffbc701bd9dd49c013e216cefa959652/libs/s25main/world/TradeRoute.cpp#L38-L39

So when this returns true we expect `route[pos]` to be valid, which would not be the case when `route` is empty.
However `RTTR_Assert(pos < route.size());` already assumes we have a valid index, as a precondition. And indeed we cannot have `pos >= size` and hence `empty` is never true, see the other discussion about handling situations _after_ OOB access.
Hence we can remove this and ensure our assumption/assertion holds.

libs/s25main/nodeObjs/noShip.cpp
Flamefire: I think you misunderstood: Simply replace the original `StartDriving` by `HandleState_ExpeditionDriving` without checking the route-index/size
This is already handled by `HandleState_ExpeditionDriving` so we can avoid the work here which is also very clean: Advance the state and call the associated state handler instead of duplicating some of its code

Noseey: Ah, I see, so that the check is not needed here either.

Flamefire:
Amazing how after all the changes to the changes and discussion the fix is a single line :smile:    
Thanks for the productive dialog!

I'll try to write a test in the next couple days to cover this. Did you try the original save/replay?

Noseey:
> I'll try to write a test in the next couple days to cover this. Did you try the original save/replay?

Sounds good! 
Yes, I did check it with the provided savegames in both debug (commented out asserts) and release build:

- https://github.com/user-attachments/files/21091348/auto_154x192.Auto-Save.zip does still crash, as we have removed the check for empty vectors in `FreePathFinder::CheckRoute` and there are still the stuck ships with positions != 0 on the map and flagged as `lost` from the previous OOB access. But like established above, with the current fix this should not lead to stuck ships in the first place.
- https://github.com/user-attachments/files/25328542/antia-getting-stuck.zip does work now and it is no longer possible to get the ship "Antia" stuck when continuing the expedition from Harbor place 8 to 9.

Flamefire:
> [github.com/user-attachments/files/25328542/antia-getting-stuck.zip](https://github.com/user-attachments/files/25328542/antia-getting-stuck.zip) does work now and it is no longer possible to get the ship "Antia" stuck when continuing the expedition from Harbor place 8 to 9.

I tried that and it fails for me:
When the ship arrives and you order it to the next harbor then `FindShipPathToHarbor` will select the same position as the ship currently is on as the (next) coast point and pass it to `FindShipPath` which doesn't expect equal start and end.

So I'd say we need to check for `pos == coast` in `FindShipPathToHarbor` and handle it as-if it is a length zero route.
What did happen for you? What do you think?

BTW: I'm done with the reproducer which is how I noticed it. Doing a larger change though, as I'm also getting rid of integers for harbor and sea IDs which made it quite confusing to create the test and navigate the affected code paths.

Noseey:
> I tried that and it fails for me: When the ship arrives and you order it to the next harbor then `FindShipPathToHarbor` will select the same position as the ship currently is on as the (next) coast point and pass it to `FindShipPath` which doesn't expect equal start and end.

Sorry, I failed to mention that I did the check with the Release build and assertions disabled. For me, it returned true by going into this path in `FreePathFinder::FindPath`:

```C
        // Ziel schon erreicht?
        if(&best == &destNode)
        {
            // Ziel erreicht!
            // Jeweils die einzelnen Angaben zurückgeben, falls gewünscht (Pointer übergeben)
            if(length)
                *length = best.curDistance;
            if(route)
                route->resize(best.curDistance);

            FreePathNode* curNode = &best;
            // Route rekonstruieren und ggf. die erste Richtung speichern, falls gewünscht
            for(unsigned z = best.curDistance; z > 0; --z)
            {
                if(route)
                    (*route)[z - 1] = curNode->dir;
                if(firstDir && z == 1)
                    *firstDir = curNode->dir;
                curNode = curNode->prev;
            }

            RTTR_Assert(curNode == &startNode);

            // Fertig, es wurde ein Pfad gefunden
            return true;
        }
```

But ofc, if we want to write tests for it that should also work with assertions enabled, which I forgot about. As a blasphemous question: removing the asserts in `FindPath` with the argumentation, that an empty route can exist in practice, is not an option? Probably not, but I just want to mention it. 😉 

> So I'd say we need to check for `pos == coast` in `FindShipPathToHarbor` and handle it as-if it is a length zero route. What did happen for you? What do you think?

I think this can work, something among those lines here, right?:

```C
    // already arrived?
    if(start == GetCoastalPoint(harborId, seaId))
    {
        route->clear();
        return true;
    }
```

Flamefire:
> As a blasphemous question: removing the asserts in `FindPath` with the argumentation, that an empty route can exist in practice, is not an option? Probably not, but I just want to mention it. 😉

I think the idea was that pathfinding was meant to find a path and start going along it to avoid issues like this: when start==goal it might need to be handled differently, e.g. searching for the next path or signalling arrival. E.g. a person arriving in a building needs to disappear or turn around.

In this case it happens to work because we don't need `firstDir`. But if the caller had passed that what would we return there?
Of course it could always be `route.front()` but then: Why construct the full route (including memory allocation) when we only need the first dir?

> I think this can work, something among those lines here, right?

Yes, just make sure all out variables are set accordingly (if requested) and document it in the header of this function and all callers with similar semantic due to calling it (e.g. `FindHarborForUnloading`). Something like "When `pos` is the coastal point of the target harbor the function returns true with an empty `route?  and `length` zero" (or a formulation you'd like better)
From a quick check it should all work because so far only `noShip` uses it and their `HandleStateX` check if they are at the end of their route which is true for an empty route. Hence the state will treat it as arrival as discussed.

Noseey:
I added the check. While having a look at the callers,  `OrderShip`, `GetJobForShip`, `FindHarborForUnloading`, `StartDrivingToHarborPlace` already have this kind of check in their methods, so no impact there. Should I still add that comment in the header file for these function?

So the check is helpful for callers `ContinueExpedition` (cause for this issue) and potentially for `DriveToHarbourPlace` (I guess, if a shiproute has changed such that we already arrived?).

libs/s25main/Pathfinding.cpp
```diff
@@ -67,7 +67,20 @@ bool GameWorldBase::FindShipPathToHarbor(const MapPoint start, unsigned harborId
     }
     // Add a few fields reserve
     maxDistance += 6;
-    return FindShipPath(start, GetCoastalPoint(harborId, seaId), maxDistance, route, length);
+
+    MapPoint harborPoint = GetCoastalPoint(harborId, seaId);
```
Flamefire: Please use `coastalPoint` as the name. We use "harborPoint" in other places for the position of the harbor, so this might be misleading. And `const`

Flamefire:
> While having a look at the callers, `OrderShip`, `GetJobForShip`, `FindHarborForUnloading`, `StartDrivingToHarborPlace` already have this kind of check in their methods, so no impact there. Should I still add that comment in the header file for these function?

I guess it makes sense to remove these checks then. E.g. `FindHarborForUnloading` can then skip the check and do `if(FindShipPathToHarbor) return true`

The others could be simplified to check the length returned as I don't really like that they get the coastal point again (duplicating logic from inside the called method) and is inconsistent with other uses, e.g. the one we care about here (`noShip::DriveToHarbourPlace`). This might hurt us later when we don't know why we sometimes do check and sometimes not.

`StartDrivingToHarborPlace` is special in that it uses some optimizations to limit pathfinding costs. Not sure if those are even worth it as they only apply when the ship starts, but maybe...

> So the check is helpful for callers `ContinueExpedition` (cause for this issue) and potentially for `DriveToHarbourPlace` (I guess, if a shiproute has changed such that we already arrived?).

Must have looked at another method. So only `FindHarborForUnloading` calls this and might return an empty vector which needs documenting.

libs/s25main/Pathfinding.cpp
```diff
@@ -67,7 +67,20 @@ bool GameWorldBase::FindShipPathToHarbor(const MapPoint start, unsigned harborId
     }
     // Add a few fields reserve
     maxDistance += 6;
-    return FindShipPath(start, GetCoastalPoint(harborId, seaId), maxDistance, route, length);
+
+    const MapPoint coastalPoint = GetCoastalPoint(harborId, seaId);
```
Flamefire, line 58: The check needs to be at the top to avoid the expensive nested loop

Noseey:
> `StartDrivingToHarborPlace` is special in that it uses some optimizations to limit pathfinding costs. Not sure if those are even worth it as they only apply when the ship starts, but maybe...

I didn't feel comfortable touching this one, but I changed the check for the others.

libs/s25main/GamePlayer.cpp
```diff
@@ -1948,17 +1948,15 @@ bool GamePlayer::OrderShip(nobHarborBuilding& hb)
...
-        if(ship.GetPos() == dest)
-        {
-            hb.ShipArrived(ship);
-            return (true);
-        }
-
         if(world.FindShipPathToHarbor(ship.GetPos(), hb.GetHarborPosID(), ship.GetSeaID(), &route, &distance))
         {
+            // ship already there?
+            if(distance == 0 && route.empty())
```
Flamefire: Do we need to check both? I would guess `distance == 0` is sufficient, possible with an `RTTR_Assert(route.empty())` if at all

libs/s25main/GamePlayer.cpp
```diff
@@ -1948,17 +1948,15 @@ bool GamePlayer::OrderShip(nobHarborBuilding& hb)
...
+            {
+                hb.ShipArrived(ship);
+                return (true);
```
Flamefire: This style is long dead ;-) I removed it already in my local branch for the test and will remove it here again when merging the branches. SO just in case you find more of those: `return true` is the way to go

libs/s25main/GamePlayer.h
```diff
@@ -256,6 +256,7 @@ class GamePlayer : public GamePlayerInfo
     unsigned GetShipsToHarbor(const nobHarborBuilding& hb) const;
     /// Sucht einen Hafen in der Nähe, wo dieses Schiff seine Waren abladen kann
     /// gibt true zurück, falls erfolgreich
+    /// If starting point equals the coastal point of target harbor, also return true.
```
Flamefire:
```suggestion
    /// If starting point equals the coastal point of target harbor, return true and an empty `route`.
```

libs/s25main/GamePlayer.cpp
Noseey: Yes, I saw the other PR now and since it is addressed there, I left it in here.

libs/s25main/GamePlayer.cpp
Noseey: It is removed now.

Flamefire:
I merged #1901 to this branch and added the reproducer.
Tested the reproducer before the merge (on the other branch) and it did fail as expected.

So from my side this is ready to merge, thank you!

Noseey:
> So from my side this is ready to merge, thank you!

Thanks aswell for adding the test-case, specifically like the Ascii sketch of the problem. 😉

PR #1898, opened 2026-02-27
Fix save crash for skinner

ottml:
Reproduce:

- Build a skinner in an area without animals
- Save game
- Crash happens because animal is not initialized


ottml:
Code coverage is no reachable at the moment therefore tests are failing

Flamefire:
True: https://status.coveralls.io/

I can't circumvent the required check so we got to wait till they are back

PR #1899, opened 2026-02-28
GUI: limit the scaling of ctrlTextButtons

### LINKED ISSUE #1896: [Suggestion] Limit stretching of Menu buttons
Since I like to play in 21:9 and fullscreen, I found that the menu buttons are stretched a bit too much looking slightly unpleasant and confusing, here an example:

Image

Would it be acceptable to limit them a bit, maybe like indicated by the red border? If it is, I could have a look into it.

Noseey:
Related Issue:
Closes https://github.com/Return-To-The-Roots/s25client/issues/1896

**Changes Summary**
This change introduces a limiting factor that slows down the X- and Y size scaling of the GUI control `ctrlTextButtons` in order to make it visually more consistent on ultrawidescreen-resolutions.

Note:
I keep the PR in draft for the moment, since I want to do some thorough testing after any review changes are applied. So far, it seems to behave correctly also when resizing, but there are some unit test failures I need to look into aswell.

### DIFF
--- a/libs/s25main/RescaleWindowProp.h
+++ b/libs/s25main/RescaleWindowProp.h
@@ -55,7 +55,7 @@ inline T_Pt RescaleWindowProp::operator()(const T_Pt& oldValue, const Extent& li
     if(limfactors.x > 0 && limfactors.x < 11 && limfactors.y > 0 && limfactors.y < 11)
     {
         T_Pt diff(oldSize - Extent(800, 600));
-        T_Pt limUnscaleValue(oldValue.x * 800 / (oldSize.x - (diff.x * limfactors.x / 10)), 
+        T_Pt limUnscaleValue(oldValue.x * 800 / (oldSize.x - (diff.x * limfactors.x / 10)),
                              oldValue.y * 600 / (oldSize.y - (diff.y * limfactors.y / 10)));
         realValue = limUnscaleValue;
     } else
--- a/libs/s25main/RescaleWindowProp.h
+++ b/libs/s25main/RescaleWindowProp.h
@@ -20,17 +20,25 @@ struct ScaleWindowPropUp
 /// Rescales a window's properties (size or positions)
 struct RescaleWindowProp
 {
-    Extent oldSize, newSize;
+    Extent oldSize, newSize, origValue;
     RescaleWindowProp(Extent oldSize, Extent newSize) : oldSize(oldSize), newSize(newSize) {}
     /// Scale the point or size from beeing relative to the oldSize to relative to the newSize
     template<typename T_Pt>
-    T_Pt operator()(const T_Pt& oldValue) const;
+    T_Pt operator()(const T_Pt& oldValue, const T_Pt& origValue) const;
 };
 
 template<typename T_Pt>
 inline T_Pt ScaleWindowPropUp::scale(const T_Pt& value, const Extent& sizeToScale)
 {
-    return T_Pt(value * sizeToScale / Extent(800, 600));
+    T_Pt scaledValue(value * sizeToScale / Extent(800, 600));
+    if(std::is_same_v<T_Pt, Extent>)
+    {
+        if(scaledValue.x > value.x << 1)
+            scaledValue.x = value.x << 1;
+        if(scaledValue.y > value.y << 1)
+            scaledValue.y = value.y << 1;
+    }
+    return scaledValue;
 }
 
 template<typename T_Pt>
@@ -40,9 +48,14 @@ inline T_Pt ScaleWindowPropUp::operator()(const T_Pt& value) const
 }
 
 template<typename T_Pt>
-inline T_Pt RescaleWindowProp::operator()(const T_Pt& oldValue) const
+inline T_Pt RescaleWindowProp::operator()(const T_Pt& oldValue, const T_Pt& origValue) const
 {
     T_Pt realValue(oldValue.x * 800 / oldSize.x, oldValue.y * 600 / oldSize.y);
+    if(std::is_same_v<T_Pt, Extent> && (origValue.x != 0 && origValue.y != 0))
+    {
+        realValue.x = origValue.x;
+        realValue.y = origValue.y;
+    }
     // Check for rounding errors
     T_Pt checkValue = ScaleWindowPropUp::scale(realValue, oldSize);
     if(checkValue.x < oldValue.x)
--- a/libs/s25main/RescaleWindowProp.h
+++ b/libs/s25main/RescaleWindowProp.h
@@ -12,7 +12,7 @@ struct ScaleWindowPropUp
     ScaleWindowPropUp(const Extent& size) : size(size) {}
 
     template<typename T_Pt>
-    static T_Pt scale(const T_Pt& value, const Extent& size, const unsigned limfactor);
+    static T_Pt scale(const T_Pt& value, const Extent& size);
     template<typename T_Pt>
     T_Pt operator()(const T_Pt& value) const;
 };
@@ -25,28 +25,18 @@ struct RescaleWindowProp
     /// Scale the point or size from beeing relative to the oldSize to relative to the newSize
     template<typename T_Pt>
     T_Pt operator()(const T_Pt& oldValue) const;
-    template<typename T_Pt>
-    T_Pt operator()(const T_Pt& trueValue, const unsigned limfactor) const;
 };
 
 template<typename T_Pt>
-inline T_Pt ScaleWindowPropUp::scale(const T_Pt& value, const Extent& sizeToScale, const unsigned limfactor)
+inline T_Pt ScaleWindowPropUp::scale(const T_Pt& value, const Extent& sizeToScale)
 {
-    T_Pt scaledValue(value * sizeToScale / Extent(800, 600));
-    if(limfactor > 0)
-    {
-        if(scaledValue.x > value.x << limfactor)
-            scaledValue.x = value.x << limfactor;
-        if(scaledValue.y > value.y << limfactor)
-            scaledValue.y = value.y << limfactor; 
-    }
-    return scaledValue;
+    return T_Pt(value * sizeToScale / Extent(800, 600));
 }
 
 template<typename T_Pt>
 inline T_Pt ScaleWindowPropUp::operator()(const T_Pt& value) const
 {
-    return scale(value, size, 0);
+    return scale(value, size);
 }
 
 template<typename T_Pt>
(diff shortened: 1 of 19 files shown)


libs/s25main/Window.h
```diff
@@ -304,6 +310,7 @@ class Window
     unsigned id_;          /// ID des Fensters.
     DrawPoint pos_;        /// Position des Fensters.
     Extent size_;          /// Höhe des Fensters.
+    Extent limit_factors_; /// X and Y scaling limiting factors from 1 (unscaled) - 10 (fully scaled)
```
Flamefire: Can you please use a new typedef for this. `Extent` is very confusing for it given that it isn't an extent at all. Even just `using LimitFactors = Point<unsigned>` here in this file would work

Also: Why 1-10? What does it mean to be "fully scaled"? And what would zero be which you check below?
You can put that description on the typedef.
And our style (except for legacy code to be updated) is `CamelCase`, i.e. `limitFactors_`

And just a heads-up: I updated the docstrings of this class in https://github.com/Return-To-The-Roots/s25client/pull/1890/commits/8c1bf5fef90e6a50b11367b263f3cd2789f720cc of #1890 which will cause (simply) conflicts

libs/s25main/Window.h
```diff
@@ -286,9 +290,11 @@ class Window
     friend constexpr auto maxEnumValue(ButtonState) { return ButtonState::Pressed; }
     using ControlMap = std::map<unsigned, Window*>;
 
-    /// scales X- und Y values to fit the screen
+    /// scales X- and Y values to fit the screen considering a limiting factor
     template<class T_Pt>
-    static T_Pt Scale(const T_Pt& pt);
+    static T_Pt Scale(const T_Pt& pt, const Extent& limfactors);
+    /// scales X- and Y values of position and size considering the current limiting factor for size
```
Flamefire: Does it consider the factor for the position? How would it know?

libs/s25main/Window.h
Noseey: I introduced LimitFactors but added it to Point, so its all in one place. The renaming of the member variable has been done and I tried to update the descriptions in the comments (hopefully for the better :grin:)

libs/s25main/controls/ctrlTextButton.cpp
```diff
@@ -11,9 +11,11 @@ static constexpr unsigned contentOffset = 2;
 
 ctrlTextButton::ctrlTextButton(Window* parent, unsigned id, const DrawPoint& pos, const Extent& size,
                                const TextureColor tc, const std::string& text, const glFont* font,
-                               const std::string& tooltip)
+                               const std::string& tooltip, const LimitFactors& limitFactors)
```
Flow86: what about making it `std::optional`?

libs/s25main/RescaleWindowProp.h
```diff
@@ -24,30 +24,50 @@ struct RescaleWindowProp
...
-    T_Pt operator()(const T_Pt& oldValue) const;
+    T_Pt operator()(const T_Pt& oldValue, const LimitFactors& limfactors) const;
 };
 
 template<typename T_Pt>
-inline T_Pt ScaleWindowPropUp::scale(const T_Pt& value, const Extent& sizeToScale)
+inline T_Pt ScaleWindowPropUp::scale(const T_Pt& value, const Extent& sizeToScale, const LimitFactors& limfactors)
 {
-    return T_Pt(value * sizeToScale / Extent(800, 600));
+    if(limfactors.x > 0 && limfactors.x < 11 && limfactors.y > 0 && limfactors.y < 11)
```
Flow86: we (should) try to avoid magic numbers, make the `11` a constant please,

also if using a `std::optional`, the test would be much simpler (is set & assertion for >0 only)

libs/s25main/Window.h
Flamefire: Thanks.
To me it's still a bit hard to understand. I thought it would be a limit, i.e. cap the change at some point. But now it looks like some kind of dampening: The scaling gets scaled by this "factor" divided by 10
E.g:

> Value 1: scales 1/10 of the difference between screensize and Base Resolution

It isn't fully clear to me what this means, i.e. what will happen to the size and or position and why.

Not sure the all-in-one place is better in this case: Could that be used outside the context of Window?
Maybe put it next to `ScaleWindowPropUp`  (not sure what the "Up" in the name meant, maybe we should drop it) where the scaling could then be described in full in a single file
And the 800x600 there should eventually become a constant like `constexpr Extent REFERENCE_RESOLUTION` or so but that's rather long and I can't think of a better name

Noseey:
I will work on the remaining findings throughout next week. There are two overall things I would also open up for discussion:

1. Currently, the `ctrlTextButtons` of `dskOptions` are also affected, but I find that it doesn't blend in well, so I would like to remove the limiting from the desktop `dskOptions` by introducing a variable `limit_` and thus make the limting dependend on both desktop and the control type. Meaning, the options will scale freely like before.
2. I would like to transform the `ScaleByFactor()` into simply `Scale()` and call that first thing in the constructors of the affected controls based upon the `scale_` and `limit_` value from the parent window. If I thought it through correctly, this would allow to remove most of the ScaleIf calls in Window.cpp and would make the limiting possible for controls that call other controls in their constructor, such as `ctrlTable`. It would unify the solution and allow to expand upon later, if needed.

Flamefire:
> * Currently, the `ctrlTextButtons` of `dskOptions` are also affected, but I find that it doesn't blend in well, so I would like to remove the limiting from the desktop `dskOptions` 

That's what I meant: It is so highly specific to a situation that doing it in general is hard.
Can you edit the issue you reported and provide a list where you'd apply such limits? Maybe we can then find what those situations have in common.

>  I would like to remove the limiting from the desktop `dskOptions` by introducing a variable `limit_` and thus make the limting dependend on both desktop and the control type. Meaning, the options will scale freely like before.

`limit_` is a bool or an `optional<LimitFactors>`? (Should be clear by the name anyway)
So limiting will only be done:
- For `ctrTextButtons` added via `AddTextButton`, not directly created
- Only if the parent window has a limit set
- That limit will only be set for the start screen

Did I got that right?
Maybe we can pass limits when creating the control, or (to avoid interface bloat) afterwards: `SetLimitFactors` does a rescale if there is a change, so the startup desktop can do `AddTextButtons(...)->SetLimitFactors({42,1337})`
But I'd still like to define those properly, e.g. I'd expect something like `(0..1]` as the range to "dampen" the extending or `>=1` to limit the extension by a factor, i.e `newValue = value + min(sizeIncrease, value * factor)` I.e. a `factor=2` means it can grow to 2 times it reference size, but that needs storing the reference size or just the maxSize which could be calculated from the reference size (`unscaledCurSize * factors`)

> * I would like to transform the `ScaleByFactor()` into simply `Scale()` and call that first thing in the constructors of the affected controls based upon the `scale_` and `limit_` value from the parent window. If I thought it through correctly, this would allow to remove most of the ScaleIf calls in Window.cpp and would make the limiting possible for controls that call other controls in their constructor, such as `ctrlTable`. It would unify the solution and allow to expand upon later, if needed.

I would actually like something like that: Handle scaling in a single place and not at every constructor invocation.
We would then only need to check **each** constructor of desktops, windows and controls if they use the position and or size passed in the constructor parameter and potentially change that to use the member instead as that is what those ctors expect right now. Forgetting one might not be immediately detected, so that should be done very thoroughly.
This also makes it easier to store/transform the original/reference size where required.

Noseey:
> Can you edit the issue you reported and provide a list where you'd apply such limits? Maybe we can then find what those situations have in common.
> Did I got that right?

I updated the issue ticket on how I currently would see the limiting needs to be applied. 

> Maybe we can pass limits when creating the control, or (to avoid interface bloat) afterwards: `SetLimitFactors` does a rescale if there is a change, so the startup desktop can do `AddTextButtons(...)->SetLimitFactors({42,1337})` But I'd still like to define those properly, e.g. I'd expect something like `(0..1]` as the range to "dampen" the extending or `>=1` to limit the extension by a factor, i.e `newValue = value + min(sizeIncrease, value * factor)` I.e. a `factor=2` means it can grow to 2 times it reference size, but that needs storing the reference size or just the maxSize which could be calculated from the reference size (`unscaledCurSize * factors`)

The current solution is really like a dampening, I did that so no storing of the reference size is required to `rescale`. If we can agree, that a storing of the reference size (or maxSize alternatively) is okay, that would give us some more options. But how would it solve cases where ctrls create sub-controls (see constructor of `ctrlTable`) and need the scaled & limited size already in their constructor? This might not be required for this specific change, as it might only focus around `ctrlTextButton` anyways, but I would like for the solution to be flexible enough to support this in the future.

> I would actually like something like that: Handle scaling in a single place and not at every constructor invocation. We would then only need to check **each** constructor of desktops, windows and controls if they use the position and or size passed in the constructor parameter and potentially change that to use the member instead as that is what those ctors expect right now. Forgetting one might not be immediately detected, so that should be done very thoroughly. This also makes it easier to store/transform the original/reference size where required.

I'm not sure what you mean by "single place" in that regard, would it be possible to get a small example?

Flamefire:
> But how would it solve cases where ctrls create sub-controls (see constructor of `ctrlTable`) and need the scaled & limited size already in their constructor?

`SetLimitFactors` would trigger a resize and e.g. ctrlTable does handle that and adjust stuff accordingly. And the constructor there uses at least `GetSize` already instead of the `size` parameter (see below)



> I'm not sure what you mean by "single place" in that regard, would it be possible to get a small example?

Currently we have:
- `new Foo(Scale(pos), Scale(size))`
- `new Bar(Scale(pos), Scale(size))`
- ...

All classes inherit from `Window`. So if the `Window` constructor was like this: `Window::Window(pos, size): pos_(Scale(pos)), size_(Scale(size))` then we can remove `Scale` from the Foo and Bar above. And the constructors of Foo and Bar can use `GetPosition` and `GetSize` to get the already scaled values.
We just need to check if there are any places where unscaled controls get added and this change would then scale them

Noseey:
So I did a bit of an update to reflect my current status:
- I largely got rid of the ScaleIfs within Window.cpp, scaling at a single point in the window constructor instead (in one spot).
- There is a bool to indicate which desktops are allowed to apply limits
- Scaling and Rescaling structs have been merged into one
- A default argument for the constructor is used for limits (not sure if an optional:: is still required)?

It is still quite rough around the edges, so don't bother with a too detailed review yet. Also, the factor/dampening itself is unchanged so far, might need to check that a bit further to make it a real factor or change it in another way. General thoughts on the approach are welcome! 😉

Noseey:
I did some further updates on it to make the scaling calculation and naming (hopefully) a bit more understandable. It is now supposed to be a scaling percentage instead of factor, reaching from 0 (no scaling) - 100% (scale fully proportional).

Noseey:
I will close it for the time being, as I don't currently have a better idea how to solve it without doing major rework on how the UI scaling / flow works.

PR #1900, opened 2026-03-02
Fix failures in UI tests

Flamefire:
`BOOST_FIXTURE_TEST_CASE` resets the fixture of the test suite so we have to add that manually when required.

In the double click tests a double click could be registered too early as the last left click time and position are initially zeroes. Do a fake click to initialize them to something that isn't relevant to the tests.


Flamefire:
@Flow86 Whats wrong with Jenkins?

Flow86:
jenkins is up again. the usual - a plugin broke

PR #1908, opened 2026-03-28
change planer algorithms to match the original

libv:
S2 planes the area to match the altitude of the flag, not the altitude of the building tile. This means both a difference in what altitude is being planed to, and also changes the calculation for deciding to plane or not. S2 changes the height of the building tile to match the flag when the planer has done his round and returns to the central building tile.

### DIFF
--- a/libs/s25main/buildings/noBuildingSite.cpp
+++ b/libs/s25main/buildings/noBuildingSite.cpp
@@ -29,7 +29,10 @@ noBuildingSite::noBuildingSite(const BuildingType type, const MapPoint pos, cons
        || GetSize() == BuildingQuality::Harbor)
     {
         // Höhe auf dem Punkt, wo die Baustelle steht
-        int altitude = world->GetNode(pos).altitude;
+        int altitude = world->GetNode(this->GetFlagPos()).altitude;
+
+        if(altitude - world->GetNode(pos).altitude != 0)
+	    state = BuildingSiteState::Planing;
 
         for(const auto dir : helpers::EnumRange<Direction>{})
         {
@@ -376,6 +379,8 @@ void noBuildingSite::PlaningFinished()
     /// Normale Baustelle
     state = BuildingSiteState::Building;
     planer = nullptr;
+
+    world->ChangeAltitude(pos, world->GetNode(this->GetFlagPos()).altitude);
 
     // Wir hätten gerne einen Bauarbeiter...
     world->GetPlayer(player).AddJobWanted(Job::Builder, this);
--- a/libs/s25main/figures/nofPlaner.cpp
+++ b/libs/s25main/figures/nofPlaner.cpp
@@ -158,7 +158,7 @@ void nofPlaner::HandleDerivedEvent(const unsigned id)
     {
         // Planieren (falls Baustelle noch existiert)
         if(building_site)
-            world->ChangeAltitude(pos, world->GetNode(building_site->GetPos()).altitude);
+            world->ChangeAltitude(pos, world->GetNode(building_site->GetFlagPos()).altitude);
         /// Sounds abmelden
         world->GetSoundMgr().stopSounds(*this);
 


Flamefire:
Just to be clear on the algorithm:
- planer goes to the building
- then walks around to each neighbour point except the flag and planes it
- while doing that, instead of going to flag (i.e. after finishing the cycle) goes back to building
- building levels too and planer goes home

I.e. it still doesn't work on the actual building site but that still gets modified.
Correct?

@Spikeone agreed?

An issue I see with this: The change is fundamental enough that existing replays will likely stop working.

libs/s25main/buildings/noBuildingSite.cpp
```diff
@@ -29,7 +29,10 @@ noBuildingSite::noBuildingSite(const BuildingType type, const MapPoint pos, cons
        || GetSize() == BuildingQuality::Harbor)
     {
         // Höhe auf dem Punkt, wo die Baustelle steht
-        int altitude = world->GetNode(pos).altitude;
+        int altitude = world->GetNode(this->GetFlagPos()).altitude;
```
Flamefire, line 32: To make this easier to understand:
```suggestion
        const auto flagAltitude = world->GetNode(this->GetFlagPos()).altitude;
```

libs/s25main/buildings/noBuildingSite.cpp
```diff
@@ -29,7 +29,10 @@ noBuildingSite::noBuildingSite(const BuildingType type, const MapPoint pos, cons
...
+
+        if(altitude - world->GetNode(pos).altitude != 0)
+	    state = BuildingSiteState::Planing;
```
Flamefire, line 35: Enhance check (same below) and indent
```suggestion
        if(altitude != world->GetNode(pos).altitude)
            state = BuildingSiteState::Planing;
```

And put the loop into an `else`-branch

libv:
> Just to be clear on the algorithm:
> 
>     * planer goes to the building
>     * then walks around to each neighbour point except the flag and planes it
>     * while doing that, instead of going to flag (i.e. after finishing the cycle) goes back to building
>     * building levels too and planer goes home

Yes. Same algorithm as rttr before, just to the level of the flag, and the building tile just pops into place as the planer finishes his round, heads to the building tile, then walks off through the flag tile.

> I.e. it still doesn't work on the actual building site but that still gets modified. Correct?

If you mean that this does not change the altitude of the actual building site, then no. This changes the altitude of the building site to also match the altitude of the flag. The whole site gets planed to the altitude of the flag.

> @Spikeone agreed?
> 
> An issue I see with this: The change is fundamental enough that existing replays will likely stop working.

Quite likely, but this is how S2 works, and i do not think i have ever known it differently. Fire up dosbox, and an original s2 game, plane a building site where the flag is a different altitude from the main buiding tile :)

libs/s25main/buildings/noBuildingSite.cpp
libv: I actually removed the tabbed indents by hand, as my editor is set to kernel style. And the very awkward calculation matches the rest of the code. So when reviewing this patch, please see the context, i have done extra work to match the rest of the codebase.

Flamefire: I now see there is a tab left, yes. You can use clang-format to auto-format. If it is found by configure there is even a build target for that. CI will tell you too once run

This wasn't criticism of your work and matching this (strange) style is certainly right in general.

My suggestion here is to fix/use the `!=` style for the new code and adapt the old code below to match this. I see no reason for keeping the old style which raises questions e.g. if it could run into UB. Doesn't here, but `!=` avoids the question and makes the intent clearer.

Flamefire:
> If you mean that this does not change the altitude of the actual building site, then no. This changes the altitude of the building site to also match the altitude of the flag. The whole site gets planed to the altitude of the flag.

The question was if the building side changes altitude as soon as the planer steps on it (at the end) or if he does some work there as he does on the other nodes.

So now the building site has the same level as all nodes around it, while the flag could be different before.

libs/s25main/buildings/noBuildingSite.cpp
libv: Well, from where I sit, once you start refactoring code, you never stop :) I would just use the old style here and let it be until someone uses lindent on whole files.

libv:
> The question was if the building side changes altitude as soon as the planer steps on it (at the end) or if he does some work there as he does on the other nodes.

It changes at the same time the planer steps on the central building tile/site/point, and no further work is done, he just walks off over the flag.

Spikeone:
> I.e. it still doesn't work on the actual building site but that still gets modified. Correct?
> 
> @Spikeone agreed?

As far as I remember thats correct yeah. When he reached the middle the terrain slightly changed.

Apart from the replay, I think this should be an addon (defaulting to this behavior, but in older replays to the "rttr" style) - simply because in some games it has become quite meta to change the terrain, which then will be less possible than before.

Flamefire:
> simply because in some games it has become quite meta to change the terrain, which then will be less possible than before.

It will just be different, won't it? It will be the height of the flag instead of the building so you could get (mostly?) the same by moving one node NW.
Addons are quite heavy implementation-wise and from UI side (more addons -> harder to find the one(s) you want) and if this isn't really a worthy enhancement I'd avoid that.

Unfortunately I don't see a easy/clean way to "rescue" older replays.
Maybe: Either use the replay-data version or the replay minor version (we should merge those on that occasion) and in the planer get the replay version if present (else use e.g. INT_MAX) and if lower than X use the old logic.

Messy because the figure should know about the replay, so an addon would be better.
This leaves the UX issue: Would people really use this addon as it does something that is better than the original?

Spikeone:
> It will just be different, won't it? It will be the height of the flag instead of the building so you could get (mostly?) the same by moving one node NW.
> Addons are quite heavy implementation-wise and from UI side (more addons -> harder to find the one(s) you want) and if this isn't really a worthy enhancement I'd avoid that.

Well - no it's not the same and changes terraforming quite a lot. Currently you can quite often create multiple new building spots by planing (for large buildings). With the original behavior that was nearly never possible - so using the original would change - or rather remove one strategy.

In normal cases I would also think, well just do it as in the original, but that behavior has been in RttR for so long, that people got quite used to it. So changing it just by default might not be the best idea, thats why I'd opt for an addon, although also not ideal.

Flamefire:
> Currently you can quite often create multiple new building spots by planing (for large buildings). With the original behavior that was nearly never possible - so using the original would change - or rather remove one strategy.

Why is that not possible? Currently you place a building and all surrounding nodes will be leveled to the same height.
Now it will still be the same height but the height of the flag. If you then place it again, but one node more NW and it will be pretty much the same as before.
The height of the flag will now be the same as the building which was unchanged before.

If that is still a large enough difference then yes, an addon would be better. That's why I'm asking you, I don't know too much about those tricks :)
How'd you name it? "Plane to height of building"?
In the replay loading code we can force-enable it for older replay versions

Spikeone:
> If that is still a large enough difference then yes, an addon would be better. That's why I'm asking you, I don't know too much about those tricks :)
> How'd you name it? "Plane to height of building"?
> In the replay loading code we can force-enable it for older replay versions

Hm - well thinking about it some days later I'm not sure if that really changes much - but then again we can terraform quite a lot in RttR but not in S2. Although I don't like it much, I count it as "advanced" strategy which is always nice to have in a game.

Maybe we should go with an addon "Planning behavior" with the options "Flag (original)" and "Building"?

PR #1909, opened 2026-04-01
Fixed clang 22 error with overloaded operator

Farmer-Markus:
Error:

```
s25client/libs/s25main/ingameWindows/iwAddons.cpp:151:65: error: use of overloaded operator '!=' is ambiguous (with operand types 'AddonGroup' and 'AddonGroup')
  151 |         const bool isVisible = (addon->getGroups() & selection) != AddonGroup(0);
      |                                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~
s25client/libs/s25main/addons/const_addons.h:90:1: note: candidate function has been explicitly deleted
   90 | MAKE_BITSET_STRONG(AddonGroup);
      | ^
s25client/external/libutil/libs/common/include/s25util/enumUtils.h:91:10: note: expanded from macro 'MAKE_BITSET_STRONG'
   91 |     bool operator!=(const Type& lhs, const Type& rhs) = delete;     \
      |          ^
s25client/libs/s25main/ingameWindows/iwAddons.cpp:151:65: note: built-in candidate operator!=(enum AddonGroup, enum AddonGroup)
  151 |         const bool isVisible = (addon->getGroups() & selection) != AddonGroup(0);
      |                                                                 ^
1 error generated.
```

### DIFF
--- a/libs/s25main/ingameWindows/iwAddons.cpp
+++ b/libs/s25main/ingameWindows/iwAddons.cpp
@@ -12,6 +12,7 @@
 #include "gameData/const_gui_ids.h"
 #include "s25util/colors.h"
 #include <utility>
+#include <bitset>
 
 namespace {
 enum
@@ -148,7 +149,7 @@ void iwAddons::UpdateView(const AddonGroup selection)
     for(unsigned i = 0; i < ggs.getNumAddons(); ++i)
     {
         const Addon* addon = ggs.getAddon(i);
-        const bool isVisible = (addon->getGroups() & selection) != AddonGroup(0);
+        const bool isVisible = bitset::any(addon->getGroups(), selection);
         auto* group = GetCtrl<ctrlGroup>(ID_grpAddonsStart + i);
 
         // Don't show addon's gui if addon is beyond selected group or is beyond current page scope


libs/s25main/ingameWindows/iwAddons.cpp
```diff
@@ -148,7 +148,7 @@ void iwAddons::UpdateView(const AddonGroup selection)
     for(unsigned i = 0; i < ggs.getNumAddons(); ++i)
     {
         const Addon* addon = ggs.getAddon(i);
-        const bool isVisible = (addon->getGroups() & selection) != AddonGroup(0);
+        const bool isVisible = static_cast<unsigned>(addon->getGroups() & selection) != 0;
```
Flamefire: The intended way was this:
```suggestion
        const bool isVisible = bitset::any(addon->getGroups(), selection);
```

I.e. bitsets shouldn't be compared but only modified or checked unless we find a use case for that.
Or do you see anything I  missed?

Farmer-Markus: Ups, missed that.
But I don't understand why the MAKE_BITSET_STRONG macro exists.
I mean what's the point of using it? Type safety or to prevent lines like the current one?

Flamefire: A bit of all:
- Type safety: Only combine enumerators of same enum
- Typing: Allow combining of enum classes at all w/o `static_cast`
- Prevent unwanted operations: `bitset != bitset` is usually not what you want as you'd likely wanted to test bits or if bitsets overlap (non-empty union)

Farmer-Markus: > A bit of all:
> 
>     * Type safety: Only combine enumerators of same enum
> 
>     * Typing: Allow combining of enum classes at all w/o `static_cast`
> 
>     * Prevent unwanted operations: `bitset != bitset` is usually not what you want as you'd likely wanted to test bits or if bitsets overlap (non-empty union)

Ok thanks.

There is no bitset any implementation in enumUtils if that is what you meant.
And I thought my compiler is doing something weird since it just crashes without error messages. Very weird

Flamefire:
Actually the issue seems to be that you updated the submodule to a different version. The change from libutil is pulled in in https://github.com/Return-To-The-Roots/s25client/pull/1890 where this issue is avoided too. Run `git submodule update` to fix your tree :)

Farmer-Markus:
> Actually the issue seems to be that you updated the submodule to a different version. The change from libutil is pulled in in #1890 where this issue is avoided too. Run `git submodule update` to fix your tree :)

I see :D
Just saw the bitset::any was available in newer commits of libutils
Thanks for your patience

PR #1911, opened 2026-04-17
Translate Campaigns and "texte"

BeWorld2018:


Flamefire (approved):
Thank you!
I added the submodule update so the other translations you did will be included too

PR #1913, opened 2026-04-19
Fix crash when asking for defenders of harbor with waiting attackers

Flamefire:
What happened:
- Sea attack starts
- Attacker arrives in start harbor and waits for ship
- target is destroyed, attacker prepares to leave to go back home
- harbor is attacked
- "active" soldiers that want to leave are stopped, including this attacker
- code wrongly considers this as belonging to this building

This was indeed introduced by #1910
Prior to that such a soldier would be ignored here.

The result is:
- Harbor is attacked
- Soldier gets notified its goal has already been destroyed
- He goes home
- Comes out and ignores the ongoing fight, YOLO

The idea is to potentially use those for defense.
However e.g. harbors might contain soldiers that belong to another
building and are about to leave.

Don't consider those as this complicates the logic too much and just revert to this behavior.

It fixes the crash which is prior to that caught by an assertion: The soldier is not in `troop_on_mission`.
Fixes #1912

Also fix an assertion failure that detected a mismatch in expected armors: There could be more armored soldiers than soldiers (logic is we have n of m soldiers with armor)
This is detected by continuing the provided savegame.

To avoid similar issues I deduplicated code handling the warehouse inventory: Boat-carrier, soldier and others have to be handled consistently

PR #1917, opened 2026-04-25
Fix sea path finding (for ships)

Flamefire:
If a harbor is too far we will not find the path even though it exists, see the previous TODO.
Issue happened in the map where there was a triangle of water with land in the middle.

A ship was moving between 2 far away harbors when one was destroyed.

When going home we limit the maximum distance by the longest connections of HomeHarbor which might not be enough

Pathfinding considered the current harbor when looking for connections which is not useful and required extra steps.
--> Removed together with `GamePlayer::AddHarborsAtSea` which was only used there and added test

I found this while debugging the [previous seafaring issue.](https://github.com/Return-To-The-Roots/s25client/pull/1913)
When letting it continue to run on the triangle-map mostly land with a triangle shaped straight of water it would run into the assert.

Additional:
`CreateTradeGraphs` simply created a cache instance but must be manually invoked or it will crash at runtime.
Remove the method and create the instance on first access for consistency

### DIFF
--- a/libs/s25main/GamePlayer.cpp
+++ b/libs/s25main/GamePlayer.cpp
@@ -2058,18 +2058,6 @@ noShip* GamePlayer::GetShipByID(const unsigned ship_id) const
         return ships[ship_id];
 }
 
-void GamePlayer::AddHarborsAtSea(std::vector<nobHarborBuilding*>& harborBuildings, const SeaId seaId) const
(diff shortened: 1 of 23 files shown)


Flamefire:
@Flow86 Merge please

Flow86 (approved):

PR #1918, opened 2026-05-01
Fix inexhaustible granite mine production

### LINKED ISSUE #918: Inexhaustible granite mines broken
Sadly that addon is broken. It is intended to allow the player to build a granite mine anywhere - regardless of actual resources. In v20160228 it was still working as expected.

DevOpsOfChaos:
## Summary
- allow granite mines to produce without a granite resource spot when `INEXHAUSTIBLE_GRANITEMINES` is enabled
- keep the existing resource requirement when the addon is disabled
- add regression coverage for both addon-disabled and addon-enabled behavior

## Root cause
The addon prevented granite resource depletion, but production still required `FindPointWithResource(ResourceType::Granite)` to find an existing resource spot. On mountain positions without granite resources, the mine therefore never started producing even with the addon enabled.

## Validation
- `Test_integration.exe --run_test=Production --report_level=short`
- `ctest --test-dir C:\dev\s25client\build-vs-x64-debug -C Debug --output-on-failure`

Fixes #918

### DIFF
--- a/libs/s25main/figures/nofMiner.cpp
+++ b/libs/s25main/figures/nofMiner.cpp
@@ -71,21 +71,30 @@ helpers::OptionalEnum<GoodType> nofMiner::ProduceWare()
 
 bool nofMiner::AreWaresAvailable() const
 {
-    return nofWorkman::AreWaresAvailable() && FindPointWithResource(GetRequiredResType()).isValid();
+    return nofWorkman::AreWaresAvailable()
+           && (CanMineWithoutResource() || FindPointWithResource(GetRequiredResType()).isValid());
 }
 
 bool nofMiner::StartWorking()
 {
-    MapPoint resPt = FindPointWithResource(GetRequiredResType());
-    if(!resPt.isValid())
-        return false;
     const GlobalGameSettings& settings = world->GetGGS();
-    bool inexhaustibleRes = settings.isEnabled(AddonId::INEXHAUSTIBLE_MINES)
-                            || (workplace->GetBuildingType() == BuildingType::GraniteMine
-                                && settings.isEnabled(AddonId::INEXHAUSTIBLE_GRANITEMINES));
-    if(!inexhaustibleRes)
-        world->ReduceResource(resPt);
+    const bool canMineWithoutResource = CanMineWithoutResource();
+    const bool inexhaustibleRes = settings.isEnabled(AddonId::INEXHAUSTIBLE_MINES) || canMineWithoutResource;
+    if(!canMineWithoutResource)
+    {
+        MapPoint resPt = FindPointWithResource(GetRequiredResType());
+        if(!resPt.isValid())
+            return false;
+        if(!inexhaustibleRes)
+            world->ReduceResource(resPt);
+    }
     return nofWorkman::StartWorking();
+}
+
+bool nofMiner::CanMineWithoutResource() const
+{
+    return workplace->GetBuildingType() == BuildingType::GraniteMine
+           && world->GetGGS().isEnabled(AddonId::INEXHAUSTIBLE_GRANITEMINES);
 }
 
 ResourceType nofMiner::GetRequiredResType() const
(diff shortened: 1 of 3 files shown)


libs/s25main/figures/nofMiner.cpp
```diff
@@ -71,23 +71,32 @@ helpers::OptionalEnum<GoodType> nofMiner::ProduceWare()
...
-        world->ReduceResource(resPt);
+    const bool canMineWithoutResource = CanMineWithoutResource();
+    const bool inexhaustibleRes = settings.isEnabled(AddonId::INEXHAUSTIBLE_MINES) || canMineWithoutResource;
+    if(!canMineWithoutResource)
+    {
+        MapPoint resPt = FindPointWithResource(GetRequiredResType());
+        if(!resPt.isValid())
+            return false;
+        if(!inexhaustibleRes)
+            world->ReduceResource(resPt);
```
Flamefire, line 89: This can be simplified: You only use `inexhaustibleRes` in the condition, so this is equivalent, isn't it?

```suggestion
    if(!CanMineWithoutResource())
    {
        MapPoint resPt = FindPointWithResource(GetRequiredResType());
        if(!resPt.isValid())
            return false;
        if(!settings.isEnabled(AddonId::INEXHAUSTIBLE_MINES))
            world->ReduceResource(resPt);
```

tests/s25Main/integration/testProduction.cpp
```diff
@@ -102,4 +102,46 @@ BOOST_FIXTURE_TEST_CASE(MetalWorkerOrders, WorldWithGCExecution1P)
...
+    const unsigned initialStones = curInventory[GoodType::Stones];
+    RTTR_EXEC_TILL(500, mine->HasWorker());
+    RTTR_SKIP_GFS(2000);
+
+    BOOST_TEST(curInventory[GoodType::Stones] == initialStones);
+}
+
+BOOST_FIXTURE_TEST_CASE(InexhaustibleGraniteMineWorksWithoutResources, WorldWithGCExecution1P)
+{
+    ggs.setSelection(AddonId::INEXHAUSTIBLE_GRANITEMINES, 1);
```
Flamefire, line 129: Those 2 tests have the same setup, which I'd like to keep in sync and clearly the same.

I see 2 ways:
1) Move the common code to a fixture
2) Change the setting of `INEXHAUSTIBLE_GRANITEMINES` in the test


I'm concerned by a (purely theoretical though) difference:

```diff
-    RTTR_EXEC_TILL(500, mine->HasWorker());
-    RTTR_SKIP_GFS(2000);
+    RTTR_EXEC_TILL(3000, curInventory[GoodType::Stones] > initialStones);
```

In 2nd case
- You don't wait for the worker to arrive
- Time is longer. So it could (theoretically) be possible that after 2501 GFs a stone would arrive making the test wrongly pass

The name `GraniteMineWithoutResourcesNeedsAddon` would be a great fit for the  2nd approach: Addon disable: No stone, enabling it yields stone being produced.

Currently not sure if the logic of when the worker starts (trying) to work again works with that though in which case the fixture needs to be used.

What do you think?

Flamefire (changes requested):
Looks good!
How much of that was created by AI? Just curious


In https://github.com/Return-To-The-Roots/s25client/issues/918 we discussed that the description of the addon would need adjustment as with this change stone mines are not (only) inexhaustible but don't need (explicit) stone resource in the first place and hence could be build anywhere on the mountain, even on e.g. gold or coal deposits
Can you (or Codex) come up with a slightly changed description to avoid confusion of that?

libs/s25main/figures/nofMiner.cpp
DevOpsOfChaos: yeah i agree , it's equivalent

tests/s25Main/integration/testProduction.cpp
DevOpsOfChaos: 70/30 AI i' use RTTR since 2024.  now i'am more familiar with programming due to ai,

DevOpsOfChaos: i will check on that later.

PR #1919, opened 2026-05-01
Add military warehouse focus

DevOpsOfChaos:
## What changed

- Adds a player-level military warehouse focus that defaults to the HQ.
- Adds a warehouse window button with a soldier icon and tooltip to select the military warehouse.
- Routes beer, swords, and shields to the selected military warehouse so recruitment goods are concentrated in one place.
- Preserves existing non-collect warehouse settings when switching the focus.
- Stores the selected military warehouse in savegames and adds a replay-safe game command.

## Why

Soldier recruitment is warehouse-local: beer, swords, shields, and helpers must be present in the same warehouse. In larger economies these goods can become scattered across multiple warehouses, which makes recruitment unreliable unless the player manually blocks or redirects goods in every warehouse.

This change gives players one explicit recruitment warehouse target without adding a new transport system.

## Validation

- `cmake --build . --config Debug --target Test_integration --parallel`
- `Test_integration.exe --run_test=GameCommandSuite/SetMilitaryWarehouseTest --catch_system_errors=no`
- `Test_integration.exe --run_test=MilitaryWarehouseAttractsRecruitmentGoods --catch_system_errors=no`
- `ctest -R Test_integration --output-on-failure -C Debug`

### DIFF
--- a/libs/s25main/GameCommand.cpp
+++ b/libs/s25main/GameCommand.cpp
@@ -15,7 +15,8 @@ unsigned Deserializer::getCurrentVersion()
     // 1: Add wine addon --> 3 new values in distribution
     // 2: Add leather addon --> 3 new values in distribution, 1 new value in transport order and transport order default
     // values changed, 3 new values in custom build order, 3 new jobs/wares in setAllInventory settings
-    return 2;
+    // 3: Add SetMilitaryWarehouse command
+    return 3;
 }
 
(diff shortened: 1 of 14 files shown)


Flamefire:
Why is that necessary? You can already do the same in a more flexible way by using the "collect" setting.

DevOpsOfChaos:
my fix also checks if you want to do it yourself. But i think it's a useful feature.

PR #1921, opened 2026-05-01
Add limited extra harbor spots addon

DevOpsOfChaos:
## Summary

- add an optional `ADDITIONAL_HARBOR_SPOTS` addon
- generate a limited deterministic set of extra harbor spots instead of using every suitable coastal castle site
- keep the default harbor-marker behavior unchanged unless the addon is enabled
- preserve existing map-defined harbor spots as the preferred source of harbor placement
- explicitly label the addon as dangerous/advanced because it can alter intended map seafaring design
- add regression coverage for capped generation, spacing, existing-harbor preservation, deterministic selection, and runtime BQ behavior

## Motivation

Some maps contain coastal locations that are physically suitable for harbor buildings, but do not define explicit harbor markers. On such maps, players may be unable to build a harbor even though the terrain would otherwise support one.

The original version of this PR generated harbor spots for all suitable coastal castle sites. Review feedback correctly pointed out that this was too aggressive: it could create harbor spam, change map design too heavily, confuse players, and make expedition behavior worse.

This version keeps the addon optional and disabled by default, but constrains the generated spots to a small deterministic set.

## Behavior

When the addon is disabled:

- default gameplay is unchanged
- map-defined harbor spots are used as before
- runtime BQ recalculation does not invent harbor spots

When the addon is enabled:

- existing explicit/map-defined harbor spots are preserved
- additional harbor spots are selected only during the one-time static `MapLoader` sea/harbor setup
- suitable markerless candidates are collected during sea/harbor initialization
- not every suitable candidate is used
- at most 4 extra harbor spots are generated
- generated spots must be at least 12 map units away from existing harbors
- generated spots must be at least 12 map units away from each other
- selection is deterministic for a given map

## Scope / safety notes

This PR intentionally keeps the behavior limited:

- the addon is disabled by default
- the addon is labelled as dangerous/advanced in the UI text
- existing map-defined harbor spots remain preferred and unchanged
- generated spots still have to satisfy the normal suitability checks for a coastal castle/harbor location
- generated spots are inserted into `harborData` during sea/harbor initialization
- generated spots receive normal `harborId`, `seaId`, coastal-point and neighbor data through the existing harbor initialization path
- `BQCalculator` only treats already initialized harbor nodes as harbors
- runtime BQ recalculation does not create markerless harbor spots
- this does not change ship movement, expedition logic, harbor building logic, or maps
- this does not add popup/warning UI

The intent is not to turn arbitrary coastal maps into seafaring maps. It is only a constrained optional fallback for maps that already have suitable coastal castle sites but no corresponding harbor markers.

## Validation

- Ran `Test_integration.exe --run_test=SeaWorldCreationSuite --report_level=short`
  - 9 test cases passed
  - 391 assertions passed
- Ran `ctest --test-dir .\build-vs-x64-debug-local -C Debug -R "^Test_integration$" --output-on-failure`
  - passed
- Ran clang-format dry-run
  - clean
  - clean

## Latest verification

Final head: `ecd316034`

- tested through a temporary local integration with current master `4b314647`
- GCC 14 Debug with warnings-as-errors: build passed, `ctest` 20/20
- Clang 18 Debug with warnings-as-errors: build passed, `ctest` 20/20
- `SeaWorldCreationSuite` passed
- earlier verification covered Release, ASan/UBSan, clang-tidy and ten deterministic repetitions
- Jenkins and MSVC were not available locally


libs/s25main/addons/AddonFreeHarborSpots.h
```diff
@@ -0,0 +1,17 @@
...
+
+#include "AddonBool.h"
+#include "mygettext/mygettext.h"
+
+class AddonFreeHarborSpots : public AddonBool
+{
+public:
+    AddonFreeHarborSpots()
+        : AddonBool(AddonId::FREE_HARBOR_SPOTS, AddonGroup::GamePlay, _("Build harbors without map markers"),
+                    _("Allows harbors on suitable coastal castle sites even if the map does not define harbor spots."))
```
Flamefire: I would clarify this a bit more:

```suggestion
        : AddonBool(AddonId::FREE_HARBOR_SPOTS, AddonGroup::GamePlay, _("Build harbors without map markers"),
                    _("Allows harbors on all suitable coastal castle sites even if the map does not define harbor spots."))
```

libs/s25main/world/BQCalculator.h
```diff
@@ -10,13 +10,16 @@
 
 struct BQCalculator
 {
-    BQCalculator(const World& world) : world(world) {}
+    BQCalculator(const World& world, const bool allowFreeHarborSpots = false)
```
Flamefire: `allowFreeHarborSpots` isn't clear enough to me. No idea right away for a better name but should probably start with "allow harbor" or "use harbor"

libs/s25main/world/BQCalculator.h
```diff
@@ -219,8 +222,24 @@ BuildingQuality BQCalculator::operator()(const MapPoint pt, T_IsOnRoad isOnRoad,
     }
 
     // If we can build a castle and this is a harbor point -> Allow harbor
-    if(curBQ == BuildingQuality::Castle && world.GetNode(pt).harborId)
-        curBQ = BuildingQuality::Harbor;
+    if(curBQ == BuildingQuality::Castle)
+    {
+        bool isHarborPoint = world.GetNode(pt).harborId.isValid();
+        if(!isHarborPoint && allowFreeHarborSpots)
```
Flamefire: Variable doesn't seem to be required, you can set `curBQ` directly

```suggestion
        if(world.GetNode(pt).harborId)
            curBQ = BuildingQuality::Harbor;
        else
```

tests/s25Main/integration/testSeaWorldCreation.cpp
```diff
@@ -128,6 +190,45 @@ BOOST_FIXTURE_TEST_CASE(HarborSpotCreation, SeaWorldWithGCExecution<>)
     }
 }
 
+BOOST_FIXTURE_TEST_CASE(FreeHarborSpotsAddonAddsCoastalHarbors, SeaWorldWithGCExecution<>)
```
Flamefire: Looks like we don't need a fixture here, especially not one "WithGCExecution"

Couldn't  you just create a `World` directly and work with that? Especially as you basically reset everything the fixture does

Flamefire (changes requested):

DevOpsOfChaos:
Thanks, addressed the review points.

* clarified the addon description with “all suitable coastal castle sites”
* renamed the `BQCalculator` option to `allowHarborsWithoutMapMarkers`
* simplified the harbor BQ branch without changing behavior
* replaced the heavier `SeaWorldWithGCExecution<>` fixture for the new tests with lighter world fixtures

Semantics are unchanged: generated harbor spots are still created only during sea/harbor initialization, and runtime BQ recalculation does not invent additional harbor spots.

Validation:

* Built `Test_integration` locally with Visual Studio 2022 / Debug
* Ran `Test_integration.exe --run_test=SeaWorldCreationSuite --report_level=short`
* Result: 5 test cases passed, 1521 assertions passed

Spikeone:
Isn't there quite a potential for breaking the game? Not sure if we correctly check if harbors are very close or not, but I think this is a dangerous option for people who do not know how the seafaring and harbors actually work.

Although I do like the idea, I'd still not use such addon and rather go for a modified version of a map.

DevOpsOfChaos:
> Isn't there quite a potential for breaking the game? Not sure if we correctly check if harbors are very close or not, but I think this is a dangerous option for people who do not know how the seafaring and harbors actually work.
> 
> Although I do like the idea, I'd still not use such addon and rather go for a modified version of a map.

That is a fair concern.

The addon is disabled by default and it only adds spots during harbor initialization, but I agree that this can still change map design quite heavily for players who do not understand the seafaring rules.

I can either make the addon text more explicit that this is an advanced/map-altering option, or close this PR if the preferred direction is to solve these cases with modified maps instead.

Spikeone:
Best would be showing a hint when enabling it I think. But we don't have that yet, not sure how hard it is to implement this. Also people sadly do not read the description in most cases.

If we can have a popup, I'd go for that.

If not, we maybe can prefix it with "Dangerous: ..."? Maybe this makes people question what this is about and they read the hint,

DevOpsOfChaos:
> Best would be showing a hint when enabling it I think. But we don't have that yet, not sure how hard it is to implement this. Also people sadly do not read the description in most cases.
> 
> If we can have a popup, I'd go for that.
> 
> If not, we maybe can prefix it with "Dangerous: ..."? Maybe this makes people question what this is about and they read the hint,

Done. I prefixed the addon with `Dangerous:` and made the description more explicit about the map/seafaring risk.

I’ll keep the popup idea separate from this PR. A small generic addon-warning popup could be worth exploring as a follow-up, then we can review whether that direction makes sense before tying it to this addon.

Flamefire:
> Isn't there quite a potential for breaking the game? Not sure if we correctly check if harbors are very close or not, but I think this is a dangerous option for people who do not know how the seafaring and harbors actually work.

Very valid concern I'd say. E.g. exploration expeditions go to the "next" unexplored harbor spot, and (founders) expedition are steered towards the next (free) harbor spot. If now suddenly there are so many of them this could become unusable.

I'm not sure if or for which maps this addon would be useful. Any example?
I'd rather not add things "just because"

DevOpsOfChaos:
> > Isn't there quite a potential for breaking the game? Not sure if we correctly check if harbors are very close or not, but I think this is a dangerous option for people who do not know how the seafaring and harbors actually work.
> 
> Very valid concern I'd say. E.g. exploration expeditions go to the "next" unexplored harbor spot, and (founders) expedition are steered towards the next (free) harbor spot. If now suddenly there are so many of them this could become unusable.
> 
> I'm not sure if or for which maps this addon would be useful. Any example? I'd rather not add things "just because"

That is a fair concern, and I agree we should not add this only because it is technically possible.

I ran a local audit with Codex to get concrete data instead of guessing. The temporary audit loaded the available `.swd`/`.wld` maps through the real `MapLoader` path, once with `FREE_HARBOR_SPOTS` disabled and once with it enabled, then compared the harbor counts. I removed the temporary audit source afterwards and kept only the generated report for review.

The scan found real examples where maps have no map-defined harbor spots but would get generated candidates:

* `DuskTillDawn.SWD`: 0 existing harbors, 10 generated with the addon, 8 significant land components
* `Insel2.swd`: 0 existing harbors, 22 generated with the addon, 2 significant land components
* `INSEL6-6.SWD`: 0 existing harbors, 39 generated with the addon, 3 significant land components
* `INSEL5-2.SWD`: 0 existing harbors, 57 generated with the addon, 4 significant land components
* `KARIBIK2.SWD`: 0 existing harbors, 66 generated with the addon, 9 significant land components


So there are existing island/coastal maps where this addon has a concrete use case.

However, the audit also confirms the risk: some maps would receive a very large number of generated harbor spots, which could indeed make expedition behavior worse or change the map too much.

Maybe the better direction is not “allow every suitable coastal castle site”, but to add stricter generation rules. For example, generate only a small number of harbor spots per land component, enforce a minimum distance between generated harbors, and maybe limit small islands to one harbor at most while larger landmasses get only a few well-spaced candidates.

I attached the audit report here so we can discuss whether the current broad version is acceptable, whether it should be constrained further, or whether this should stay out of upstream.

[free_harbor_spots_map_audit.md](https://github.com/user-attachments/files/27315463/free_harbor_spots_map_audit.md)

Spikeone:
> So there are existing island/coastal maps where this addon has a concrete use case.

Those are very few maps aren't they? Also those "Insel" Maps are rarely played. Thinking about this, I now also fear that people to expect this to work on most maps (e.g. they may think they can make "The green island" a seafaring map) but it fails most of the time. 5 / 444 maps are only some rare edge cases I think. 

If those maps are truly meant for seafaring, we should modify them I think.

Flamefire:
I feel similar: If there is a seafaring map then harbors will be placed by the author. Very likely in places where it makes sense and on purpose.
"Spamming" the map with harbor spots doesn't seem sensible.

Was something like this ever requested? If so in which situation?

DevOpsOfChaos:
> I feel similar: If there is a seafaring map then harbors will be placed by the author. Very likely in places where it makes sense and on purpose. "Spamming" the map with harbor spots doesn't seem sensible.
> 
> Was something like this ever requested? If so in which situatio

> I feel similar: If there is a seafaring map then harbors will be placed by the author. Very likely in places where it makes sense and on purpose. "Spamming" the map with harbor spots doesn't seem sensible.
> 
> Was something like this ever requested? If so in which situation?

I’ve noticed this on a few maps myself, so the idea came from my own observation rather than a specific request. In some of those cases, I think additional harbor spots could have been useful.

That said, I agree with your overall point: on a seafaring map, harbor placement is usually intentional and adding more indiscriminately would likely reduce map quality. It also probably wouldn’t have as much positive impact as one might expect.

Given that this only affects a small number of maps, adjusting the system to avoid “harbor spam” doesn’t seem worth the effort.

Flamefire:
Ok, so it was basically your own request.

Well, if we add this we should at least be sensible, i.e. this makes sense:
> Maybe the better direction is not “allow every suitable coastal castle site”, but to add stricter generation rules. For example, generate only a small number of harbor spots per land component, enforce a minimum distance between generated harbors, and maybe limit small islands to one harbor at most while larger landmasses get only a few well-spaced candidates.

However this might be significant effort to identify landmasses/islands. There is already a quite elaborate algorithm for the map generator which works on "coastlines" but that can't be used here directly.

Not sure if this is worth the effort.
But if we want something "cheap" then maybe: Collect all possible new harbor spots (coasts with castle BQ), determine an amount of new harbors (maybe based on map size? sea size? existing number of harbors?) and randomly select some rejecting those that are too close to an existing one. This would however not place ones on different islands opposite a small strip of water but further away. Should be OK though.
The randomness has the advantage that this makes games different for each start, which could be nice.
Or seed the RNG by something map specific to select random harbors but the same/deterministic for each map to avoid having to create some policy for that.

@Spikeone 

> they may think they can make "The green island" a seafaring map

That wouldn't work? Why?

> 5 / 444 maps are only some rare edge cases I think.

That's the maps that had no harbors at all. So on others there were harbors, and now are more.

That would also be an option: Limit it to maps with original harbors.

Either way the add-on could then be described as "(add) extra harbors"

Just some thoughts.
What do you 2 think? And maybe @Flow86 has an opinion too?

Spikeone:
> That wouldn't work? Why?

```
| `C:/Program Files/Spiele/Siedler\DATA\MAPS3\AKARTE03.WLD` | 96x96 | 4 | land components 1 (1 significant), seas 3 |
```
So this is a map that feels quite coastal (but has no large building spots). I just wanted to say, that players might expect such maps to have seafaring with this addon enabled.



> Not sure if this is worth the effort.
> But if we want something "cheap" then maybe: Collect all possible new harbor spots (coasts with castle BQ), determine an amount of new harbors (maybe based on map size? sea size? existing number of harbors?) and randomly select some rejecting those that are too close to an existing one. This would however not place ones on different islands opposite a small strip of water but further away. Should be OK though.
> The randomness has the advantage that this makes games different for each start, which could be nice.
> Or seed the RNG by something map specific to select random harbors but the same/deterministic for each map to avoid having to create some policy for that.

I like this idea to spread them somewhat evenly on a landmass. Also this should have some validation, if a harbor can connect any roads at all - maybe the harbor would block itself or trees are blocking buildings, so you'd have a harbor that can not be used. That even spread is important so you may not have most harbors on one side. But implementing this could be hard in some cases (e.g. if one side of the island is a mountain area and the other is flat for harbors).

Could we preview the amount of generated harbors? Or is it smarted to only generate them after the game actually starts so disabling/enabling the addon does not cause that generation?

As long as things are an addon, I'm fine with any addition - just want to make sure those addons don't confuse people or cause false positive bug reports because results don't meet expectation :)

Flamefire:
> So this is a map that feels quite coastal (but has no large building spots). I just wanted to say, that players might expect such maps to have seafaring with this addon enabled.

So it just doesn't make (much) sense as there is only a single island. But you could still use harbors as means of fast(er?) transport, can't you?

> I like this idea to spread them somewhat evenly on a landmass. Also this should have some validation, if a harbor can connect any roads at all - maybe the harbor would block itself or trees are blocking buildings, so you'd have a harbor that can not be used. That even spread is important so you may not have most harbors on one side. But implementing this could be hard in some cases (e.g. if one side of the island is a mountain area and the other is flat for harbors).

That might then be too much already. "spread even on a landmass" requires determining what a "landmass" is first. Possible but not trivial. Hence the idea to use randomness and distance thresholds to get (likely) something close to that.
But if there is a castle building spot, I'd guess that we don't need additional checks for blockage.

> Could we preview the amount of generated harbors? Or is it smarted to only generate them after the game actually starts so disabling/enabling the addon does not cause that generation?

Preview where? But yes as addons are part of a game they will be generated when loading the map for the game.

> As long as things are an addon, I'm fine with any addition - just want to make sure those addons don't confuse people or cause false positive bug reports because results don't meet expectation :)

True. But we could make that clearer in the description. If it has something like "randomly converts some coastal spots to harbor spots" it would convey that results might not be perfect. Or even have that in the tooltip that not all such harbors might be useful or even usable.

Spikeone:
> So it just doesn't make (much) sense as there is only a single island. But you could still use harbors as means of fast(er?) transport, can't you?

Honestly, on a map like the green island seafaring makes more sense in multiplayer than on some other maps. Simply because it is not necessary, but adds a second front with multiple players (if sea attacks are enabled. So normally you may not attack from the upper left to the lower right - but with seafaring you could. 

> But if there is a castle building spot, I'd guess that we don't need additional checks for blockage.

I think we should, simpyl because even our random maps sometimes generated such spots and having a harbor you can not attack is pretty annyoing - or if you can not build anything next to it.

> Preview where? But yes as addons are part of a game they will be generated when loading the map for the game.

Like a message saying "X - Harbor spots generated on Y islands". So at least players know that something happened.



> True. But we could make that clearer in the description. If it has something like "randomly converts some coastal spots to harbor spots" it would convey that results might not be perfect. Or even have that in the tooltip that not all such harbors might be useful or even usable.

👍 maybe I'm to fearful, that people don't read those^^

Flamefire:
> Like a message saying "X - Harbor spots generated on Y islands". So at least players know that something happened.

That could happen when starting the game as a message printed to the log. Or do you want a message box?
The latter requires determining islands which, mentioned above, isn't trivial. Do we really need that?
Depends on the approach we choose on the distribution.
Shall we go with that landmass approach? What would you suggest? Maybe:

 Place harbors on all spots but keep distance X to harbors on the same island and distance Y to any other harbor in "lexical" order (L->R, T->B)

> 👍 maybe I'm to fearful, that people don't read those^^

Not our fault if they don't ;-)

DevOpsOfChaos:
Thanks for the feedback and suggestions.

I have my final exams this week, so I won’t be able to properly continue the discussion or rework this PR until after that.

My current feeling is that the addon should probably be constrained more than the current broad version. The audit shows real markerless island/coastal map cases where generated harbor spots can help, but it also shows that allowing every suitable coastal castle site can create too many harbor spots on some maps.

So I think the better direction might be stricter generation rules, for example limiting generated harbor spots per landmass and enforcing some spacing, instead of simply accepting every suitable candidate.

I’ll take another look after my exams.

DevOpsOfChaos:
After rereading the discussion, I agree that the current broad version is probably too aggressive.

The audit shows real candidate maps, but it also shows the main problem: generating every suitable coastal castle site can create far too many harbor spots and may hurt expedition behavior.

If this PR should continue, I would rework it toward a much more limited “extra harbor spots” addon instead of “free harbor spots”: collect candidate coastal castle sites, keep existing map harbors preferred, reject candidates too close to existing/generated harbors, and cap the number of generated spots. I would also keep the selection deterministic, not random per game, to avoid replay/network surprises.

If that direction is still considered too niche for upstream, I’m also fine with closing this PR and leaving such cases to modified maps.

DevOpsOfChaos:
Updated the PR to address the harbor-spam/design risk.

The addon no longer generates every suitable coastal castle site as a harbor spot. Instead it now performs a deterministic limited selection:

- at most 4 generated extra harbor spots
- minimum distance 12 from existing map-defined harbors
- minimum distance 12 between generated harbors
- existing explicit/map-defined harbors are preserved
- runtime BQ recalculation still does not create markerless harbor spots
- addon text now says `Dangerous: Add limited extra harbor spots`

This keeps the feature optional and still dangerous/advanced, but avoids the previous “turn every suitable coastal castle site into a harbor” behavior.

The implementation intentionally does not add popup UI or expedition/ship logic changes in this PR.

Validation:

- built `Test_integration`
- `Test_integration.exe --run_test=SeaWorldCreationSuite --report_level=short`
  - 7 test cases
  - 366 assertions
- `ctest --test-dir .\build-vs-x64-debug-local -C Debug -R "^Test_integration$" --output-on-failure`
  - passed
  - clean

libs/s25main/world/BQCalculator.h
```diff
@@ -219,8 +222,23 @@ BuildingQuality BQCalculator::operator()(const MapPoint pt, T_IsOnRoad isOnRoad,
     }
 
     // If we can build a castle and this is a harbor point -> Allow harbor
-    if(curBQ == BuildingQuality::Castle && world.GetNode(pt).harborId)
-        curBQ = BuildingQuality::Harbor;
+    if(curBQ == BuildingQuality::Castle)
```
Flamefire: This should not be in here as it slows down processing during the game. This can be fully done at the place where this information is required and it makes the "keep in sync" part easier as both will be in the same file

libs/s25main/world/MapLoader.cpp
```diff
@@ -420,7 +422,52 @@ bool MapLoader::PlaceHQs(GameWorldBase& world, const std::vector<MapPoint>& hqPo
     return true;
 }
 
-bool MapLoader::InitSeasAndHarbors(World& world, const std::vector<MapPoint>& additionalHarbors)
+namespace {
+bool hasHarborAt(const World& world, const MapPoint pt)
+{
+    for(const auto harborId : helpers::idRange<HarborId>(world.GetNumHarborPoints()))
```
Flamefire: This might be unnecessarily slow: You can simply collect all harbor positions in a vector and use `helpers::contains`, can't you?

libs/s25main/world/MapLoader.cpp
```diff
@@ -420,7 +422,52 @@ bool MapLoader::PlaceHQs(GameWorldBase& world, const std::vector<MapPoint>& hqPo
...
+    {
+        if(world.GetHarborPoint(harborId) == pt)
+            return true;
+    }
+    return false;
+}
+
+bool isFarEnoughFromHarbors(const World& world, const MapPoint pt, const std::vector<MapPoint>& generatedHarbors)
+{
+    for(const auto harborId : helpers::idRange<HarborId>(world.GetNumHarborPoints()))
```
Flamefire, line 553: Similar as above: If you have a single vector of all harbor points this is much faster and shorter

Flamefire (commented):
There is one big issue: This will basically create 4 harbors in the top of the map which doesn't sound like much of an improvement. I'd suggest to pick all viable spots and choose the "best" ones. That could be done fully randomly seeded by something map-specific for reproducibility or some "smart" algorithm. E.g.: Pick a random initial point and then add one-by-one points that have the largest min-distance to all existing ones >= MIN_GENERATED_HARBOR_DISTANCE
Or maybe someone has a better idea, but currently it just leads to a (very small) cluster of harbors

Additionally: If this is an addon already, why not let the user choose how many harbors to add? E.g. "None", "few", "many" (and 1-3 more options?) which then limit to some (internal) number which we could adjust in the future instead of letting users select e.g. "5 extra harbors". That could even scale with the map size and/or number/size of seas, possibly adjusted if there are many harbors already.

DevOpsOfChaos:
I tried to open the Jenkins job, but `jenkins.mytrap.de` redirects me to a GitLab Community Edition login page, and I do not seem to have access to that instance.

I revalidated the current head locally and could not reproduce a failure:

- built `Test_integration`
- `SeaWorldCreationSuite`: 7 test cases / 366 assertions passed
- `ctest -R "^Test_integration$"` passed
- `clang-format --dry-run --Werror` on the changed files was clean
- `git diff --check upstream/master...HEAD` was clean

Could someone with Jenkins/GitLab access please paste the first failing console-log section or rerun the job?

DevOpsOfChaos:
Applied the requested structural rework.

- Removed addon-specific markerless-harbor handling from `BQCalculator`; runtime BQ recalculation now only recognizes already initialized harbor nodes.
- Moved extra harbor spot selection into the one-time `MapLoader` sea/harbor setup path.
- The setup now maintains one local list of existing and selected harbor positions, using it for duplicate checks and minimum-distance checks. Each selected candidate is appended immediately, preserving deterministic selection.
- Kept the existing cap of four generated spots and the minimum distance rules.
- Extended the sea-world tests for disabled behavior, cap and spacing, deterministic selection, and runtime BQ behavior.

Validation passed locally:

- `SeaWorldCreationSuite`: 9 test cases / 391 assertions
- clang-format dry-run: clean

I also investigated the previously failing CI. The Linux/clang-tidy failure came from the former inline addon-specific logic in `BQCalculator.h`, where required declarations/types were unavailable in that header context. The new structure removes that runtime/header path.

DevOpsOfChaos:
Updated the branch against current master and reworked the generated harbor selection.

Additional harbor candidates are now collected first and then selected deterministically by maximum minimum distance to existing and previously selected harbor spots. This avoids filling all slots from an early scan-order cluster while keeping the existing minimum-distance rule intact.

The selection now stops when the remaining candidates are too close, instead of forcing the configured maximum number of generated spots.

Validated locally:

- `SeaWorldCreationSuite`: 11 test cases / 431 assertions
- `FreeHarborSpotsAddonStopsWhenRemainingCandidatesAreTooClose`: 30 assertions

libs/s25main/world/MapLoader.cpp
```diff
@@ -523,7 +526,86 @@ bool MapLoader::PlaceHQs(GameWorldBase& world, const std::vector<MapPoint>& hqPo
...
+    BQCalculator calcBQ(world);
+    RTTR_FOREACH_PT(MapPoint, world.GetSize())
+    {
+        if(helpers::contains(harborPositions, pt))
+            continue;
+        if(calcBQ(pt, [](const MapPoint&) { return false; }) != BuildingQuality::Castle)
+            continue;
+        if(!hasEligibleHarborCoast(world, pt))
+            continue;
+        if(!isFarEnoughFromHarbors(world, pt, harborPositions))
```
Flamefire: This is a wrapper around a raw loop only used here: Always consider using algorithms instead of raw loops. Here I'd rather have:

```suggestion
        if(!helpers::contains_if(harborPositions, [&](const MapPoint harborPt){ return world.CalcDistance(pt, harborPt) < MapLoader::MIN_GENERATED_HARBOR_DISTANCE; }))
```
Or a named var for the lambda if you consider that not readable enough. Or put that into the function instead of the raw loop

libs/s25main/world/MapLoader.cpp
```diff
@@ -523,7 +526,86 @@ bool MapLoader::PlaceHQs(GameWorldBase& world, const std::vector<MapPoint>& hqPo
     return true;
 }
 
-bool MapLoader::InitSeasAndHarbors(World& world, const std::vector<MapPoint>& additionalHarbors)
+namespace {
+bool hasEligibleHarborCoast(const World& world, const MapPoint pt)
+{
+    for(const auto dir : helpers::EnumRange<Direction>{})
+    {
+        if(dir != Direction::NorthWest && world.GetSeaFromCoastalPoint(world.GetNeighbour(pt, dir)))
```
Flamefire, line 535: needs a comment on the !=NW

libs/s25main/world/MapLoader.cpp
```diff
@@ -523,7 +526,86 @@ bool MapLoader::PlaceHQs(GameWorldBase& world, const std::vector<MapPoint>& hqPo
...
+
+        candidates.push_back(pt);
+    }
+
+    std::vector<MapPoint> generatedHarbors;
+    while(!candidates.empty() && generatedHarbors.size() < MapLoader::MAX_GENERATED_HARBOR_SPOTS)
+    {
+        auto bestCandidate = candidates.begin();
+        unsigned bestDistance = getMinimumHarborDistance(world, *bestCandidate, harborPositions);
+        for(auto it = std::next(candidates.begin()); it != candidates.end(); ++it)
```
Flamefire: I'd say the single saved check per generated harbor isn't worth the readability-hit compared to range-based-for (compare to `getMinimumHarborDistance`)
So use that instead please. If you start with `bestDistance = 0` you don't even need to check `candidates.empty()` above

libs/s25main/addons/AddonFreeHarborSpots.h
```diff
@@ -0,0 +1,18 @@
...
+#include "AddonBool.h"
+#include "mygettext/mygettext.h"
+
+class AddonFreeHarborSpots : public AddonBool
+{
+public:
+    AddonFreeHarborSpots()
+        : AddonBool(AddonId::FREE_HARBOR_SPOTS, AddonGroup::GamePlay, _("Dangerous: Add limited extra harbor spots"),
+                    _("Advanced option. Adds only a small deterministic set of suitable coastal castle sites as extra "
+                      "harbor spots. May alter intended map seafaring design."))
```
Flamefire: I'd suggest:

```suggestion
        : AddonBool(AddonId::FREE_HARBOR_SPOTS, AddonGroup::GamePlay, _("Dangerous: Add extra harbor spots"),
                    _("Advanced option. Converts a small set of suitable coastal castle sites to harbor spots. "
                      "Caution: May alter intended map seafaring design."))
```

The "limited" isn't required as there is no other option and "only a small" also seems like comparing against something.
Could also replace "suitable coastal castle sites" by "suitable castle sites" or even "suitable points", as e.g. I can't find a good German translation that sounds good and "suitable" should be specific enough

Flamefire (changes requested):
Looks good, only minor comments on the code and:

1. What about letting the user control the amount of harbors added? For a large map without harbors, 4 additional ones would likely go unnoticed while for a very small map they might be enough, or when harbors exist already
2. "Free harbor spots" doesn't really capture what it does, it doesn't free (existing) spots or so, hence maybe one of "Additional", "new", "generate" or similar. What do you think?

libs/s25main/world/MapLoader.cpp
DevOpsOfChaos: Addressed using `helpers::contains_if` directly. The condition intentionally discards a candidate when any existing or already selected harbor is closer than `MIN_GENERATED_HARBOR_DISTANCE`; the sense was explicitly checked to avoid inversion. An empty harbor list correctly allows candidates through.

libs/s25main/world/MapLoader.cpp
DevOpsOfChaos: Added a comment explaining the `Direction::NorthWest` exclusion. It mirrors the existing coast-assignment rule in `InitSeasAndHarbors`, keeping candidate collection and final harbor initialization consistent.

libs/s25main/world/MapLoader.cpp
DevOpsOfChaos: Addressed with a range-based loop and `bestDistance = 0`; the separate empty-candidate check is gone. Empty candidates break immediately, `MapPoint::Invalid()` cannot be inserted, selected candidates cannot be selected twice, and the loop remains bounded by `MAX_GENERATED_HARBOR_SPOTS`. The deterministic sea-world test was also run ten consecutive times with identical results.

libs/s25main/addons/AddonFreeHarborSpots.h
DevOpsOfChaos: Applied the suggested text and renamed the addon from “Free” to “Additional” throughout. The numeric ID remains `0x01000004`. Include and registration positions were also updated to match the project’s ordering conventions, and no old class, header or enum reference remains.

DevOpsOfChaos:
Final verification update for head `ecd316034`.

The branch was tested through a temporary local integration with current `master` (`4b314647`). GCC 14 and Clang 18 Debug builds with warnings-as-errors both pass, complete `ctest` is 20/20, and `SeaWorldCreationSuite` passes.

The previous include-order workaround was removed. `BQCalculator.h` now includes the declarations it uses and was verified with a minimal translation unit that includes only `world/BQCalculator.h`, compiling under both GCC and Clang.

Regarding configurable harbor counts: I propose keeping the conservative fixed cap of four in this PR. “Few” and “many” do not yet have agreed semantics, and their serialized selection indices would become compatibility-relevant. The current algorithm already spreads selections by maximum minimum distance and stops early when remaining candidates are too close. A separately specified follow-up can add user-selectable levels once the map-size/sea-size scaling is defined.

Earlier verification additionally covered Release, ASan/UBSan, clang-tidy and ten consecutive deterministic sea-world runs. Jenkins and MSVC were not available locally.

PR #1922, opened 2026-05-01
Add recruitment goods collect shortcut

DevOpsOfChaos:
## Summary

- add a warehouse window shortcut for collecting recruitment goods
- sets the existing `Collect` inventory setting for beer, swords, and shields on the selected warehouse
- keeps the existing warehouse collection, distribution, and routing mechanics unchanged

## Motivation

Recruitment depends on beer, swords, and shields being available in the same warehouse. This can already be configured manually through the existing inventory `Collect` setting, but it is easy to miss and repetitive to set up every time.

This is intentionally a small convenience feature. It may be unnecessary for players who already manage warehouse inventory settings directly. The goal is not to introduce a new military warehouse mode, but only to make an existing configuration easier to discover and apply.

## Scope / safety notes

This PR deliberately keeps the behavior narrow:

- no new warehouse state is added
- no new routing logic is added
- no special military warehouse behavior is added
- no recruitment rules are changed
- no ware distribution priorities are changed
- the shortcut only applies existing `Collect` settings for the selected warehouse

In other words, this is a UI shortcut for an existing mechanic, not a new economy system.

## Validation

- Built locally with Visual Studio 2022 Debug
- Ran the affected local build/test target successfully

### DIFF
--- a/libs/s25main/ingameWindows/iwBaseWarehouse.cpp
+++ b/libs/s25main/ingameWindows/iwBaseWarehouse.cpp
@@ -154,6 +154,28 @@ void iwBaseWarehouse::Msg_Group_ButtonClick(const unsigned group_id, const unsig
     }
 }
 
+
+void iwBaseWarehouse::CollectRecruitmentGoods()
+{
+    if(GAMECLIENT.IsReplayModeOn())
+        return;
+
+    for(const GoodType good : RECRUITMENT_GOODS)
+    {
+        InventorySetting state = wh->GetInventorySettingVisual(good);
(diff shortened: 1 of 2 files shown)


Spikeone:
Well, I think this should either be an addon - but even then I dislike it for a simple reason: Settlers 2 is quite graphic, there are only few shortcuts and all things are possible using the UI.

Adding a somewhat hidden feature (read the keyboard layout or you don't know it) feels counterintuitive. So to counter this we should have an UI option - but that would clutter the UI.

Therefore I'd opt for an addon that also does explain the shortcut then. But not sure if one really needs this feature.

PR #1923, opened 2026-05-01
Send password rejection before kicking client

### LINKED ISSUE #1057: Direct IP and passwords
When you open a direct IP game with a password and try to join without entering one, you get a message "connection to host lost". Which is simply incorrect as it has been refused as you entered no / wrong password.

DevOpsOfChaos:
## Summary

- send the password rejection response synchronously before kicking the client
- avoid reporting a generic connection-lost error when joining a password-protected direct-IP game with a wrong password
- keep the existing `WrongPassword` client error path unchanged
- add deterministic client-side regression coverage for the wrong-password response

## Root cause

The server queued the password response with `sendMsgAsync()` and immediately kicked the client afterwards. For wrong passwords this could close the connection before the queued `"false"` response was sent, so the client only observed a connection loss.

This PR sends the `"false"` password response synchronously before kicking the player. Successful password responses keep the existing async send path.

The regression test intentionally avoids asserting real socket delivery-vs-disconnect timing, because that would be brittle. Instead, it verifies that receiving `GameMessage_Server_Password("false")` triggers the existing `ClientError::WrongPassword` path and stops the client.

## Validation

- Built `Test_network` locally with Visual Studio 2022 / Debug
- Ran `Test_network.exe --run_test=GameClientTests/ClientReportsWrongPasswordResponse --report_level=short`
  - 1 test case passed
  - 9 assertions passed
- Ran `Test_network.exe --report_level=short`
  - 14 test cases passed
  - 228 assertions passed
- Ran `git diff --check`

Fixes #1057

### DIFF
--- a/libs/s25main/network/GameServer.cpp
+++ b/libs/s25main/network/GameServer.cpp
@@ -938,10 +938,12 @@ bool GameServer::OnGameMessage(const GameMessage_Server_Password& msg)
     } else
         playerInfos[msg.senderPlayerID].isHost = false;
 
-    player->sendMsgAsync(new GameMessage_Server_Password(passwordok));
-
     if(passwordok == "false")
+    {
+        player->sendMsg(GameMessage_Server_Password(passwordok));
         KickPlayer(msg.senderPlayerID, KickReason::WrongPassword, __LINE__);
+    } else
+        player->sendMsgAsync(new GameMessage_Server_Password(passwordok));
     return true;
 }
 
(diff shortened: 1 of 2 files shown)


Flow86 (dismissed):
This was probably a missed refactor thing, I remember that in an early version the KickPlayer flushed the queue of the disconnected player, but now its not doing that.

DevOpsOfChaos:
Added deterministic client-side coverage for the wrong-password response path.

I avoided a socket timing test here; the new test verifies that receiving `GameMessage_Server_Password("false")` triggers `ClientError::WrongPassword` and stops the client.

Validation:
- targeted wrong-password test passed
- full `Test_network` passed: 14 tests, 228 assertions
- `git diff --check`

Flamefire (approved):
LGTM.

PR #1924, opened 2026-05-01
Account for static blockers in frontier reachability

### LINKED ISSUE #979: Frontier distance addon not working correctly
The addon "Frontier Distance checks rechability" seems to not work correctly as it does not take all static objects into account. For example on wasteland maps stalagmites are not taken into account, thus indicating wrong distance.

image

I'm not sure if other map type specific objects (e.g. dead tree / skeletton on wasteland) or other specific objects (waterstone, broken ship etc.) are handled correctly.

DevOpsOfChaos:
## Summary

- account for blocking static map objects in frontier-distance reachability checks
- keep terrain-based reachability behavior otherwise unchanged
- add regression coverage for static blockers in the frontier-distance path check

## Motivation

The `FRONTIER_DISTANCE_REACHABLE` addon previously checked terrain reachability only. This meant blocking static map objects, such as decorative objects used as permanent path blockers, could be ignored when determining whether enemy military buildings were reachable.

This keeps the check intentionally narrow: it only treats blocking `Staticobject`s as blockers and does not switch to full human pathing behavior.

## Validation

- Ran `Test_integration.exe --run_test=FrontierDistanceSuite --report_level=short`
- Result: 8 test cases passed, 68 assertions passed

Fixes #979


Flamefire (dismissed):

Flamefire (dismissed):

Spikeone:
Just asking: Does this by chance also take into account, that objects can change (due to scripts) so an object that is static, could become env.

Also not sure, pineapple trees are also static, are they? They can be passed though.

DevOpsOfChaos:
> Just asking: Does this by chance also take into account, that objects can change (due to scripts) so an object that is static, could become env.
> 
> Also not sure, pineapple trees are also static, are they? They can be passed though.

Good point, I added a small regression test for that.

Passable size-0 static objects and environment objects are still treated as passable. Only current static objects with actual blocking behavior affect this reachability check.

No production code change in this follow-up, just test coverage.

Validation:
- built `Test_integration`
- ran `FrontierDistanceSuite`: 8 test cases, 70 assertions passed
- `git diff --check`

tests/s25Main/integration/testFrontierDistance.cpp
```diff
@@ -347,4 +351,55 @@ BOOST_FIXTURE_TEST_CASE(FrontierDistanceBug_815, WorldBig)
...
+
+            MapNode& node = world.GetNodeWriteable(pt);
+            node.t1 = tWater;
+            node.t2 = tWater;
+        }
+    }
+
+    BOOST_TEST_REQUIRE(DoesReachablePathExist(world, startPt, endPt, maxPathLen));
+
+    for(unsigned x = startPt.x + 1; x < endPt.x; ++x)
```
Flamefire:
```suggestion
    for(const unsigned x: helpers::range(startPt.x + 1, endPt.x))
```

tests/s25Main/integration/testFrontierDistance.cpp
```diff
@@ -347,4 +351,55 @@ BOOST_FIXTURE_TEST_CASE(FrontierDistanceBug_815, WorldBig)
...
+    {
+        const MapPoint passablePt(x, corridorY);
+        world.DestroyNO(passablePt, false);
+        world.SetNO(passablePt, new noStaticObject(passablePt, 0, 0, 0));
+    }
+
+    BOOST_TEST_REQUIRE(DoesReachablePathExist(world, startPt, endPt, maxPathLen));
+
+    for(unsigned x = startPt.x + 1; x < endPt.x; ++x)
```
Flamefire:
```suggestion
    for(const unsigned x: helpers::range(startPt.x + 1, endPt.x))
```

tests/s25Main/integration/testFrontierDistance.cpp
```diff
@@ -347,4 +351,55 @@ BOOST_FIXTURE_TEST_CASE(FrontierDistanceBug_815, WorldBig)
...
+    {
+        const MapPoint passablePt(x, corridorY);
+        world.DestroyNO(passablePt, false);
+        world.SetNO(passablePt, new noEnvObject(passablePt, 0, 0));
+    }
+
+    BOOST_TEST_REQUIRE(DoesReachablePathExist(world, startPt, endPt, maxPathLen));
+
+    for(unsigned x = startPt.x + 1; x < endPt.x; ++x)
```
Flamefire:
```suggestion
    for(const unsigned x: helpers::range(startPt.x + 1, endPt.x))
```

tests/s25Main/integration/testFrontierDistance.cpp
```diff
@@ -347,4 +351,55 @@ BOOST_FIXTURE_TEST_CASE(FrontierDistanceBug_815, WorldBig)
...
 
+BOOST_FIXTURE_TEST_CASE(FrontierDistanceReachabilityConsidersStaticObjectBlockers, FrontierWorldStaticBlocker)
+{
+    const DescIdx<TerrainDesc> tWater = GetWaterTerrain(world.GetDescription());
+    const unsigned corridorY = 6;
+    const MapPoint startPt(2, corridorY);
+    const MapPoint endPt(10, corridorY);
+    constexpr unsigned maxPathLen = 8;
+
+    for(const auto y : helpers::range(0, +world.GetHeight()))
```
Flamefire: You can use `RTTR_FOREACH_PT` here

libs/s25main/pathfinding/FindPathReachable.cpp
```diff
@@ -6,10 +6,29 @@
 #include "pathfinding/FreePathFinderImpl.h"
 #include "pathfinding/PathConditionReachable.h"
 #include "world/GameWorldBase.h"
+#include "nodeObjs/noBase.h"
+
+namespace {
+
+struct PathConditionReachableWithStaticBlockers : PathConditionReachable
```
Flamefire, line 15: Can you add a comment here on why this is required, i.e. what the difference to `PathConditionHuman` is. I.e. that this excludes nodes that will never be accepted by `PathConditionHuman` as static objects can't be removed

It might even make sense to put it in the same file of `PathConditionHuman` so one can see that easier

Flamefire (dismissed):
Double checked and looks good. Just a few minor comments.

The first on is because it confused me to now have 3 classes:
PathConditionHuman, PathConditionReachable, PathConditionReachableWithStaticBlockers and why the code of this new one can't be in the existing one(s).
So this needs at least a comment

DevOpsOfChaos:
Addressed the remaining minor review comments.

I added a short explanation for `PathConditionReachableWithStaticBlockers`, simplified the repeated corridor loops with `helpers::range(...)`, and used `RTTR_FOREACH_PT` for the full-map setup loop.

Validation:
- built `Test_integration`
- ran `FrontierDistanceSuite`: 8 test cases, 70 assertions passed

libs/s25main/pathfinding/FindPathReachable.cpp
```diff
@@ -6,10 +6,31 @@
 #include "pathfinding/FreePathFinderImpl.h"
 #include "pathfinding/PathConditionReachable.h"
 #include "world/GameWorldBase.h"
+#include "nodeObjs/noBase.h"
+
+namespace {
+
+// Keep this narrower than PathConditionHuman, but reject permanently blocking static objects:
+// Human pathing can never accept them and they cannot be removed.
```
Flamefire, line 14: Not sure, but "narrower" sounds like "more strict" to me.
Based on this maybe:
```suggestion
/// More permissive version of PathConditionHuman that rejects only permanently blocking objects,
/// instead of any blocking object, to find paths which may be usable at some point.
```

Or maybe your AI can come up with something better

DevOpsOfChaos: Adjusted the wording.

The comment now describes this condition as more permissive than `PathConditionHuman`, while still rejecting permanently blocking static objects.

Validation:
- ran clang-format on `FindPathReachable.cpp`

Flamefire (approved):

DevOpsOfChaos:
Updated the branch against current master.

The frontier-distance behavior remains unchanged: permanently blocking static objects are now considered during reachable-path checks, while passable or temporary objects remain allowed.

Validated locally:

- `FrontierDistanceReachabilityConsidersStaticObjectBlockers`: 6 assertions
- `FrontierDistanceSuite`: 8 test cases / 70 assertions

PR #1925, opened 2026-05-01
Ignore isolated fish resources for fisheries

### LINKED ISSUE #1171: Fisheries work in 1-node water
I'm currently testing if the range is correct for fisheries. So I first checked if the produce the same amount of fish. I created a 1 node water pond - in Settlers 2 (left) the fishery is exhausted without producing any fish, while in RttR there are still some produced:

image

DevOpsOfChaos:
## Summary

- remove fish resources from non-water and isolated one-node water during map resource setup
- keep connected/larger water with fish usable for fisheries
- keep fisher runtime checks simple by relying on normalized resource data
- move resource setup/normalization into `MapLoader`
- add regression coverage for isolated fish water versus connected water

## Motivation

Fisheries could still produce fish from an isolated one-node water pond if that node had a fish resource. In the original behavior described in #1171, such isolated water should be exhausted without producing fish.

This change keeps larger/connected water usable, but removes fish resources from unusable water during map resource setup.

## Implementation details

Resource setup is now handled by `MapLoader::SetupResources(GameWorldBase&)`.

The setup step normalizes resources after map loading:

- converts mine resource types as before
- places/fixes water as before
- removes fish resources from:
  - non-water points
  - isolated one-node water points

The fisher runtime path remains simple and only checks whether fish resources exist.

The fish cleanup iterates row-by-row and uses a row-local shortcut: if the previous point in the same row was kept as fish-on-water, the current water fish point does not need the full neighbor-water scan. This shortcut is not carried across row boundaries.

## Validation

- Ran `Test_integration.exe --run_test=BuildingSuite/FisherIgnoresIsolatedFishWater --report_level=short`
  - Result: passed, 11 assertions
- Ran `Test_integration.exe --run_test=BuildingSuite --report_level=short`
  - Result: passed, 1738 assertions

Fixes #1171


tests/s25Main/integration/testBuilding.cpp
```diff
@@ -476,4 +482,45 @@ BOOST_FIXTURE_TEST_CASE(RoadRemovesObjs, EmptyWorldFixture1P)
...
+{
+    const DescIdx<TerrainDesc> tWater = GetWaterTerrain(world.GetDescription());
+
+    const MapPoint fisheryPos = world.MakeMapPoint(world.GetPlayer(0).GetHQPos() - Position(6, 6));
+    auto* fishery = dynamic_cast<nobUsual*>(
+      BuildingFactory::CreateBuilding(world, BuildingType::Fishery, fisheryPos, 0, Nation::Romans));
+    BOOST_TEST_REQUIRE(fishery);
+
+    nofFisher fisher(fisheryPos, 0, fishery);
+    const nofFarmhand& farmhand = fisher;
```
Flamefire: Why this variable? That just makes the interpretation of `static_cast<unsigned>(farmhand.GetPointQuality(workPt))` below harder

tests/s25Main/integration/testBuilding.cpp
```diff
@@ -476,4 +482,45 @@ BOOST_FIXTURE_TEST_CASE(RoadRemovesObjs, EmptyWorldFixture1P)
...
+        neNode.t1 = tWater;
+        curNode.t1 = tWater;
+        curNode.t2 = tWater;
+        wNode.t2 = tWater;
+    };
+
+    makeWaterPoint(fishPt);
+    world.SetResource(fishPt, Resource(ResourceType::Fish, 4));
+
+    BOOST_TEST_REQUIRE(static_cast<unsigned>(farmhand.GetPointQuality(workPt))
```
Flamefire: Casts are not required as the type already matches, please remove

DevOpsOfChaos:
Updated. I cleaned up the test as requested.

The extra `nofFarmhand` reference is now only kept where it is needed to call `GetPointQuality`, and the unnecessary `unsigned` casts are gone. I also fixed the small formatting/include-order cleanup before pushing.

Validation:

* Built `Test_integration`
* Ran `BuildingSuite/FisherIgnoresIsolatedFishWater`
* Ran `BuildingSuite`

tests/s25Main/integration/testBuilding.cpp
Flamefire: I still don't understand why you don't use `fisher` directly. Can you explain?

DevOpsOfChaos: > I still don't understand why you don't use `fisher` directly. Can you explain?

`nofFisher::GetPointQuality(MapPoint, bool)` is a private override, so calling `fisher.GetPointQuality(workPt)` directly is not accessible / gets hidden on the derived type.

The test uses the public `nofFarmhand` wrapper `GetPointQuality(MapPoint)`, which then dispatches to the fisher implementation. That is why the object is viewed as `nofFarmhand` here.

I can add a short comment or use an inline `static_cast<const nofFarmhand&>(fisher)` if that makes the intent clearer.

Flamefire: I see, makes sense
It looks odd, that it is public at all but it's useful for tests, so I'd say it doesn't hurt.
For this case I would make the method public in nofFisher and use it. It being private seems to go against the LSP.

@Flow86 What is your opinion on this?

DevOpsOfChaos: Done, I made `nofFisher::GetPointQuality(...)` public and updated the test to call the fisher directly.

No behavior change; this only removes the awkward `nofFarmhand` view from the regression test.

Flamefire:
Thinking about this again: Wouldn't it make more sense to "fix" the map during loading?
I.e. remove fish resources in <=1-water places. This would simplify the runtime check to "is there fish?" which then improves performance.
The terrain cannot change at runtime, can it?  so we'd only need to do those checks once.

@Spikeone ?

DevOpsOfChaos:
> Thinking about this again: Wouldn't it make more sense to "fix" the map during loading? I.e. remove fish resources in <=1-water places. This would simplify the runtime check to "is there fish?" which then improves performance. The terrain cannot change at runtime, can it? so we'd only need to do those checks once.
> 
> @Spikeone ?

I reworked this in the direction you suggested.

The isolated/non-water fish cleanup now happens once in `GameWorld::SetupResources()`, so the fisher runtime path can stay simple and only check for fish resources again.

Connected water with fish is unchanged; isolated one-node water and non-water fish resources are normalized away during world resource setup.

Validation:
- built `Test_integration`
- ran `BuildingSuite/FisherIgnoresIsolatedFishWater`
- ran `BuildingSuite`

libs/s25main/world/GameWorld.cpp
```diff
@@ -1375,6 +1376,21 @@ void GameWorld::PlaceAndFixWater()
...
 }
 
+void GameWorld::RemoveUnusableFishResources()
+{
+    RTTR_FOREACH_PT(MapPoint, GetSize())
+    {
+        if(!GetNode(pt).resources.has(ResourceType::Fish))
+            continue;
+
+        if(!IsWaterPoint(pt)
```
Flamefire: This might be slow for large maps, even if just done once. 

Because currently: IsWaterPoint reads 3 nodes and does (up to) 6 checks, each with a memory access.
So for each point with fish we do `6*3+6*6*3=126` checks

I have an idea for a simple optimization: If the previous node has fish, we do not need to check neighbors, do we?

Usually we have bodies of water with fish on consecutive nodes, so we can avoid 108 checks in the best case and have only one check of a local variable more

This certainly works within a row: If (x,y) has fish, then (x+1,y) has a neighbor water point
It doesn't work across rows though: The SE neighbor of the last node in a row is the first node in the next row only for odd rows. So we can't use `RTTR_FOREACH_PT`

Or am I missing anything why this wouldn't work?

Flamefire (changes requested):
Please no force-pushes during a review. If necessary a revert works too. This way I have to check more changes.
Doesn't matter much for this small PR though.

With the added method in `GameWorld`: I think `GameWorld::SetupResources()` belongs more to `MapLoader`. Because that is the only context where it makes sense to call it, doesn't it?
So if I'm not missing anything please move those methods. `SetupResources` can be a public method, and the others can be private.

DevOpsOfChaos:
> Please no force-pushes during a review. If necessary a revert works too. This way I have to check more changes. Doesn't matter much for this small PR though.
> 
> With the added method in `GameWorld`: I think `GameWorld::SetupResources()` belongs more to `MapLoader`. Because that is the only context where it makes sense to call it, doesn't it? So if I'm not missing anything please move those methods. `SetupResources` can be a public method, and the others can be private.

Sorry about the force-push. I’ll avoid force-pushing this PR from here on and use normal follow-up commits instead.

DevOpsOfChaos:
I moved the resource setup logic from `GameWorld` into `MapLoader` as suggested. `MapLoader::SetupResources(GameWorldBase&)` now owns the setup step, with the resource-normalization helpers kept private there.

The fish cleanup behavior is unchanged: fish on non-water and isolated one-node water is removed, while connected/larger water keeps fish. The row-local optimization for consecutive fish water is still in place.

Validation:
- built `Test_integration`
- ran `BuildingSuite/FisherIgnoresIsolatedFishWater`: 11 assertions passed
- ran `BuildingSuite`: 1738 assertions passed

Flamefire (dismissed):

Flow86 (dismissed):
pipeline is failing - did this break exiting replays/savegames?

Flamefire:
> pipeline is failing - did this break exiting replays/savegames?

Some, yes: For replays starting at zero, i.e. with a map as the base it will now remove some fish so fishers may not go out or somewhere else instead.

Double-checking and will update the replays in this branch if required. Maybe in a way that won't break them in similar ways if possible

Flamefire:
Ok, done. The async in those replays is avoided now, I also changed the test to include a replay started from a savegame to cover the 2 different cases that are exposed here: Starting a replay that was started from a save game would have worked

DevOpsOfChaos:
Pushed a small follow-up fix for the CI warning.

The Linux clang build treated the intentionally unused named parameter in the autoplay `MockGameState::SystemChat` override as an error, and clang-tidy surfaced the same diagnostic. The follow-up commit only removes the unused parameter name and keeps the override signature unchanged.

Flamefire:
And hopefully the last one: The Boost version we use on Jenkins doesn't handle output of `std::unique_ptr` yet

Flamefire (approved):

PR #1931, opened 2026-05-01
Validate troop limit commands

DevOpsOfChaos:
## Summary

- validate serialized soldier ranks before executing the affected rank-based game commands
- reject invalid rank values during command deserialization instead of reaching building-state access with unchecked input
- currently this applies to `SetTroopLimit` and `ChangeReserve`, both using `popSoldierRank`
- add regression coverage for invalid serialized rank values and unchanged troop-limit state after rejection

## Motivation

Troop limits and reserves are normally changed through UI paths where rank values are bounded.

However, the affected game commands store the rank as a raw `uint8_t` and deserialize it from the command stream before forwarding it into game-state handling.

Review feedback pointed out that silently ignoring invalid values inside `nobMilitary::SetTroopLimit` is not the preferred style. This version validates the deserialized command value instead, matching the existing approach used for invalid/out-of-range enum values.

## Behavior

- valid rank-based commands keep their existing behavior
- invalid serialized rank values now throw `std::range_error` via `helpers::makeOutOfRange`
- invalid ranks are rejected during deserialization before building state can be changed
- `SetTroopLimit` and `ChangeReserve` now both use `popSoldierRank`
- `nobMilitary::SetTroopLimit` keeps only an internal assert

## Validation

- Ran `Test_integration.exe --run_test=GameCommandSuite --report_level=short`
  - 30 test cases passed
  - 3930 assertions passed
- Ran `Test_integration.exe --run_test=AttackSuite --report_level=short`
  - 23 test cases passed
  - 1262 assertions passed
- Ran full `Test_integration.exe --report_level=short`
  - 194 test cases passed
  - 93363 assertions passed
- Ran `ctest -R "^Test_integration$" --output-on-failure`
  - passed
  - clean


Flamefire:
Even though a limit > soldier count might not look right, it doesn't cause any issues: A large limit is just the same as no limit.

So if this doesn't fix an actual wrong behavior resulting from that I'd keep it, especially as it is virtually impossible to trigger/use the wrong count, so nothing will be gained by the added code.

libs/s25main/buildings/nobMilitary.cpp
```diff
@@ -661,7 +661,10 @@ unsigned nobMilitary::CalcRequiredNumTroops(FrontierDistance assumedFrontierDist
 
 void nobMilitary::SetTroopLimit(const unsigned rank, const unsigned limit)
 {
-    troop_limits[rank] = limit;
+    if(rank >= NUM_SOLDIER_RANKS)
+        return;
```
Flow86: well, this part is the only thing I would think i useful. Accessing a `rank` outside the `troop_limits` range is UB.
but may be I would generally say to use a `RTTR_Assert` here instead (to "crash" the game if someone manipulates that)

Flamefire: Sure this could have an assert like `rank < troop_limits.size()` but then: Why here and not in one of the 100s other places?

IIRC MSVC already does range checks in debug mode which (partially) covers that.

DevOpsOfChaos:
> Even though a limit > soldier count might not look right, it doesn't cause any issues: A large limit is just the same as no limit.
> 
> So if this doesn't fix an actual wrong behavior resulting from that I'd keep it, especially as it is virtually impossible to trigger/use the wrong count, so nothing will be gained by the added code.

Good point. I removed the count clamp.

Large troop-limit values now keep the existing behavior. The PR only keeps the invalid-rank guard before indexing the troop-limit array.

libs/s25main/buildings/nobMilitary.cpp
DevOpsOfChaos: I kept this as a runtime guard instead of assert-only because the value comes through the game command path, not just a direct UI-only call.

An `RTTR_Assert(rank < NUM_SOLDIER_RANKS)` would be fine as an additional debug check, but assert-only would still leave release builds indexing out of range if a malformed command gets through.

So my intention here is not to introduce broad validation everywhere, only to avoid UB at this command boundary. If you prefer the project style to be assert + return here, I can adjust it that way.

Flamefire: > So my intention here is not to introduce broad validation everywhere, only to avoid UB at this command boundary.

If the intention is **not** to introduce it everywhere (which I wouldn't do anyway) then the question is, why here? What makes this place special?

So unless there is some actual situation where this is an issue I'd just leave it as-is.

Otherwise: Simply ignoring the caller isn't great either. So if we would use a runtime-check then I'd simply use `troop_limits.at(rank) = limit;`
But then we'd need to basically do this everywhere just for assuming some issue that never happened before.

DevOpsOfChaos: I checked the path again.

The normal UI path seems bounded: the military building window only creates rank values from controls in the `0..NUM_SOLDIER_RANKS - 1` range.

The reason I considered this command boundary special is that `SetTroopLimit` is also deserialized from the game-command stream: the command stores `rank` as a raw `uint8_t`, `SetTroopLimit(Serializer& ser)` reads it via `PopUnsignedChar()`, and `Execute()` forwards it directly to `nobMilitary::SetTroopLimit(rank, count)`.

So the current guard is not meant to handle normal UI-generated input, but to avoid UB if a malformed command stream reaches this boundary.

That said, I do not currently have a reproduced normal-gameplay path where such an invalid rank is generated. If that makes the guard too speculative for this codebase, I’m fine with leaving the current behavior unchanged.

Flamefire: IMO there are 3 better paths:
- RTTR_Assert
- `.at()`
- Validate and throw when deserializing the GC.

The last one is more in line with what we do already: E.g. invalid/out-of-range enum values cause an exception.
It might then make sense to check the other commands if they have similar, critical issues. I.e. not just "might cause odd results", but actual, always-wrong OOB accesses

Flamefire:
See my [prior comment](https://github.com/Return-To-The-Roots/s25client/pull/1931#discussion_r3247009442)

DevOpsOfChaos:
Updated this PR according to the deserialization-boundary feedback.

The invalid rank is no longer silently ignored in `nobMilitary::SetTroopLimit`. Instead, `gc::SetTroopLimit` now validates the deserialized rank before `Execute()` can call into the building code.

Invalid serialized ranks now throw `std::range_error` via `helpers::makeOutOfRange`, matching the existing style used for invalid enum deserialization. `nobMilitary::SetTroopLimit` keeps only an internal assert, and large valid troop-limit counts remain accepted unchanged.

Validation:

- built `Test_integration`
- `Test_integration.exe --run_test=GameCommandSuite/SetTroopLimitRejectsInvalidSerializedRank --report_level=short`
  - 1 test case
  - 1 assertion
- `Test_integration.exe --run_test=AttackSuite --report_level=short`
  - 23 test cases
  - 1244 assertions
  - clean

libs/s25main/buildings/nobMilitary.cpp
```diff
@@ -696,6 +696,7 @@ unsigned nobMilitary::CalcRequiredNumTroops(FrontierDistance assumedFrontierDist
 
 void nobMilitary::SetTroopLimit(const unsigned rank, const unsigned limit)
 {
+    RTTR_Assert(rank <= MAX_MILITARY_RANK);
```
Flamefire: I'd say this is clearer:
```suggestion
    RTTR_Assert(rank <= troop_limits.size());
```

libs/s25main/GameCommands.h
```diff
@@ -25,6 +26,16 @@ class GameWorld;
 
 namespace gc {
 
+namespace detail {
+    inline uint8_t popTroopLimitRank(Serializer& ser)
```
Flamefire: Are there other places where this can be used?

And a more generic name makes sense. It isn't a "troop limit" rank but simply a "soldier rank"
```suggestion
    inline uint8_t popSoldierRank(Serializer& ser)
```

Flamefire (changes requested):

DevOpsOfChaos:
Applied the two review suggestions with one correction.

- renamed the helper to `popSoldierRank`
- kept the validation scoped to `SetTroopLimit`
- updated the internal assert to `rank < troop_limits.size()`

I used `<` instead of the suggested `<=` because `troop_limits[rank]` uses 0-based indexing, so `rank == troop_limits.size()` would already be out of range.

I also noticed `ChangeReserve` deserializes a soldier rank directly, but I left it unchanged here to avoid broadening this PR beyond the reviewed `SetTroopLimit` command path.

tests/s25Main/integration/testAttacking.cpp
```diff
@@ -418,6 +418,18 @@ BOOST_FIXTURE_TEST_CASE(StartAttack, AttackFixture<>)
...
+BOOST_FIXTURE_TEST_CASE(TroopLimitCommandRejectsInvalidSerializedRank, AttackFixture<>)
+{
+    const unsigned maxTroops = milBld0->GetMaxTroopsCt();
+    const unsigned limit = maxTroops + 10;
+
+    this->SetTroopLimit(milBld0Pos, 0, limit);
+    BOOST_TEST_REQUIRE(milBld0->GetTroopLimit(0) == limit);
+
+    BOOST_REQUIRE_THROW(this->SetTroopLimit(milBld0Pos, NUM_SOLDIER_RANKS, 0), std::range_error);
+    BOOST_TEST_REQUIRE(milBld0->GetTroopLimit(0) == limit);
```
Flamefire: This test seems incomplete: Why test the limit for (only) `0`? 
Shouldn't the test be:
- Use invalid command
- Assert that nothing changed

And instead of a fixed initial value(s) it could use random values to avoid reliance on some specific one unless testing for a specific corner case.

tests/s25Main/integration/testGameCommands.cpp
```diff
@@ -55,6 +57,17 @@ static void dummySuppressUnused(std::ostream& out)
 
 BOOST_AUTO_TEST_SUITE(GameCommandSuite)
 
+BOOST_AUTO_TEST_CASE(SetTroopLimitRejectsInvalidSerializedRank)
+{
+    gc::Deserializer ser;
+    helpers::pushEnum<uint8_t>(ser, gc::GCType::SetTroopLimit);
+    helpers::pushPoint(ser, MapPoint(0, 0));
+    ser.PushUnsignedChar(static_cast<uint8_t>(MAX_MILITARY_RANK + 1u));
```
Flamefire: This test seems odd: Why test for essentially the same thing as `TroopLimitCommandRejectsInvalidSerializedRank` but with `+1`?

Do we need this test at all or does it provide any additional benefit?

Flamefire (changes requested):
> I also noticed ChangeReserve deserializes a soldier rank directly, but I left it unchanged here to avoid broadening this PR beyond the reviewed SetTroopLimit command path.

I'd be fine with using the new  function in all command deserializations where it applies, at least for consistency, possibly  without (new) tests. In the future we might otherwise wonder why it wasn't used.
Of course just where it is obviously correct, i.e. also had the same `rank(ser.PopUnsignedChar())`

DevOpsOfChaos:
Applied the remaining review feedback.

- `ChangeReserve` now uses the same `popSoldierRank` validation helper as `SetTroopLimit`; these are the only directly deserialized command fields that are clearly soldier ranks.
- Consolidated the invalid-rank coverage: the deserialize test now covers both rank-based commands.
- Strengthened the military-state regression test with deterministic, distinct initial troop limits for every rank and verifies that an invalid command leaves every value unchanged.
- Removed the redundant near-duplicate invalid-rank coverage.

Validation passed locally:

- `GameCommandSuite`: 30 test cases / 3930 assertions
- `AttackSuite`: 23 test cases / 1262 assertions
- full `Test_integration`: 194 test cases / 93363 assertions

I also checked the previously reported GitHub Actions failure. The Linux test job itself completed with `100% tests passed, 0 tests failed out of 20`; the later failure was the Codecov upload step reporting that its signature could not be verified. That appears unrelated to this PR's test behavior.

tests/s25Main/integration/testAttacking.cpp
```diff
@@ -418,6 +418,25 @@ BOOST_FIXTURE_TEST_CASE(StartAttack, AttackFixture<>)
     TestFailingAttack(gwv, milBld1Pos, attackSrc, 1u);
 }
 
+BOOST_FIXTURE_TEST_CASE(TroopLimitCommandRejectsInvalidSerializedRank, AttackFixture<>)
+{
+    std::array<unsigned, NUM_SOLDIER_RANKS> expectedLimits{};
+    for(unsigned rank = 0; rank < expectedLimits.size(); ++rank)
+    {
+        expectedLimits[rank] = (rank + 1) * 3;
```
Flamefire: Instead of using magic numbers in test, please use random numbers, i.e. `rttr::test::random*`
That a) shows that the value doesn't matter for the test and b) possibly increases test coverage by catching cases missed in explicit tests. Similar the `+7` below: You could add a (small) random number or use a completely new random one.

tests/s25Main/integration/testGameCommands.cpp
```diff
@@ -55,6 +57,21 @@ static void dummySuppressUnused(std::ostream& out)
...
 BOOST_AUTO_TEST_SUITE(GameCommandSuite)
 
+BOOST_AUTO_TEST_CASE(SoldierRankCommandsRejectInvalidSerializedRank)
+{
+    const gc::GCType commands[] = {gc::GCType::SetTroopLimit, gc::GCType::ChangeReserve};
+    for(const auto command : commands)
+    {
+        gc::Deserializer ser;
+        helpers::pushEnum<uint8_t>(ser, command);
+        helpers::pushPoint(ser, MapPoint(0, 0));
```
Flamefire: `randomPoint` can be used. But I'm not a fan of this test basically copying the serialize function of a class. Why not use that class and call serialize on it?

Flamefire (changes requested):
Looks good now, only 2 minor things in the test.

> Consolidated the invalid-rank coverage: the deserialize test now covers both rank-based commands.

How so? Because both commands would use the same function that triggers the error? Or am I missing something in the tests?

tests/s25Main/integration/testAttacking.cpp
DevOpsOfChaos: Updated this test to use bounded random values: `firstLimit` is in [1, 20] and `limitStep` in [1, 5]. With the five military ranks this guarantees a positive step, no unsigned overflow, and distinct initial limits. I also added an explicit uniqueness check so the test fails immediately if that setup invariant ever changes.

tests/s25Main/integration/testGameCommands.cpp
DevOpsOfChaos: Updated this test to create the commands through the regular factory API, serialize them through the real command `Serialize()` implementation, and deserialize them through `gc::GameCommand::Deserialize`. The invalid rank is now passed through the command API rather than being assembled manually in a raw serializer payload.

tests/s25Main/integration/testGameCommands.cpp
```diff
@@ -53,8 +56,50 @@ static void dummySuppressUnused(std::ostream& out)
 }
 // LCOV_EXCL_STOP
 
+namespace {
+class CommandCaptureFactory : public GameCommandFactory
```
Flamefire: This looks obscure and overly complex.
Why was this added?

tests/s25Main/integration/testGameCommands.cpp
```diff
@@ -53,8 +56,50 @@ static void dummySuppressUnused(std::ostream& out)
...
+    factory.lastCommand->Serialize(ser);
+    return ser;
+}
+} // namespace
+
 BOOST_AUTO_TEST_SUITE(GameCommandSuite)
 
+BOOST_AUTO_TEST_CASE(SoldierRankCommandsRejectInvalidSerializedRank)
+{
+    const auto invalidRank = static_cast<unsigned char>(MAX_MILITARY_RANK + 1u);
```
Flamefire:
```suggestion
    constexpr auto invalidRank = static_cast<uint8_t>(MAX_MILITARY_RANK + 1u);
```
If you need/want the cast at all

tests/s25Main/integration/testGameCommands.cpp
```diff
@@ -53,8 +56,50 @@ static void dummySuppressUnused(std::ostream& out)
...
+    const auto pt = rttr::test::randomPoint<MapPoint>(0, 20);
+    const auto count = rttr::test::randomValue<unsigned>(0u, 100u);
```
Flamefire: Do the ranges matter?

```suggestion
    const auto pt = rttr::test::randomPoint<MapPoint>();
    const auto count = rttr::test::randomValue<unsigned>();
```

tests/s25Main/integration/testGameCommands.cpp
```diff
@@ -53,8 +56,50 @@ static void dummySuppressUnused(std::ostream& out)
...
+    auto serializedCommands = std::array{
```
Flamefire: Just create the commands directly

tests/s25Main/integration/testAttacking.cpp
```diff
@@ -418,6 +419,31 @@ BOOST_FIXTURE_TEST_CASE(StartAttack, AttackFixture<>)
...
+    std::array<unsigned, NUM_SOLDIER_RANKS> expectedLimits{};
+    const auto firstLimit = rttr::test::randomValue<unsigned>(1u, 20u);
+    const auto limitStep = rttr::test::randomValue<unsigned>(1u, 5u);
+    for(unsigned rank = 0; rank < expectedLimits.size(); ++rank)
+    {
+        expectedLimits[rank] = firstLimit + rank * limitStep;
+        const bool isUniqueLimit =
+          std::find(expectedLimits.begin(), expectedLimits.begin() + rank, expectedLimits[rank])
+          == expectedLimits.begin() + rank;
+        BOOST_TEST_REQUIRE(isUniqueLimit);
```
Flamefire:
```suggestion
        // Use unique values to detect overwriting
        expectedLimits[rank] = firstLimit + rank * limitStep;
```

tests/s25Main/integration/testAttacking.cpp
```diff
@@ -418,6 +419,31 @@ BOOST_FIXTURE_TEST_CASE(StartAttack, AttackFixture<>)
     TestFailingAttack(gwv, milBld1Pos, attackSrc, 1u);
 }
 
+BOOST_FIXTURE_TEST_CASE(TroopLimitCommandRejectsInvalidSerializedRank, AttackFixture<>)
+{
+    std::array<unsigned, NUM_SOLDIER_RANKS> expectedLimits{};
+    const auto firstLimit = rttr::test::randomValue<unsigned>(1u, 20u);
+    const auto limitStep = rttr::test::randomValue<unsigned>(1u, 5u);
```
Flamefire:
```suggestion
    const auto firstLimit = rttr::test::randomValue(1u, 20u);
    const auto limitStep = rttr::test::randomValue(1u, 5u);
```

Flamefire (changes requested):
> I also added an explicit uniqueness check so the test fails immediately if that setup invariant ever changes.

While not wrong I think this is where your AI overshot the target:
a) The uniqueness isn't strictly required, fully random values would do
b) There is no invariant as far as I can see. If you want to state that unique values shall be used then rather state the reason than adding complexity

See inline suggestions

tests/s25Main/integration/testGameCommands.cpp
DevOpsOfChaos: You’re right. The helper was added to avoid manually reproducing the serialized payload after the earlier feedback, but it makes this small test needlessly indirect.

To be transparent, I did not independently review the AI-assisted follow-up to the standard it needed before posting it. I’ll simplify the test to construct and serialize the concrete commands directly, apply the remaining test-cleanup suggestions, and validate the relevant suites before pushing an update. Thanks for the detailed review.

DevOpsOfChaos: Updated in 567045a.

The test now serializes the two concrete command types directly through their real `Serialize()` implementations. The small local test types only expose the protected constructors; the former capture factory and generic serialization helper are gone.

Validated with the focused command and troop-limit tests, both full suites, and `ctest`.

tests/s25Main/integration/testGameCommands.cpp
```diff
@@ -53,8 +56,39 @@ static void dummySuppressUnused(std::ostream& out)
 }
 // LCOV_EXCL_STOP
 
+namespace {
```
Flamefire, line 59: What is the point of those classes?

Flamefire (changes requested):
> To be transparent, I did not independently review the AI-assisted follow-up to the standard it needed before posting it.

That really is an issue for us. We currently get a lot of AI assisted/generated PRs but AI is still bad at understanding such a complex code base and producing clean, maintainable code.
We work on this project in our spare time so if it feels like we spend too much time on AI slop that hasn't been cleaned up before submission we'll need to ban use of AI as other projects already do.

So please help us here and make sure the changes are clean and reviewed before pushing. It would be bad if we couldn't profit from the increased speed enabled by AI due to its shortcomings.

tests/s25Main/integration/testGameCommands.cpp
DevOpsOfChaos: They only expose the protected constructors of `SetTroopLimit` and `ChangeReserve`.

I added them so the test could call the concrete commands’ real `Serialize()` implementations instead of manually reproducing their serialized representation. I reviewed that approach as a way to avoid the earlier capture helper, but it is still an extra workaround solely for constructor access.

I agree that this does not make the test as direct or clear as intended. I will revisit the test instead of defending the additional types.

DevOpsOfChaos:
> > To be transparent, I did not independently review the AI-assisted follow-up to the standard it needed before posting it.
> 
> That really is an issue for us. We currently get a lot of AI assisted/generated PRs but AI is still bad at understanding such a complex code base and producing clean, maintainable code. We work on this project in our spare time so if it feels like we spend too much time on AI slop that hasn't been cleaned up before submission we'll need to ban use of AI as other projects already do.
> 
> So please help us here and make sure the changes are clean and reviewed before pushing. It would be bad if we couldn't profit from the increased speed enabled by AI due to its shortcomings.

You are right to raise this.

My earlier wording was imprecise: I did review the generated follow-up and validate it locally before pushing it, but I did not review the resulting test structure deeply enough against the existing codebase patterns. That distinction does not change the unnecessary review work it caused.

I understand that AI assistance is only useful here when it reduces maintainer effort rather than shifting cleanup work to reviewers. I will treat generated output as a draft, review the final diff and surrounding project patterns more critically, and only push follow-ups that I can clearly explain and justify. Sorry for the unnecessary churn.

tests/s25Main/integration/testGameCommands.cpp
Flamefire: Double-checked: I forgot that those ctors are protected to enforce only creating those in the factories.
TBH: I don't see a better way than this one or the previous factory-override, although that might have been possible in a slightly cleaner way.

I'm fine with either way now as using the serialize method of them is certainly best.
If you prefer the earlier version feel free to revert. otherwise we can merge as-is, but maybe add a comment that this is (only) to make the ctors public for tests to avoid future confusion

Flamefire (approved):

PR #1933, opened 2026-05-01
Remove empty replay recordings

### LINKED ISSUE #984: Do not record 0 GF replays
When testing LUA scripts you have to start the game quite often - if the script crashes right at the begining (loading routine for example) you create a 0 GF replay

image

would be nice if you could set a limit, e.g. keep no replays under 25000GF or something like that.

DevOpsOfChaos:
## Summary

- remove replay files when recording stops without any recorded game frame
- keep normal replay recording unchanged once the replay has a non-zero last game frame
- add regression coverage for empty replay recording cleanup

## Motivation

Starting a game and aborting during early setup, for example after a script error, can leave behind 0-GF replay files. These files are not useful as actual replays and clutter the replay folder.

This keeps completed/non-empty replay recordings unchanged, but removes empty recordings when recording is stopped cleanly.

Fixes #984

## Validation

- Built `Test_integration` locally with Visual Studio 2022 Debug
- Ran `Serialization/EmptyReplayRecordingIsRemoved`
- Ran `Serialization/ReplayWithMap`
- Ran `Serialization`

### DIFF
--- a/libs/s25main/Replay.cpp
+++ b/libs/s25main/Replay.cpp
@@ -72,6 +72,15 @@ bool Replay::StopRecording()
     isRecording_ = false;
     file_.Close();
 
+    // Remove empty replay recordings. They are mostly produced when a game is aborted
+    // during startup, for example after an early script error.
+    if(lastGF_ == 0)
+    {
+        boost::system::error_code ec;
+        boost::filesystem::remove(filepath_, ec);
+        return !ec;
+    }
+
     BinaryFile file;
     if(!file.Open(filepath_, OpenFileMode::Read))
         return false;
--- a/tests/s25Main/integration/testSerialization.cpp
+++ b/tests/s25Main/integration/testSerialization.cpp
@@ -398,6 +398,26 @@ struct ReplayMapFixture
     }
 };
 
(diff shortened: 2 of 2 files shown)


Flow86 (approved):
sounds reasonable to me. cleaning up clutter is always good

Spikeone:
Doesn't this potentially also mean, that if during game setup something wents wrong, we won't be able to find this by loading a replay? Maybe there are some special addon combinations that would break the game on startup so no GF is run. Not sure so guess you know better @Flamefire

Flamefire:
I'd say in those cases we could as well start the game from whatever was chosen there.
So far we never needed that but you were annoyed by those 0 GF replays. So I'd use this until we see actual uses where those replays would help.

DevOpsOfChaos:
> I'd say in those cases we could as well start the game from whatever was chosen there. So far we never needed that but you were annoyed by those 0 GF replays. So I'd use this until we see actual uses where those replays would help.

Thanks, that matches my reasoning as well.

I mainly wanted to remove the useless 0-GF replay clutter for cleanly stopped recordings. If we later find a real debugging case where such empty replays are useful, we can revisit this, but for now I would keep the cleanup as-is.

PR #1935, opened 2026-05-01
Store received savegames in save folder

### LINKED ISSUE #1846: some sav files are stored in ~/.s25rttr/MAPS/
I just had a look into the dir where the userdata is stored, which under linux is .s25rttr in /home/user/ ; to delete old logs and reviews mostly - and I noticed a lot of *.sav files in the MAPS folder: 

cd ~/.s25rttr/MAPS/ 

ls -ls | grep sav

```
336 -rw-r--r-- 1 user user 340003 30. Apr 2023  attacke.sav
 52 -rw-r--r-- 1 user user  52033 25. Mär 2023  1.sav
304 -rw-r--r-- 1 user user 309288 26. Mär 2023  123.sav
 96 -rw-r--r-- 1 user user  96982 20. Nov 08:52 2.sav
 64 -rw-r--r-- 1 user user  61874 16. Jun 2024  222.sav
 60 -rw-r--r-- 1 user user  59075 16. Jun 2024  2222222.sav
 64 -rw-r--r-- 1 user user  62145 16. Jun 2024  2222222222.sav
120 -rw-r--r-- 1 user user 119924 20. Nov 12:39 3.sav
 56 -rw-r--r-- 1 user user  53844 15. Jun 2024  33.sav
 64 -rw-r--r-- 1 user user  63540 16. Jun 2024  333.sav
132 -rw-r--r-- 1 user user 131674 20. Nov 12:58 4.sav
124 -rw-r--r-- 1 user user 122967 19. Nov 16:33 7.sav
380 -rw-r--r-- 1 user user 386765 28. Mär 2023  Archipelag Sumary (Automatisches Speichern).sav
380 -rw-r--r-- 1 user user 386765 28. Mär 2023  Archipelag_Sumary__Automatisches_Speichern_.sav
 92 -rw-r--r-- 1 user user  90614 15. Nov 09:26 Australien (Automatisches Speichern).sav
 92 -rw-r--r-- 1 user user  90614 15. Nov 09:26 Australien__Automatisches_Speichern_.sav
344 -rw-r--r-- 1 user user 348259  1. Mai 2023  Europa (Automatisches Speichern).sav
344 -rw-r--r-- 1 user user 348259  1. Mai 2023  Europa__Automatisches_Speichern_.sav
132 -rw-r--r-- 1 user user 134394 20. Nov 14:41 Glueck.sav
 52 -rw-r--r-- 1 user user  49844  9. Okt 14:28 Japan (Automatisches Speichern).sav
 52 -rw-r--r-- 1 user user  49844  9. Okt 14:28 Japan__Automatisches_Speichern_.sav
548 -rw-r--r-- 1 user user 557153 29. Nov 09:46 Migration (Automatisches Speichern).sav
548 -rw-r--r-- 1 user user 557153 29. Nov 09:46 Migration__Automatisches_Speichern_.sav
116 -rw-r--r-- 1 user user 117373 20. Nov 10:24 Nordasien (Automatisches Speichern).sav
116 -rw-r--r-- 1 user user 117373 20. Nov 10:24 Nordasien__Automatisches_Speichern_.sav
 88 -rw-r--r-- 1 user user  87251  8. Nov 2023  Totes Land (Automatisches Speichern).sav
 88 -rw-r--r-- 1 user user  87251  8. Nov 2023  Totes_Land__Automatisches_Speichern_.sav
144 -rw-r--r-- 1 user user 145593 18. Jun 2024  V - Im Ödland (Automatisches Speichern).sav
144 -rw-r--r-- 1 user user 145593 18. Jun 2024  V_-_Im___dland__Automatisches_Speichern_.sav
 68 -rw-r--r-- 1 user user  68681 18. Okt 12:42 Weihnachtsinseln (Automatisches Speichern).sav
 68 -rw-r--r-- 1 user user  68681 25. Feb 2025  Weihnachtsinseln__Automatisches_Speichern_.sav
336 -rw-r--r-- 1 user user 340003 30. Apr 2023  _attacke.sav
516 -rw-r--r-- 1 user user 527733 25. Nov 19:41 attacke.sav
208 -rw-r--r-- 1 user user 212796 26. Mär 2023  der  Beginn.sav
208 -rw-r--r-- 1 user user 212796 26. Mär 2023  der__Beginn.sav
132 -rw-r--r-- 1 user user 134721 24. Nov 07:41 overun.sav
520 -rw-r--r-- 1 user user 532097 30. Nov 08:28 sav
336 -rw-r--r-- 1 user user 341720 26. Mär 2023  testrun.sav

```

DevOpsOfChaos:
## Summary

- store received savegame transfers in the savegame folder instead of the played-maps folder
- keep regular map and Lua transfers in the played-maps folder
- add focused regression coverage for received savegame path selection

## Motivation

Received map data is currently written to the played-maps folder unconditionally. This also affects `MapType::Savegame`, leaving `.sav` files in `MAPS`.

Savegame transfers should not be stored as played maps.

Fixes #1846

## Validation

- Built `Test_network` locally with VS2022 Debug
- Ran `GameClientTests/ClientStoresReceivedSavegamesInSaveFolder`
- Ran `GameClientTests`

### DIFF
--- a/tests/s25Main/network/testGameClient.cpp
+++ b/tests/s25Main/network/testGameClient.cpp
@@ -244,6 +244,36 @@ BOOST_DATA_TEST_CASE(ClientFollowsConnectProtocol, usesLuaScriptValues, usesLuaS
     }
 }
 
+BOOST_AUTO_TEST_CASE(ClientStoresReceivedSavegamesInSaveFolder)
+{
+    rttr::test::TmpFolder testUserData;
+    rttr::test::ConfigOverride userDataOverride("USERDATA", testUserData);
+
+    GameClient client;
+    GameMessageInterface& clientMsgInterface = client;
+    TestServer server;
+    const auto serverPort = server.tryListen();
+    BOOST_TEST_REQUIRE(serverPort >= 0);
+
+    BOOST_TEST_REQUIRE(client.Connect("localhost", rttr::test::randString(10), rttr::test::randomEnum<ServerType>(),
+                                      serverPort, false, false));
+    clientMsgInterface.OnGameMessage(GameMessage_Player_Id(1));
+    client.GetMainPlayer().sendQueue.clear();
+    clientMsgInterface.OnGameMessage(GameMessage_Server_TypeOK(GameMessage_Server_TypeOK::StatusCode::Ok, ""));
+    client.GetMainPlayer().sendQueue.clear();
+    clientMsgInterface.OnGameMessage(GameMessage_Server_Password("true"));
+    client.GetMainPlayer().sendQueue.clear();
+
+    const auto expectedSavePath = RTTRCONFIG.ExpandPath(s25::folders::save) / "received.sav";
+
+    clientMsgInterface.OnGameMessage(GameMessage_Map_Info("received.sav", MapType::Savegame, 1, 1, 0, 0));
+    const auto msg = boost::dynamic_pointer_cast<GameMessage_MapRequest>(client.GetMainPlayer().sendQueue.pop());
+    BOOST_TEST_REQUIRE(msg);
+    BOOST_TEST(!msg->requestInfo);
+    BOOST_TEST(client.GetMapType() == MapType::Savegame);
+    BOOST_TEST(client.GetMapPath() == expectedSavePath);
+}
+
 BOOST_AUTO_TEST_CASE(ClientDetectsMapBufferOverflow)
 {
     rttr::test::LogAccessor _suppressLogOutput;
--- a/libs/s25main/network/GameClient.cpp
+++ b/libs/s25main/network/GameClient.cpp
@@ -928,7 +928,10 @@ bool GameClient::OnGameMessage(const GameMessage_Map_Info& msg)
         OnError(ClientError::InvalidMap);
         return true;
     }
-    mapinfo.filepath = RTTRCONFIG.ExpandPath(s25::folders::mapsPlayed) / portFilename;
+    const auto targetPath =
+      RTTRCONFIG.ExpandPath((msg.mt == MapType::Savegame) ? s25::folders::save : s25::folders::mapsPlayed);
+    bfs::create_directories(targetPath);
+    mapinfo.filepath = targetPath / portFilename;
     mapinfo.type = msg.mt;
 
     // lua script file path


Flow86 (dismissed):
I like the solution. simple and effective

DevOpsOfChaos:
Tightened the implementation and regression test.

The target folder is now created before assigning the received map/savegame path, so savegame transfers do not depend on the SAVE directory already existing.

The regression test now sends actual compressed transfer data through `GameMessage_Map_Data` and verifies that the received savegame payload is written to the save folder and not to the played-maps folder. The transferred payload intentionally uses existing map test data, so the client reports an error after the storage path has been verified.

Validation:
- Built `Test_network` locally with VS2022 Debug
- Ran `GameClientTests/ClientStoresReceivedSavegamesInSaveFolder`
- Ran `GameClientTests/ClientFollowsConnectProtocol`
- Ran `Test_network`

tests/s25Main/network/testGameClient.cpp
```diff
@@ -244,6 +244,65 @@ BOOST_DATA_TEST_CASE(ClientFollowsConnectProtocol, usesLuaScriptValues, usesLuaS
     }
 }
 
+BOOST_AUTO_TEST_CASE(ClientStoresReceivedSavegamesInSaveFolder)
+{
+    GameClient client;
+    GameMessageInterface& clientMsgInterface = client;
+    MockClientInterface callbacks;
+    client.SetInterface(&callbacks);
```
Flamefire: Is this required for this test? The previous test already does that, so this one can focus on the save game path.

tests/s25Main/network/testGameClient.cpp
```diff
@@ -244,6 +244,65 @@ BOOST_DATA_TEST_CASE(ClientFollowsConnectProtocol, usesLuaScriptValues, usesLuaS
...
+    const boost::filesystem::path testMapPath =
+      rttr::test::rttrBaseDir / "tests" / "testData" / "maps" / "LuaFunctions.SWD";
+    MapInfo mapInfo;
+    mapInfo.mapData.CompressFromFile(testMapPath, &mapInfo.mapChecksum);
+    const auto mapDataSize = mapInfo.mapData.data.size();
+
+    clientMsgInterface.OnGameMessage(GameMessage_Map_Info("received.sav", MapType::Savegame,
+                                                          mapInfo.mapData.uncompressedLength, mapDataSize, 0, 0));
+
+    const auto expectedSavePath = RTTRCONFIG.ExpandPath(s25::folders::save) / "received.sav";
```
Flamefire, line 267: If I understand this correctly, this will modify the current users $HOME, won't it?
Use `rttr::test::ConfigOverride` to redirect it

Flamefire (changes requested):
Should be ok, not sure if we should overwrite the users savegames with that received in multiplayer, but I guess that's rare enough

DevOpsOfChaos:
Thanks for the review.

Good point. I’ll tighten the test so it focuses on the savegame path instead of duplicating the full connect protocol already covered by `ClientFollowsConnectProtocol`.

I’ll also make the path isolation explicit with `rttr::test::ConfigOverride` / a test-owned temp folder, so the test cannot write to the user’s real save/map folders.

The production change should stay unchanged; this should only need a test cleanup.

DevOpsOfChaos:
Addressed both review points.

The savegame-path test now uses an explicit `TmpFolder` plus `rttr::test::ConfigOverride("USERDATA", ...)`, and it asserts that both the expected SAVE path and the unexpected MAPS path are under that test-owned folder.

I also moved the minimal connection setup into `AdvanceClientToMapInfoRequest()` so the test no longer reasserts the full connect protocol already covered by `ClientFollowsConnectProtocol`. The production change is unchanged.

Validation:

* Built `Test_network` locally with VS2022 Debug
* Ran `GameClientTests/ClientStoresReceivedSavegamesInSaveFolder`
* Ran `GameClientTests`

Flamefire:
> I also moved the minimal connection setup into `AdvanceClientToMapInfoRequest()` so the test no longer reasserts the full connect protocol already covered by `ClientFollowsConnectProtocol`. The production change is unchanged.

This is still done, isn't it? `MockClientInterface` doesn't seem necessary and the `AdvanceClientToMapInfoRequest` overkill

DevOpsOfChaos:
Updated again.

I removed the helper and the `MockClientInterface` setup. The test now only brings the client to `QueryMapInfo`, because `GameMessage_Map_Info` is ignored before that state.

After that it only checks the savegame path handling:

* savegames go to `SAVE`
* the played-maps path is not used
* both paths are under the test `USERDATA` override

Validation:

* Built `Test_network`
* Ran `GameClientTests/ClientStoresReceivedSavegamesInSaveFolder`
* Ran `GameClientTests`

Flamefire (dismissed):

tests/s25Main/network/testGameClient.cpp
```diff
@@ -244,6 +244,37 @@ BOOST_DATA_TEST_CASE(ClientFollowsConnectProtocol, usesLuaScriptValues, usesLuaS
...
+    client.GetMainPlayer().sendQueue.clear();
+    clientMsgInterface.OnGameMessage(GameMessage_Server_Password("true"));
+    client.GetMainPlayer().sendQueue.clear();
+
+    clientMsgInterface.OnGameMessage(GameMessage_Map_Info("received.sav", MapType::Savegame, 0, 0, 0, 0));
+
+    const auto expectedSavePath = RTTRCONFIG.ExpandPath(s25::folders::save) / "received.sav";
+    const auto unexpectedMapsPath = RTTRCONFIG.ExpandPath(s25::folders::mapsPlayed) / "received.sav";
+    BOOST_TEST(expectedSavePath.parent_path() == testUserData / "SAVE");
+    BOOST_TEST(unexpectedMapsPath.parent_path() == testUserData / "MAPS");
```
Flow86: this doesnt test anything besides filesystem methods.

I think it should be a `.exists()` vs not `.exists()`?

DevOpsOfChaos: Updated the test to send actual transfer data again and check that the received savegame file exists in `SAVE` but not in `MAPS`.

Flamefire: These 2 tests are only to verify the `ExpandPath` works as expected with the override.

The `exists` is actually not required. IMO we can stop at `BOOST_TEST(client.GetMapPath() == expectedSavePath);`

Or do we actually want/need to transfer something and test files?
If not we don't even need to load and compress anything which makes the test much smaller and faster.

Flow86:
now basically only a proper git squash/rebase is necessary to clean the git history

tests/s25Main/network/testGameClient.cpp
```diff
@@ -244,6 +244,44 @@ BOOST_DATA_TEST_CASE(ClientFollowsConnectProtocol, usesLuaScriptValues, usesLuaS
...
+    clientMsgInterface.OnGameMessage(GameMessage_Player_Id(1));
+    client.GetMainPlayer().sendQueue.clear();
+    clientMsgInterface.OnGameMessage(GameMessage_Server_TypeOK(GameMessage_Server_TypeOK::StatusCode::Ok, ""));
+    client.GetMainPlayer().sendQueue.clear();
+    clientMsgInterface.OnGameMessage(GameMessage_Server_Password("true"));
+    client.GetMainPlayer().sendQueue.clear();
+
+    const auto expectedSavePath = RTTRCONFIG.ExpandPath(s25::folders::save) / "received.sav";
+
+    clientMsgInterface.OnGameMessage(GameMessage_Map_Info("received.sav", MapType::Savegame, 1, 1, 0, 0));
```
Flow86, line 269: do we really need a server part? basically the mocked game messages should be enough?

DevOpsOfChaos: The `TestServer` is only used so `client.Connect(...)` succeeds and the client reaches `QueryMapInfo` through the normal public flow.

After that, the test only uses mocked incoming game messages. Without the small server setup, I don’t see a clean public way to reach that state without adding test-only access to the client’s private connect state.

Flamefire:
> now basically only a proper git squash/rebase is necessary to clean the git history

@Flow86 Can you enable the squash-merge feature for this repo? IMO it is better to use that, than requiring force-pushes for (very) small changes where a single commit makes sense.

For others an update is not (always) necessary and could be done by a merge too. As mentioned elsewhere this is quite common and doesn't break reviews, so I'd prefer that if updating at all.

Flamefire (approved):

PR #1937, opened 2026-05-01
Keep foresters off potential farm fields

DevOpsOfChaos:
## Summary

This adds an optional gameplay addon for farm/forester interaction.

When the addon is enabled, foresters no longer plant trees on spots that a completed own farm could currently use as a new grain field.

The default gameplay behavior is unchanged because the addon is disabled by default.

The change is intentionally narrow:

- it only affects forester planting target selection when the addon is enabled
- it only protects potential new grain field spots for the same player
- it mirrors the existing farmer field-placement constraints
- it does not reserve farm land globally
- it does not change how farmers place fields
- it does not change how foresters plant outside potential farm field spots
- it does not affect enemy farms or cross-player land usage

## Motivation

Foresters can currently plant trees on otherwise valid farm field spots before the farmer gets to them.

This can be annoying in normal play because a forester may occupy land that a nearby farm could otherwise use productively. However, this is also part of the original gameplay feel for some players, since foresters blocking useful spots is expected behavior in classic play.

For that reason, this PR no longer changes the default behavior directly. Instead, the behavior is now optional through a new addon.

## Implementation details

This PR adds a new addon:

- `AddonForesterFarmFieldAvoidance`
- `AddonId::FORESTER_FARM_FIELD_AVOIDANCE = 0x01100000`
- addon namespace entry: `011 DevOpsOfChaos`
- default status: disabled

The new check in `nofForester::GetPointQuality()` is gated behind:

```cpp
world->GetGGS().isEnabled(AddonId::FORESTER_FARM_FIELD_AVOIDANCE)

### DIFF
--- a/libs/s25main/figures/nofForester.cpp
+++ b/libs/s25main/figures/nofForester.cpp
@@ -8,12 +8,52 @@
 #include "GamePlayer.h"
 #include "Loader.h"
 #include "SoundManager.h"
+#include "buildings/noBaseBuilding.h"
 #include "network/GameClient.h"
 #include "ogl/glArchivItem_Bitmap_Player.h"
 #include "random/Random.h"
 #include "world/GameWorld.h"
 #include "nodeObjs/noTree.h"
 #include <boost/container/static_vector.hpp>
+
+namespace {
+bool IsPotentialNewFieldForOwnFarm(GameWorld& world, const MapPoint pt, const unsigned char player)
+{
+    constexpr unsigned FARM_FIELD_RADIUS = 2;
+
+    for(const auto dir : helpers::EnumRange<Direction>{})
+    {
+        if(world.GetPointRoad(pt, dir) != PointRoad::None)
+            return false;
+    }
+
+    if(!world.IsOfTerrain(pt, [](const auto& desc) { return desc.IsVital(); }))
+        return false;
+
+    const NodalObjectType noType = world.GetNO(pt)->GetType();
+    if(noType != NodalObjectType::Environment && noType != NodalObjectType::Nothing)
+        return false;
+
+    for(const MapPoint nb : world.GetNeighbours(pt))
+    {
+        const NodalObjectType nbType = world.GetNO(nb)->GetType();
+        if(nbType == NodalObjectType::Grainfield || nbType == NodalObjectType::Grapefield
+           || nbType == NodalObjectType::Building || nbType == NodalObjectType::Buildingsite)
+            return false;
+    }
+
+    return world.CheckPointsInRadius(
+      pt, FARM_FIELD_RADIUS,
+      [&world, player](const MapPoint farmPt, unsigned) {
+          if(world.GetNO(farmPt)->GetType() != NodalObjectType::Building)
+              return false;
+
+          const auto* building = world.GetSpecObj<noBaseBuilding>(farmPt);
+          return building && building->GetPlayer() == player && building->GetBuildingType() == BuildingType::Farm;
+      },
+      false);
+}
+} // namespace
 
 nofForester::nofForester(const MapPoint pos, const unsigned char player, nobUsual* workplace)
     : nofFarmhand(Job::Forester, pos, player, workplace)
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

**Comment by stefson** [comment]
can you still use the forester to spam a big forest to block your enemy at bottlenecks?

**Comment by DevOpsOfChaos** [comment]
Yes, this PR intentionally does not change that.

Foresters can still plant normal forests as before, including in places that may be tactically annoying or blocking. This change is only about one narrow case: a forester should not choose a point that a completed own farm could currently use as a new grain field.

So this is not an anti-forest-spam or anti-bottleneck-blocking change. It does not add global land reservation, enemy-distance checks, chokepoint logic, or any broader restriction on forester planting.

The only newly rejected planting spots are spots that match the existing farmer field-placement rules for the same player.

**Comment by Spikeone** [comment]
I think this should be an addon since it does change the default behavior of the game. Part of the original feeling and gameplay is foresters blocking stuff you hate him doing.

**Comment by DevOpsOfChaos** [comment]
Fair point. I agree that this changes default gameplay behavior, even if the change is intentionally narrow.

I’ll put this on my TODO list and look into converting it into an addon instead of changing the default behavior directly.

File: libs/s25main/figures/nofForester.cpp [hunk 1]
```diff
@@ -110,6 +150,10 @@ nofFarmhand::PointQuality nofForester::GetPointQuality(const MapPoint pt, bool /
             return PointQuality::NotPossible;
     }
 
+    // Avoid occupying spots that an own farm could use for a new grain field.
+    if(IsPotentialNewFieldForOwnFarm(*world, pt, player))
```
Flow86: making it an addon is quiet easy, add a check for "is addon enabled" additionally here.

DevOpsOfChaos: Done, I converted it into an addon.

Default behavior stays unchanged because the addon is disabled by default. When enabled, foresters avoid potential own farm field spots.

I also updated the test to cover both cases: addon disabled and addon enabled.

Validation passed locally, including clang-format 10.0.0, targeted integration tests, and `ctest -R "^Test_integration$"`.

libs/s25main/figures/nofForester.cpp
```diff
@@ -6,15 +6,57 @@
...
+#include "buildings/noBaseBuilding.h"
 #include "network/GameClient.h"
 #include "ogl/glArchivItem_Bitmap_Player.h"
 #include "random/Random.h"
 #include "world/GameWorld.h"
 #include "nodeObjs/noTree.h"
 #include <boost/container/static_vector.hpp>
 
+namespace {
+bool IsPotentialNewFieldForOwnFarm(GameWorld& world, const MapPoint pt, const unsigned char player)
```
Flamefire, line 23: What I don't like about this is the duplication of code that lives elsewhere. This is (almost) `nofFarmer::GetPointQuality` isn't it?

I'd say extract the code for the (max)radius to a static (protected?) method of nofFarmhand and a static variant of `nofFarmer::GetPointQuality` as a public method, called from the override there

DevOpsOfChaos: Done. I refactored it so the forester check reuses the farmer field-placement logic instead of duplicating it.

I also replaced the hardcoded farmer radius with `nofFarmhand::GetWorkRadius(Job::Farmer)`.

No intended behavior change, just less duplicated logic. Local validation passed.

tests/s25Main/integration/testFarmer.cpp
```diff
@@ -33,6 +34,31 @@ struct FarmerFixture : public WorldFixture<CreateEmptyWorld, 1>
...
+    const MapPoint fieldPt = world.GetNeighbour2(farmPt, 0);
+    BOOST_TEST_REQUIRE(isPointAvailable(farmer, fieldPt));
+
+    const MapPoint foresterPt = world.MakeMapPoint(world.GetPlayer(0).GetHQPos() - Position(5, 0));
+    auto* foresterBuilding = dynamic_cast<nobUsual*>(
+      BuildingFactory::CreateBuilding(world, BuildingType::Forester, foresterPt, 0, Nation::Romans));
+    BOOST_TEST_REQUIRE(foresterBuilding);
+
+    nofForester foresterWorker(world.GetNeighbour(foresterPt, Direction::SouthEast), 0, foresterBuilding);
+    const nofFarmhand* forester = &foresterWorker;
```
Flamefire: @Flow86 Similar to the Fisher issue: I'd make `GetPointQuality` public in all descendants of `nofFarmhand` for consistency. This avoids "hacks" like this.

@DevOpsOfChaos This shouldn't be a pointer. And if the "cast/decay" is required it could be done in the parameter type of `isPointAvailable` instead to avoid an extra, potentially confusing variable.
But that depends on the resolution of the above.

DevOpsOfChaos: Done. I only cleaned up the test pointer.

Flamefire (approved):

PR #1941, opened 2026-05-20
Adjust working ranges of forester, woodcutter and stonemason (3 Addons)

MichalLabuda:
# **Summary**

This PR adds three gameplay addons that allow players to customize the working radius of **forester**, **woodcutter**, and **stonemason**.

The default gameplay behavior remains unchanged because all three addons default to the original radii.

These addons introduce selectable radius presets expressed in percentages that increase the worker’s effective reach, making it possible to interact with resources that were previously just out of range on certain maps or simply for fun. Each addon provides several percentage‑based increments, with the default option matching the original game values.

Forester and Woodcutter share identical radius presets:

```
Default (6), +33% (8), +66% (10), +100% (12), +150% (15), +200% (18)
```

Stonemason uses slightly different values:

```
Default (8), +25% (10), +50% (12), +75% (14), +100% (15), +150% (20), +200% (24)
```

---
# **Motivation**

On some maps, it can be frustrating when a worker cannot reach a single tree or a lone granite deposit that is only slightly outside the default radius.

For personal gameplay, I previously increased these radii locally. This PR formalizes that idea by implementing it as proper optional addons so that all players can benefit from the feature without altering the default gameplay balance.

---
# **Implementation details**

- Addon author list entry: `012 MichalLabuda`
- New addon IDs:
    - `AddonId::FORESTER_REACH_RADIUS = 0x01200000`
    - `AddonId::WOODCUTTER_REACH_RADIUS = 0x01200001`
    - `AddonId::STONEMASON_REACH_RADIUS = 0x01200002`
- New addon classes:
    - `AddonForesterReachRadius.h`
    - `AddonWoodcutterReachRadius.h`
    - `AddonStonemasonReachRadius.h`
- `nofFarmhand::GetWorkRadius()` was updated to use the selected addon values instead of hardcoded constants.
- Original gameplay is preserved when the default option is selected.


libs/s25main/addons/AddonStonemasonReachRadius.h
```diff
@@ -0,0 +1,37 @@
...
+
+#include "AddonList.h"
+#include "mygettext/mygettext.h"
+
+const std::array<unsigned, 7> stonemasonRadiusValues = {
+    8,  // Default
+    10, // +25%
+    12, // +50%
+    14, // +75%
+    15, // +100%
```
Flamefire: One off?
```suggestion
    16, // +100%
```

libs/s25main/figures/nofFarmhand.cpp
```diff
@@ -37,9 +42,21 @@ unsigned nofFarmhand::GetWorkRadius(const Job job)
         case Job::Winegrower: return 2;
         case Job::CharBurner: return 3;
         case Job::Woodcutter:
-        case Job::Forester: return 6;
+        {
+            unsigned sel = world->GetGGS().getSelection(AddonId::WOODCUTTER_REACH_RADIUS);
```
Flamefire:
```suggestion
            const unsigned sel = world->GetGGS().getSelection(AddonId::WOODCUTTER_REACH_RADIUS);
```

Could even omit the variable and make it a single return-line without braces, but why not. Just const-correct

libs/s25main/figures/nofFarmhand.cpp
```diff
@@ -37,9 +42,21 @@ unsigned nofFarmhand::GetWorkRadius(const Job job)
...
+            return woodcutterRadiusValues[sel];
+        }
+        case Job::Forester:
+        {
+            unsigned sel = world->GetGGS().getSelection(AddonId::FORESTER_REACH_RADIUS);
```
Flamefire:
```suggestion
            const unsigned sel = world->GetGGS().getSelection(AddonId::FORESTER_REACH_RADIUS);
```

libs/s25main/figures/nofFarmhand.cpp
```diff
@@ -37,9 +42,21 @@ unsigned nofFarmhand::GetWorkRadius(const Job job)
...
+            return foresterRadiusValues[sel];
+        }
         case Job::Fisher: return 7;
-        case Job::Stonemason: return 8;
+        case Job::Stonemason:
+        {
+            unsigned sel = world->GetGGS().getSelection(AddonId::STONEMASON_REACH_RADIUS);
```
Flamefire:
```suggestion
            const unsigned sel = world->GetGGS().getSelection(AddonId::STONEMASON_REACH_RADIUS);
```

Flamefire (changes requested):
Thanks for the PR, looks good, just minor nits which you can apply from the UI here.

I'm thinking about making the description more explicit to avoid just repeating the name, i.e. change "Adjust the stonemason's mining range" to something like "Increase the radius in which the stonemason searches for stone" or "(possible) work area", but that doesn't sound much easier so I guess the "xxx range" is enough to understand the effect and we can keep that, unless you have an idea.

libs/s25main/addons/AddonForesterReachRadius.h
```diff
@@ -0,0 +1,35 @@
...
+    18  // +200%
+};
+
+class AddonForesterReachRadius : public AddonList
+{
+public:
+    AddonForesterReachRadius()
+        : AddonList(AddonId::FORESTER_REACH_RADIUS, AddonGroup::GamePlay,
+                    _("Adjust forester's range"),
+                    _("Adjust the forester's planting range."),
```
MichalLabuda:
```suggestion
                    _("Increase the radius in which the forester plants trees."),
```

libs/s25main/addons/AddonStonemasonReachRadius.h
```diff
@@ -0,0 +1,37 @@
...
+    24  // +200%
+};
+
+class AddonStonemasonReachRadius : public AddonList
+{
+public:
+    AddonStonemasonReachRadius()
+        : AddonList(AddonId::STONEMASON_REACH_RADIUS, AddonGroup::GamePlay,
+                    _("Adjust stonemason's range"),
+                    _("Adjust the stonemason's mining range."),
```
MichalLabuda:
```suggestion
                    _("Increase the radius in which the stonemason searches for stone."),
```

libs/s25main/addons/AddonWoodcutterReachRadius.h
```diff
@@ -0,0 +1,35 @@
...
+    18  // +200%
+};
+
+class AddonWoodcutterReachRadius : public AddonList
+{
+public:
+    AddonWoodcutterReachRadius()
+        : AddonList(AddonId::WOODCUTTER_REACH_RADIUS, AddonGroup::GamePlay,
+                    _("Adjust woodcutter's range"),
+                    _("Adjust the woodcutter's cutting range."),
```
MichalLabuda:
```suggestion
                    _("Increase the radius in which the woodcutter searches for trees."),
```

libs/s25main/addons/AddonForesterReachRadius.h
```diff
@@ -0,0 +1,35 @@
...
+    15, // +150%
+    18  // +200%
+};
+
+class AddonForesterReachRadius : public AddonList
+{
+public:
+    AddonForesterReachRadius()
+        : AddonList(AddonId::FORESTER_REACH_RADIUS, AddonGroup::GamePlay,
+                    _("Adjust forester's range"),
```
MichalLabuda:
```suggestion
                    _("Adjust forester's planting range"),
```

libs/s25main/addons/AddonStonemasonReachRadius.h
```diff
@@ -0,0 +1,37 @@
...
+    20, // +150%
+    24  // +200%
+};
+
+class AddonStonemasonReachRadius : public AddonList
+{
+public:
+    AddonStonemasonReachRadius()
+        : AddonList(AddonId::STONEMASON_REACH_RADIUS, AddonGroup::GamePlay,
+                    _("Adjust stonemason's range"),
```
MichalLabuda:
```suggestion
                    _("Adjust stonemason's mining range"),
```

libs/s25main/addons/AddonWoodcutterReachRadius.h
```diff
@@ -0,0 +1,35 @@
...
+    15, // +150%
+    18  // +200%
+};
+
+class AddonWoodcutterReachRadius : public AddonList
+{
+public:
+    AddonWoodcutterReachRadius()
+        : AddonList(AddonId::WOODCUTTER_REACH_RADIUS, AddonGroup::GamePlay,
+                    _("Adjust woodcutter's range"),
```
MichalLabuda:
```suggestion
                    _("Adjust woodcutter's chopping range"),
```

MichalLabuda: Not sure about this one. Other options would be: cutting, felling, harvesting or just working.

MichalLabuda (commented):
I've committed the suggestions regarding the const and good catch with the stonemason's 100% range - thanks.

I've also modified the descriptions according to your suggestion.

I added the actual work names to the addon names for clarity.

Flamefire:
I'm fine with the current short addon names like "Adjust xxx's range", clear for those who know and if in doubt the tooltip provides a longer variant. I think this makes the addon list easier to read.

You need to run clang-format, e.g. via `make clangFormat`, see the failing CI check

Flamefire (approved):

PR #1942, opened 2026-05-21
Configurable mine resource behavior

DevOpsOfChaos:
## Summary

This PR implements configurable mine resource behavior as a cleaned-up replacement path for the broader #1501 direction.

It adds per-mine resource behavior settings for coal, iron, gold, and granite mines, while reusing existing inexhaustible mine addon IDs where appropriate instead of stacking additional addons on top.

## Details

- Add configurable mine resource behavior per mine type:
  - Default
  - S4-like exhaustion
  - Inexhaustible
  - Work everywhere
- Reuse existing `INEXHAUSTIBLE_*` addon IDs as per-mine behavior list settings where possible
- Map old boolean `ON` values to the new `Inexhaustible` behavior
- Migrate legacy global `INEXHAUSTIBLE_MINES` settings to per-mine inexhaustible behavior when loading old settings
- Remove the separate granite work-everywhere addon path
- Treat `Work everywhere` as ignoring ground-resource requirements instead of creating resources in the world
- Add configurable no-output fallback behavior for failed S4-like production cycles
- Teach AI resource rating and mine planning about configured mine behavior
- Adjust displayed mine productivity for S4-like resource chance
- Add integration coverage for production behavior, AI behavior, serialization/migration, and productivity display

## S4-like behavior

For S4-like behavior:

- each mine type can be configured independently
- a mine attempts production based on remaining matching resources in its mining radius
- 20 remaining matching resources is the explicit full-productivity reference
- below that reference, production chance is reduced proportionally
- successful production depletes resources down to 1, not to 0
- failed production runs a no-output cycle
- the no-output cycle can either produce nothing or use the configured fallback production behavior

## Work-everywhere behavior

`Work everywhere` means that the mine skips the ground-resource requirement.

It does not create temporary resources in the world and does not deplete such generated resources.

## Validation

- Built `Test_integration`
- Built `Test_UI`
- Passed `Test_integration --run_test=Production`
- Passed `Test_integration --run_test=AI`
- Passed `Test_integration --run_test=Serialization`
- Passed `ctest --test-dir .\build-vs-x64-debug-local -C Debug -R "^Test_integration$" --output-on-failure`
- Passed `Test_UI.exe --report_level=short`
- `git diff --check upstream/master...HEAD` passed


Flamefire (changes requested):
Thanks for taking care of that.
Please remove duplications and check back with what was discussed before. This includes:

- There is no need for multiple addons for a single mine type: `GRANITEMINES_WORK_EVERYWHERE` is superflous and all the existing "inexhaustible*" addons should be integrated into this ones and use the same IDs: The OFF setting should be "Default" and ON (the new) "inexhaustible"
- `FindPointWithResourceQuiet` duplicates code already present -> Reuse existing one, can be done by simply adding a default bool parameter `notify`
- `INEXHAUSTIBLE_MINES` addon should be removed: When detected during loading it should set the new addons accordingly, keep ID for that with a TODO for gamedata version increasing
- Explain `CalcAverageDisplayProductivity` please, why the extra methods?
- Isn't there already a method to convert AIResource to Resource so `GetMineBuildingType` doesn't need to be (fully) implemented twice?
- `CanCreateWorkEverywhereResource` looks odd: "Work everywhere" should be just that: Ignore any resource in the ground, not add resources to the world, doesn't it? So just skip the search
- `IsMineResourceDepletable` shouldn't use `GetConfiguredMineResourceBehavior`, in fact it seems only a single function is required, i.e. both Get... combined.

Maybe more.

Please give a quick summary of how "s4 like behavior" is supposed to work, so @Spikeone can verify it is what he intended in https://github.com/Return-To-The-Roots/s25client/pull/1501 and we can verify it is implemented according to that spec.

DevOpsOfChaos:
> Thanks for taking care of that. Please remove duplications and check back with what was discussed before. This includes:
> 
> * There is no need for multiple addons for a single mine type: `GRANITEMINES_WORK_EVERYWHERE` is superflous and all the existing "inexhaustible*" addons should be integrated into this ones and use the same IDs: The OFF setting should be "Default" and ON (the new) "inexhaustible"
> * `FindPointWithResourceQuiet` duplicates code already present -> Reuse existing one, can be done by simply adding a default bool parameter `notify`
> * `INEXHAUSTIBLE_MINES` addon should be removed: When detected during loading it should set the new addons accordingly, keep ID for that with a TODO for gamedata version increasing
> * Explain `CalcAverageDisplayProductivity` please, why the extra methods?
> * Isn't there already a method to convert AIResource to Resource so `GetMineBuildingType` doesn't need to be (fully) implemented twice?
> * `CanCreateWorkEverywhereResource` looks odd: "Work everywhere" should be just that: Ignore any resource in the ground, not add resources to the world, doesn't it? So just skip the search
> * `IsMineResourceDepletable` shouldn't use `GetConfiguredMineResourceBehavior`, in fact it seems only a single function is required, i.e. both Get... combined.
> 
> Maybe more.
> 
> Please give a quick summary of how "s4 like behavior" is supposed to work, so @Spikeone can verify it is what he intended in #1501 and we can verify it is implemented according to that spec.

Thanks, that direction makes sense.

My understanding of the requested cleanup is:

- fold the existing `INEXHAUSTIBLE_*` addon IDs into the new per-mine behavior list settings
- keep `OFF == Default` and map the old `ON` value to the new `Inexhaustible` behavior
- remove the extra granite-specific work-everywhere addon path
- remove the old global `INEXHAUSTIBLE_MINES` behavior as an active setting and migrate it during loading to the new per-mine settings, keeping its ID only for compatibility/TODO until the gamedata version can be increased
- avoid helper duplication such as `FindPointWithResourceQuiet`
- treat `WorkEverywhere` as “ignore ground resources”, not “create resources in the world”
- simplify the effective behavior/depletion helper logic

My current S4-like behavior implementation is intended to work like this:

- each mine type can be configured independently
- for S4-like behavior, a mine can attempt production based on the remaining matching resources in its mining radius
- the production chance is derived from remaining resource amount compared to the relevant possible resource capacity around the mine
- resources are depleted only down to 1, not to 0
- if the chance check fails, the mine runs a no-output cycle
- the no-output cycle can either produce nothing or use the configured fallback production behavior

I’ll rework the implementation toward the existing-ID/list-setting approach and remove the duplicated/intermediate pieces instead of stacking this on top of the old addon set.

Flamefire:
> * for S4-like behavior, a mine can attempt production based on the remaining matching resources in its mining radius
> 
> * the production chance is derived from remaining resource amount compared to the relevant possible resource capacity around the mine
> * resources are depleted only down to 1, not to 0
> * if the chance check fails, the mine runs a no-output cycle
> * the no-output cycle can either produce nothing or use the configured fallback production behavior

@Spikeone Looks correct to me. Can you confirm?

"possible resource capacity" isn't fully clear to me. There are 7 points considered and currently each resource gives 5% chance. I.e. when all have at least 3 resources we will have 100%. IIRC the max amount of resources per node is 7 (we should have a constant for that if we don't already have one)
The code should be clear here, the 5% looks random. The chance could be e.g. directly calculated from the points and use the number of points and max amount. Or the constant derived from something like that.

DevOpsOfChaos:
> > * for S4-like behavior, a mine can attempt production based on the remaining matching resources in its mining radius
> > * the production chance is derived from remaining resource amount compared to the relevant possible resource capacity around the mine
> > * resources are depleted only down to 1, not to 0
> > * if the chance check fails, the mine runs a no-output cycle
> > * the no-output cycle can either produce nothing or use the configured fallback production behavior
> 
> @Spikeone Looks correct to me. Can you confirm?
> 
> "possible resource capacity" isn't fully clear to me. There are 7 points considered and currently each resource gives 5% chance. I.e. when all have at least 3 resources we will have 100%. IIRC the max amount of resources per node is 7 (we should have a constant for that if we don't already have one) The code should be clear here, the 5% looks random. The chance could be e.g. directly calculated from the points and use the number of points and max amount. Or the constant derived from something like that.

Good point. The current implementation intentionally models the existing S4-like “full productivity around ~20 remaining resources” behavior rather than deriving 100% from the absolute theoretical map capacity.

The current `5% per remaining resource` effectively means:

* 20 remaining matching resources => 100% production chance
* depletion continues reducing the chance below that point

I agree the current formulation makes the `5%` look arbitrary in code and in the PR description.

I will rework this to express the behavior through named constants / derived calculation so the intended reference point becomes explicit instead of looking like a magic number.

Flamefire:
> Good point. The current implementation intentionally models the existing S4-like “full productivity around ~20 remaining resources” behavior rather than deriving 100% from the absolute theoretical map capacity.

Yes, its 100% at (exactly) 20 resources, but why 20 and not 10, 30 or 40 when 49 is the theoretical maximum? Or even as low as 7 which would be 1 node with full resources.
I'm not decided on what makes sense, but it needs to be explained in the code where the constant is defined with a short sentence.
And as the initial version was by @Spikeone I'd follow his preference if he has any.

DevOpsOfChaos:
Updated the PR to address the requested architecture cleanup and the S4-like chance clarity point.

Main changes:

- removed the separate `GRANITEMINES_WORK_EVERYWHERE` addon path
- reused `INEXHAUSTIBLE_GRANITEMINES` as the granite mine behavior list ID
- old boolean value `1` now maps to `Inexhaustible`
- removed active registration/use of global `INEXHAUSTIBLE_MINES`
- legacy enabled global `INEXHAUSTIBLE_MINES` settings are migrated to per-mine `Inexhaustible` selections when loading/deserializing settings
- replaced the duplicated quiet resource lookup with `FindPointWithResource(..., false)`
- simplified mine behavior helper logic
- corrected `Work everywhere` semantics so it skips resource checks instead of creating/depleting world resources
- kept S4-like behavior and no-output fallback behavior
- made the S4-like chance reference explicit via a named 20-resource full-productivity reference instead of a buried “5% per resource” rule
- updated Production, AI, Serialization, and UI/productivity tests

S4-like behavior after the rework:

- each mine type is configured independently
- production chance is based on remaining matching resources in the mining radius
- 20 remaining matching resources is the explicit full-productivity reference
- successful production depletes resources down to 1, not 0
- failed production runs the configured no-output/fallback cycle

Validation:

- built `Test_integration`
- built `Test_UI`
- `Test_integration --run_test=Production` passed
- `Test_integration --run_test=AI` passed
- `Test_integration --run_test=Serialization` passed
- `ctest -R "^Test_integration$"` passed
- `Test_UI.exe --report_level=short` passed
- `git diff --check upstream/master...HEAD` clean

libs/s25main/addons/AddonInexhaustibleGraniteMines.h
```diff
@@ -4,17 +4,20 @@
...
 
 /**
- *  Addon for allowing to have unlimited resources.
+ *  Granite mine resource behavior list.
+ *
+ *  Reuses the old boolean INEXHAUSTIBLE_GRANITEMINES id: saved value 0 still means default behavior and saved value 1
+ *  now selects the inexhaustible behavior.
  */
-class AddonInexhaustibleGraniteMines : public AddonBool
+class AddonInexhaustibleGraniteMines : public AddonMineResourceBehaviorBase
```
Flamefire: Please move this class to the other ones to avoid missing it. And you can rename the addon id to match the new ones if you keep the value.

libs/s25main/addons/AddonInexhaustibleMines.h
```diff
@@ -8,7 +8,11 @@
 #include "mygettext/mygettext.h"
 
 /**
- *  Addon for allowing to have unlimited resources.
+ *  Deprecated global mine setting.
```
Flamefire, line 11: Add `TODO(Replay) TODO(Savegame)` so it can be found when updating

libs/s25main/addons/const_addons.h
```diff
@@ -38,6 +38,7 @@
 //
 // Add the #include for your AddonXXX.h in Addons.h!
 //
+// TODO: INEXHAUSTIBLE_MINES is kept only for loading old settings/savegames until the gamedata version can be raised.
```
Flamefire:
```suggestion
// TODO(Replay) TODO(Savegame): INEXHAUSTIBLE_MINES is kept only for loading old settings/savegames until the gamedata version can be raised.
```

libs/s25main/gameTypes/MineResourceBehavior.cpp
```diff
@@ -0,0 +1,80 @@
...
+    // S4-like productivity is intentionally based on a 20-resource reference amount, not the theoretical maximum
+    // resources in the mine radius. Below that amount, the production chance degrades linearly.
+    const unsigned chancePercent =
+      remainingMatchingResources * MAX_PRODUCTION_PERCENT / GetS4LikeMineFullProductivityResourceAmount();
+    return std::min(MAX_PRODUCTION_PERCENT, chancePercent);
+}
+
+MineResourceBehavior GetMineResourceBehavior(const GlobalGameSettings& settings, const BuildingType buildingType)
+{
+    switch(static_cast<MineResourceBehavior>(settings.getSelection(GetMineResourceBehaviorAddonId(buildingType))))
```
Flamefire: What is the reason for this switch? I should at least be explained by a short comment.
If it is only validation (why necessary?) then I'd rather have a `auto result = ...; switch(result){case X:caseY:case Z:break /*or return*/; default: result=default}`-like code to avoid having to read all this C&P

libs/s25main/gameTypes/MineResourceBehavior.cpp
```diff
@@ -0,0 +1,80 @@
...
+    {
+        case ResourceType::Gold: return BuildingType::GoldMine;
+        case ResourceType::Iron: return BuildingType::IronMine;
+        case ResourceType::Coal: return BuildingType::CoalMine;
+        case ResourceType::Granite: return BuildingType::GraniteMine;
+        default: return boost::none;
+    }
+}
+
+unsigned GetS4LikeMineFullProductivityResourceAmount()
```
Flamefire: Why have a function that returns a constant?

libs/s25main/buildings/nobUsual.cpp
```diff
@@ -519,6 +522,25 @@ bool nobUsual::HasWorker() const
     return worker && worker->GetState() != nofBuildingWorker::State::FigureWork;
 }
 
+unsigned short nobUsual::GetDisplayProductivity() const
+{
```
Flamefire, line 526: Why do we need this? Shouldn't the regular productivity reflect the addon state by counting produced-cycles correctly?

libs/s25main/buildings/nobUsual.h
```diff
@@ -102,6 +102,7 @@ class nobUsual : public noBuilding
     /// Gibt Pointer auf Produktivität zurück
     const unsigned short* GetProductivityPointer() const { return &productivity; }
     unsigned short GetProductivity() const { return productivity; }
+    unsigned short GetDisplayProductivity() const;
```
Flamefire: See above: Needs a doc string to explain what this is and why/when/how it might be different to `GetProductivity` if it is required at all

libs/s25main/ingameWindows/iwBuildingProductivities.cpp
```diff
@@ -100,7 +100,7 @@ iwBuildingProductivities::iwBuildingProductivities(const GamePlayer& player)
 
 void iwBuildingProductivities::UpdatePercents()
 {
-    percents = player.GetBuildingRegister().CalcProductivities();
+    percents = player.GetBuildingRegister().CalcDisplayProductivities();
```
Flamefire: Why would we need `CalcProductivities` if we need `CalcDisplayProductivities`?

libs/s25main/figures/nofMiner.cpp
```diff
@@ -10,7 +10,104 @@
...
+
+namespace {
+constexpr unsigned MAX_PRODUCTION_PERCENT = 100;
+constexpr unsigned GRANITE_FALLBACK_25_PERCENT = 25;
+constexpr unsigned GRANITE_FALLBACK_50_PERCENT = 50;
+constexpr unsigned S4LIKE_MIN_RESOURCE_AMOUNT = 1;
+
+MineNoOutputFallback GetConfiguredNoOutputFallback(const GlobalGameSettings& settings)
+{
+    switch(static_cast<MineNoOutputFallback>(settings.getSelection(AddonId::MINE_NO_OUTPUT_FALLBACK)))
```
Flamefire: Similar to the other method: At least avoid C&P by using a variable and directly returning or doing nothing in the "success" case

libs/s25main/figures/nofMiner.cpp
```diff
@@ -10,7 +10,104 @@
...
+
+    if(fallback == MineNoOutputFallback::ProduceLowerGradeResource)
+        return GetLowerGradeFallbackGood(buildingType);
+
+    return boost::none;
+}
+
+std::vector<MapPoint> GetPointsWithResource(const GameWorld& world, const MapPoint pos, const ResourceType type)
+{
+    return world.GetMatchingPointsInRadius<1>(
```
Flamefire: This will only return a single point, which is not what you want, do you?

libs/s25main/figures/nofMiner.cpp
```diff
@@ -71,30 +184,39 @@ helpers::OptionalEnum<GoodType> nofMiner::ProduceWare()
...
+
+    const MineResourceBehavior behavior = GetMineResourceBehavior(world->GetGGS(), workplace->GetBuildingType());
+    if(behavior == MineResourceBehavior::WorkEverywhere)
+        return true;
+
+    if(FindPointWithResource(GetRequiredResType(), false).isValid())
+        return true;
+
+    workplace->OnOutOfResources();
+    return false;
```
Flamefire: Why not

```suggestion
   return FindPointWithResource(GetRequiredResType()).isValid();
```

Wasn't that the reason for the `notify` param?  Or is that not required and we could always notify in the caller (which might be cleaner anyway)?

libs/s25main/gameTypes/MineNoOutputFallback.h
```diff
@@ -0,0 +1,14 @@
+// Copyright (C) 2005 - 2026 Settlers Freaks (sf-team at siedler25.org)
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#pragma once
+
+enum class MineNoOutputFallback : unsigned
```
Flamefire: do you need this to have an explicit type? Or is the default enough, i.e.:

```suggestion
enum class MineNoOutputFallback
```

Only cases where I can imagine explicit types are serialized enums that need a fixed-size type or when you rely on overflow etc.

libs/s25main/gameTypes/MineResourceBehavior.h
```diff
@@ -0,0 +1,30 @@
...
+#include <cstdint>
+
+class GlobalGameSettings;
+enum class AddonId;
+enum class BuildingType : unsigned char;
+enum class ResourceType : uint8_t;
+
+enum class MineResourceBehavior : unsigned
+{
+    Default = 0,
```
Flamefire: Why the explicit type and values?

Flamefire (changes requested):
Most things are now readability:
- `TODO(Replay)` and/or `TODO(Savegame)` allows to quickly find all places
- "display"productivity seems odd
- less code is usually better: Reduce C&P and avoid enum constraints where not necessary or document why
- The new `notify` param for `FindPointWithResource` now looks redundant so notification might get moved out which is better for SRP as a "find" should not "notify"

DevOpsOfChaos:
Updated the branch against current master and addressed the remaining mine-resource review points.

The follow-up now uses one productivity path for the expected primary-resource yield, removes the parallel display-productivity layer, and keeps fallback goods outside that primary-resource metric.

It also fixes the S4-like resource calculation to consider all matching resource points within the mine radius, with regression coverage for multiple points, wrong resource types, out-of-range points, empty resources, and rounding.

Validated locally:

- Production suite: 16 test cases / 179 assertions
- AI suite
- targeted mine-productivity and legacy serialization scenarios
- Test_integration via ctest
- Test_UI build
- git diff --check: clean

DevOpsOfChaos:
Followed up on the remaining review-level cleanup points.

- The legacy `INEXHAUSTIBLE_MINES` compatibility marker now explicitly covers both replay and savegame migration.
- Removed the public getter that only exposed the internal S4-like full-productivity reference constant; the named constant is now local to the implementation.
- Kept `FindPointWithResource(..., notify)` unchanged because the former quiet duplicate is already removed and the current callers intentionally use the shared function.

Validated locally:

- Production: 16 test cases / 179 assertions
- Serialization: 10 test cases / 15917 assertions
- AI: 11 test cases / 21376 assertions
- Test_integration ctest: passed
- Test_UI: 73 test cases / 11106 assertions

Flamefire:
There are a couple minor points left that need to be addressed so we can get this merged. Just in case they got missed

libs/s25main/addons/AddonInexhaustibleGraniteMines.h
DevOpsOfChaos: Addressed. The granite behavior class now lives in `AddonMineResourceBehavior.h` as `AddonGraniteMineResourceBehavior`, next to the coal, iron and gold variants. The ID is now `GRANITEMINE_RESOURCE_BEHAVIOR`, while its numeric value remains `0x00800000`, preserving settings, savegame and replay compatibility. No stale `INEXHAUSTIBLE_GRANITEMINES` class or ID reference remains.

libs/s25main/addons/AddonInexhaustibleMines.h
DevOpsOfChaos: Added `TODO(Replay) TODO(Savegame)` to the legacy compatibility class and enum marker. Verification also uncovered and fixed a migration edge case: old global inexhaustible-mine data now migrates every mine type still set to `Default`, including granite, without overriding explicit per-mine behavior. This is covered by a serialization regression test.

libs/s25main/figures/nofMiner.cpp
DevOpsOfChaos: Went with the caller-notification approach: `FindPointWithResource` is now side-effect free and the callers notify explicitly. A complete call-site audit found that the two well-worker paths also required notification; those are now covered as well. `Production/WellWithoutWaterReportsOutOfResources` was confirmed failing before the fix and passes afterward.

DevOpsOfChaos:
Final verification update for head `9305cea8a`.

In addition to addressing the remaining inline review points, the branch was tested through a temporary local integration with current `master` (`4b314647`). That exposed five new `boost::none` returns which no longer compiled after master migrated `helpers::OptionalEnum` to `std::optional`; they now use `{}`, compiling both standalone and with current master.

Lua backward compatibility was also preserved. `ADDON_INEXHAUSTIBLE_GRANITEMINES` remains available as an alias for `ADDON_GRANITEMINE_RESOURCE_BEHAVIOR`, and `rttr:SetAddon(ADDON_INEXHAUSTIBLE_MINES, true)` now uses the same legacy migration as settings/savegames without re-registering the obsolete addon. A Lua regression test covers the alias, all four mine types, explicit-setting preservation, the false case and the unregistered legacy addon.

Locally verified with GCC 14 and Clang 18, Debug and warnings-as-errors: complete build, `ctest` 20/20, Production, AI, Serialization and Lua settings suites. Earlier verification additionally covered Release, ASan/UBSan, clang-tidy and Test_UI. Jenkins and MSVC were not available locally, so this comment makes no claim about them.

tests/s25Main/integration/testAI.cpp
```diff
@@ -51,6 +55,37 @@ inline bool playerHasBld(const GamePlayer& player, BuildingType type)
...
+    RTTR_FOREACH_PT(MapPoint, world.GetSize())
+    {
+        MapNode& node = world.GetNodeWriteable(pt);
+        node.t1 = node.t2 = mineableTerrain;
+        node.resources = Resource();
+    }
+    world.InitAfterLoad();
+}
+
+void makeMineNodesUsableForSearch(AIJH::AIPlayerJH& aijh, const GameWorld& world, const unsigned player)
```
Flamefire: The name should clarify this is affecting AI nodes, e.g:
```suggestion
void initAIJhNodes(AIJH::AIPlayerJH& aijh, const GameWorld& world, const unsigned player)
```
It's also overly specific: "MineNodes" has no relevance here, has it?

tests/s25Main/integration/testAI.cpp
```diff
@@ -109,6 +144,66 @@ BOOST_FIXTURE_TEST_CASE(AIChat, EmptyWorldFixture2P)
...
+
+BOOST_FIXTURE_TEST_CASE(MineWorkEverywhereAffectsMatchingAIResourceOnly, EmptyWorldFixture1P)
+{
+    makeWorldMineable(world);
+    ggs.setSelection(AddonId::COALMINE_RESOURCE_BEHAVIOR, static_cast<unsigned>(MineResourceBehavior::WorkEverywhere));
+
+    AIJH::AIPlayerJH ai(0, world, AI::Level::Hard);
+    makeMineNodesUsableForSearch(ai, world, 0);
+
+    const MapPoint around = world.GetPlayer(0).GetHQPos();
```
Flamefire: Better name here? "around" what? Why the HQ? Maybe just a short comment in this line

tests/s25Main/integration/testAI.cpp
```diff
@@ -109,6 +144,66 @@ BOOST_FIXTURE_TEST_CASE(AIChat, EmptyWorldFixture2P)
...
+    const MapPoint otherResourcePos = world.MakeMapPoint(around + Position(8, 0));
+    world.GetNodeWriteable(otherResourcePos).resources = Resource(ResourceType::Iron, 4);
+    BOOST_TEST(ai.getAIInterface().GetResourceRating(otherResourcePos, AIResource::Coal)
+               == static_cast<int>(RES_RADIUS[AIResource::Coal]));
+    BOOST_TEST(ai.FindBestPosition(around, AIResource::Coal, BuildingQuality::Mine, 5).isValid());
+    BOOST_TEST(!ai.FindBestPosition(around, AIResource::Ironore, BuildingQuality::Mine, 5).isValid());
```
Flamefire: What is the motivation for those tests? You place iron in some distance but AI finds a position for coal but not for Iron? If that is intentional add comments

tests/s25Main/integration/testAI.cpp
```diff
@@ -109,6 +144,66 @@ BOOST_FIXTURE_TEST_CASE(AIChat, EmptyWorldFixture2P)
...
+    ggs.setSelection(AddonId::COALMINE_RESOURCE_BEHAVIOR,
+                     static_cast<unsigned>(MineResourceBehavior::S4LikeExhaustion));
+    const int s4LikeRating = ai.getAIInterface().GetResourceRating(resourcePos, AIResource::Coal);
+
+    BOOST_TEST(defaultRating == static_cast<int>(RES_RADIUS[AIResource::Coal]));
+    BOOST_TEST(s4LikeRating > 0);
+    BOOST_TEST(s4LikeRating < defaultRating);
+}
+
+BOOST_FIXTURE_TEST_CASE(MineWorkEverywhereAffectsMatchingAIResourceOnly, EmptyWorldFixture1P)
```
Flamefire:
```suggestion
BOOST_FIXTURE_TEST_CASE(MineWorkEverywhereAffectsMatchingResourceOnly, EmptyWorldFixture1P)
```

tests/s25Main/integration/testAI.cpp
```diff
@@ -109,6 +144,66 @@ BOOST_FIXTURE_TEST_CASE(AIChat, EmptyWorldFixture2P)
...
+    MockAI ai(0, world, AI::Level::Easy);
+    const int defaultRating = ai.getAIInterface().GetResourceRating(resourcePos, AIResource::Coal);
+
+    ggs.setSelection(AddonId::COALMINE_RESOURCE_BEHAVIOR,
+                     static_cast<unsigned>(MineResourceBehavior::S4LikeExhaustion));
+    const int s4LikeRating = ai.getAIInterface().GetResourceRating(resourcePos, AIResource::Coal);
+
+    BOOST_TEST(defaultRating == static_cast<int>(RES_RADIUS[AIResource::Coal]));
+    BOOST_TEST(s4LikeRating > 0);
+    BOOST_TEST(s4LikeRating < defaultRating);
```
Flamefire, line 169: Maybe include setting the resource behavior to unlimited and inexhaustible too when checking the rating?

tests/s25Main/integration/testAI.cpp
```diff
@@ -109,6 +144,66 @@ BOOST_FIXTURE_TEST_CASE(AIChat, EmptyWorldFixture2P)
...
+BOOST_FIXTURE_TEST_CASE(InexhaustibleGraniteDoesNotImplyWorkEverywhereForAI, EmptyWorldFixture1P)
+{
+    makeWorldMineable(world);
+    ggs.setSelection(AddonId::GRANITEMINE_RESOURCE_BEHAVIOR, 1);
+
+    AIJH::AIPlayerJH ai(0, world, AI::Level::Hard);
+    makeMineNodesUsableForSearch(ai, world, 0);
+
+    const MapPoint around = world.GetPlayer(0).GetHQPos();
+    BOOST_TEST(!ai.FindBestPosition(around, AIResource::Granite, BuildingQuality::Mine, 5).isValid());
```
Flamefire: I'd combine this test with the one above: change from "WorkEverywhere" to "Inexhaustible" changes what is found. That highlights the difference and avoids the test setup for a trivial test.
Instead of the test name a comment would work and test name could be "Granite mine setting affects AI" or so.

tests/s25Main/integration/testAI.cpp
```diff
@@ -109,6 +144,66 @@ BOOST_FIXTURE_TEST_CASE(AIChat, EmptyWorldFixture2P)
...
+
+    const MapPoint around = world.GetPlayer(0).GetHQPos();
+    BOOST_TEST(ai.FindBestPosition(around, AIResource::Granite, BuildingQuality::Mine, 5).isValid());
+    BOOST_TEST(!ai.FindBestPosition(around, AIResource::Coal, BuildingQuality::Mine, 5).isValid());
+}
+
+BOOST_FIXTURE_TEST_CASE(InexhaustibleGraniteDoesNotImplyWorkEverywhereForAI, EmptyWorldFixture1P)
+{
+    makeWorldMineable(world);
+    ggs.setSelection(AddonId::GRANITEMINE_RESOURCE_BEHAVIOR, 1);
```
Flamefire: Use `MineResourceBehavior` as above not "1"

tests/s25Main/integration/testGamePlayer.cpp
```diff
@@ -55,6 +58,25 @@ RTTR_ATTRIBUTE_NO_UBSAN(vptr) void setProductivity(nobUsual* bld, unsigned short
 } // namespace
 
 using WorldFixtureEmpty1P = WorldFixture<CreateEmptyWorld, 1, 2 * helpers::MaxEnumValue_v<BuildingType> + 14, 4>;
+using WorldFixtureMineRadius1P = WorldFixture<CreateEmptyWorld, 1, 20, 12>;
+
+namespace {
+MapPoint FindMinePosition(const WorldFixtureMineRadius1P& fixture)
+{
+    for(MapCoord y = MINER_RADIUS; y + MINER_RADIUS < fixture.world.GetSize().y; ++y)
```
Flamefire: Why the weird bounds? Don't you just need any point?

tests/s25Main/integration/testGamePlayer.cpp
```diff
@@ -55,6 +58,25 @@ RTTR_ATTRIBUTE_NO_UBSAN(vptr) void setProductivity(nobUsual* bld, unsigned short
...
+    {
+        for(MapCoord x = MINER_RADIUS; x + MINER_RADIUS < fixture.world.GetSize().x; ++x)
+        {
+            const MapPoint pt(x, y);
+            if(fixture.world.GetNode(pt).bq == BuildingQuality::Castle)
```
Flamefire: Wouldn't you need `BuildingQuality::Mine` for a mine?

tests/s25Main/integration/testGamePlayer.cpp
```diff
@@ -121,6 +143,90 @@ BOOST_FIXTURE_TEST_CASE(ProductivityStats, WorldFixtureEmpty1P)
     BOOST_TEST(buildingRegister.CalcAverageProductivity() == avgProd);
 }
 
+BOOST_FIXTURE_TEST_CASE(MineProductivityAccountsForS4LikeResourceChance, WorldFixtureEmpty1P)
+{
+    MapPoint minePos(0, 0);
+    while(world.GetNode(minePos).bq != BuildingQuality::Castle)
```
Flamefire: Why does this not use the method above? And why use at all? `BuildingFactory::CreateBuilding` doesn't even care about BQs

tests/s25Main/integration/testGamePlayer.cpp
```diff
@@ -121,6 +143,90 @@ BOOST_FIXTURE_TEST_CASE(ProductivityStats, WorldFixtureEmpty1P)
...
+        BOOST_TEST_REQUIRE((++minePos.x) < world.GetSize().x);
+
+    auto* coalMine = static_cast<nobUsual*>(
+      BuildingFactory::CreateBuilding(world, BuildingType::CoalMine, minePos, 0, Nation::Romans));
+    setProductivity(coalMine, 100);
+
+    world.SetResource(minePos, Resource(ResourceType::Coal, 1));
+    BOOST_TEST(coalMine->GetProductivity() == 100u);
```
Flamefire, line 181: What does this test? You set the productivity 3 lines above so this looks like it cannot fail.
This test at least needs explanations what exactly is tested and why the values change

Maybe with less examples and hard-coded values. E.g. after `setProductivity(coalMine, 50);` the result should be half the previous value (and a comment why)
Similar doubling the resources should double productivity (does it?)
And/Or add a comment that full productivity is at 20 resources

As you set resource on the mine and neighbor I'd make a lambda taking both values: `setCoalAmounts(15, 5)` reads easier than having to check what the neighbor resource was previously set to.

tests/s25Main/integration/testGamePlayer.cpp
```diff
@@ -121,6 +143,90 @@ BOOST_FIXTURE_TEST_CASE(ProductivityStats, WorldFixtureEmpty1P)
...
+    BOOST_TEST(coalMine->GetProductivity() == 75u);
+
+    world.SetResource(minePos, Resource());
+    BOOST_TEST(coalMine->GetProductivity() == 0u);
+
+    ggs.setSelection(AddonId::COALMINE_RESOURCE_BEHAVIOR, static_cast<unsigned>(MineResourceBehavior::Inexhaustible));
+    BOOST_TEST(coalMine->GetProductivity() == 100u);
+}
+
+BOOST_FIXTURE_TEST_CASE(MineProductivityUsesAllMatchingResourcesWithinMineRadius, WorldFixtureMineRadius1P)
```
Flamefire: It looks like this test doesn't add anything to the one above: You also only add resources to 2 points.
You could use a lambda that distributes a given amount randomly over all mine points (set all others to zero)

tests/s25Main/integration/testGamePlayer.cpp
```diff
@@ -121,6 +143,90 @@ BOOST_FIXTURE_TEST_CASE(ProductivityStats, WorldFixtureEmpty1P)
...
+
+    const MapPoint outOfRangePt = world.GetNeighbour(
+      world.GetNeighbour(world.GetNeighbour(minePos, Direction::East), Direction::East), Direction::East);
+    const bool isOutOfRange = std::find(inRangePts.begin(), inRangePts.end(), outOfRangePt) == inRangePts.end();
+    BOOST_TEST_REQUIRE(isOutOfRange);
+    world.SetResource(outOfRangePt, Resource(ResourceType::Coal, 15));
+    BOOST_TEST(GetRemainingMineResources(world, minePos, ResourceType::Coal) == 10u);
+    BOOST_TEST(coalMine->GetProductivity() == 50u);
+
+    setProductivity(coalMine, 99);
```
Flamefire: Why that? Setting it to 100 at the start is mostly clear, although could use a comment. But this looks random

tests/s25Main/integration/testProduction.cpp
```diff
@@ -23,6 +26,76 @@ static std::ostream& operator<<(std::ostream& os, const PostCategory& cat)
 
 BOOST_AUTO_TEST_SUITE(Production)
 
+namespace {
+GoodType GetMineGoodType(const BuildingType mineType)
```
Flamefire: This duplicates a method in nofMiner. Maybe make that public or constexpr in header?

tests/s25Main/integration/testProduction.cpp
```diff
@@ -23,6 +26,76 @@ static std::ostream& operator<<(std::ostream& os, const PostCategory& cat)
...
+        MapPoint minePos = hqPos + MapPoint(2, 0);
+        const auto* mine = static_cast<nobUsual*>(
+          BuildingFactory::CreateBuilding(world, BuildingType::GraniteMine, minePos, curPlayer, Nation::Romans));
+        BuildRoad(world.GetNeighbour(minePos, Direction::SouthEast), false, std::vector<Direction>(2, Direction::West));
+        RTTR_EXEC_TILL(500, mine->HasWorker());
+        return minePos;
+    }
+};
+
+struct MineProductionFixture : WorldWithGCExecution1P
```
Flamefire, line 38: Can you use this as the base for `GraniteMineWithoutResourcesFixture` to use the functions for simplicity and consistency. Especially `CreateGraniteMineWithoutResources` looks like it should be code in a ctor. Isn't that the point for having a fixture?
Similar for this one: Shouldn't it setup something?

tests/s25Main/integration/testProduction.cpp
```diff
@@ -102,4 +175,227 @@ BOOST_FIXTURE_TEST_CASE(MetalWorkerOrders, WorldWithGCExecution1P)
...
+
+BOOST_FIXTURE_TEST_CASE(CoalMineS4LikeNoOutputGraniteFallback25ProducesStones, MineProductionFixture)
+{
+    ggs.setSelection(AddonId::COALMINE_RESOURCE_BEHAVIOR,
+                     static_cast<unsigned>(MineResourceBehavior::S4LikeExhaustion));
+    ggs.setSelection(AddonId::MINE_NO_OUTPUT_FALLBACK, static_cast<unsigned>(MineNoOutputFallback::ProduceGranite25));
+    const MapPoint minePos = CreateMine(BuildingType::CoalMine, Resource(ResourceType::Coal, 1));
+    const Inventory& curInventory = world.GetPlayer(curPlayer).GetInventory();
+    const unsigned initialCoal = curInventory[GoodType::Coal];
+    const unsigned initialStones = curInventory[GoodType::Stones];
```
Flamefire: Looks like those could be set in the fixture instead of all tests

With the [above comment ](https://github.com/Return-To-The-Roots/s25client/pull/1942/changes#r3652367523) the tests could be:
- Set resources
- Skip X GFs
- Assert number of stones and coal with a comment what/why that number is expected
- Opt. Assert number of resources in ground

Point is that currently CoalMineS4LikeNoOutputGraniteFallback25ProducesStones and CoalMineS4LikeNoOutputGraniteFallback50ProducesStones seem to be not related: They settings could be handled the same and both tests would still succeed.

This could even be a dataset-testcase: Run literally the same code for all the settings instead of copying it and then do assertions based on the setting. That would be the clearest to highlight the expected differences which is currently hard to see

tests/s25Main/integration/testProduction.cpp
```diff
@@ -23,6 +26,76 @@ static std::ostream& operator<<(std::ostream& os, const PostCategory& cat)
...
+        AddMinerSupplies();
+        MapPoint minePos;
+        const nobUsual* mine = PlaceMine(mineType, minePos);
+        if(initialResource.getType() != ResourceType::Nothing)
+            world.GetNodeWriteable(minePos).resources = initialResource;
+        ConnectMineAndWaitForWorker(minePos, mine);
+        return minePos;
+    }
+
+    void ResetMineProductionRng(const unsigned seed) { RANDOM.Init(seed); }
```
Flamefire: Maybe also set this in the ctor

tests/s25Main/integration/testProduction.cpp
```diff
@@ -102,4 +175,227 @@ BOOST_FIXTURE_TEST_CASE(MetalWorkerOrders, WorldWithGCExecution1P)
...
+{
+    ggs.setSelection(AddonId::GRANITEMINE_RESOURCE_BEHAVIOR,
+                     static_cast<unsigned>(MineResourceBehavior::WorkEverywhere));
+    const MapPoint minePos = CreateMine(BuildingType::GraniteMine, Resource(ResourceType::Coal, 4));
+    const Inventory& curInventory = world.GetPlayer(curPlayer).GetInventory();
+    const unsigned initialStones = curInventory[GoodType::Stones];
+
+    RTTR_EXEC_TILL(2000, curInventory[GoodType::Stones] > initialStones);
+    BOOST_TEST(world.GetNode(minePos).resources.has(ResourceType::Coal));
+    BOOST_TEST(world.GetNode(minePos).resources.getAmount() == 4u);
```
Flamefire: If you don't set that in the ctor/`CreateMine`, this could be like:
```
const Resource initCoal(ResourceType::Coal, 4);
world...setResource(minePos, initCoal);
...
BOOST:TEST(...resources == initCoal)
```
That avoids the magic number and is clearer in "this is unchanged"

tests/s25Main/integration/testProduction.cpp
```diff
@@ -102,4 +175,227 @@ BOOST_FIXTURE_TEST_CASE(MetalWorkerOrders, WorldWithGCExecution1P)
...
+    const MapPoint minePos = CreateMine(BuildingType::CoalMine, Resource(ResourceType::Coal, 4));
+    const Inventory& curInventory = world.GetPlayer(curPlayer).GetInventory();
+    const unsigned initialCoal = curInventory[GetMineGoodType(BuildingType::CoalMine)];
+
+    RTTR_EXEC_TILL(5000, curInventory[GoodType::Coal] > initialCoal);
+
+    BOOST_TEST(world.GetNode(minePos).resources.getAmount() == 4u);
+}
+
+BOOST_FIXTURE_TEST_CASE(CoalMineS4LikeExhaustionCanProduceNothing, MineProductionFixture)
```
Flamefire, line 244: This doesn't checkout with the above:
- in 5000 GFs something is produced
- in 2000 GFs with a different seed and amount nothing is produced

So this should at least be the same number (e.g. constant in fixture) and seed (e.g. set in ctor with a comment that it is mean for comparison between test cases)

tests/s25Main/integration/testProduction.cpp
```diff
@@ -102,4 +175,227 @@ BOOST_FIXTURE_TEST_CASE(MetalWorkerOrders, WorldWithGCExecution1P)
...
+    const unsigned initialStones = curInventory[GoodType::Stones];
+
+    ResetMineProductionRng(21);
+    RTTR_EXEC_TILL(5000, curInventory[GoodType::Coal] > initialCoal);
+
+    BOOST_TEST(curInventory[GoodType::Stones] == initialStones);
+    BOOST_TEST(world.GetNode(minePos).resources.getAmount() == 14u);
+}
+
+BOOST_FIXTURE_TEST_CASE(CoalMineDefaultProductionIgnoresNoOutputFallback, MineProductionFixture)
```
Flamefire, line 323: Similar: Tests seem to test for differences but are so different in the setup that it isn't clear why/when they'd succeed

tests/s25Main/integration/testSerialization.cpp
```diff
@@ -201,6 +216,66 @@ BOOST_AUTO_TEST_CASE(SerializeGGS)
...
+        {BuildingType::GraniteMine, BuildingType::CoalMine, BuildingType::IronMine, BuildingType::GoldMine})
+        BOOST_TEST(GetMineResourceBehavior(ggsLoaded, mineType) == MineResourceBehavior::Inexhaustible);
+    BOOST_TEST(ggsLoaded.getSelection(AddonId::INEXHAUSTIBLE_MINES) == 0u);
+}
+
+BOOST_AUTO_TEST_CASE(LegacyInexhaustibleMinesDeserializeAlsoMigratesLegacyGraniteAddonValue)
+{
+    // Old data always contains a value for 0x00800000 (formerly INEXHAUSTIBLE_GRANITEMINES, now
+    // GRANITEMINE_RESOURCE_BEHAVIOR). A disabled granite addon must not prevent the legacy global setting from
+    // making granite mines inexhaustible, which is what the old code did.
```
Flamefire: The magic number doesn't make sense in the comment as it isn't in the test.

Also which "old code"?

Flamefire (changes requested):
Ok, now mostly some changes to make the test more self-documenting and clean. A few things look odd there.

Also `LuaInterfaceGameBase::GetFeatureLevel` must be increased to cater for the new constants and this change should be documented too

DevOpsOfChaos:
Addressed all review points:

- **GetFeatureLevel** bumped 6 → 7 with documenting comment
- **nofMiner::ProduceWare** extracted shared  — test and production now reuse it
- **testAI.cpp** — renamed , added motivation comments, combined granite tests, replaced magic  with 
- **testGamePlayer.cpp** — dropped dead BQ search, introduced  with / lambdas, added  constant
- **testProduction.cpp** — merged fixtures into , named constants replace magic numbers, fallback tests collapsed into , exhaustion tests paired with shared constants
- **testSerialization.cpp** — rewrote misleading magic-number comment to reference actual symbol

Locally verified with GCC 14 and Clang 18 as before — ctest run pending submodule init. Ready for re-review.

tests/s25Main/integration/testAI.cpp
```diff
@@ -109,6 +146,83 @@ BOOST_FIXTURE_TEST_CASE(AIChat, EmptyWorldFixture2P)
...
+
+    // WorkEverywhere lets the AI place a granite mine on any mineable node, but only for granite, not coal.
+    BOOST_TEST(ai.FindBestPosition(searchCenter, AIResource::Granite, BuildingQuality::Mine, 5).isValid());
+    BOOST_TEST(!ai.FindBestPosition(searchCenter, AIResource::Coal, BuildingQuality::Mine, 5).isValid());
+
+}
+
+BOOST_FIXTURE_TEST_CASE(InexhaustibleGraniteDoesNotImplyWorkEverywhereForAI, EmptyWorldFixture1P)
+{
+    // Same setup as above, but the AI is constructed AFTER Inexhaustible is selected.
```
Flamefire: This comment doesn't make sense: It is the same setup except for MineResourceBehavior::Inexhaustible

What about my earlier idea of combining those tests to show the difference of the 2 settings? Didn't that work?

tests/s25Main/integration/testGamePlayer.cpp
```diff
@@ -55,6 +58,55 @@ RTTR_ATTRIBUTE_NO_UBSAN(vptr) void setProductivity(nobUsual* bld, unsigned short
...
+using WorldFixtureMineRadius1P = WorldFixture<CreateEmptyWorld, 1, 20, 12>;
+
+namespace {
+// S4-like mine productivity reaches the mine's base productivity once this many matching resources remain in the
+// mine radius, degrading linearly below it (see GetS4LikeMineProductionChance).
+constexpr unsigned S4LIKE_FULL_PRODUCTIVITY_AMOUNT = 20;
+
+// Places a coal mine and drives its S4-like productivity purely through the resources in its radius.
+// The 20x12 map is larger than 2*MINER_RADIUS in each dimension, so the mine radius never wraps onto itself.
+struct MineProductivityFixture : WorldFixtureMineRadius1P
```
Flamefire: Inline `WorldFixtureMineRadius1P` instead of making it a typedef above as you reference the size here it makes it clearer to have it here and the alias is only used here, isn't it? You could even use 2*MINER_RADIUS in defining the size which makes the comment shorter: Size chosen such that mine radius never wraps onto itself.

tests/s25Main/integration/testGamePlayer.cpp
```diff
@@ -55,6 +58,55 @@ RTTR_ATTRIBUTE_NO_UBSAN(vptr) void setProductivity(nobUsual* bld, unsigned short
...
+{
+    nobUsual* coalMine;
+    MapPoint minePos;
+
+    MineProductivityFixture()
+    {
+        // BuildingFactory::CreateBuilding ignores the building quality, so any node works; offset from the HQ keeps
```
Flamefire: The first part sounds like an AI update comment and can be removed, can't it?
The 2nd part "offset from the HQ keeps the mine radius clear of the HQ": What does that mean and why is this relevant?
Same above

tests/s25Main/integration/testGamePlayer.cpp
```diff
@@ -55,6 +58,55 @@ RTTR_ATTRIBUTE_NO_UBSAN(vptr) void setProductivity(nobUsual* bld, unsigned short
...
+        // the mine radius clear of the HQ. The empty world has no resources, but clear the radius to be explicit.
+        minePos = world.MakeMapPoint(world.GetPlayer(0).GetHQPos() + Position(4, 0));
+        coalMine = static_cast<nobUsual*>(
+          BuildingFactory::CreateBuilding(world, BuildingType::CoalMine, minePos, 0, Nation::Romans));
+        for(const MapPoint pt : world.GetPointsInRadiusWithCenter(minePos, MINER_RADIUS))
+            world.SetResource(pt, Resource());
+    }
+
+    // Puts coal on the mine node and its eastern neighbor (0 == none), leaving the rest of the radius empty.
```
Flamefire:
```suggestion
    // Puts given amounts of coal on the mine node and its eastern neighbor
```
0=none is clear, and it does NOT leave the rest **empty**, it doesn't touch it at all which is not expected anyway

tests/s25Main/integration/testGamePlayer.cpp
```diff
@@ -121,6 +173,63 @@ BOOST_FIXTURE_TEST_CASE(ProductivityStats, WorldFixtureEmpty1P)
...
+    ggs.setSelection(AddonId::COALMINE_RESOURCE_BEHAVIOR,
+                     static_cast<unsigned>(MineResourceBehavior::S4LikeExhaustion));
+
+    // A different resource type in range and matching coal just outside the radius must not count as coal.
+    world.SetResource(world.GetNeighbour(minePos, Direction::NorthWest), Resource(ResourceType::Iron, 15));
+    const MapPoint outOfRangePt = world.GetNeighbour(
+      world.GetNeighbour(world.GetNeighbour(minePos, Direction::East), Direction::East), Direction::East);
+    const auto inRangePts = world.GetPointsInRadiusWithCenter(minePos, MINER_RADIUS);
+    const bool outOfRange = std::find(inRangePts.begin(), inRangePts.end(), outOfRangePt) == inRangePts.end();
+    BOOST_TEST_REQUIRE(outOfRange);
```
Flamefire: Why not:

How about simply making this out of range by construction?
```suggestion
    const MapPoint outOfRangePt = world.MakeMapPoint(minePos + Position(MINER_RADIUS + 1, 0));
```
BTW: We have `helpers::find` and `helpers::contains` to easily work with containers

tests/s25Main/integration/testProduction.cpp
```diff
@@ -102,4 +168,208 @@ BOOST_FIXTURE_TEST_CASE(MetalWorkerOrders, WorldWithGCExecution1P)
...
+    BOOST_TEST(world.GetNode(minePos).resources.getAmount() == foreignDeposit.getAmount());
+}
+
+BOOST_FIXTURE_TEST_CASE(CoalMineInexhaustibleBehaviorDoesNotDepleteResource, MineProductionFixture)
+{
+    ggs.setSelection(AddonId::COALMINE_RESOURCE_BEHAVIOR, static_cast<unsigned>(MineResourceBehavior::Inexhaustible));
+    const Resource initCoal(ResourceType::Coal, 4);
+    const MapPoint minePos = CreateMine(BuildingType::CoalMine, initCoal);
+    const Inventory& curInventory = world.GetPlayer(curPlayer).GetInventory();
+    const unsigned initialCoal = curInventory[GetMineOutput(BuildingType::CoalMine)];
```
Flamefire: You ask for coal, so you can use coal directly, especially given that you compare this against coal
```suggestion
    const unsigned initialCoal = curInventory[GoodType::Coal];
```

tests/s25Main/integration/testSerialization.cpp
```diff
@@ -201,6 +216,67 @@ BOOST_AUTO_TEST_CASE(SerializeGGS)
...
+
+    for(const BuildingType mineType :
+        {BuildingType::GraniteMine, BuildingType::CoalMine, BuildingType::IronMine, BuildingType::GoldMine})
+        BOOST_TEST(GetMineResourceBehavior(ggsLoaded, mineType) == MineResourceBehavior::Inexhaustible);
+    BOOST_TEST(ggsLoaded.getSelection(AddonId::INEXHAUSTIBLE_MINES) == 0u);
+}
+
+BOOST_AUTO_TEST_CASE(LegacyInexhaustibleMinesDeserializeAlsoMigratesLegacyGraniteAddonValue)
+{
+    // Old data always contains an explicit (usually disabled) value for GRANITEMINE_RESOURCE_BEHAVIOR, because that
```
Flamefire:
```suggestion
    // Old data always contains an explicit value for GRANITEMINE_RESOURCE_BEHAVIOR, because that
```

Flamefire (changes requested):
Sorry for the delay, vacation time and such...
I went over the changes and resolved adressed comments. A few are still open or new, but should be quick to resolve.
And it needs a clang-format run, see CI

DevOpsOfChaos:
No worries about the delay, hope the vacation was good.

All the open points are in, plus the clang-format run.

On the AI tests: combining them does work, it just needs a fresh AI per setting. `AIPlayerJH` rates the nodes once on construction, so flipping the addon afterwards had no effect and the assertion was silently testing nothing — that was the reason for the split. It is one test again now, with a small lambda that builds its own AI per behavior, so the WorkEverywhere/Inexhaustible difference sits side by side.

One thing I left alone: the map size in `MineProductivityFixture`. I inlined the alias and shortened the comment, but kept the literal 20x12 — with `MINER_RADIUS == 2` anything expressed in terms of it either gets too small for the HQ offset the test relies on, or needs fairly arbitrary multipliers. Happy to change it if you had a particular form in mind.

One caveat: I could not build locally this time (the machine I am on has neither MSVC nor Boost set up), so the test run is on CI here.

DevOpsOfChaos:
Small correction to my previous comment: I did get a local build working after all, so the caveat at the end of it is outdated.

Built and ran on Linux with GCC 15.2 and Boost 1.90:

- `ctest`: 20/20 passed (incl. Test_UI, Test_lua, Test_drivers)
- `Test_integration` on its own: 215/215 test cases, 93537 assertions
- Production 17/17, AI 10/10, Serialization 11/11

The combined AI test and the reworked out-of-range check in `testGamePlayer` both pass — so combining the two AI tests really does work as long as each behavior gets its own AI instance.

One thing you may want to know: the Actions runs for b066a8ea5 are currently sitting in `action_required`, so they need your approval before CI picks the commit up.

tests/s25Main/integration/testAI.cpp
```diff
@@ -109,6 +146,72 @@ BOOST_FIXTURE_TEST_CASE(AIChat, EmptyWorldFixture2P)
...
+        AIJH::AIPlayerJH ai(0, world, AI::Level::Hard);
+        initAIJhNodes(ai, world, 0);
+        return ai.FindBestPosition(searchCenter, res, BuildingQuality::Mine, 5).isValid();
+    };
+
+    // WorkEverywhere lets the AI place a granite mine on any mineable node, but only for granite, not coal.
+    BOOST_TEST(findsMineSpotFor(MineResourceBehavior::WorkEverywhere, AIResource::Granite));
+    BOOST_TEST(!findsMineSpotFor(MineResourceBehavior::WorkEverywhere, AIResource::Coal));
+    // Inexhaustible does not imply "work everywhere": without an actual deposit there is still no spot.
+    BOOST_TEST(!findsMineSpotFor(MineResourceBehavior::Inexhaustible, AIResource::Granite));
```
Flamefire, line 212: For reference with the above asserts maybe:
```suggestion
    BOOST_TEST(!findsMineSpotFor(MineResourceBehavior::Inexhaustible, AIResource::Granite));
    BOOST_TEST(!findsMineSpotFor(MineResourceBehavior::Inexhaustible, AIResource::Coal));
```

And the opposite is missing: Have a single split with granite. Or is there another AI test for `Inexhaustible` granite with some granite somewhere?

If not I'd suggest to place granite somewhere , e.g. neighbor to searchCenter and test:
```
    BOOST_TEST(findsMineSpotFor(MineResourceBehavior::Inexhaustible, AIResource::Granite));
    BOOST_TEST(!findsMineSpotFor(MineResourceBehavior::WorkEverywhere, AIResource::Coal));
```

tests/s25Main/integration/testGamePlayer.cpp
```diff
@@ -55,6 +58,54 @@ RTTR_ATTRIBUTE_NO_UBSAN(vptr) void setProductivity(nobUsual* bld, unsigned short
...
+// Places a coal mine and drives its S4-like productivity purely through the resources in its radius.
+// Size chosen such that the mine radius never wraps onto itself.
+struct MineProductivityFixture : WorldFixture<CreateEmptyWorld, 1, 20, 12>
+{
+    nobUsual* coalMine;
+    MapPoint minePos;
+
+    MineProductivityFixture()
+    {
+        // Offset > MINER_RADIUS so no node of the mine radius is covered by the HQ and each of them can hold coal.
```
Flamefire, line 76: Why not use MINER_RADIUS here? This assumption (Is it `4 > MINER_RADIUS`?) may break in the future
And I don't understand why this is a requirement at all: What does the HQ has to do with this and why wouldn't a node be able to hold coal? I.e. what is the difference between the nodes at `hq + 4` and `hq + 3`?

Flamefire (changes requested):
> On the AI tests: combining them does work, it just needs a fresh AI per setting. AIPlayerJH rates the nodes once on construction, so flipping the addon afterwards had no effect and the assertion was silently testing nothing — that was the reason for the split. It is one test again now, with a small lambda that builds its own AI per behavior, so the WorkEverywhere/Inexhaustible difference sits side by side.

Great! Makes sense of course but isn't immediately clear. Can you put a comment with that finding there? E.g. `AI assumes game settings do not change, so need to create it (a new instance) after the settings are changed`

I left a previous comment there on also testing `Inexhaustible`. That is currently incomplete: It only tests that there are no places when inexhaustible is set on NO resources, but not that there are places if there are resources. That's a trivial extension to the test. So can you add that? Unless I missed that this is already tested somewhere else. But it surely does make sense to add the 3-4 lines there to highlight the differences.

> One thing I left alone: the map size in MineProductivityFixture. I inlined the alias and shortened the comment, but kept the literal 20x12 — with MINER_RADIUS == 2 anything expressed in terms of it either gets too small for the HQ offset the test relies on, or needs fairly arbitrary multipliers. Happy to change it if you had a particular form in mind.

I would have imagined something like `MINER_RADIUS * 2 + x`. Otherwise you'd need an assertion that `size.x/y > MINER_RADIUS * 2` (if that's what you meant) to ensure your assumption does not break.
It basically comes down to non-local assumptions: Someone changing `MINER_RADIUS` might not be aware of tests relying on it. See also the new comment below that.

> One thing you may want to know: the Actions runs for https://github.com/Return-To-The-Roots/s25client/commit/b066a8ea5c3beb56836099b7cbc4a224f6514471 are currently sitting in action_required, so they need your approval before CI picks the commit up.

@Flow86 I think it makes sense to disable that requirement until we see the need for that by e.g. a flood of trash PRs. Currently it seems more like a nuisance.

PR #1944, opened 2026-06-01
Update languages submodule to latest commit

MichalLabuda:
Update languages submodule to latest commit (2b500b9)

Latest languages commit resolves a buggy polish translation which in turn causes Test_Integration failure on a Windows PC with Polish regional settings.


Spikeone (approved):

MichalLabuda:
@Flamefire Hey! Unit test failed on Codecov trying to check GPG signature (log below). From what I could read, it happens from time to time but according to the last activity in https://github.com/codecov/codecov-action/issues/1876 some fix was introduced. Anyway, I'm unable to restart jobs or actually do anything with this so I'm just letting you know. Thanks.

`gpg: Can't check signature: No public key
==> Could not verify signature. Please contact Codecov if problem continues
    Exiting...
`

Issue #129, opened 2015-08-01
[addon] Visible workarea for buildings


Spikeone:
How it should work:
- when you click on a building (woodcutter's house, quarry, farm, catapult, etc) you should see the 'workarea' of that building
- workarea is an option which should show you the range of each building
- you can see but you can't change the workarea

How to enable this addon:
- this addon should be optional. Should be available like other RttR addons (in options).

Why this addon:
Many times you are curious what is the range of your catapult, how far your woodcutter can go, etc. This option is missing from the very beginning. Should be very helpful for each player.

https://bugs.launchpad.net/s25rttr/+bug/1394118

K4T:
Awesome idea, I would like to see this in game too.

Flow86:
+1

MarcusSt:
Is clicking the right way? Shouldn't it be on mouse over or sth? Otherwise, when will the area disappear? Is it associated with the respective window? Do I want so see the areas of all military buildings I have windows open for?

I like the general idea, but we should discuss about the way this can be implemented.

Spikeone:
Well, there are several good ideas I guess but two are best for me:
Hotkey for displaying workareas (just like space for BQ) so for example CTRL + SPACE = Civil buildings working space, ALT + SPACE = Military building working space.
or
You have a button in each building where you can manually enable/disable this.

Maybe both can be  implemented at the same time.

MarcusSt:
How do you define "military building working space"?

Spikeone:
As combat reach (full and max), for building there is already an addon "military aid"

RazrFalcon:
This will be very useful for farms.

Spikeone:
From #1228:
I assume that the work area for a particular building is centered on the building itself, and then extends out to a certain radius. In other words, the tree planter will plant trees __ meters from his building. It would be nice if one could move the work area. If enough trees have been planted with the building in the center, move the center of the work area off of the building to a new location (within reason) so he can keep working.

Thank you for considering this. :)

Flamefire:
Well, this is Settlers 2. So while I'd agree in showing the work area as a convenience changing it would be to much change to the game especially as the work required would be huge. So similar to allowing settlers to walk freely I'd refer players wanting this to settlers 3 and 4.

Spikeone:
Guess we can all agree, that this is more settlers3/4 than 2 where everything is indirect.
[...]

PR #1946, opened 2026-06-15
Radius overlay

morganchristiansson:
Add tooltip and overlay for building ranges

image
image
image
image

### DIFF
--- a/libs/s25main/GlobalGameSettings.cpp
+++ b/libs/s25main/GlobalGameSettings.cpp
@@ -156,7 +156,7 @@ const GlobalGameSettings::AddonWithState* GlobalGameSettings::getAddon(AddonId i
 bool GlobalGameSettings::isEnabled(AddonId id) const
 {
     const auto* addon = getAddon(id);
-    return addon && addon->status != 0;
+    return addon && addon->status != addon->addon->getDefaultStatus();
 }
 
 unsigned GlobalGameSettings::getSelection(AddonId id) const
--- a/libs/s25main/GlobalGameSettings.cpp
+++ b/libs/s25main/GlobalGameSettings.cpp
@@ -156,7 +156,7 @@ const GlobalGameSettings::AddonWithState* GlobalGameSettings::getAddon(AddonId i
 bool GlobalGameSettings::isEnabled(AddonId id) const
 {
     const auto* addon = getAddon(id);
-    return addon && addon->status != addon->addon->getDefaultStatus();
+    return addon && addon->status != 0;
 }
 
 unsigned GlobalGameSettings::getSelection(AddonId id) const
--- a/libs/s25main/GlobalGameSettings.cpp
+++ b/libs/s25main/GlobalGameSettings.cpp
@@ -108,7 +108,8 @@ void GlobalGameSettings::registerAllAddons()
         AddonForesterFarmFieldAvoidance,
         AddonForesterReachRadius,
         AddonWoodcutterReachRadius,
-        AddonStonemasonReachRadius
+        AddonStonemasonReachRadius,
+        AddonBuildingRadius
     >;
     // clang-format on
     using namespace boost::mp11;
--- a/libs/s25main/addons/AddonBuildingRadius.h
+++ b/libs/s25main/addons/AddonBuildingRadius.h
@@ -16,6 +16,7 @@ public:
     AddonBuildingRadius()
         : AddonBool(AddonId::BUILDING_RADIUS, AddonGroup::GamePlay, _("Show building radius"),
                     _("Shows the working radius of buildings in the build menu tooltip and as an overlay on the map "
-                      "when hovering over a building icon or selecting a building."))
+                      "when hovering over a building icon or selecting a building."),
+                    1) // Enabled by default
     {}
 };
--- a/libs/s25main/addons/AddonBuildingRadius.h
+++ b/libs/s25main/addons/AddonBuildingRadius.h
@@ -1,0 +1,21 @@
+// Copyright (C) 2005 - 2026 Settlers Freaks (sf-team at siedler25.org)
+//
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#pragma once
+
+#include "AddonBool.h"
+#include "mygettext/mygettext.h"
+
+/**
+ *  Show building radius information in tooltips and as map overlay
+ */
+class AddonBuildingRadius : public AddonBool
+{
+public:
+    AddonBuildingRadius()
+        : AddonBool(AddonId::BUILDING_RADIUS, AddonGroup::GamePlay, _("Show building radius"),
+                    _("Shows the working radius of buildings in the build menu tooltip and as an overlay on the map "
+                      "when hovering over a building icon or selecting a building."))
+    {}
+};
--- a/libs/s25main/desktops/dskGameInterface.cpp
+++ b/libs/s25main/desktops/dskGameInterface.cpp
@@ -750,6 +750,9 @@ bool dskGameInterface::Msg_LeftUp(const MouseCoords& mc)
 
 bool dskGameInterface::Msg_MouseMove(const MouseCoords& mc)
 {
+    // Update radius preview for buildings under the cursor (event-driven, not in draw loop)
+    gwv.UpdateRadiusPreviewForMousePos(mc.pos);
+
     if(!isScrolling)
     {
         if(mc.num_tfingers == 1)
--- a/libs/s25main/figures/nofHunter.cpp
+++ b/libs/s25main/figures/nofHunter.cpp
@@ -132,16 +132,14 @@ void nofHunter::HandleDerivedEvent(unsigned /*id*/)
 void nofHunter::TryStartHunting()
 {
     // Find animals in a square around building (actually should be circle, but animals are moving anyway)
-    const int SQUARE_SIZE = 19;
-
     // Liste mit den gefundenen Tieren
     std::vector<noAnimal*> available_animals;
 
     // Durchgehen und nach Tieren suchen
     Position curPos;
-    for(curPos.y = pos.y - SQUARE_SIZE; curPos.y <= pos.y + SQUARE_SIZE; ++curPos.y)
+    for(curPos.y = pos.y - HUNTER_SEARCH_HALFSIDE; curPos.y <= pos.y + HUNTER_SEARCH_HALFSIDE; ++curPos.y)
     {
-        for(curPos.x = pos.x - SQUARE_SIZE; curPos.x <= pos.x + SQUARE_SIZE; ++curPos.x)
+        for(curPos.x = pos.x - HUNTER_SEARCH_HALFSIDE; curPos.x <= pos.x + HUNTER_SEARCH_HALFSIDE; ++curPos.x)
         {
             MapPoint curMapPos = world->MakeMapPoint(curPos);
 
--- a/libs/s25main/gameData/BuildingConsts.cpp
+++ b/libs/s25main/gameData/BuildingConsts.cpp
@@ -440,15 +440,23 @@ unsigned GetBuildingRadius(BuildingType bld)
         case BuildingType::CoalMine:
         case BuildingType::IronMine:
         case BuildingType::GoldMine: return MINER_RADIUS;
-        // For farmhand-based buildings, delegate to nofFarmhand::GetWorkRadius via the BLD_WORK_DESC
-        // job mapping. This is the single source of truth for worker reach radii, including
-        // addon-adjustable ranges (woodcutter, forester, stonemason).
-        default:
+        // Farmhand-based buildings — worker goes out to gather resources from the map.
+        // Map each building type to its job via BLD_WORK_DESC, then query the work
+        // radius from nofFarmhand::GetWorkRadius.
+        case BuildingType::Woodcutter:
+        case BuildingType::Forester:
+        case BuildingType::Fishery:
+        case BuildingType::Quarry:
+        case BuildingType::Farm:
+        case BuildingType::Vineyard:
+        case BuildingType::Charburner:
         {
             const auto job = BLD_WORK_DESC[bld].job;
             if(job)
                 return nofFarmhand::GetWorkRadius(*job);
             return 0;
         }
+        // All other building types have no relevant radius overlay
+        default: return 0;
     }
 }
--- a/libs/s25main/gameData/BuildingConsts.cpp
+++ b/libs/s25main/gameData/BuildingConsts.cpp
@@ -6,11 +6,10 @@
 #include "mygettext/mygettext.h"
 #include "gameData/MilitaryConsts.h"
 #include "gameData/GameConsts.h"
-#include "GlobalGameSettings.h"
-#include "addons/AddonWoodcutterReachRadius.h"
-#include "addons/AddonForesterReachRadius.h"
-#include "addons/AddonStonemasonReachRadius.h"
-#include "addons/const_addons.h"
+#include "figures/nofFarmhand.h"
+#include "figures/nofCatapultMan.h"
+#include "figures/nofHunter.h"
+#include "gameTypes/BuildingTypes.h"
 #include <type_traits>
 
 const helpers::EnumArray<const char*, BuildingType> BUILDING_NAMES = {
@@ -417,11 +416,11 @@ const helpers::MultiEnumArray<DrawPoint, Nation, BuildingType> BUILDING_ARMOR_SI
     return result;
 }();
 
-unsigned GetBuildingRadius(BuildingType bld, const GlobalGameSettings& ggs)
+unsigned GetBuildingRadius(BuildingType bld)
 {
     switch(bld)
     {
-        // Military buildings (territory influence radius)
+        // Military buildings (territory influence radius) — from MilitaryConsts.h
         case BuildingType::Barracks: return MILITARY_RADIUS[0];
         case BuildingType::Guardhouse: return MILITARY_RADIUS[1];
         case BuildingType::Watchtower: return MILITARY_RADIUS[2];
@@ -430,35 +429,26 @@ unsigned GetBuildingRadius(BuildingType bld, const GlobalGameSettings& ggs)
         case BuildingType::Headquarters: return HQ_RADIUS;
         // Harbor building
         case BuildingType::HarborBuilding: return HARBOR_RADIUS;
-        // Production buildings with (adjustable) reach radius
-        case BuildingType::Woodcutter:
-        {
-            const unsigned sel = ggs.getSelection(AddonId::WOODCUTTER_REACH_RADIUS);
-            return woodcutterRadiusValues[sel];
-        }
-        case BuildingType::Forester:
-        {
-            const unsigned sel = ggs.getSelection(AddonId::FORESTER_REACH_RADIUS);
-            return foresterRadiusValues[sel];
-        }
-        case BuildingType::Quarry:
-        {
-            const unsigned sel = ggs.getSelection(AddonId::STONEMASON_REACH_RADIUS);
-            return stonemasonRadiusValues[sel];
-        }
-        case BuildingType::Fishery: return FISHERY_RADIUS;
-        case BuildingType::Catapult: return CATAPULT_RANGE;
-        case BuildingType::LookoutTower: return VISUALRANGE_LOOKOUTTOWER; // Scouting range
-        // Farm, Charburner, and Hunter have fixed working/search radii
-        case BuildingType::Farm: return FARMER_RADIUS;
-        case BuildingType::Charburner: return CHARBURNER_RADIUS;
-        case BuildingType::Hunter: return HUNTER_SEARCH_RADIUS; // Search radius (square half-side 19)
-        // Mines extract resources from within a 2-tile radius (miner stays inside building)
+        // Lookout tower — scouting visibility range
+        case BuildingType::LookoutTower: return VISUALRANGE_LOOKOUTTOWER;
+        // Catapult attack range
+        case BuildingType::Catapult: return CATAPULT_MAX_TARGET_RANGE;
+        // Hunter searches for animals in a square of this half-side length
+        case BuildingType::Hunter: return HUNTER_SEARCH_HALFSIDE;
+        // Mines — miner stays inside and extracts from adjacent tiles
         case BuildingType::GraniteMine:
         case BuildingType::CoalMine:
         case BuildingType::IronMine:
         case BuildingType::GoldMine: return MINER_RADIUS;
-        // Buildings that don't have a notable radius
-        default: return 0;
+        // For farmhand-based buildings, delegate to nofFarmhand::GetWorkRadius via the BLD_WORK_DESC
+        // job mapping. This is the single source of truth for worker reach radii, including
+        // addon-adjustable ranges (woodcutter, forester, stonemason).
+        default:
+        {
+            const auto job = BLD_WORK_DESC[bld].job;
+            if(job)
+                return nofFarmhand::GetWorkRadius(*job);
+            return 0;
+        }
     }
 }
--- a/libs/s25main/gameData/BuildingConsts.cpp
+++ b/libs/s25main/gameData/BuildingConsts.cpp
@@ -5,9 +5,12 @@
 #include "BuildingConsts.h"
 #include "mygettext/mygettext.h"
 #include "gameData/MilitaryConsts.h"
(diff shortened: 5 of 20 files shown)


libs/s25main/gameData/BuildingConsts.h
```diff
@@ -119,3 +119,16 @@ constexpr std::array<DrawPoint, 4> SUPPRESS_UNUSED NUBIAN_MINE_FIRE = {{
 
 /// Hilfetexte für Gebäude
 extern const helpers::EnumArray<const char*, BuildingType> BUILDING_HELP_STRINGS;
+
+/// Reach radius for various production buildings (in tiles)
+constexpr unsigned FISHERY_RADIUS = 7;       ///< Fisher searches for fish within this radius
```
Flamefire: If we copy the values here the constants should be used by the respective classes too.
As most (all?) are in `figures` it might make sense to put them in `JobConsts` instead.

libs/s25main/gameData/BuildingConsts.h
```diff
@@ -119,3 +119,16 @@ constexpr std::array<DrawPoint, 4> SUPPRESS_UNUSED NUBIAN_MINE_FIRE = {{
...
+constexpr unsigned HUNTER_SEARCH_RADIUS = 19;  ///< Hunter searches for animals in a square of this half-side length
+constexpr unsigned FARMER_RADIUS = 2;         ///< Farmer plants and harvests crops within this radius
+constexpr unsigned CHARBURNER_RADIUS = 3;     ///< Charburner places and harvests charcoal piles within this radius
+constexpr unsigned CATAPULT_RANGE = 12;       ///< Catapult attack range
```
Flamefire: I see a `if(distance < 14)` check there. So why 12 here?

libs/s25main/ingameWindows/iwAction.h
```diff
@@ -7,6 +7,9 @@
 #include "IngameWindow.h"
 #include "variant.h"
 #include "gameTypes/MapCoordinates.h"
+#include <boost/optional.hpp>
```
Flamefire: Unsed?

libs/s25main/world/GameWorldView.cpp
```diff
@@ -3,9 +3,12 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 
 #include "world/GameWorldView.h"
+#include <boost/optional.hpp>
```
Flamefire: We are moving to `std::optional`, I opened a PR replacing the remaining boost::optional uses to avoid the confusion: https://github.com/Return-To-The-Roots/s25client/pull/1947

So use that instead please.

libs/s25main/world/GameWorldView.cpp
```diff
@@ -220,6 +223,36 @@ void GameWorldView::Draw(const RoadBuildState& rb, const MapPoint selected, bool
     if(show_names || show_productivity)
         DrawNameProductivityOverlay(terrainRenderer);
 
+    // Draw radius preview outline (if set via action window hover)
+    if(radiusPreview_)
+        DrawRadiusOutline(radiusPreview_->first, radiusPreview_->second);
+
+    // Draw radius outline for the building under the mouse cursor
+    if(!radiusPreview_ && mousePos.x >= 0 && mousePos.x < static_cast<int>(size_.x) && mousePos.y >= 0
```
Flamefire: Why the check for mousePos which isn't used below? And should be the `else` branch for the above

Flamefire (changes requested):
Good idea, makes it a bit S4-like.

@Spikeone Should this be an addon too?

libs/s25main/gameData/BuildingConsts.h
morganchristiansson: Google said 12. :-( I started with my own numbers.
This is fixed.
image

libs/s25main/gameData/BuildingConsts.h
morganchristiansson: I agree. There should be single source of truth for these radiuses.
This is fixed.

Catapult and Hunter did not have re-usable constants and they have now been extracted to constants.

libs/s25main/gameData/BuildingConsts.h
morganchristiansson: Also catapult was not available as constant, I extracted it to constant.

libs/s25main/world/GameWorldView.cpp
morganchristiansson: Replaced boost::optional → std::optional, boost::none → std::nullopt, and remove unused #include <boost/optional.hpp>

libs/s25main/world/GameWorldView.cpp
```diff
@@ -713,6 +746,41 @@ void GameWorldView::RemoveDrawNodeCallback(IDrawNodeCallback* callbackToRemove)
...
+        for(int dw : {-w, 0, w})
+        {
+            for(int dh : {-h, 0, h})
+            {
+                const MapPoint copyPt = MakeMapPoint(Position(basePt.x + dw, basePt.y + dh), mapSize);
+                const auto alt = world.GetNode(copyPt).altitude;
+                const DrawPoint scr = GetNodePos(copyPt) - DrawPoint(0, HEIGHT_FACTOR * alt) - offset;
+                Window::DrawRectangle(Rect(scr - DrawPoint(2, 2), Extent(5, 5)), BORDER_COLOR);
+            }
+        }
```
morganchristiansson: Noone commented on this yet.
The drawn radius was being clipped where map repeats and this is a workaround. Did not find any cleaner way advice appreciated.

Image

libs/s25main/world/GameWorldView.cpp
morganchristiansson: I simplified the UI code it was doing things that weren't needed.

I'm very new to this project thank you for feedback.

libs/s25main/figures/nofHunter.cpp
```diff
@@ -132,16 +132,14 @@ void nofHunter::HandleDerivedEvent(unsigned /*id*/)
 void nofHunter::TryStartHunting()
 {
     // Find animals in a square around building (actually should be circle, but animals are moving anyway)
```
morganchristiansson, line 134: Uhm hunters look for animals in a SQUARE?

Should fix to be circle? Kinda out of scope but can include or do separate PR. Cause right now the range outline doesn't match what game uses.

morganchristiansson: Created PR #1948 to make Hunters use circle instead of square.

libs/s25main/world/GameWorldView.cpp
Flamefire: That is indeed hard. You can use the `GetNodePos` overload taking the height as the base point. Then use `const Position mapDrawSize = world->GetSize() * Position(TR_W, TR_H);` as the offsets

I don't think it gets better than that: You have to draw possibly all copies when the zoom factor is large which it seems is also not yet taken into account. Possibly even that is not enough.
Maybe rather use a (flat_)set to store the radius points and move drawing to `DrawGUI` where you can check with `contains` whether to draw the mark

Flamefire:
Given #129 was a request for an addon, can you add one and guard the code changing the tooltip and setting the draw-circle by the addon setting? We usually keep the default behavior to be the same as in S2

BTW: Unless you just applied a suggested code snippet directly please don't resolve conversations as then we miss your reply

libs/s25main/gameData/BuildingConsts.h
Flamefire: Extracting constants is always good. You just have to make sure you actually use them after defining it.

@Spikeone I guess we should move to actually using radius for all jobs, not rectangles.

Spikeone: Probably yes - can check in the original but pretty sure thats it's circular

Spikeone:
> @Spikeone Should this be an addon too?

Yes, I think so

libs/s25main/gameData/BuildingConsts.h
Spikeone: > I see a `if(distance < 14)` check there. So why 12 here?

Isn't it the same reason our building range is +1? So the target node is included or something like that?

morganchristiansson:
I think it should be player toggleable during game, and per player in multiplayer.
It's just a fancy measuring tool so I think this should be fine.
Also I think the list of addons is getting large.

But most ppl said it should be addon so maybe best to proceed with that. 

But should tooltip also be guarded by addon?

> BTW: Unless you just applied a suggested code snippet directly please don't resolve conversations as then we miss your reply

Ok I will leave resolving to original commenter from now on. I just did it to track what I resolved.

libs/s25main/gameData/BuildingConsts.h
Flamefire: Which building range? `if(distance < 14)` means the building on the 13th node away can still be hit, so there are 12 nodes in-between. So maybe that is what is meant by google. And maybe the other part you are referring to has a corner case which required that. But would need to check

stefson:
nice idea, but please make it an optional addon. 

also, the red isn't very much in the style of S2,

morganchristiansson:
> nice idea, but please make it an optional addon.

it is now an addon

> also, the red isn't very much in the style of S2,

not sure what graphics to use, please provide/suggest.

libs/s25main/world/GameWorldView.cpp
morganchristiansson: I found that s25edit draws a brush under the cursor which is resizeable by numpad +/- and based the radius outline drawing on that!

morganchristiansson:
Ok all issues solved.
There's been calls for better graphics. I'm not a good artist and noone has stepped up. I think for measuring tool/hud the red rectangles are fine.

And made the addon enabled by default - mainly so I could use my existing save for testing.

morganchristiansson:
Found an edge case: it hovers game world buildings thru windows and will draw outline

morganchristiansson:
I have played many games with this enabled and it works perfectly. Hard to play without it.

Only remaining issue is to use custom graphics instead of red rectangles.

libs/s25main/gameData/BuildingConsts.cpp
```diff
@@ -409,3 +415,48 @@ const helpers::MultiEnumArray<DrawPoint, Nation, BuildingType> BUILDING_ARMOR_SI
...
+        case BuildingType::Farm:
+        case BuildingType::Vineyard:
+        case BuildingType::Charburner:
+        {
+            const auto job = BLD_WORK_DESC[bld].job;
+            if(job)
+                return nofFarmhand::GetWorkRadius(*job);
+            return 0;
+        }
+        // All other building types have no relevant radius overlay
```
Flamefire:
```suggestion
        // All other building types have no relevant work radius (e.g. only work inside the building)
```

libs/s25main/addons/AddonBuildingRadius.h
```diff
@@ -0,0 +1,22 @@
...
+ *  Show building radius information in tooltips and as map overlay
+ */
+class AddonBuildingRadius : public AddonBool
+{
+public:
+    AddonBuildingRadius()
+        : AddonBool(AddonId::BUILDING_RADIUS, AddonGroup::GamePlay, _("Show building radius"),
+                    _("Shows the working radius of buildings in the build menu tooltip and as an overlay on the map "
+                      "when hovering over a building icon or selecting a building."),
+                    1) // Enabled by default
```
Flamefire, line 20:
```suggestion
                      "when hovering over a building icon or selecting a building."), 0)
```

The convention is that the default is S2-like behavior, so you can "disable" all addons and get vanilla S2

libs/s25main/ingameWindows/iwAction.cpp
```diff
@@ -526,6 +536,30 @@ void iwAction::Msg_Group_TabChange(const unsigned /*group_id*/, const unsigned c
...
     IngameWindow::Msg_PaintAfter();
+
+    // Resolve building icon hover preview after all mouse events are processed
+    auto* mainTab = GetCtrl<ctrlTab>(0);
+    auto* buildTabCtrl =
+      (mainTab && mainTab->GetCurrentTab() == TAB_BUILD) ? mainTab->GetGroup(TAB_BUILD)->GetCtrl<ctrlTab>(1) : nullptr;
+    auto* bldGroup = buildTabCtrl ? buildTabCtrl->GetGroup(buildTabCtrl->GetCurrentTab()) : nullptr;
+
+    bool hasHoveredIcon = false;
+    if(bldGroup && gwv.GetWorld().GetGGS().isEnabled(AddonId::BUILDING_RADIUS))
```
Flamefire: This looks inefficient doing on every single draw call. What was wrong with the MouseMove-hovering?

libs/s25main/ingameWindows/iwAction.cpp
```diff
@@ -526,6 +536,30 @@ void iwAction::Msg_Group_TabChange(const unsigned /*group_id*/, const unsigned c
...
+    {
+        for(auto* icon : bldGroup->GetCtrls<ctrlBuildingIcon>())
+        {
+            const unsigned radius = GetBuildingRadius(icon->GetType());
+            if(icon->IsMouseOver() && radius > 0)
```
Flamefire: Shouldn't you break once you found ANY mouseover?

libs/s25main/world/GameWorldView.cpp
```diff
@@ -713,6 +748,60 @@ void GameWorldView::RemoveDrawNodeCallback(IDrawNodeCallback* callbackToRemove)
     drawNodeCallbacks.erase(itPos);
 }
 
+// -----------------------------------------------------------------------------
+// Snap a point to the nearest toroidal copy relative to a reference position.
+// Same formula as s25edit's correctMouseBlit():
```
Flamefire: Please don't include such references. The client should be self-contained and the code correct on its own. So rather explain why this specific code is correct as-is.
Also the arguments are not clear to me. Why pass 2 by const-ref? Should all be values.
Can mapPxSize ever be zero? When and why?
Without that it gets much easier to read as you can use e.g. `pt -ref` etc.

libs/s25main/world/GameWorldView.cpp
```diff
@@ -713,6 +748,60 @@ void GameWorldView::RemoveDrawNodeCallback(IDrawNodeCallback* callbackToRemove)
...
+void GameWorldView::DrawRadiusOutline(const MapPoint& center, unsigned radius)
+{
+    const auto& world = GetWorld();
+    auto pts = world.GetPointsInRadius(center, radius, ReturnMapPointWithRadius{});
+
+    const MapExtent mapSize = world.GetSize();
+    constexpr unsigned BORDER_COLOR = 0xFFFF0000;
+    const DrawPoint mapPxSize(mapSize.x * TR_W, mapSize.y * TR_H);
+
+    // Reference: screen position of the center vertex (with seam offset).
```
Flamefire: Why "reference"? Isn't this simply a position in screen coordinates? I.e. a `DrawPoint` from a `MapPoint`?

libs/s25main/world/GameWorldView.cpp
```diff
@@ -713,6 +748,60 @@ void GameWorldView::RemoveDrawNodeCallback(IDrawNodeCallback* callbackToRemove)
...
+    const auto centerAlt = world.GetNode(center).altitude;
+    const DrawPoint ref = GetNodePos(center) - DrawPoint(0, HEIGHT_FACTOR * centerAlt) + selPtOffset;
```
Flamefire: Use the overload:

```suggestion
    const DrawPoint ref = GetNodePos(center, world.GetNode(center).altitude) + selPtOffset;
```
Also where does `selPtOffset` come from? Would need a comment

libs/s25main/world/GameWorldView.cpp
```diff
@@ -713,6 +748,60 @@ void GameWorldView::RemoveDrawNodeCallback(IDrawNodeCallback* callbackToRemove)
...
+
+    for(const auto& [basePt, dist] : pts)
```
Flamefire:
```suggestion
    for(const auto& [pt, dist] : pts)
```
Or `mapPt` or similar, I suspected a function parameter or similar especially with `scr` not containing "Point" which I guess it should. Maybe `mapPt` and `screenPt`?

libs/s25main/GlobalGameSettings.cpp
```diff
@@ -155,7 +156,7 @@ const GlobalGameSettings::AddonWithState* GlobalGameSettings::getAddon(AddonId i
 bool GlobalGameSettings::isEnabled(AddonId id) const
 {
     const auto* addon = getAddon(id);
-    return addon && addon->status != addon->addon->getDefaultStatus();
```
Flamefire, line 158: This doesn't seem correct. Please revert

libs/s25main/world/GameWorldView.cpp
```diff
@@ -220,6 +224,37 @@ void GameWorldView::Draw(const RoadBuildState& rb, const MapPoint selected, bool
     if(show_names || show_productivity)
         DrawNameProductivityOverlay(terrainRenderer);
 
+    // Draw radius preview outline
+    if(radiusPreview_)
+        DrawRadiusOutline(radiusPreview_->first, radiusPreview_->second);
+
+    // Auto-detect radius for the building under the cursor.
+    // Do not trigger hover when the mouse is over an ingame window.
+    if(!radiusPreview_ && GetWorld().GetGGS().isEnabled(AddonId::BUILDING_RADIUS)
```
Flamefire: Shouldn't that be rather on the outside, e.g. in dskGameInterface where it could be more efficiently in a MouseMove handler. This way it is opt-in and e.g. doesn't enable for observation windows (suspicion only, haven't checked)
And whereever this is done I'd say it makes sense to cache `GetWorld().GetGGS().isEnabled(AddonId::BUILDING_RADIUS)` (effectively const) instead of getting this on every draw call.

Flamefire (changes requested):

libs/s25main/addons/AddonBuildingRadius.h
morganchristiansson: Yeah can revert this. Making it default enabled allowed me to continue using my save game for testing tho so it's been nice. Also playing campaign with it enabled is lovely.

I'm addicted I can't play without it.

Let's revert to default disable last thing before merge ok?

libs/s25main/GlobalGameSettings.cpp
morganchristiansson: This is to allow default enabled addons.
Ok I find it very useful to have it default enabled for testing. Can we revert this last thing before merge.

libs/s25main/ingameWindows/iwAction.cpp
morganchristiansson: There's an issue with icon->SetOnHoverChanged()
When hovering icons in forward direction, the leave event fires after enter event.

Maybe there should be 2 passes, first send leave event, then send enter.

morganchristiansson:
Addressed all feedback.
Ran clang-format.
24 commits so far getting a bit messy maybe could squash it.
Or I would've liked to split it but it's only usable and testable as a whole.

libs/s25main/desktops/dskGameInterface.cpp
```diff
@@ -750,6 +750,9 @@ bool dskGameInterface::Msg_LeftUp(const MouseCoords& mc)
 
 bool dskGameInterface::Msg_MouseMove(const MouseCoords& mc)
 {
+    // Update radius preview for buildings under the cursor (event-driven, not in draw loop)
```
Flamefire, line 753: Make sure to clean such "intermediate" comments from AI
```suggestion
    // Update radius preview for buildings under the cursor
```

libs/s25main/ingameWindows/iwAction.cpp
```diff
@@ -175,6 +177,13 @@ iwAction::iwAction(GameInterface& gi, GameWorldView& gwv, const Tabs& tabs, MapP
                 std::stringstream tooltip;
                 tooltip << _(BUILDING_NAMES[bld]);
 
+                // Radius anzeigen falls vorhanden
```
Flamefire, line 180: English comments only (at least for new/changed ones)

libs/s25main/ingameWindows/iwAction.cpp
```diff
@@ -175,6 +177,13 @@ iwAction::iwAction(GameInterface& gi, GameWorldView& gwv, const Tabs& tabs, MapP
...
+                unsigned radius = 0;
+                if(showBuildingRadius)
+                    radius = GetBuildingRadius(bld);
```
Flamefire, line 183: Prefer const variables:

```suggestion
                const unsigned radius = showBuildingRadius ? GetBuildingRadius(bld) : 0;
```

libs/s25main/ingameWindows/iwAction.cpp
```diff
@@ -186,8 +195,26 @@ iwAction::iwAction(GameInterface& gi, GameWorldView& gwv, const Tabs& tabs, MapP
                 }
 
                 DrawPoint iconPos((k % 5) * 36, (k / 5) * 36 + 45);
-                build_tab->GetGroup(static_cast<int>(bt))
-                  ->AddBuildingIcon(k, iconPos, bld, player.nation, 36, tooltip.str());
+                ctrlBuildingIcon* icon = build_tab->GetGroup(static_cast<int>(bt))
+                                           ->AddBuildingIcon(k, iconPos, bld, player.nation, 36, tooltip.str());
+
+                // Store hover callback; activeHoveredIcon_ guards against stale leaves from
+                // reversed child iteration order in Msg_MouseMove dispatch
```
Flamefire, line 202: Similar AI comment unless I'm missing something this is completely redundant

```suggestion
```

libs/s25main/world/GameWorldView.cpp
```diff
@@ -220,6 +225,10 @@ void GameWorldView::Draw(const RoadBuildState& rb, const MapPoint selected, bool
     if(show_names || show_productivity)
         DrawNameProductivityOverlay(terrainRenderer);
 
+    // Draw radius preview outline (set by icon-hover or map-hover)
```
Flamefire, line 228: this comment is about the caller and hence can easily be outdated. Avoid please

```suggestion
    // Draw radius preview outline if enabled
```

libs/s25main/world/GameWorldView.cpp
```diff
@@ -713,6 +722,81 @@ void GameWorldView::RemoveDrawNodeCallback(IDrawNodeCallback* callbackToRemove)
...
+        DrawPoint screenPt = GetNodePos(mapPt, world.GetNode(mapPt).altitude);
+        screenPt = SnapToNearestCopy(screenPt, centerScr, mapPxSize) - offset;
+
+        Window::DrawRectangle(Rect(screenPt - DrawPoint(2, 2), Extent(5, 5)), BORDER_COLOR);
+    }
+}
+
+void GameWorldView::UpdateRadiusPreviewForMousePos(const Position& mousePos)
+{
+    if(!isBuildingRadiusEnabled_ || WINDOWMANAGER.FindWindowAtPos(mousePos))
```
Flamefire, line 766: Hm, this looks like this check should rather be handled by the caller: You update by `selPt` not `mousePos` and that is only set/updated in the `Draw` call so likely to be stale by 1 frame anyway.
Passing a value into this class just to check it using a different layer (world vs windows) feels wrong

Maybe we can turn it around: Add a method that gets the building type at the selection which we can reuse in ` dskGameInterface::ContextClick` too, see existing code.
Do we really want to handle non-owned buildings? Feels like a cheat... If we do the code will be different so then I'd say a method `getSelectedBuildingType` here and handling the mouse pos/window-manager and buildings-enabled in the caller is cleaner.

libs/s25main/world/GameWorldView.cpp
```diff
@@ -713,6 +722,81 @@ void GameWorldView::RemoveDrawNodeCallback(IDrawNodeCallback* callbackToRemove)
     drawNodeCallbacks.erase(itPos);
 }
 
+// Snap a point to the nearest toroidal copy (map wrapping) relative to a reference.
+// k = round((reference - pt) / mapSize), pt += k * mapSize.
```
Flamefire, line 726: You can (pretty much) directly express this below like `k = PointF(ref - pt) / mapPxSize` (maybe a better name than k) then `pt.x += DrawPoint(round(k.x), round(k.y)) * mapPxSize`

And a short comment why this formula is correct/used would be good. I.e. what is this k and what does it mean. This way would keep us wondering later.

Flamefire (changes requested):

PR #1947, opened 2026-06-15
Replace `boost::optional` by `std::optional`

Flamefire:
Unify uses now that we require C++17

### DIFF
--- a/extras/ai-battle/main.cpp
+++ b/extras/ai-battle/main.cpp
@@ -16,8 +16,14 @@
 #include <boost/nowide/args.hpp>
 #include <boost/nowide/filesystem.hpp>
 #include <boost/nowide/iostream.hpp>
-#include <boost/optional.hpp>
 #include <boost/program_options.hpp>
+#if BOOST_VERSION >= 109000
+#    include <optional>
+using std::optional;
+#else
+#    include <boost/optional.hpp>
+using boost::optional;
+#endif
 
 namespace bnw = boost::nowide;
 namespace bfs = boost::filesystem;
@@ -28,8 +34,8 @@ int main(int argc, char** argv)
     bnw::nowide_filesystem();
     bnw::args _(argc, argv);
 
-    boost::optional<std::string> replay_path;
-    boost::optional<std::string> savegame_path;
+    optional<std::string> replay_path;
+    optional<std::string> savegame_path;
     unsigned random_init = static_cast<unsigned>(std::chrono::high_resolution_clock::now().time_since_epoch().count());
     unsigned random_ai_init = random_init;
 
--- a/libs/s25main/figures/nofDefender.h
+++ b/libs/s25main/figures/nofDefender.h
@@ -5,6 +5,7 @@
 #pragma once
 
 #include "nofActiveSoldier.h"
+#include <stdexcept>
 
 class nofAttacker;
 class nofPassiveSoldier;
--- a/libs/s25main/gameTypes/BuildingTypes.h
+++ b/libs/s25main/gameTypes/BuildingTypes.h
@@ -43,9 +43,9 @@ public:
 struct BldWorkDescription
 {
     /// Worker belonging to the building, if any
-    helpers::OptionalEnum<Job> job = boost::none;
+    helpers::OptionalEnum<Job> job = std::nullopt;
     /// Ware produced, if any
-    boost_variant2<GoodType, Job, boost::none_t> producedWare = boost::none;
+    boost_variant2<GoodType, Job, std::nullopt_t> producedWare = std::nullopt;
     // Required for use in aggregate initialization
     // NOLINTBEGIN(readability-redundant-member-init)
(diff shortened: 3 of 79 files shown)


extras/ai-battle/main.cpp
```diff
@@ -16,8 +16,14 @@
...
 #include <boost/nowide/iostream.hpp>
-#include <boost/optional.hpp>
 #include <boost/program_options.hpp>
+#if BOOST_VERSION >= 109000
+#    include <optional>
+using std::optional;
+#else
+#    include <boost/optional.hpp>
+using boost::optional;
+#endif
```
Flow86, line 26: do we really need this backward support?

libs/s25main/figures/nofDefender.h
Flow86: maybe move implementation into cpp file, then we dont need the `stdexcept`  include here

libs/s25main/gameTypes/BuildingTypes.h
```diff
@@ -43,9 +43,9 @@ class WaresNeeded
 struct BldWorkDescription
 {
     /// Worker belonging to the building, if any
-    helpers::OptionalEnum<Job> job = boost::none;
+    helpers::OptionalEnum<Job> job = std::nullopt;
     /// Ware produced, if any
-    boost_variant2<GoodType, Job, boost::none_t> producedWare = boost::none;
+    boost_variant2<GoodType, Job, std::nullopt_t> producedWare = std::nullopt;
```
Flow86, line 48: can't we use std::variant  too?

libs/s25main/ingameWindows/iwSettings.cpp
```diff
@@ -133,11 +133,10 @@ void iwSettings::Msg_ButtonClick(unsigned ctrl_id)
             SETTINGS.video.displayMode = DisplayMode(*GetCtrl<ctrlComboBox>(ID_cbDisplayMode)->GetSelection());
             SETTINGS.video.displayMode.resizeable = !grpWindowSize.GetCtrl<ctrlCheck>(ID_cbLockWindowSize)->isChecked();
             SETTINGS.interface.mapScrollMode =
-              static_cast<MapScrollMode>(GetCtrl<ctrlComboBox>(ID_cbMapScrollMode)->GetSelection().get());
+              static_cast<MapScrollMode>(*GetCtrl<ctrlComboBox>(ID_cbMapScrollMode)->GetSelection());
             SETTINGS.video.fullscreenSize = supportedResolutions_
-              [GetCtrl<ctrlGroup>(ID_grpResolution)->GetCtrl<ctrlComboBox>(ID_cbResolution)->GetSelection().get()];
-            const unsigned windowSizeSelection =
-              grpWindowSize.GetCtrl<ctrlComboBox>(ID_cbWindowSize)->GetSelection().get();
+              [*GetCtrl<ctrlGroup>(ID_grpResolution)->GetCtrl<ctrlComboBox>(ID_cbResolution)->GetSelection()];
```
Flow86: the `*` is ambigious, maybe use explicit `.value()` here, also can we make sure  a selection does exist? (clang-tidy with specific settings would cause a `unchecked optional access` warning here)

libs/s25main/world/GameWorld.cpp
```diff
@@ -994,7 +994,7 @@ bool GameWorld::ValidWaitingAroundBuildingPoint(const MapPoint pt, const MapPoin
             return false;
     }
     // object wall or impassable terrain increasing my path to target length to a higher value than the direct distance?
-    return FindHumanPath(pt, center, CalcDistance(pt, center)) != boost::none;
+    return FindHumanPath(pt, center, CalcDistance(pt, center)) != std::nullopt;
```
Flow86: shouldn't this be a `.has_value()`?

extras/ai-battle/main.cpp
Flamefire: Unless we want to require Boost 1.90+. Prior to that Boost.programOptions didn't support std::optional. I wanted to make it clear when we can switch and remove "old" stuff

libs/s25main/gameTypes/BuildingTypes.h
Flamefire: `boost_variant2` is never valueless which makes it easier to work with

libs/s25main/world/GameWorld.cpp
Flamefire: Only a matter of style. But sure

libs/s25main/ingameWindows/iwSettings.cpp
Flamefire: Yeah, I'm missing a `get` or similar unchecked access method for `std::optional`
But speed doesn't matter here anyway so `.value()` it is

libs/s25main/figures/nofDefender.h
Flamefire: 👍

PR #1948, opened 2026-06-15
Hunter: Use GetPointsInRadius() instead of square

morganchristiansson:
Hunter currently uses square, comments says it should be circle.
Update it to use new [GetPointsInRadius() pattern from nofSkinner.cpp](https://github.com/morganchristiansson/s25client/blob/hunter-radius/libs/s25main/figures/nofSkinner.cpp#L207) to find animals.
Also extract ANIMAL_RADIUS to shared constant used by Hunter, AI Hunter builder and Skinner.

This supports #1946 which displays building radius as outline.

### DIFF
--- a/libs/s25main/ai/aijh/AIPlayerJH.cpp
+++ b/libs/s25main/ai/aijh/AIPlayerJH.cpp
@@ -1965,51 +1965,26 @@ bool AIPlayerJH::HuntablesinRange(const MapPoint pt, unsigned min)
     // check first if no other hunter(or hunter buildingsite) is nearby
     if(aii.isBuildingNearby(BuildingType::Hunter, pt, 14))
         return false;
-    unsigned maxrange = 25;
-    unsigned short fx, fy, lx, ly;
-    const unsigned short SQUARE_SIZE = 19;
-    unsigned huntablecount = 0;
-    if(pt.x > SQUARE_SIZE)
-        fx = pt.x - SQUARE_SIZE;
-    else
-        fx = 0;
-    if(pt.y > SQUARE_SIZE)
-        fy = pt.y - SQUARE_SIZE;
-    else
-        fy = 0;
-    if(pt.x + SQUARE_SIZE < gwb.GetWidth())
-        lx = pt.x + SQUARE_SIZE;
-    else
-        lx = gwb.GetWidth() - 1;
-    if(pt.y + SQUARE_SIZE < gwb.GetHeight())
-        ly = pt.y + SQUARE_SIZE;
-    else
-        ly = gwb.GetHeight() - 1;
-    // Durchgehen und nach Tieren suchen
-    for(MapPoint p2(0, fy); p2.y <= ly; ++p2.y)
-    {
-        for(p2.x = fx; p2.x <= lx; ++p2.x)
+    const unsigned maxrange = 25;
+
+    const auto pointToAnimal = [this](const MapPoint pt, unsigned) -> const noAnimal* {
+        for(const noBase& fig : gwb.GetFigures(pt))
         {
-            // Search for animals
-            for(const noBase& fig : gwb.GetFigures(p2))
-            {
-                if(fig.GetType() == NodalObjectType::Animal)
-                {
-                    // Ist das Tier überhaupt zum Jagen geeignet?
-                    if(!static_cast<const noAnimal&>(fig).CanHunted())
-                        continue;
-                    // Und komme ich hin?
-                    if(gwb.FindHumanPath(pt, static_cast<const noAnimal&>(fig).GetPos(), maxrange))
-                    // Dann nehmen wir es
-                    {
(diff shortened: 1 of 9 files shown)


libs/s25main/figures/nofHunter.cpp
```diff
@@ -131,39 +131,24 @@ void nofHunter::HandleDerivedEvent(unsigned /*id*/)
 
 void nofHunter::TryStartHunting()
 {
-    // Find animals in a square around building (actually should be circle, but animals are moving anyway)
```
morganchristiansson, line 134: Literally says it should be a circle

libs/s25main/ai/aijh/AIPlayerJH.cpp
```diff
@@ -1965,51 +1965,26 @@ bool AIPlayerJH::HuntablesinRange(const MapPoint pt, unsigned min)
...
-            }
+            if(fig.GetType() == NodalObjectType::Animal)
+                return static_cast<const noAnimal*>(&fig);
         }
-    }
-    return false;
+        return nullptr;
+    };
+
+    const auto canAnimalBeHunted = [this, pt, maxrange](const noAnimal* const animal) {
```
Flamefire: This is almost exactly the code used in nofHunter, isn't it? Having to change it in 2 locations means we should have used a common method. Can you introduce a (probably static) method in `nofHunter`  for getting animals?

libs/s25main/gameData/GameConsts.h
```diff
@@ -41,6 +41,9 @@ constexpr auto gfs_to_duration(const unsigned gfs)
 /// Reichweite der Bergarbeiter
 constexpr unsigned MINER_RADIUS = 2;
 
+/// Suchradius für Tiere (Jäger, Gerber, KI)
```
Flamefire: Please no more new German comments, at some point all should be English

morganchristiansson: Changed to English. Most entries in this file have comments, and it adds context where it's used, so maybe worth keeping comment.

libs/s25main/ai/aijh/AIPlayerJH.cpp
morganchristiansson: Alright it's refactored to use nofHunter::GetAnimalsInRange()
nofSkinner.cpp also uses nofHunter::GetAnimalsInRange() but it uses CanBeSkinned() instead of CanHunted()

morganchristiansson:
Need some help here.
Test is failing on auto replay tests where AI behaviour has changed due to square->radius.

AI started bumping replay versions to keep legacy square for backwards compatibility. Is that a solution? But then we leave the old ugly square code in place.
https://github.com/Return-To-The-Roots/s25client/actions/runs/27752624950/job/82121998496?pr=1948

Flamefire:
> AI started bumping replay versions to keep legacy square for backwards compatibility. Is that a solution? But then we leave the old ugly square code in place.

I'd prefer that, it is valuable to keep replay working to recreate failures

morganchristiansson:
Ran the auto replay test locally and it passes now.
Not entirely happy about adding GameWorldBase::GetReplayMinorVersion() but some method to make current replay file version available is needed.

libs/s25main/world/GameWorldBase.h
```diff
@@ -67,6 +67,8 @@ class GameWorldBase : public World
     GameInterface* gi;
     std::unique_ptr<EconomyModeHandler> econHandler;
     std::unique_ptr<TradePathCache> tradePathCache;
+    /// Replay minor version when replaying. Set to current minor version for live games.
```
Flamefire, line 70: > Set to current minor version for live games.

Where?

tests/s25Main/autoplay/main.cpp
```diff
@@ -92,6 +92,7 @@ static void playReplay(const boost::filesystem::path& replayPath, const bool isS
             gameWorld.GetPlayer(i).MakeStartPacts();
     }
 
+    gameWorld.SetReplayMinorVersion(replay.GetMinorVersion());
```
morganchristiansson, line 95: here.

Replay is not accessible from hunter AI so that's why it's mirrored to GameWorldBase. Not ideal is it.

Flamefire: This is not the "live game" part. It is only ever set for replays.

morganchristiansson: Ah for live games it's default initialized to 255...

morganchristiansson:
Also, my memory may be wrong, but I recall original S2 the hunter would approach prey to within ~2 tiles and shoot at close range.
Observing the hunter in RttR he can be quite far away when shooting. Code says 4 tiles. It's not always clear which animal he's shooting, and sometimes there's other animals closer.

Not really in scope, except changing this might also affect replay compatibility.

PR #1949, opened 2026-06-18
Fix and extend AI-Battle

MichalLabuda:
## Summary

Fixes several bugs in the `ai-battle` headless runner and extends it with new CLI options for wares, addons, and Lua scripting.

### Bug fixes

- **Player colors were unassigned** - `GeneratePlayerInfo` never set `pi.color`, so all AI players used a default/zero color. Now assigns colors from `PLAYER_COLORS` round-robin.
- **Map resources not initialized** - `MapLoader::SetupResources(world_)` was missing after `Load()`, which could cause out-of-sync errors when playing back replays created using ai-battle.
- **`--objective` flag was ignored** - `ggs.objective` was hardcoded to `TotalDomination` regardless of the `--objective` CLI argument. The hardcoded line is removed so the parsed option takes effect.

### New features

- **`--wares`** - controls starting wares for all players: `vlow` | `low` | `normal` (default) | `alot`.
    
- **`--settings`** - accepts an INI file with an `[addons]` section to configure addon settings. Section is the same as in CONFIG.INI file. Non-default selections are logged to stdout after loading. Requires `AddonList::getOptionName()` added to `AddonList` to display human-readable option names.
    
- **`--lua`** - loads and runs a Lua script via `MapLoader::LoadLuaScript`. The script is also embedded into the replay file when `--replay` is used. Requires a new `HeadlessLocalState` struct (implements `ILocalGameState`) to satisfy the interface; ~~Lua `SystemChat` messages are forwarded to stdout. Messages printed by Lua script are messing up the status table - future fix maybe)~~
    
- **`<RTTR_X>` path expansion** - all path options (`--map`, `--settings`, `--lua`, `--replay`, `--save`) now run through `RTTRCONFIG.ExpandPath()`, enabling `<RTTR_X>` placeholders and leading `~`. Previously `--replay` and `--save` passed paths through as-is.
    

### Help system

- Running without arguments or with invalid arguments now prints ~~a compact one-line usage hint instead of~~ the full option list.
- `--help` / `-h` prints full per-option descriptions plus a path expansion appendix ~~documenting all `<RTTR_X>` placeholders, their per-platform default values, and an example invocation.~~

Edit:
- LuaInterfaceBase::setSuppressStdout to suppress lua rttr:log stdout output, writes to log file only

### DIFF
--- a/extras/ai-battle/HeadlessGame.cpp
+++ b/extras/ai-battle/HeadlessGame.cpp
@@ -57,6 +57,7 @@ HeadlessGame::HeadlessGame(const GlobalGameSettings& ggs, const bfs::path& map, 
     {
         if(!loader.LoadLuaScript(game_, localState_, luaPath))
             throw std::runtime_error("Failed to load Lua script: " + luaPath.string());
+        world_.GetLua().setSuppressStdout(true);
         luaPath_ = luaPath;
         bnw::cout << "Lua script loaded: " << luaPath << '\n';
     }
--- a/extras/ai-battle/HeadlessGame.cpp
+++ b/extras/ai-battle/HeadlessGame.cpp
@@ -5,7 +5,6 @@
 #include "HeadlessGame.h"
 #include "EventManager.h"
 #include "GlobalGameSettings.h"
-#include "ILocalGameState.h"
 #include "PlayerInfo.h"
 #include "Savegame.h"
 #include "factories/AIFactory.h"
@@ -30,15 +29,6 @@ std::string HumanReadableNumber(unsigned num);
 namespace bfs = boost::filesystem;
 namespace bnw = boost::nowide;
 
-namespace {
-struct HeadlessLocalState : ILocalGameState
-{
-    unsigned GetPlayerId() const override { return 0; }
-    bool IsHost() const override { return true; }
-    std::string FormatGFTime(unsigned) const override { return ""; }
-    void SystemChat(const std::string& msg) override { bnw::cout << "[lua] " << msg << "\n"; }
-};
-} // namespace
 using bfs::canonical;
 
 #ifdef WIN32
@@ -166,9 +156,8 @@ void HeadlessGame::RecordReplay(const bfs::path& path, unsigned random_init)
 
 void HeadlessGame::LoadLuaScript(const bfs::path& luaPath)
 {
-    HeadlessLocalState state;
     MapLoader loader(world_);
-    if(!loader.LoadLuaScript(game_, state, luaPath))
+    if(!loader.LoadLuaScript(game_, localState_, luaPath))
         throw std::runtime_error("Failed to load Lua script: " + luaPath.string());
     luaPath_ = luaPath; // remember for embedding in the replay
     bnw::cout << "Lua script loaded: " << luaPath << '\n';
--- a/extras/ai-battle/HeadlessGame.cpp
+++ b/extras/ai-battle/HeadlessGame.cpp
@@ -1,10 +1,11 @@
-// Copyright (C) 2005 - 2024 Settlers Freaks (sf-team at siedler25.org)
+// Copyright (C) 2005 - 2026 Settlers Freaks (sf-team at siedler25.org)
 //
 // SPDX-License-Identifier: GPL-2.0-or-later
 
 #include "HeadlessGame.h"
 #include "EventManager.h"
 #include "GlobalGameSettings.h"
+#include "ILocalGameState.h"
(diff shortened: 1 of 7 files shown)


extras/ai-battle/HeadlessGame.cpp
```diff
@@ -27,6 +29,16 @@ std::string HumanReadableNumber(unsigned num);
...
 namespace bfs = boost::filesystem;
 namespace bnw = boost::nowide;
+
+namespace {
+struct HeadlessLocalState : ILocalGameState
+{
+    unsigned GetPlayerId() const override { return 0; }
+    bool IsHost() const override { return true; }
+    std::string FormatGFTime(unsigned) const override { return ""; }
+    void SystemChat(const std::string& msg) override { bnw::cout << "[lua] " << msg << "\n"; }
```
Flamefire: As you noticed this messes up the table. I'd say we don't need the output

extras/ai-battle/HeadlessGame.cpp
```diff
@@ -138,13 +151,29 @@ void HeadlessGame::RecordReplay(const bfs::path& path, unsigned random_init)
...
         replay_.AddPlayer(world_.GetPlayer(playerId));
     replay_.ggs = game_.ggs_;
     if(!replay_.StartRecording(path, mapInfo, random_init))
         throw std::runtime_error("Replayfile could not be opened!");
 }
 
+void HeadlessGame::LoadLuaScript(const bfs::path& luaPath)
+{
+    HeadlessLocalState state;
+    MapLoader loader(world_);
```
Flamefire: That looks odd:
- Why do we need the loader?
- Does the local `HeadlessLocalState` cause a dangling reference?

Flamefire (changes requested):
> <RTTR_X> path expansion - all path options (--map, --settings, --lua, --replay, --save) now run through RTTRCONFIG.ExpandPath(), enabling <RTTR_X> placeholders and leading ~. Previously --replay and --save passed paths through as-is.

This looks a bit overkill to me. What is your use case for that?

extras/ai-battle/HeadlessGame.cpp
MichalLabuda: Removed LoadLuaScript entirely. Moved logic to the HeadlessGame constructor.

MichalLabuda:
> > <RTTR_X> path expansion - all path options (--map, --settings, --lua, --replay, --save) now run through RTTRCONFIG.ExpandPath(), enabling <RTTR_X> placeholders and leading ~. Previously --replay and --save passed paths through as-is.
> 
> This looks a bit overkill to me. What is your use case for that?

I don't exactly know if you're referring to the admittedly a bit excessive help text printout or to the use of ExpandPath() itself.
However I didn't like the help text my self so I reverted back to the original style.
About ExpandPath(), I used only RTTR_USERDATA. I find it very useful here for easy access to MAPS, REPLAYS, PRESETS etc. 
In help text I mention only about this single placeholder.

extras/ai-battle/HeadlessGame.cpp
MichalLabuda: I've changed the override and introduced `suppressStdout` flag to `LuaInterfaceBase` to redirect `rttr:log` output to file-only (`LuaInterfaceBase::log`)

Flamefire (dismissed):

MichalLabuda:
@Flamefire  Clang-tidy failed on one line so I had to push a commit which disabled your auto-merge. Please re-enable it. Thanks.

Flamefire (dismissed):

Flamefire:
Merge conflict to resolve

MichalLabuda:
> Merge conflict to resolve

Missed that. I've merged master.

Flamefire (approved):

extras/ai-battle/main.cpp
```diff
@@ -29,13 +35,47 @@ namespace bnw = boost::nowide;
...
+        bnw::cout << "Note: no [addons] section in " << iniPath << ", using defaults.\n";
+        return;
+    }
+
+    unsigned loaded = 0;
+    for(const auto& entry : *addons)
+    {
+        try
+        {
+            const auto id = static_cast<AddonId>(std::stoul(entry.first));
```
Flamefire: Just seen this small but possibly important part: Use `s25util::*FromStringClassic` to avoid locale-dependent behavior when reading config files.

Flamefire (changes requested):

Flamefire (approved):

PR #1950, opened 2026-06-18
Headless replay player

MichalLabuda:
## Summary

Adds a new headless `replay-player` tool that loads and plays back `.rpl` replay files without a GUI, reporting live progress and detecting desyncs.

### New files

- **`extras/replay-player/CMakeLists.txt`** - build target linking against `s25Main` and Boost; includes DLL gathering on Windows.
- **`extras/replay-player/HeadlessReplay.h/.cpp`** - supporting types and helpers:
    - `HeadlessGameState` - minimal `ILocalGameState` implementation (player 0 / host); ~~forwards Lua `SystemChat` messages to stdout.~~
    - `ReplayStatus` - tracks game reference, total GF count, wall-clock start time, and last-reported GF for progress display.
    - `printConsole()` - cross-platform console output: uses `WriteConsoleA` with VT100 mode enabled on Windows, `boost::nowide::cout` elsewhere.
    - `printTable()` - live-updating stats table using ANSI cursor-up escape codes, refreshed every second. Shows current GF / total GFs, game clock, wall clock, GF/sec, and per-player Country / Buildings / Military / Gold statistics. Defeated players are shown with strikethrough.
    - `printInitialInfo()` - prints replay metadata (map name, version, type, size, seed, GF range) and a player roster before playback begins.
- **`extras/replay-player/main.cpp`** - main program:
    - Accepts the replay path via `--replay` / `-r` or as a positional argument.
    - Supports both new-game replays (decompresses embedded map data and Lua script to temp files) and savegame-based replays.
    - Applies the fish-fix skip for replays predating version 8.3.
    - Executes all game commands frame by frame, verifying async checksums on each `GameCommand`.
    - Reports the first desync with GF number and actual-vs-stored checksum; `--verbose` / `-V` additionally dumps the full async log.
    - Exits with code `0` if no desyncs were found, `1` otherwise (also `1` on any load/parse error).
    - Replay path argument supports `<RTTR_X>` placeholders and leading `~` via `RTTRCONFIG.ExpandPath()`.
    - ~~Two-tier help: brief usage on no-args or parse error; full option list with path expansion reference on `--help` / `-h`.~~
    - `--version` / `-v` prints title, version, revision, compiler, and OS.

Edit:
- LuaInterfaceBase::setSuppressStdout to suppress lua rttr:log stdout output, writes to log file only


MichalLabuda:
Replay-Player is very similar to AI-Battle and theoretically could be implemented as an additional mode to it instead of completely separate app. I've decided to implement it like this but I'm not entirely convinced that this is the way to go.

Flamefire (commented):
Looks good but I'm wondering if we can unify loading the replay. Especially with the fish fix it becomes a maintenance burden. Can you investigate this?

And it doesn't make sense to count asyncs: Every NWF after an async will be async too. So you can stop right there. Or at least report it only once if you see any use for letting it continue.

wichern:
Great idea having a headless replay! I have been wanting this feature myself.

My suggestion would be to make HeadlessReplay a class instead of just a collection of types and helper functions.
Then it could be used for other tools as well.

The binary in this PR checks replays. I would like to be able to create another binary, that uses the same class and extracts training data from replays.

How about something like this:

```cpp
class HeadlessReplay
{
public:
    HeadlessReplay(const boost::filesystem::path& replay);
    bool Run();
};
```

PR #1951, opened 2026-06-18
Addon presets

### LINKED ISSUE #587: Saving of addons/game settings
There was a bug, that addon settings are reset if you open the options w/o opening also the addons, which is now fixed in https://github.com/Return-To-The-Roots/s25client/commit/fd9af5f24ef79abdc475262db4dd0fb2f1448ead.
Although I think the fix would be better to save/load only in the addon window.

However game settings (speed etc) is never saved. Also addons are not saved when set when hosting a game. IMO it would be more user-friendly if the settings (addons and others) would be saved when the host starts a normal game (no lua file) or even if he just sets them. Then you can have your preferred settings and keep them every time.
ATM it is also not intuitive that the addon settings are saved for the options but not ingame although the dialog looks the same.

TLDR: When and what settings should be saved?

ccing @Flow86 @Spikeone @MarcusSt @jhkl

### LINKED ISSUE #665: [wish] Add Addon presets
It'd be nice to add addon presets, that way you may have a preset for small maps (less catapults, but inexhaustible mines) etc.
So you may create your own presets and load them on demand.

MichalLabuda:
## Summary

Implements addon preset save/load functionality, as requested in issue https://github.com/Return-To-The-Roots/s25client/issues/665.
Closes https://github.com/Return-To-The-Roots/s25client/issues/665
Players can now save their current addon configuration to a named `.ini` file and reload it later, instead of manually re-configuring addons each time.

### Changes

- **New `iwAddonPresets.h/.cpp`** - Two new dialog windows (`iwSaveAddonPreset`, `iwLoadAddonPreset`) built on a shared base class (`iwAddonPresetsBase`). The dialogs show a sortable list of saved presets, a name edit field, and buttons for the primary action and deletion (with confirmation). Double-clicking a row triggers the action directly.
- **`iwAddons`** - Added **Save** and **Load** buttons in a new button row above the existing Apply/Abort row. Window height increased from 500 to 530 px to accommodate the extra row. New `applyAddonStates()` method applies a loaded preset back into the GUI, respecting read-only addons and falling back to defaults for unknown/out-of-range values.
- **`files.h`** - New folder constant `addonPresets` pointing to `<RTTR_USERDATA>/PRESETS`.
- **`const_gui_ids.h`** - New `CGI_ADDON_PRESETS` GUI ID for the preset dialogs.

Presets are stored as `.ini` files (using `libsiedler2`) in `<RTTR_USERDATA>/PRESETS/`. The save dialog validates the name (strips path components, rejects Windows reserved device names, prompts before overwrite). Load validates the file format and falls back to addon defaults for any missing entries.

---

This PR probably also allows to solve issue https://github.com/Return-To-The-Roots/s25client/issues/587, which reports that addon configurations configured while hosting a game are not preserved - players can now explicitly save a preset before a session and reload it next time.

Edit: Resolves https://github.com/Return-To-The-Roots/s25client/issues/587

---

image


libs/s25main/ingameWindows/iwAddonPresets.cpp
```diff
@@ -0,0 +1,294 @@
...
+constexpr unsigned ID_msgboxDelete = 0;
+constexpr unsigned ID_msgboxOverwrite = 1;
+} // namespace
+
+static bfs::path GetPresetsDir()
+{
+    return RTTRCONFIG.ExpandPath(s25::folders::addonPresets);
+}
+
+static std::optional<std::map<unsigned, unsigned>> LoadStatesFromFile(const bfs::path& filePath)
```
Flamefire: Why "states"? Isn't it "preset(s)"?

libs/s25main/ingameWindows/iwAddonPresets.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+// iwSaveAddonPreset
+iwSaveAddonPreset::iwSaveAddonPreset(std::map<unsigned, unsigned> states)
+    : iwAddonPresetsBase(_("Save Addon Preset"), _("Save")), states_(std::move(states))
+{}
+
+bfs::path iwSaveAddonPreset::GetSaveFilePath() const
+{
+    std::string name = GetCtrl<ctrlEdit>(ID_edtName)->GetText();
+    name = bfs::path(name).filename().string();
```
Flamefire: Isn't the edit supposed to contain a name only? Then why the "detour" over a filepath?

libs/s25main/ingameWindows/iwAddonPresets.cpp
```diff
@@ -0,0 +1,294 @@
...
+    const bfs::path filePath = GetSaveFilePath();
+    if(filePath.empty())
+        return;
+
+    // Reject Windows reserved device names (NUL, CON, COM1, etc.) — these cause failures on Windows
+    // even with an extension, and presets should be portable across platforms.
+    static constexpr std::array reservedNames{"con",  "prn",  "aux",  "nul",  "com1", "com2", "com3", "com4",
+                                              "com5", "com6", "com7", "com8", "com9", "lpt1", "lpt2", "lpt3",
+                                              "lpt4", "lpt5", "lpt6", "lpt7", "lpt8", "lpt9"};
+    if(helpers::contains(reservedNames, s25util::toLower(filePath.stem().string())))
```
Flamefire: Would `makePortableName` work?

Flamefire (changes requested):
Looks good, thanks!
Just some questions

libs/s25main/ingameWindows/iwAddonPresets.cpp
MichalLabuda: Yes it should, and now it is.

libs/s25main/ingameWindows/iwAddonPresets.cpp
MichalLabuda: The intention is to disallow saving/loading from different paths. Now, it rejects slash and backslash with error message.

libs/s25main/ingameWindows/iwAddonPresets.cpp
MichalLabuda: The idea for handling file names was different from what makePortableName (makePortableFileName) does.
However, I could change this to use makePortableFileName and adjust the rest of the code accordingly - let me know.

We could (or more precisely you could ;-) ) also decide if makePortableName would benefit from adding those restrictions which I could implement.

Flamefire: Yes, move it there, I'd say any of those names is not a portable name so should be handled there.

libs/s25main/ingameWindows/iwAddonPresets.cpp
```diff
@@ -0,0 +1,302 @@
...
+    if(name.empty() || name == "." || name == "..")
+        return {};
+    if(s25util::toLower(bfs::path(name).extension().string()) != ".ini")
+        name += ".ini";
+    return GetPresetsDir() / name;
+}
+
+void iwSaveAddonPreset::DoAction()
+{
+    const std::string rawName = GetCtrl<ctrlEdit>(ID_edtName)->GetText();
```
Flamefire: Why is this check not in `GetSaveFilePath`? Looks incomplete to get the name twice from (seemingly) different sources.

MichalLabuda:
Last push contained:

- external/libutil bumped to latest
- Replace ad-hoc checks in iwSave/iwAddonPresets with isValidFileName()/isValidFileNameChar() from libutil.
- Add fileNameOnly_ mode to ctrlEdit filtering invalid chars on both AddChar() and SetText().
- Fix double-space bug: remove KeyType::Space from Msg_KeyDown, space already arrives via char/text-input event.
- Set maxLength=251 on filename edits (255-byte limit minus 4-char extension).
- Skip makePortableFileName() sanitization for ServerType::Local+MapType::Savegame to prevent renamed duplicates.
- Migrate iwAddonPresets Msg_TableSelectItem from boost::optional to std::optional.
- Add unit and integration tests for all new behaviours.

tests/s25Main/UI/testControls.cpp
```diff
@@ -239,6 +239,61 @@ BOOST_FIXTURE_TEST_CASE(EditShowsCorrectChars, uiHelper::Fixture)
...
+
+    for(char32_t c : {U'a', U'b', U'1', U'2', U' '})
+        edt.Msg_KeyDown(KeyEvent(c));
+    BOOST_TEST(edt.GetText() == "ab12 ");
+
+    for(char32_t c : {U'<', U'>', U':', U'"', U'/', U'\\', U'|', U'?', U'*'})
+        edt.Msg_KeyDown(KeyEvent(c));
+    BOOST_TEST(edt.GetText() == "ab12 "); // all rejected, text unchanged
+
+    // SetText() has its own erase_if filter (separate code path from AddChar - doesn't go through
```
Flamefire: This comment doesn't belong in the test: Tests should not describe implementation, rather expected behavior. But AI is prone to such stuff, so please clean up

tests/s25Main/UI/testControls.cpp
```diff
@@ -239,6 +239,61 @@ BOOST_FIXTURE_TEST_CASE(EditShowsCorrectChars, uiHelper::Fixture)
...
+    mc.ldown = true;
+    edt.Msg_LeftDown(mc); // give focus
+
+    for(char32_t c : {U'a', U'b', U'1', U'2', U' '})
+        edt.Msg_KeyDown(KeyEvent(c));
+    BOOST_TEST(edt.GetText() == "ab12 ");
+
+    for(char32_t c : {U'<', U'>', U':', U'"', U'/', U'\\', U'|', U'?', U'*'})
+        edt.Msg_KeyDown(KeyEvent(c));
+    BOOST_TEST(edt.GetText() == "ab12 "); // all rejected, text unchanged
```
Flamefire: Put some valid char(s) somewhere in the list so it tests that those are still allowed too

libs/s25main/network/GameClient.cpp
```diff
@@ -917,8 +917,13 @@ bool GameClient::OnGameMessage(const GameMessage_Map_Info& msg)
     if(!VerifyState(ConnectState::QueryMapInfo))
         return true;
 
-    // full path
-    const std::string portFilename = makePortableFileName(msg.filename);
+    // For local savegame loads the filename comes from our own server and is already a valid
+    // filesystem name — sanitizing it (spaces -> underscores) would create a renamed duplicate
+    // alongside the original save file.
+    const std::string portFilename =
```
Flamefire: I actually don't agree with this one, do we really need a special case? Does `makePortableFileName` modify a name accepted by `isValidFilename`? If so why was this not a problem before?

libs/s25main/ingameWindows/iwAddonPresets.cpp
```diff
@@ -176,45 +180,21 @@ iwSaveAddonPreset::iwSaveAddonPreset(std::map<unsigned, unsigned> states)
...
-
-    // Trim leading/trailing whitespace
-    const auto first = name.find_first_not_of(" \t\r\n");
-    if(first == std::string::npos)
-        return {};
-    name = name.substr(first, name.find_last_not_of(" \t\r\n") - first + 1);
-
-    if(name.empty() || name == "." || name == "..")
-        return {};
+    boost::algorithm::trim_if(name, [](char c) { return c == ' ' || c == '\t' || c == '\r' || c == '\n'; });
```
Flamefire: This might be surprising to users as this hadn't been there before.
To avoid the duplication we should put this in a function so it is at least consistent. Like `getFilenameWithExtension`. As `ctrlEdit` now understands to accept only filenames, it could be a member to be called instead of `GetText` returning an optional string: empty on error, trimmed string otherwise so we could distinguish an empty user input from an invalid input

libs/s25main/controls/ctrlEdit.h
```diff
@@ -26,6 +26,7 @@ class ctrlEdit : public Window
     void SetDisabled(bool disabled = true) { this->isDisabled_ = disabled; }
     void SetNotify(bool notify = true) { this->notify_ = notify; }
     void SetNumberOnly(const bool activated) { this->numberOnly_ = activated; }
+    void SetFileNameOnly(const bool activated) { this->fileNameOnly_ = activated; }
```
Flamefire: This is problematic: It doesn't make sense to e.g. enable both.
How about an enum `EditType::Text,Number,Filename`?

tests/s25Main/UI/testWindows.cpp
```diff
@@ -1,34 +1,56 @@
...
 
 //-V:MOCK_METHOD:813
 //-V:MOCK_EXPECT:807
 
+namespace bfs = boost::filesystem;
 using SmallWorldFixture = WorldFixture<CreateEmptyWorld, 1, 10, 10>;
 
+// For iwSave/iwAddonPresets: builds a name from N 2-byte UTF-8 chars (U+00E9 'é'). With a 4-byte
+// extension (".ini"/".sav"), isValidFileName's 255-byte limit falls between 125 and 126 of these:
+// 125*2+4=254 bytes (valid), 126*2+4=256 bytes (invalid) - one char is the difference.
```
Flamefire: Again this is describing the use, not the function, but I'd say it is just fine here as it has the constants below

tests/s25Main/UI/testWindows.cpp
```diff
@@ -86,6 +108,48 @@ BOOST_AUTO_TEST_CASE(AddonWindow)
...
+{
+    rttr::test::TmpFolder tmp;
+    rttr::test::ConfigOverride userDataOverride("USERDATA", tmp);
+
+    iwSaveAddonPreset wnd(std::map<unsigned, unsigned>{{1, 2}});
+    Window& base = wnd; // upcast: GetCtrls/Msg_EditEnter are public on Window, unlike the overrides here
+    const auto presetsDir = RTTRCONFIG.ExpandPath(s25::folders::addonPresets);
+
+    base.GetCtrls<ctrlEdit>().at(0)->SetText("myPreset.ini"); // user already typed the extension
+    base.Msg_EditEnter(0); // ctrl_id is ignored by Msg_EditEnter, triggers DoAction()
```
Flamefire: Useless comment. It doesn't matter for the test what happens with `ctrl_id`

tests/s25Main/UI/testWindows.cpp
```diff
@@ -86,6 +108,48 @@ BOOST_AUTO_TEST_CASE(AddonWindow)
...
+    BOOST_TEST(bfs::exists(presetsDir / "myPreset.ini"));
+    BOOST_TEST(!bfs::exists(presetsDir / "myPreset.ini.ini")); // extension must not be duplicated
+
+    // Right at the byte limit: 125 chars * 2 bytes + ".ini" = 254 bytes, must succeed.
```
Flamefire: Now you are duplicating things/comments again. There is no use of 125,2 or 254 here.

I had a famous comment chain in a project:
```
CHUNK_SIZE_2k = 1024*4;
...
// creates a chunk of size 8k
function create_chunk(){
  return chunk(CHUNK_SIZE_2k);
}
```

So make sure you only add reasonable comments that cannot run into this. `justFitsName` and `kMaxValidTwoByteCharCount` pretty much capture it already.

tests/s25Main/UI/testWindows.cpp
```diff
@@ -86,6 +108,48 @@ BOOST_AUTO_TEST_CASE(AddonWindow)
...
+    const std::string justFitsName = makeMultiByteName(kMaxValidTwoByteCharCount);
+    base.GetCtrls<ctrlEdit>().at(0)->SetText(justFitsName);
+    base.Msg_EditEnter(0);
+    BOOST_TEST(bfs::exists(presetsDir / (justFitsName + ".ini")));
+
+    // One more character tips it over: 126 * 2 + 4 = 256 bytes, must be rejected.
```
Flamefire: So why not use "one more character"? Like `oneOverName = justFitsName + twoByteCharStr`
Then the comment could be: `// With one more character the name is rejected`

tests/s25Main/UI/testWindows.cpp
```diff
@@ -86,6 +108,48 @@ BOOST_AUTO_TEST_CASE(AddonWindow)
...
+BOOST_AUTO_TEST_CASE(SaveGameRejectsNameThatWouldOverflowAfterExtension)
+{
+    rttr::test::TmpFolder tmp;
+    rttr::test::ConfigOverride userDataOverride("USERDATA", tmp);
+
+    iwSave wnd;
+    Window& base = wnd;
+    // One char past the byte limit (126 * 2 + ".sav" = 256 bytes); can't test the "just fits" side
+    // here since a successful save would call GAMECLIENT.SaveToFile(), which needs a live game.
+    const std::string oneOverName = makeMultiByteName(kMinInvalidTwoByteCharCount);
```
Flamefire: Same as above

tests/s25Main/UI/testWindows.cpp
```diff
@@ -86,6 +108,48 @@ BOOST_AUTO_TEST_CASE(AddonWindow)
...
+    base.GetCtrls<ctrlEdit>().at(0)->SetText(oneOverName);
+    base.Msg_EditEnter(0); // SaveLoad() -> GetSaveFilePath() rejects -> returns before GAMECLIENT is touched
```
Flamefire: Why does GAMECLIENT matter for the test? That is what the comment should explain, not what methods are called

Flamefire (changes requested):
Ok, thanks. Most comments are pretty much about the quality of the AI code. I tried to provide enough reasoning so you can catch that for future PRs too

libs/s25main/network/GameClient.cpp
MichalLabuda: I guess people just accepted the way it worked or were used to saving games with very short names (since the column was so narrow anyway) consisting only of letters and numbers. However, if for your save game filename, you use spaces, brackets or any other valid filename character that is replaced with underscore by `makePortableFileName`, then you get a duplicate filename when you load that save game. Examples below.

Image

Image

tests/s25Main/UI/testControls.cpp
MichalLabuda: Valid chars are tested few lines up. Do you mean to combine both tests so one will contain both valid and invalid chars?

Flamefire: No I meant to test more something real world: A user types in something that contains invalid chars. The current test would wrongly succeed if upon seeing an invalid char the input rejects or swallows following valid chars.
So I'd put a "c" and "d" inside and an "f" at the end (or better: some non-consecutive ones like f, t, y) and check that only valid ones are added and all invalids rejected.
I'm wary of having the test simply copying the implementation instead of capturing behavior

MichalLabuda: Thanks for the explanation, it makes sense to do so. Added some valid special chars too.

libs/s25main/ingameWindows/iwAddonPresets.cpp
MichalLabuda: I added `ctrlEdit::GetFileName(const std::string& ext)`. Parameter `ext` is optional and if omitted, user input is validated as is but if provided then extension is ensured and validation is performed on entire input including extension. Trim is performed before validation in both cases and all-whitespace input is treated as empty.

libs/s25main/controls/ctrlEdit.h
MichalLabuda: In this case (both enabled) my thinking was that it won't crash and it will just work with the most restrictive one. Also I usually try to minimize changes to existing code but maybe it isn't the best approach... Anyway, I did as you suggested and added an enum, updated code referencing old flags.

tests/s25Main/UI/testWindows.cpp
MichalLabuda: Since I moved the tests into a single one of ctrlEdit, these consts including comments are removed.

tests/s25Main/UI/testWindows.cpp
MichalLabuda: I've removed this code, but point taken - comments should explain why something matters for the test, not which methods get called. I'll keep that in mind.

MichalLabuda:
Last few commits contained:
- `ctrlEdit`: replaced bool flags with an `EditType` enum (`Text`/`Number`/`Filename`)
- added `ctrlEdit::GetFileName()`: trims, optionally appends missing extension, validates - replaces the duplicated `GetSaveFilePath()` in `iwSave`/`iwAddonPresets`
- `RTTR_Assert` that `GetFileName()` is only called on `Filename` edits
- preset window now closes together with the addon window
- preset folder creation no longer throws from the window ctor
- tests:
	- freed leaked overwrite msgbox, msgbox IDs as constants
	- removed the `iwSave` filename test and the preset extension/length test (behavior moved into `GetFileName()`, unit-tested by `EditGetFileName` in testControls.cpp)
	- new `AddonPresetSaveLoadAndOverwrite` integration test
	- comment cleanups

libs/s25main/ingameWindows/iwAddonPresets.cpp
```diff
@@ -0,0 +1,276 @@
...
+// iwAddonPresetsBase
+iwAddonPresetsBase::iwAddonPresetsBase(const std::string& title, const std::string& actionLabel)
+    : IngameWindow(CGI_ADDON_PRESETS, IngameWindow::posLastOrCenter, Extent(440, 330), title,
+                   LOADER.GetImageN("resource", 41))
+{
+    using SRT = ctrlTable::SortType;
+    AddTable(ID_tblPresets, DrawPoint(20, 30), Extent(400, 200), TextureColor::Green2, NormalFont,
+             ctrlTable::Columns{{_("Preset Name"), 400, SRT::String}, {}});
+
+    const bfs::path presetsDir = GetPresetsDir();
```
Flamefire: Move below right before where it is used. Avoids needing to scroll to find where it is used/where it is set

libs/s25main/ingameWindows/iwAddonPresets.cpp
```diff
@@ -0,0 +1,276 @@
...
+    bfs::remove(filePath, ec);
+    if(ec)
+    {
+        LOG.write("Failed to delete addon preset %1%: %2%\n") % filePath % ec.message();
+        WINDOWMANAGER.Show(std::make_unique<iwMsgbox>(_("Delete Failed"), _("Failed to delete the selected preset."),
+                                                      this, MsgboxButton::Ok, MsgboxIcon::ExclamationRed));
+        // Refresh so the list reflects the actual filesystem state (e.g. file became a directory)
+        RefreshTable();
+        GetCtrl<ctrlEdit>(ID_edtName)->SetText("");
+        return;
```
Flamefire: Same code as below, so this can be removed

```suggestion
```

Maybe add a word to the comment that this is intentional. Like " Refresh the list in both cases..."

libs/s25main/ingameWindows/iwAddonPresets.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+// iwSaveAddonPreset
+iwSaveAddonPreset::iwSaveAddonPreset(std::map<unsigned, unsigned> states)
+    : iwAddonPresetsBase(_("Save Addon Preset"), _("Save")), states_(std::move(states))
+{}
+
+void iwSaveAddonPreset::DoAction()
+{
+    const auto fileNameResult = GetCtrl<ctrlEdit>(ID_edtName)->GetFileName(".ini");
+    if(fileNameResult.status == FileNameStatus::Empty)
```
Flamefire: Maybe a switch is better here when handling all cases of an enum. That allows the compiler to verify that all case are indeed handled.

libs/s25main/ingameWindows/iwAddonPresets.cpp
```diff
@@ -0,0 +1,276 @@
...
+    archive.push(std::move(iniItem));
+
+    if(libsiedler2::Write(filePath, archive) != 0)
+    {
+        LOG.write("Failed to save addon preset to %1%\n") % filePath;
+        WINDOWMANAGER.Show(std::make_unique<iwMsgbox>(
+          _("Save Failed"), _("Failed to save the preset. Please check the filename and try again."), this,
+          MsgboxButton::Ok, MsgboxIcon::ExclamationRed));
+        RefreshTable();
+        return;
```
Flamefire, line 253: The `return` seems a bit hidden in the larger handling. Maybe just `else Close`? Or better: `if(success) Close else LongHandleErrorCode` I.e. I try to keep the short branch first so you don't have to go up to find what it belonged to

libs/s25main/ingameWindows/iwAddonPresets.cpp
```diff
@@ -0,0 +1,276 @@
...
+    {
+        if(mbr == MsgboxResult::Yes)
+        {
+            const auto fileNameResult = GetCtrl<ctrlEdit>(ID_edtName)->GetFileName(".ini");
+            if(fileNameResult.status == FileNameStatus::Valid)
+                SaveToPath(GetPresetsDir() / fileNameResult.name);
+        }
+        return;
+    }
+    iwAddonPresetsBase::Msg_MsgBoxResult(msgbox_id, mbr);
```
Flamefire: Similar: else

libs/s25main/ingameWindows/iwAddonPresets.cpp
```diff
@@ -0,0 +1,276 @@
...
+        return std::nullopt;
+
+    std::map<unsigned, unsigned> states;
+    for(unsigned i = 0; i < ini->size(); ++i)
+    {
+        const auto* item = dynamic_cast<const libsiedler2::ArchivItem_Text*>(ini->get(i));
+        if(!item)
+            return std::nullopt;
+        unsigned id, status;
+        if(!s25util::tryFromStringClassic(item->getName(), id)
```
Flamefire, line 55: Maybe add a LOG entry here which option failed to load? Makes it easier for the user
Note that you could use a range-loop if you want, but both work

libs/s25main/ingameWindows/iwAddonPresets.cpp
```diff
@@ -0,0 +1,276 @@
...
+    boost::system::error_code ec;
+    bfs::create_directories(presetsDir, ec);
+    if(ec)
+        LOG.write("Failed to create addon preset folder %1%: %2%\n") % presetsDir % ec.message();
+    RefreshTable();
+}
+
+void iwAddonPresetsBase::RefreshTable()
+{
+    auto* table = GetCtrl<ctrlTable>(ID_tblPresets);
```
Flamefire: Nit: When you don't check for NULL use references please.

libs/s25main/ingameWindows/iwAddonPresets.cpp
```diff
@@ -0,0 +1,276 @@
...
+    const auto* table = GetCtrl<ctrlTable>(ID_tblPresets);
+    GetCtrl<ctrlEdit>(ID_edtName)->SetText(selection ? table->GetItemText(*selection, 0) : "");
+}
+
+void iwAddonPresetsBase::ConfirmDelete()
+{
+    if(GetSelectedFilePath().empty())
+        return;
+    WINDOWMANAGER.Show(std::make_unique<iwMsgbox>(_("Delete Preset"),
+                                                  _("Are you sure you want to delete the selected preset?"), this,
```
Flamefire: An issue I can imagine:
User selects an entry, then types a name into the edit and then presses delete -> Unexpected delete
As selecting an entry changes the edit but not vice-versa, maybe always consider the edit and/or show the name that will be deleted here

libs/s25main/ingameWindows/iwAddonPresets.cpp
```diff
@@ -0,0 +1,276 @@
...
+        RefreshTable();
+        return;
+    }
+
+    Close();
+}
+
+void iwSaveAddonPreset::Msg_TableChooseItem(const unsigned /*ctrl_id*/, const unsigned /*selection*/)
+{
+    DoAction();
```
Flamefire: This can be in the base class can't it?

libs/s25main/ingameWindows/iwAddons.cpp
```diff
@@ -32,24 +36,28 @@ constexpr unsigned AddonGuiLineHeight = 30;
 
 iwAddons::iwAddons(GlobalGameSettings& ggs, Window* parent, AddonChangeAllowed policy,
                    std::vector<AddonId> whitelistedAddons)
-    : IngameWindow(CGI_ADDONS, IngameWindow::posLastOrCenter, Extent(700, 500), _("Addon Settings"),
-                   LOADER.GetImageN("resource", 41), true, CloseBehavior::Custom, parent),
+    : IngameWindow(CGI_ADDONS, IngameWindow::posLastOrCenter, Extent(700, 530), _("Addon Settings"),
+                   LOADER.GetImageN("resource", 41), false, CloseBehavior::Custom, parent),
```
Flamefire: Why did you make this non-Modal? The reason here was that you have to choose an action (Apply or Discard)

tests/s25Main/UI/testControls.cpp
```diff
@@ -239,6 +239,121 @@ BOOST_FIXTURE_TEST_CASE(EditShowsCorrectChars, uiHelper::Fixture)
...
+
+    MouseCoords mc(edt.GetPos());
+    mc.ldown = true;
+    edt.Msg_LeftDown(mc); // give focus
+
+    for(char32_t c : {U'a', U'B', U'1', U'2', U' '}) // valid chars
+        edt.Msg_KeyDown(KeyEvent(c));
+    BOOST_TEST(edt.GetText() == "aB12 "); // all accepted
+
+    for(char32_t c : {U'c', U'<', U'D', U'>', U'-', U':', U'"', U'3', U'/', U'\\', U'4', U'|', U'?', U'!',
```
Flamefire, line 257: small trick if you want to make this shorter: Use a string view (to avoid the terminating NULL)
```
using namespace std::literals::string_view_literals;

    for(const auto c: U"1234"sv){
        std::cout << unsigned(c);
    }
```
Existing code (e.g. `createMockFont`) doesn't use it because it predates C++17, but you can use it if you want

tests/s25Main/UI/testWindows.cpp
```diff
@@ -1,17 +1,22 @@
...
 #include "controls/ctrlCheck.h"
 #include "controls/ctrlComboBox.h"
+#include "controls/ctrlEdit.h"
 #include "controls/ctrlGroup.h"
 #include "controls/ctrlImage.h"
 #include "controls/ctrlMultiline.h"
+#include "controls/ctrlTable.h"
 #include "controls/ctrlTextButton.h"
 #include "desktops/Desktop.h"
+#include "files.h"
```
Flamefire, line 18: This is unused, isn't it?

tests/s25Main/UI/testWindows.cpp
```diff
@@ -100,6 +107,59 @@ BOOST_FIXTURE_TEST_CASE(JumpWindow, SmallWorldFixture)
...
+    const std::map<unsigned, unsigned> states2{{3, 4}};
+
+    const auto save = [](const std::map<unsigned, unsigned>& states, const std::string& name) {
+        iwSaveAddonPreset wnd(states);
+        Window& base = wnd;
+        base.GetCtrls<ctrlEdit>().at(0)->SetText(name);
+        base.Msg_EditEnter(0);
+    };
+
+    const auto load = [](std::map<unsigned, unsigned>& out) {
```
Flamefire: Nit: Use a return value instead of out parameters and prefer consts where useful: `const std::map<unsigned, unsigned> loaded = load();` and then maybe `BOOST_TEST(load() == states1)`
Or is there a reason this is used like that?

Flamefire (changes requested):
Small polishing only. I would have used a `std::optional` but I guess `GetFileNameResult` is more explicit, so good idea there!

BTW: Please don't resolve conversations during a review. I usually use them to check what I reviewed before and how that was implemented. If you wan't it to keep track for yourself you could e.g. add a thumbs-up reaction for the done-comments.
Exceptions are where the comment is a simple suggestion with not further text which you applied as-is. GitHub auto-resolves those anyway when using the UI.

libs/s25main/ingameWindows/iwAddonPresets.cpp
MichalLabuda: Done. Left the indexed loop so failed option index number can be logged.

libs/s25main/ingameWindows/iwAddonPresets.cpp
MichalLabuda: Done.
Short branch first, got it.

tests/s25Main/UI/testControls.cpp
MichalLabuda: Good to know for the future. Thanks.

libs/s25main/ingameWindows/iwAddonPresets.cpp
MichalLabuda: I actually changed it so that if folder creation fails, the user gets notified and the presets window is immediately closed.

libs/s25main/ingameWindows/iwAddonPresets.cpp
MichalLabuda: Removed it and added an explanation comment.

libs/s25main/ingameWindows/iwAddonPresets.cpp
MichalLabuda: I reworked it as you suggested. I made the name in the edit the single source of truth for both load and delete. I also added the name of the preset in the delete confirmation box. If preset is not found, the user gets not found message.

I also changed how `ctrlEdit::GetFileName` treats `ext`: when `ext` is non-empty it now **ALWAYS** appends it. My first version only appended the extension if it wasn't already present, but that got messy once a name already contained the extension (double-extension edge cases). Always appending is simpler and keeps things consistent between what the user types and what they see: if they name a preset `myPreset.ini`, the file on disk is `myPreset.ini.ini`, but everywhere in the window and in messages it shows as `myPreset.ini` - exactly what they typed.

tests/s25Main/UI/testWindows.cpp
MichalLabuda: It is used. It provides: `s25::folders::addonPresets`

tests/s25Main/UI/testWindows.cpp
MichalLabuda: Changed it as suggested. Also extracted it (with `save`/`numPresets`) into the new `AddonPresetFixture`

libs/s25main/ingameWindows/iwAddons.cpp
MichalLabuda: The problem was that `WindowManager::DoShow` inserts every new window (including modal) behind all previously open modals (FIFO for modals). So, modal `iwAddons` trapped its own Save/Load preset child behind it - no input, uncloseable directly. Making iwAddons non-modal was the workaround.

However, I'd rather fix that than keep the workaround, which has a real bug: with both windows non-modal you can open the Save preset window, change addon settings afterwards, and pressing Save then writes the settings captured when the window was opened, not the current ones.

Simplest fix but app-wide change: **LIFO modal stacking** - in `DoShow`, insert modal windows at `windows.end()` (non-modals keep current behavior). Newest modal is topmost, closing it reveals the previous one. This flips the FIFO ordering tested in `ModalWindowPlacement` so this test would have to be rewritten unless there was some reason behind the intentional FIFO stacking that I don't see. With this change addons and presets windows could be made modal.
I asked the AI (Fable 5) to perform a deep analysis for potential breaking consequences of changing the stacking model from FIFO to LIFO, full analysis report is at the bottom of this message. In general it found that one change would be in case when there would be multiple `MissionStatement`s from LUA open at once (a same-frame burst, or in unpaused multiplayer a new one arriving while an earlier is still open) - then messages would be in reverse order (newest first). Second change would be possible focus-stealing if some async modal would pop-up. Positive change, apart from the modal windows chaining required for addons presets, would be the reduced deadlock potential.

So in general, I would be willing to introduce this change and update the `ModalWindowPlacement` test accordingly.
 
Alternatives I considered:

- an optional `opener` argument on `Show()`, used once at insertion to place a child modal on top of its opener - keeps FIFO for unrelated modals, but adds API surface and every spawning site must remember to opt in.
- `Window::parent_` can't be used to set parent window because of `GetDrawPos()` parent relative position calculation. Using this member would require a lot of work to redesign this and could quickly get messy.
- adding new member `parentWindow_` - would help but then it requires some good explanations in the comments why this member is introduced when you already have `parent_` available, it just isn't pretty, and as a stored raw pointer it can dangle if the opener closes first.
- change nothing, keep iwAddons non-modal - works, but loses the forced Apply/Discard choice and also introduces the stale-settings save bug.


# Full report: consequences of FIFO → LIFO modal stacking

## TL;DR

LIFO preserves the invariant everything else keys off ("if any modal is open, `windows.back()` is modal"), so input routing, activation and close handling are untouched; msgbox-result handlers are id-keyed and order-independent. The only semantic change is that _independent_ queued modals surface newest-first — practically relevant only if multiple lua `MissionStatement`s are open at once (a same-frame burst, or in unpaused multiplayer a new one arriving while an earlier is still open; reading order reverses — shipped campaigns show one per event). Bonus: chained windows (msgbox→msgbox, error over `iwPleaseWait`) currently spend a frame hidden behind the dying modal with the mouse warped onto a hidden button — LIFO makes those immediately correct. Test-wise only `ModalWindowPlacement` and one assertion in the replace test need updating.

## The change

In `DoShow` (WindowManager.cpp): modal windows get inserted at `windows.end()` instead of before the first modal. Non-modal windows keep the current rule (inserted before the first modal). One-line change.

## What provably does NOT change

The critical invariant — _if any modal is open, `windows.back()` is modal_ — holds under both schemes, because non-modals are still filed behind all modals. Everything that depends on it was checked and is untouched:

- **Input routing**: `Msg_LeftDown`/`Msg_RightDown`/`findAndActivateWindow` only test `windows.back()->IsModal()` — they don't care _which_ modal is at the back.
- **`SetActiveWindowImpl`**: still refuses to activate non-topmost windows while a modal exists; under LIFO the newly shown modal _is_ topmost, so it activates cleanly.
- **`DoClose`**: activates the new `windows.back()` when the topmost closes — this is what unwinds the LIFO stack correctly.
- **`ReplaceWindow`/`ToggleWindow`**: use `FindNonModalWindow`, so modals were never replaced/toggled; unchanged.
- **`dskGameInterface::SetActive`** (dskGameInterface.cpp:212): checks `GetTopMostWindow()->IsModal()` — invariant preserved.
- **Esc / Alt+W / right-click close**: operate on the topmost window; behavior is identical whenever at most one modal is open, which is every case these paths were designed for.
- **`iwMissionStatement`'s pause logic** (iwMissionStatement.cpp:59): pauses on every activation, unpauses on Continue — works under both orders, since each stacked statement re-pauses when it becomes active.
- **Msgbox result handlers**: all `Msg_MsgBoxResult` implementations are keyed by `msgboxid` and mutually independent; the ones that switch desktops on acknowledgment destroy all remaining windows anyway (`DoDesktopSwitch` clears the list), so ack-order reversal can't break logic.

## What changes

1. **The fix itself**: a modal opened from a modal is now on top and usable.
2. **Transient chains get _better_**: msgbox→msgbox chains, `iwConnecting::CI_Error`, and dskSelectMap's please-wait→error currently spend one frame with the new window behind the dying modal (and `iwMsgbox` even warps the mouse cursor onto its still-hidden default button, iwMsgbox.cpp:91). Under LIFO the new window is immediately topmost — strictly more correct.
3. **Independent async modals surface newest-first instead of oldest-first.** Two sub-cases:
    - _Reading order_: the only realistic producer of a modal burst is Lua's `MissionStatement` (LuaInterfaceGame.cpp:308) — likely why the `ModalWindowPlacement` test uses `CGI_MISSION_STATEMENT`. If a script fires two statements in one game-frame batch, the player now reads them in reverse order. Shipped campaign scripts (`data/RTTR/campaigns/roman/*.lua`) show one statement per event, and in paused single-player statements can't accumulate at all (no game frames while paused), so the exposure is: multiplayer or multi-event-per-frame scripts, rare.
    - _Focus stealing_: an async error msgbox now pops **over** a modal the user is interacting with (e.g. typing in `iwDirectIPConnect`) instead of hiding behind it; a stray Enter could hit the msgbox's default button. This is standard toolkit behavior, and FIFO already half-misbehaves here via the mouse-warp above.
4. **Deadlock/soft-lock surface shrinks.** Today a msgbox spawned behind a button-less `CloseBehavior::Custom` modal (`iwPleaseWait`) is unreachable until code closes the modal; under LIFO it's clickable immediately. No new lock scenario is introduced: the topmost modal is always the newest, which its creator by definition just chose to show.

## Test impact

- `ModalWindowPlacement` (testWindowManager.cpp:399): must be rewritten — the modal segment of the expected order reverses (`{wnd3, wnd2, wnd, wnd6, wnd4}`-style instead of `{wnd, wnd2, wnd3, ...}`, with wnd5 topmost).
- The "modal windows are not replaced" case (testWindowManager.cpp:385-395): one assertion flips (`REQUIRE_WINDOW_ACTIVE(wnd)` → `wnd2`).
- `EscClosesWindow` and `RightclickClosesWindow` pass unchanged — they never stack two modals.

## Verdict

No hidden coupling to FIFO exists anywhere in the input, activation, or close machinery — it all keys off "is the back window modal", which LIFO preserves. The only genuine semantic trade-off is newest-first surfacing of _independent_ modal bursts, which in practice means rare multi-`MissionStatement` scripts read in reverse; everything else is neutral or an improvement.

MichalLabuda:
@Flamefire I think I addressed all of your recent comments except for the addons window modal/non-modal in which I await your reply.

Last commits contained:

- Resolve the target preset from the name edit, not the table selection,
  so a typed/edited name wins over a stale selection (edit = source of truth)
- Replace GetSelectedFilePath() with GetTargetFilePath() (name -> existing
  file, else empty) and GetTargetFileOrNotify() (shows "Preset Not Found"
  for a non-empty unresolved name; empty name is a silent no-op)
- Create the presets folder in the ctor before building controls; on failure
  show "Addon Presets Unavailable" and close the window instead of only logging
- Move Msg_TableChooseItem (double-click acts) into the base class
- Move control/msgbox IDs into a class enum; rename ID_msgbox* to ID_mb*
- iwSaveAddonPreset: switch over FileNameStatus; return early on save success
- LoadPresetsFromFile: log non-text entries and unparseable options
- ctrlEdit::GetFileName: always append a non-empty ext (a name already ending
  in it becomes a distinct ".ini.ini" file); drop now-unused includes
- ListDir: guard on is_directory() rather than exists()
- Tests: add AddonPresetFixture and cases for distinct-extension names,
  edit-overrides-selection, delete + named confirmation, target-not-found,
  empty-name no-op, and folder-unavailable

MichalLabuda:
Corrected the clang-format and GCC issues that my MSVC build didn't catch. I've set up a local Linux dev env with clang-format, clang-tidy, and the gcc build at the same versions CI uses, so I should be able to catch these kinds of errors before pushing in the future.

libs/s25main/ingameWindows/iwAddonPresets.cpp
```diff
@@ -0,0 +1,296 @@
...
+    RefreshTable();
+}
+
+void iwAddonPresetsBase::RefreshTable()
+{
+    auto& table = *GetCtrl<ctrlTable>(ID_tblPresets);
+    table.DeleteAllItems();
+
+    for(const auto& file : ListDir(GetPresetsDir(), "ini"))
+        table.AddRow({file.stem().string(), file.string()});
```
Flamefire, line 111: Do you still use the 2nd column anywhere?

tests/s25Main/UI/testWindows.cpp
```diff
@@ -100,6 +111,239 @@ BOOST_FIXTURE_TEST_CASE(JumpWindow, SmallWorldFixture)
...
+
+    void save(const std::map<unsigned, unsigned>& states, const std::string& name)
+    {
+        iwSaveAddonPreset wnd(states);
+        Window& base = wnd;
+        base.GetCtrls<ctrlEdit>().at(0)->SetText(name);
+        base.Msg_EditEnter(0);
+    }
+
+    // Returns the states passed to the callback, or empty if it was not invoked (name not found).
```
Flamefire: AI described "how"... The caller doesn't care about any callback. This loads the given preset and returns its settings/values/...

libs/s25main/ListDir.cpp
```diff
@@ -15,7 +15,7 @@ std::vector<bfs::path> ListDir(const bfs::path& path, std::string extension, boo
 {
     std::vector<bfs::path> result;
 
-    if(!bfs::exists(path))
+    if(!bfs::is_directory(path))
```
Flamefire: Not sure, what happens if it is a non-directory without it? Probably an exception, isn't it?
I'd say that is fine as we do not handle the case where we expect a directory where the user put a file.

I actually use a trick based on that in other software: Add a read-only file where an app would place a folder to ensure it doesn't (usually env-vars can be set, this ensures they take effect)

Flamefire:
> I've set up a local Linux dev env with clang-format, clang-tidy, and the gcc build at the same versions CI uses, so I should be able to catch these kinds of errors before pushing in the future.

There are static clang-tools binaries also for Windows if you work rather there: https://github.com/muttleyxd/clang-tools-static-binaries/releases
No real need to test with all compilers if that is much work as that's what CI is for. At least for cases that are not easy to spot. So don't worry too much about that :)

> The only genuine semantic trade-off is newest-first surfacing of independent modal bursts, which in practice means rare multi-MissionStatement scripts read in reverse; everything else is neutral or an improvement.

@Spikeone I'd say it makes sense that the last opened modal-window will be the top window as I'd say this is the most intuitive. Lua-msg-windows (mission statements) pause the game so I don't really see where this would cause an issue there. Could there even be a situation with multiple statements in a GF? Do we have any? Would it be an issue if they get reversed, i.e. newest/last-added first?
If not I'd say we do make the change. Especially e.g. for the case like: 1. Show progress/wait/settings window 2. show an error message --> You do indeed want the last one first/top.

Flamefire (changes requested):

libs/s25main/ListDir.cpp
MichalLabuda: I changed it because before, when it was only LOG-on-presets-folder-creation-failed-and-continue, the code reached  ListDir and it crashed because for testing I added a file named PRESETS. Thought that it would prevent crashes for the future but also would silently allow for someone to miss potential error situation because it just returns empty now. On the other hand it's a protection against the crash if someone missed something... You're choice - for my case in this PR it is no longer necessary as with current state it doesn't even open presets window if folder creation fails.

Spikeone:
> @Spikeone I'd say it makes sense that the last opened modal-window will be the top window as I'd say this is the most intuitive. Lua-msg-windows (mission statements) pause the game so I don't really see where this would cause an issue there. Could there even be a situation with multiple statements in a GF? Do we have any? Would it be an issue if they get reversed, i.e. newest/last-added first?
> If not I'd say we do make the change. Especially e.g. for the case like: 1. Show progress/wait/settings window 2. show an error message --> You do indeed want the last one first/top.

I agree that this is what I'd expect.

libs/s25main/ListDir.cpp
Flamefire: > On the other hand it's a protection against the crash if someone missed something...

I'm Ok with the crash: If this fires then we are in a situation that is unhandled, like what you just mentioned: We shouldn't try to list a directory which is not a directory and/or a user shouldn't create a file where we expect a directory. I mean: That would need to be a deliberate act, not something that could "just happen"

libs/s25main/ingameWindows/iwAddonPresets.cpp
MichalLabuda: I do now :-) I've updated how the preset name for Load and Delete is retrieved (again). Now the text input is compared to the first column and, if found, the second column is used to load/delete the file. Now whatever file finds its way into the PRESETS folder, if it shows in the table, it will be possible to load/delete it.

tests/s25Main/UI/testWindows.cpp
MichalLabuda: Changed.

libs/s25main/ListDir.cpp
MichalLabuda: Reverted to the original.

libs/s25main/ingameWindows/iwAddons.cpp
MichalLabuda: Change applied. New modal windows are now stacked LIFO (newest modal on top); non-modal placement (always behind all modals) is unchanged. `iwAddons` and the addon preset windows are now modal, and are opened via `.Show` instead of `.ReplaceWindow` (which only closes an existing *non-modal* window by ID, so it was a no-op for these once they became modal). Tests adjusted to the new modal stacking order.

MichalLabuda:
My latest commit includes:

- New modal windows now appear on top; non-modal windows stay behind all modals
- Addon presets and Addons windows are now modal
- Preset lookup now matches by the table's displayed name
- Overwrite prompt now names the preset being overwritten
- Addons uses Show instead of ReplaceWindow for preset windows
- ctrlEdit trims trailing whitespace only when no extension is appended
- ListDir reverts to an exists() check
- iwSave switches on FileNameStatus
- Added tests covering modal stacking, mouse blocking, and preset matching

MichalLabuda:
@Flamefire I think all the comments are addressed. Any chance for a review?

libs/s25main/ingameWindows/iwAddonPresets.cpp
```diff
@@ -0,0 +1,300 @@
...
+
+bfs::path iwAddonPresetsBase::GetTargetFileOrNotify() const
+{
+    bfs::path path = GetTargetFilePath();
+    if(!path.empty())
+        return path;
+
+    const std::string name = GetCtrl<ctrlEdit>(ID_edtName)->GetText();
+    if(name.empty())
+        return {};
```
Flamefire, line 139: This is confusing. I had to check multiple times in which situation name and/or path can be empty. I think it is better to pass the name to `GetTargetFilePath()`
The only other usage is the deletion in the msgbox result where you then do `GetCtrl<ctrlEdit>(ID_edtName)->SetText("")`

If you get the text from there first that line makes much more sense.
I'd also assert that only non-empty names are passed to `GetTargetFilePath` so an empty result is always: name not found in table.
I guess `GetFilePath(name)` is then enough: "Convert" a name to a path using the table.

libs/s25main/ingameWindows/iwAddons.cpp
```diff
@@ -165,6 +198,21 @@ void iwAddons::UpdateView(const AddonGroup selection)
...
+{
+    for(unsigned i = 0; i < ggs.getNumAddons(); ++i)
+    {
+        const Addon* addon = ggs.getAddon(i);
+        if(!isReadOnly(addon->getId()))
+        {
+            const auto it = states.find(static_cast<unsigned>(addon->getId()));
+            const unsigned rawStatus = (it != states.end()) ? it->second : addon->getDefaultStatus();
+            const unsigned status = (rawStatus < addon->getNumOptions()) ? rawStatus : addon->getDefaultStatus();
+            addonGuis_[i]->setStatus(*GetCtrl<ctrlGroup>(ID_grpAddonsStart + i), status);
```
Flamefire, line 211: This looks odd, although it already exists... 
Can you do a small cleanup commit and store the window/group in `AddonGui` and remove it from the set/getStatus functions?
Avoids the error-prone use of `ID_grpAddonsStart +i` in multiple cases. We'd need a `getWindow` (or so) function for `UpdateView` then.

Flamefire (changes requested):
Ok thanks!
Just 2 more changes, one optional to not introduce another instance of `addonGuis_[i]->setStatus(*GetCtrl<ctrlGroup>(ID_grpAddonsStart + i), status)`:
the assumption there is that the same group/window is always passed as used in the constructor. We should instead store a reference to it and not pass it again. Also replace the existing usages too.

I'd like to see the `Get[Target]FilePath(name)`  change to avoid the hidden dependency on ` GetCtrl<ctrlEdit>(ID_edtName)` and make it clear why clearing that edit afterwards makes sense. Or what do you think?

libs/s25main/ingameWindows/iwAddonPresets.cpp
MichalLabuda: I agree that there is a lot of confusion with this edit box and list and when is what actually used. I have an idea for simplifying it considerably - I would remove Delete option from Save dialog and have it only available on Load. This way I can remove edit box completely from Load screen and now it's obvious that what you select is what you're loading/deleting and when you save you save to what you have in edit box. When saving, only confusing part to decide would be how to treat extension added to the name in the edit box - should edit box be ALWAYS treated as stem or omit adding extension if user provided it already. Your opinion?

libs/s25main/ingameWindows/iwAddons.cpp
MichalLabuda: I'll look into it.

libs/s25main/ingameWindows/iwAddonPresets.cpp
Flamefire: I'd say we don't show the extension to users. For savegames this is already done. The replay dialog does show extensions which is odd. Likely an oversight.

MichalLabuda: Ok. List in the table - without extension. What happens when saving, indicated by the table below.
| User input | Resulting filename | Name on the list |
| :----------- | :------------------- | :---------- |
| `My preset` | `My preset.ini` | `My preset` |
| `My preset.ini` | `My preset.ini` | `My preset` |
| `My preset.ini.ini ` | `My preset.ini.ini` | `My preset.ini` |

Flamefire: Not sure if it even makes sense to allow users to enter an extension. I.e. we could disallow the dot. I cannot see a use case for that as all dialogues are specific to a single type of file.
But doesn't hurt a least, the proposal looks good!

Flamefire (approved):

Flamefire:
I added the suggested change for simplified Addon GUI component handling as it's only remotely related to the work on presets you made, and so we can get this in before forgetting about it.

Thanks a lot!

PR #1952, opened 2026-06-18
Address PR feedback

morganchristiansson:
Address s25edit PR feedback Return-To-The-Roots/s25edit#38 as changes there are based on this repo.

### DIFF
--- a/libs/s25main/SignalHandler.cpp
+++ b/libs/s25main/SignalHandler.cpp
@@ -37,9 +37,9 @@ void ConsoleSignalHandler(int sig)
     if(sig == SIGINT)
     {
         if(!killme)
-            boost::nowide::cout << "Do you really want to terminate the program (y/n) : " << std::flush;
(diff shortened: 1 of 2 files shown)


Flow86 (approved):

Flow86 (approved):

libs/s25main/SignalHandler.cpp
```diff
@@ -37,9 +37,9 @@ void ConsoleSignalHandler(int sig)
     if(sig == SIGINT)
     {
         if(!killme)
-            boost::nowide::cout << "Do you really want to terminate the program (y/n) : " << std::flush;
+            boost::nowide::cerr << "Do you really want to terminate the program (y/n) : ";
```
Flamefire, line 40: @Flow86 IMO outputting a prompt to stderr isn't the right thing (TM) to do, it is not an error, just "regular" output, isn't it?

PR #1955, opened 2026-06-21
Disable zoom of DrawNameProductivityOverlay

morganchristiansson:
Submitting this for discussion as Draft.

`DrawNameProductivityOverlay()` draws building text overlays adjusted for zoom.
This makes it look garbled when zooming as it's a bitmapped font.
image

This does make it feel like font is changing size with zoom, it's an illusion due to zoom it doesn't change.
Also it feels smaller at full zoom but I think it's the pixel perfect original size.
image

Also it feels really small on 4K monitor now, which is maybe just because 4K monitor.

Better would be to use vector (non-bitmapped) font that can be properly rescaled.
Do we know what font is used? Does the building productivity&status even need correct font?

### DIFF
--- a/libs/s25main/world/GameWorldView.cpp
+++ b/libs/s25main/world/GameWorldView.cpp
@@ -217,9 +217,6 @@ void GameWorldView::Draw(const RoadBuildState& rb, const MapPoint selected, bool
             between_line.obj.Draw(between_line.pos);
     }
 
-    if(show_names || show_productivity)
-        DrawNameProductivityOverlay(terrainRenderer);
-
     DrawGUI(rb, terrainRenderer, selected, drawMouse);
 
     // Draw catapult stones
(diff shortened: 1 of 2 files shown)


Flow86:
> Better would be to use vector (non-bitmapped) font that can be properly rescaled.

we do use the original font though.

morganchristiansson:
> > Better would be to use vector (non-bitmapped) font that can be properly rescaled.
> 
> we do use the original font though.

Ya I realize that. That's why it's bitmap font. :-)

(PR is un-reviewed AI code, it works but can usually be cleaned up).

Spikeone:
I think we should at least zoom the text at steps - so it is zoomed at (don't know, 125/150/175/200 steps) some levels - not zooming it looks funny and somewhat strange for screenshots imho

morganchristiansson:
Yeah not zooming it does look funny which I didn't expect.

Also text is tiny on my 4K monitor and I usually play zoomed in.
Maybe zooming text in steps that will not look garbled is a good idea.

I tried to find the font with some whatthefont websites but got garbage results. Maybe it's custom font made for this game.
Maybe a different vectored font could be used for building name&productivity tho.

morganchristiansson:
Closing as I'm not happy with result and prefer old behaviour. Vectorized font would be best, or maybe some zooming is still needed, new size is too small on 4K monitor.

PR #1956, opened 2026-06-28
Seed AI RNG deterministically on game start for reproducibility

FlexApex:
### Problem

The AI uses a non-deterministic random source, which breaks replay determinism. Two distinct issues:

1. **`AI::getRandomGenerator()` is never seeded from the game seed.** It is lazily seeded from `std::random_device` on first use, so in replay mode it starts from a *different* state than during the original match. Any AI decision that draws from it diverges, desyncing the replay.
2. **`AIPlayerJH::TrySeaAttack()` allocates its own generator** — `std::mt19937(std::random_device()())` — on every call, which is non-deterministic by construction.

The net effect: **any replay in which an AI evaluates a sea attack desyncs** ("The played replay is not in sync with the original match"), even though the simulation is otherwise correct.

### Fix

- Seed `AI::getRandomGenerator()` from `random_init` in `GameClient::StartGame()`, right where `RANDOM.Init(random_init)` is already called — so replay and live runs share the same AI RNG sequence.
- Replace the local `std::mt19937` in `TrySeaAttack()` with the shared `AI::getRandomGenerator()`.

3 lines across 2 files; no behavioural change to a live game beyond making the AI RNG reproducible.

### How it was found

Surfaced while running headless replay verification on AI-battle recordings (related to the headless replay player in #1950) — a desync verifier flagged AI sea-attack frames as the first point of divergence, with the RNG checksum being the differing field.


Flamefire:
> any replay in which an AI evaluates a sea attack desyncs 

There is no AI in replays. So how exactly have you seen errors?

libs/s25main/ai/aijh/AIPlayerJH.cpp
```diff
@@ -1661,7 +1661,7 @@ void AIPlayerJH::TrySeaAttack()
             // \n",gwb.GetHarborPoint(i).x,gwb.GetHarborPoint(i).y);
         }
     }
-    auto prng = std::mt19937(std::random_device()());
+    auto& prng = AI::getRandomGenerator();
```
Flamefire, line 1664: This one looks correct though

FlexApex:
> > any replay in which an AI evaluates a sea attack desyncs
> 
> There is no AI in replays. So how exactly have you seen errors?

In my own fork I use ai-battle with a small eval framework to improve the AI:
See recent commits here: https://github.com/FlexApex/s25client/commits/master
There it lead to an .rpl desync.

Flamefire:
Can you explain a bit what you are doing there so I don't have to search your code? Because replays replay recorded actions. So there is no AI running there doing actions and hence no possibility for asyncs in replays.

Note that this gives a false impression of safety: You seed the AI at the start. Then you save the game. The AI actions from the continued game at that point and from the loaded game won't match.
It does look like they will match for replays if the AI would do actions there, which it doesn't (currently).

FlexApex:
I use extras/ai-battle for self-play benchmarking (challenger vs baseline over maps and seeds), record those games with --replay, and play the recordings back headless through the GameClient path, using the headless replay player from #1950 to compare the stored checksums.

Those replays do desync, but not because of the AI, you are right. HeadlessGame's ctor does MapLoader::Load then InitAfterLoad, while GameClient::StartGame does MakeStartPacts, Load, SetupResources, InitAfterLoad. So playback happens on a world with different resources (and different pacts when teams are used) and the rand checksum diverges early. So ai-battle recordings aren't replayable in the client at all right now.

So this PR isn't a desync fix. What's left is reproducibility. TrySeaAttack() built a std::mt19937(std::random_device()()) per call and never touched AI::getRandomGenerator(), which a36241c21 seems to have missed, so not even ai-battle's --random_ai_init (147d082d5) reaches it. 

About the savegame: agreed. It's the same guarantee RANDOM has though. random_init is generated fresh at every start, no RNG state goes into the savegame, and AI players are rebuilt from the world on load, so continuing vs save+load can't match anyway. Seeding the AI generator from random_init just puts it under the same rule. Doing it properly would mean serialising the AI RNG and the AI state into the save, which I'd leave out of scope.

Happy to drop the StartGame part and keep only the TrySeaAttack fix, rewrite the title and description for both, or close this. Whatever you prefer.

Flamefire:
> I use extras/ai-battle for self-play benchmarking (challenger vs baseline over maps and seeds), record those games with --replay, and play the recordings back headless through the GameClient path, using the headless replay player from #1950 to compare the stored checksums.
> 
> Those replays do desync, but not because of the AI, you are right. HeadlessGame's ctor does MapLoader::Load then InitAfterLoad, while GameClient::StartGame does MakeStartPacts, Load, SetupResources, InitAfterLoad. So playback happens on a world with different resources (and different pacts when teams are used) and the rand checksum diverges early. So ai-battle recordings aren't replayable in the client at all right now.

Sounds like worth fixing although I do not fully understand it: Does it have a wrong sequence even for replays?

> So this PR isn't a desync fix. What's left is reproducibility. TrySeaAttack() built a std::mt19937(std::random_device()()) per call and never touched AI::getRandomGenerator(), which [a36241c](https://github.com/Return-To-The-Roots/s25client/commit/a36241c21ac0fcb2567bbfed4fe4d4e1787251f6) seems to have missed, so not even ai-battle's --random_ai_init ([147d082](https://github.com/Return-To-The-Roots/s25client/commit/147d082d5bda46fe2eaf638c5e312abd53947b86)) reaches it.

Fully agreed, maybe update title and description for future reference.

> About the savegame: agreed. It's the same guarantee RANDOM has though. random_init is generated fresh at every start, no RNG state goes into the savegame, and AI players are rebuilt from the world on load, so continuing vs save+load can't match anyway. Seeding the AI generator from random_init just puts it under the same rule. Doing it properly would mean serialising the AI RNG and the AI state into the save, which I'd leave out of scope.

It is actually worse. We intended to save the RNG state in savegames and restore it on load, see
https://github.com/Return-To-The-Roots/s25client/blob/5aef739b0bdf71a5893b4f039e55a9f75e7b10b4/libs/s25main/network/GameServer.cpp#L534-L539

However this never happened. So for every savegame the RNG state is set to a fixed value: `0`.
That warrants its own PR, so agreed there too.

> Happy to drop the StartGame part and keep only the TrySeaAttack fix, rewrite the title and description for both, or close this. Whatever you prefer.

Seems to be working:
1. New game starts
  - RNG gets seeded randomly
  - AI RNG gets seeded with same seed
  - seed is stored in replay
2. Replay load
  - RNGs gets seeded with (same) stored seed
3. Savegame load
  - RNGs gets seeded with 0

So just update description & title to reflect this is about reproducibility after introduction of the deterministic AI.
I'll trigger CI and merge once passed

Flamefire (dismissed):

libs/s25main/network/GameClient.cpp
Flamefire:
```suggestion
#include "CreateServerInfo.h"
#include "EventManager.h"
#include "Game.h"
#include "GameEvent.h"
#include "GameLobby.h"
#include "GameManager.h"
#include "GameMessage_GameCommand.h"
#include "JoinPlayerInfo.h"
#include "LeatherLoader.h"
#include "Loader.h"
#include "NWFInfo.h"
#include "PlayerGameCommands.h"
#include "RTTR_Version.h"
#include "ReplayInfo.h"
#include "RttrConfig.h"
#include "Savegame.h"
#include "SerializedGameData.h"
#include "Settings.h"
#include "ai/AIPlayer.h"
#include "ai/random.h"
```

Flamefire (approved):

PR #1957, opened 2026-06-28
ai-battle: support team assignment (--teams) and establish team start-pacts

FlexApex:
# ai-battle: support team assignment (`--teams`) and establish team start-pacts

Builds on #1949 (uses its `HeadlessGame`). Two commits:

### 1. `--teams` option for team assignment

Adds a `--teams` option to the headless ai-battle harness so the AI can be
verified under real team rules instead of a free-for-all:

```
ai-battle -m Map.swd --ai aijh --ai aijh --ai aijh --ai aijh --teams "0,1;2,3"
```

Groups are separated by `;`, player indices by `,`. The first group becomes
`Team1`, the second `Team2`, and so on; unlisted players stay teamless. The
assignment is threaded through `HeadlessGame` into `GeneratePlayerInfo`, which
sets `PlayerInfo::team`.

### 2. Establish team start-pacts in `HeadlessGame` to match `GameClient` (avoid desyncs in replays)

`GameClient::StartGame` calls `GamePlayer::MakeStartPacts()` for every player on
a fresh map, creating the ally + non-aggression pacts that bind a team together.
`HeadlessGame` never did this, so a recorded ai-battle with teams had no pacts:
`GamePlayer::IsAttackable()` returned `true` for teammates and the AIs attacked
their own team. On replay `GameClient` *does* set up the pacts, so the recorded
inter-team attack commands hit allied players and were resolved differently —
producing an object-count divergence (replay desync) **without** any RANDOM
divergence.

`MakeStartPacts()` early-returns for `Team::None`, so the call is a no-op until
players are actually assigned to teams via `--teams`. Together with the
`MapLoader::SetupResources()` call already in `HeadlessGame`, this makes the
headless world setup match the `GameClient` replay path.

### Note
- Depends on #1949 — best merged after it (or rebased onto master once #1949 lands).

### DIFF
--- a/extras/ai-battle/main.cpp
+++ b/extras/ai-battle/main.cpp
@@ -14,6 +14,7 @@
 #include "ai/random.h"
 #include "files.h"
 #include "random/Random.h"
+#include "s25util/StringConversion.h"
 #include "s25util/System.h"
 
 #include <boost/filesystem.hpp>
@@ -55,7 +56,7 @@ static void loadAddonsFromIni(GlobalGameSettings& ggs, const bfs::path& iniPath)
     {
         try
         {
-            const auto id = static_cast<AddonId>(std::stoul(entry.first));
+            const auto id = static_cast<AddonId>(s25util::fromStringClassic<unsigned>(entry.first));
             const auto v = entry.second.get_value<unsigned>();
             ggs.setSelection(id, v);
             ++loaded;
--- a/extras/ai-battle/main.cpp
+++ b/extras/ai-battle/main.cpp
@@ -29,35 +29,6 @@ namespace bnw = boost::nowide;
 namespace bfs = boost::filesystem;
 namespace po = boost::program_options;
 
-static void printBriefUsage(const char* prog)
-{
-    bnw::cerr << "Usage: " << prog
-              << " -m <map> --ai aijh|dummy [--ai ...] [options]\n"
-                 "Run with --help or -h for a full list of options and examples.\n";
-}
-
-static const char PATH_EXPANSION_HELP[] =
-  "\nPath expansion (all path options):\n"
-  "  All paths support <RTTR_X> placeholders and a leading ~.\n"
-  "  ~ is expanded by the application: %USERPROFILE%\\Saved Games on Windows,\n"
-  "  $HOME on Linux/macOS.\n"
-  "  Available placeholders:\n"
-  "    <RTTR_USERDATA>   user data dir\n"
-  "                        Windows : %USERPROFILE%\\Saved Games\\Return To The Roots\n"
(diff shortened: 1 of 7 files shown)


extras/ai-battle/main.cpp
```diff
@@ -116,15 +170,79 @@ int main(int argc, char** argv)
             return 1;
         }
 
-        ggs.objective = GameObjective::TotalDomination;
-        HeadlessGame game(ggs, mapPath, ais);
+        const auto wares = options["wares"].as<std::string>();
```
Flamefire, line 179: Any specific reason to use this construct here but have lua/settings_path "auto-filled"?

extras/ai-battle/main.cpp
```diff
@@ -29,13 +38,47 @@ namespace bnw = boost::nowide;
...
+    {
+        try
+        {
+            const auto id = static_cast<AddonId>(s25util::fromStringClassic<unsigned>(entry.first));
+            const auto v = entry.second.get_value<unsigned>();
+            ggs.setSelection(id, v);
+            ++loaded;
+        } catch(const std::exception&)
+        {
+            // Unknown or invalid entry - skip silently
```
Flamefire: I'd say this should be a hard error: The user intended to do something which isn't done.
A missing addon section would also be an error, maybe unless the file is fully empty. Otherwise an accidental spelling mistake like "adon" goes unnoticed.

extras/ai-battle/main.cpp
FlexApex: No deliberate reason beyond following the file's two pre-existing conventions: the always-present, default-valued options are read inline with options[...].as<>() (this is how objective was already handled), while the genuinely-optional file paths are bound to optional<std::string> via po::value(&var) so their presence can be checked with if(var) (this is how replay/save were already handled). I kept wares like objective, and lua/settings like replay/save.

extras/ai-battle/main.cpp
FlexApex: Agreed. I'll make it strict: fail on an unparseable entry, on an unknown addon id, and on a missing [addons] section (except for a fully-empty file).

extras/ai-battle/main.cpp
Flamefire: I checked the history and it seems this was already in  the initial version after some revisions of the PR. This seems confusing, especially as e.g. other optionals (with default values) are not used as references. Could you "fix" that please and always use the reference approach? Maybe order the variable declarations the same as the options definitions.
This avoids missing to initialize some due to that mixed approach and gets rid of the templates in `po::value` and type duplications. And I'd like to avoid confusion in future PRs again

PR #1959, opened 2026-07-12
Update submodules and resize save/load window

Flamefire:
Make the window 100px wider and reduce with of the time columns to allow for more space for map and file names

{39CA1994-365C-4BEC-A79E-B9194B2A1B83}

Submodule updates pull in new translations, faster renderer for s25edit and an update to turtle avoiding the flood of warnings for mocked methods


Flamefire:
@Flow86 CI passed

PR #1960, opened 2026-07-15
Add leather/wine addon goods to economy mode if addons activated

ottml:
When the addon leather or wine is active we use the goods from this addons for the enconomy mode too


Flamefire (approved):
LGTM

Issue #953, opened 2018-11-27
AI doesn't build plank production with Very Low resources


MelTraX:
I started a game on the map "Teamisland1" (OTHER) with 5 Hard AIs with Very Low starting resources. Only one of the AIs managed to build a sawmill.

My guess is that harbor buildings have a high priority and because on that map you can build it immediately, the AI does.

It's also interesting that the AIs seem to start an expedition immediately. That's a minor issue but maybe it should only care about harbors and ships after having at least 2 sawmills or something?

Return To The Roots v20181123-db01993df545504b3e8a30bd6f13a0e9905fe80f
Compiled with Gnu GCC C/C++ (5.3.1) for Linux, Unix Environment 64 Bit

MelTraX:
Hmm, the AI can't handle Very Low resources on other maps either? So is this normal behaviour? Shouldn't the AI have an equivalent to the Emergency Program?

Spikeone:
Well, the AI was simply not made for such special cases. Im not sure if the ai ignores the emergency program - at least on very low goods players do / did by storing out wares which then could be used for building (e.g. first iron smelter).

By watching the replay you can find out what the ai exactly does.

Flamefire:
From @88ernie in #1198:
> isn't a solution here to add maybe 2 boards of wood for a computer that he could build sawmill and then he would be able to work properly?

It is not that easy. The AI need to be enhanced to follow some better rules. Like making sure it priories stone/wood production and if not possible right now take the remaining wares into account when building new buildings.   
If you'd just gave it 2 wood it would use it for another barracks or so...

Generally: I'd like an AI which is mainly resource driven: Check what wares it has least and work on that. This should take into account wares it has in stores, wares on the road and existing building sites.

88ernie:
I see... like I siad I'm a noob in programming :D Anyway, good luck trying to improve AI !

Flamefire:
Fixed by #1961

PR #1961, opened 2026-07-16
fixed emergency protocol by removing sheduled wares requested before ...

tost11:
When emergency protocol is activated and many buildings are already planned, the wares already scheduled for delivery are ignoring the protocol.

So now when emergency protocol is activated, those wares get removed (if not for the sawmill or woodcutter).

This also solved the issue for AI enemies on very low resources where they're not able to start because the first sawmill never got built.

### DIFF
--- a/libs/s25main/GamePlayer.cpp
+++ b/libs/s25main/GamePlayer.cpp
@@ -889,7 +889,7 @@ void GamePlayer::FindWarehouseForAllJobs(const Job job)
     }
 }
 
-bool GamePlayer::IsWareFineWithEmergencyProtocol(GoodType goodType, const noBaseBuilding& goal){
+static bool IsWareFineWithEmergencyProtocol(GoodType goodType, const noBaseBuilding& goal){
     return (goodType != GoodType::Boards && goodType != GoodType::Stones) || goal.GetBuildingType() == BuildingType::Woodcutter || goal.GetBuildingType() == BuildingType::Sawmill;
 }
 
@@ -2097,7 +2097,7 @@ void GamePlayer::CancelWaresForEmergencyProtocol()
     for(auto it = ware_list.begin(); it != ware_list.end();){
         Ware * ware = *it;
         // checks if this ware is
-        if(ware->IsWaitingInWarehouse() && !IsWareFineWithEmergencyProtocol(ware->type,*ware->GetGoal()))
+        if(ware->IsWaitingInWarehouse() && ware->GetGoal() != nullptr && !IsWareFineWithEmergencyProtocol(ware->type,*ware->GetGoal()))
         {
             ware->NotifyGoalAboutLostWare();
             static_cast<nobBaseWarehouse*>(ware->GetLocation())->CancelWare(ware);
--- a/libs/s25main/GamePlayer.cpp
+++ b/libs/s25main/GamePlayer.cpp
@@ -2089,6 +2089,25 @@ bool GamePlayer::FindHarborForUnloading(noShip* ship, const MapPoint start, Harb
     return false;
 }
 
+void GamePlayer::CancelWaresForEmergencyProtocol()
+{
+    for(auto it = ware_list.begin(); it != ware_list.end();){
+        Ware * ware = *it;
+        if(ware->IsWaitingInWarehouse())
+        {
+            auto* goal = ware->GetGoal();
+            if(goal != nullptr && goal->GetBuildingType() != BuildingType::Sawmill && goal->GetBuildingType() != BuildingType::Woodcutter)
+            {
+                ware->NotifyGoalAboutLostWare();
+                static_cast<nobBaseWarehouse*>(ware->GetLocation())->CancelWare(ware);
+                it = ware_list.erase(it);
+                continue;
+            }
+        }
+        it++;
+    }
+}
+
 void GamePlayer::TestForEmergencyProgramm()
 {
     // we are already defeated, do not even think about an emergency program - it's too late :-(
@@ -2119,6 +2138,9 @@ void GamePlayer::TestForEmergencyProgramm()
             SendPostMessage(std::make_unique<PostMsg>(
               world.GetEvMgr().GetCurrentGF(), _("The emergency program has been activated."), PostCategory::Economy));
         }
+
+        //remove all existing ware deliveries to buildings not sawmill or woodcutter
+        CancelWaresForEmergencyProtocol();
     } else
     {
         // Sobald Notfall vorbei, Notfallprogramm beenden, evtl. Baustellen wieder mit Kram versorgen
--- a/libs/s25main/GamePlayer.cpp
+++ b/libs/s25main/GamePlayer.cpp
@@ -2097,7 +2097,7 @@ void GamePlayer::CancelWaresForEmergencyProtocol()
     for(auto it = ware_list.begin(); it != ware_list.end();){
         Ware * ware = *it;
         // checks if this ware is
-        if(ware->IsWaitingInWarehouse() && ware->GetGoal() != nullptr && !IsWareFineWithEmergencyProtocol(ware->type,*ware->GetGoal()))
+        if(ware->IsWaitingInWarehouse() && ware->GetGoal() && !IsWareFineWithEmergencyProtocol(ware->type,*ware->GetGoal()))
         {
             ware->NotifyGoalAboutLostWare();
             static_cast<nobBaseWarehouse*>(ware->GetLocation())->CancelWare(ware);
--- a/libs/s25main/GamePlayer.cpp
+++ b/libs/s25main/GamePlayer.cpp
@@ -889,6 +889,10 @@ void GamePlayer::FindWarehouseForAllJobs(const Job job)
     }
 }
 
+bool GamePlayer::IsWareFineWithEmergencyProtocol(GoodType goodType, const noBaseBuilding& goal){
+    return (goodType != GoodType::Boards && goodType != GoodType::Stones) || goal.GetBuildingType() == BuildingType::Woodcutter || goal.GetBuildingType() == BuildingType::Sawmill;
+}
+
 Ware* GamePlayer::OrderWare(const GoodType ware, noBaseBuilding& goal)
 {
     /// Gibt es ein Lagerhaus mit dieser Ware?
@@ -902,8 +906,7 @@ Ware* GamePlayer::OrderWare(const GoodType ware, noBaseBuilding& goal)
         else
         {
             // Wenn Notfallprogramm aktiv nur an Holzfäller und Sägewerke Bretter/Steine liefern
-            if((ware != GoodType::Boards && ware != GoodType::Stones)
-               || goal.GetBuildingType() == BuildingType::Woodcutter || goal.GetBuildingType() == BuildingType::Sawmill)
+            if(IsWareFineWithEmergencyProtocol(ware,goal))
                 return wh->OrderWare(ware, goal);
             else
                 return nullptr;
@@ -2093,16 +2096,13 @@ void GamePlayer::CancelWaresForEmergencyProtocol()
 {
     for(auto it = ware_list.begin(); it != ware_list.end();){
         Ware * ware = *it;
-        if(ware->type == GoodType::Boards && ware->IsWaitingInWarehouse())
+        // checks if this ware is
+        if(ware->IsWaitingInWarehouse() && !IsWareFineWithEmergencyProtocol(ware->type,*ware->GetGoal()))
         {
-            auto* goal = ware->GetGoal();
-            if(goal != nullptr && goal->GetBuildingType() != BuildingType::Sawmill && goal->GetBuildingType() != BuildingType::Woodcutter)
-            {
-                ware->NotifyGoalAboutLostWare();
-                static_cast<nobBaseWarehouse*>(ware->GetLocation())->CancelWare(ware);
-                it = ware_list.erase(it);
-                continue;
-            }
+            ware->NotifyGoalAboutLostWare();
+            static_cast<nobBaseWarehouse*>(ware->GetLocation())->CancelWare(ware);
+            it = ware_list.erase(it);
+            continue;
         }
         it++;
     }
@@ -2137,10 +2137,10 @@ void GamePlayer::TestForEmergencyProgramm()
             emergency = true;
             SendPostMessage(std::make_unique<PostMsg>(
               world.GetEvMgr().GetCurrentGF(), _("The emergency program has been activated."), PostCategory::Economy));
-        }
 
-        //remove all existing ware deliveries to buildings not sawmill or woodcutter
-        CancelWaresForEmergencyProtocol();
+            //Handle wares already ordered
+            CancelWaresForEmergencyProtocol();
+        }
     } else
     {
         // Sobald Notfall vorbei, Notfallprogramm beenden, evtl. Baustellen wieder mit Kram versorgen
--- a/libs/s25main/GamePlayer.cpp
+++ b/libs/s25main/GamePlayer.cpp
@@ -889,8 +889,10 @@ void GamePlayer::FindWarehouseForAllJobs(const Job job)
     }
 }
 
-static bool IsWareFineWithEmergencyProtocol(GoodType goodType, const noBaseBuilding& goal){
-    return (goodType != GoodType::Boards && goodType != GoodType::Stones) || goal.GetBuildingType() == BuildingType::Woodcutter || goal.GetBuildingType() == BuildingType::Sawmill;
+static bool IsWareFineWithEmergencyProtocol(GoodType goodType, const noBaseBuilding& goal)
+{
+    return (goodType != GoodType::Boards && goodType != GoodType::Stones)
+           || goal.GetBuildingType() == BuildingType::Woodcutter || goal.GetBuildingType() == BuildingType::Sawmill;
 }
 
 Ware* GamePlayer::OrderWare(const GoodType ware, noBaseBuilding& goal)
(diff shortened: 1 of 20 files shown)


libs/s25main/GamePlayer.cpp
```diff
@@ -2089,6 +2089,25 @@ bool GamePlayer::FindHarborForUnloading(noShip* ship, const MapPoint start, Harb
...
 }
 
+void GamePlayer::CancelWaresForEmergencyProtocol()
+{
+    for(auto it = ware_list.begin(); it != ware_list.end();){
+        Ware * ware = *it;
+        if(ware->IsWaitingInWarehouse())
+        {
+            auto* goal = ware->GetGoal();
+            if(goal != nullptr && goal->GetBuildingType() != BuildingType::Sawmill && goal->GetBuildingType() != BuildingType::Woodcutter)
```
Flamefire: Shouldn't this apply only to boards and stones? Maybe add a function that checks the ware type and goal and use it here and in `OrderWare` so it is consistent

libs/s25main/GamePlayer.h
```diff
@@ -331,6 +331,8 @@ class GamePlayer : public GamePlayerInfo
     const Statistic& GetStatistic(StatisticTime time) const { return statistic[time]; };
     unsigned GetStatisticCurrentValue(StatisticType idx) const { return statisticCurrentData[idx]; }
 
+    // Lösht alle waren die bereits zur Auslieferung vorbereitet sind aber dem Notfallprogramm wiedersprechen
```
Flamefire: All (new) comments/docstrings must be English

libs/s25main/GamePlayer.cpp
```diff
@@ -2119,6 +2138,9 @@ void GamePlayer::TestForEmergencyProgramm()
             SendPostMessage(std::make_unique<PostMsg>(
               world.GetEvMgr().GetCurrentGF(), _("The emergency program has been activated."), PostCategory::Economy));
         }
+
+        //remove all existing ware deliveries to buildings not sawmill or woodcutter
```
Flamefire: Shouldn't it be enough to do this once? I.e. in the branch above. And comments shouldn't be so explicit: Which wares are cancelled doesn't matter here. So maybe something like "Handle wares already ordered"

Flamefire (changes requested):

libs/s25main/GamePlayer.h
```diff
@@ -331,7 +331,8 @@ class GamePlayer : public GamePlayerInfo
     const Statistic& GetStatistic(StatisticTime time) const { return statistic[time]; };
     unsigned GetStatisticCurrentValue(StatisticType idx) const { return statisticCurrentData[idx]; }
 
-    // Lösht alle waren die bereits zur Auslieferung vorbereitet sind aber dem Notfallprogramm wiedersprechen
+    bool IsWareFineWithEmergencyProtocol(GoodType goodType, const noBaseBuilding& goal);
```
Flamefire: Can be a static or free function (inside the cpp file)

libs/s25main/GamePlayer.cpp
```diff
@@ -2093,16 +2096,13 @@ void GamePlayer::CancelWaresForEmergencyProtocol()
 {
     for(auto it = ware_list.begin(); it != ware_list.end();){
         Ware * ware = *it;
-        if(ware->type == GoodType::Boards && ware->IsWaitingInWarehouse())
+        // checks if this ware is
+        if(ware->IsWaitingInWarehouse() && !IsWareFineWithEmergencyProtocol(ware->type,*ware->GetGoal()))
```
Flamefire: Needs a NULL check for the goal as here it isn't guaranteed to be non_NULL so this could be UB

Flamefire (changes requested):

libs/s25main/GamePlayer.cpp
```diff
@@ -2089,6 +2092,22 @@ bool GamePlayer::FindHarborForUnloading(noShip* ship, const MapPoint start, Harb
     return false;
 }
 
+void GamePlayer::CancelWaresForEmergencyProtocol()
+{
+    for(auto it = ware_list.begin(); it != ware_list.end();){
+        Ware * ware = *it;
+        // checks if this ware is
+        if(ware->IsWaitingInWarehouse() && ware->GetGoal() != nullptr && !IsWareFineWithEmergencyProtocol(ware->type,*ware->GetGoal()))
```
Flamefire:
```suggestion
        if(ware->IsWaitingInWarehouse() && ware->GetGoal() && !IsWareFineWithEmergencyProtocol(ware->type,*ware->GetGoal()))
```

Flamefire (approved):
LGTM,
trivial style issue in the NULL check which I fixed directly. `git pull` if you need to do further changes

Flamefire (changes requested):
You need to run clang-format first. If you have it installed when configuring you can run e.g. `make clangFormat`
Or use the diff as shown in the CI job.

Can you add a test for that? Might be a bit tricky though

Flamefire (dismissed):

Spikeone:
Well, this is original behavior - you were also able to store out wares to get them going anway (a tactic used with very low goods frequently). This fix breaks this original behavior, doesn't it?

This sounds reading your description more like a problem the AI has, and should  be fixed for the AI instead of for players (maybe the AI can use that system which is slight cheating but I'd opt for a smarter AI instead of an AI playing only by rules)

Flamefire:
> Well, this is original behavior - you were also able to store out wares to get them going anway (a tactic used with very low goods frequently). This fix breaks this original behavior, doesn't it?

What we have:
- When emergency program is active newly built buildings don't get wares except woodcutter or sawmills
- With this: Additionally wares ordered for such "blocked" buildings which are still in a warehouse, i.e. not carried out yet, are stopped, wares already on the way are not.

Not fully sure how the original handled that as it seems like a minor corner case happening rarely.
If in the original wares for other buildings are still carried out, then I'd also say that we guard this behavior for AI only. If it helps then I don't see an issue there as it is such a minimal QoL change for rare cases that it likely won't even be noticed in most cases.

tost11:
The issue can also occur for the player, like On low resources, start when starting with multiple buildings (but the sawmill first, but on hilly terrain) and then multiple other buildings (fast before all boards are given out). Then all boards are scheduled for building. The sawmill and woodcutter are already planed. Then some time later the Terain Blow sawmill is leveled, and it requests boards. But all boards are already given out to other buildings. That results in no boards for the sawmill.

But yes, it is an edge case, and yes, I mainly built it for the AI to start proper play on some maps where it gets stuck on low-resource starts.

But actually it's an issue in the game, not only an AI issue.

So what now? I see multiple options that could be implemented to not block fast start for players.

- Fix it only on AI, but behavior stays broken on edge cases.
- Check what the behavior was in the original game.
- Improve emergency protocol with:

-- Only hold back needed boards (2 for the sawmill and 2 for the woodcutter) depending on what buildings are missing. Also maybe 1 more if the street gets removed and one is lost by the player or AI.
-- Unlock emergency protocol if the boards reach the building, not when finished (but on the other hand, without those building boards, production isn't possible anyway).
-- Map boards to exactly one needed building; for example, if two woodcutters build a target, one gets the boards. so other boards can be normally used.

What do you prefer?

For me it would be the last option, and I also would implement that more complicated but better emergency program behavior if wanted.

Flamefire:
> But yes, it is an edge case, and yes, I mainly built it for the AI to start proper play on some maps where it gets stuck on low-resource starts.
> 
> But actually it's an issue in the game, not only an AI issue.

That depends on the definition of "issue". We want (at least the default) behavior to match that of S2, so if "emergency mode" there meant that only newly built buildings do not get boards/stones then that is (usually) what we want.
AI may "cheat" and that's fine as a player could invoke this behavior by removing roads, so that's a shortcut.

> So what now? I see multiple options that could be implemented to not block fast start for players.
> 
>     * Fix it only on AI, but behavior stays broken on edge cases.
No issue with that in any case
>     * Check what the behavior was in the original game.
See above: That would be the most important thing to check first although it may be difficult to get into exactly this situation. But it looks like your description above should be reasonably easy to check. Could you do that?

>     * Improve emergency protocol with:

I think that is too complicated to avoid such a rare issue. In the end this is a resource management game and the emergency mode already a QoL improvement only. So it could be "incomplete" leaving more responsibility to the player.

However as queued wares do not turn up for the player anywhere in the UI it might seem like emergency mode is ignored for a short time. I didn't fully understand @Spikeone 

> Well, this is original behavior - you were also able to store out wares to get them going anway (a tactic used with very low goods frequently). This fix breaks this original behavior, doesn't it?

What exactly does this break? You build a building, wares for this building are queued, emergency mode activates, wares for building stop being carried out as-if the building was built/connected right after the activation.
If you can "force" delivering wares in emergency mode by out-storing then this isn't blocked by this change. Not sure if it was possible in RttR before, but if it was it still will be.

So I'd lean into accepting this change to enforce the _visible_ semantics: When emergency mode is active no ware will leave the WHs (with the purpose) of going to a "disallowed" building site.
Or what do you thing @Spikeone ?

tost11:
I have tested the behavior in the Siedler 2 Gold Edition.

It seems to be the cleverest approach. Boards are stored until only 4 are left. If a woodcutter builds it, it gets two boards. Another one would not receive any boards (like it's currently here); only the sawmill gets them (see screenshots).

When both buildings are finished, the protocol is deactivated.

I have also checked the emergency protocol for stones; if you run out of stones, you are doomed—no emergency protocol for it.

image_1
image_2

libs/s25main/GamePlayer.cpp
```diff
@@ -2089,6 +2094,24 @@ bool GamePlayer::FindHarborForUnloading(noShip* ship, const MapPoint start, Harb
     return false;
 }
 
+void GamePlayer::CancelWaresForEmergencyProtocol()
+{
+    for(auto it = ware_list.begin(); it != ware_list.end();)
+    {
+        Ware* ware = *it;
+        // checks if this ware is
```
Flamefire: incomplete comment?

libs/s25main/GamePlayer.h
```diff
@@ -331,6 +331,8 @@ class GamePlayer : public GamePlayerInfo
     const Statistic& GetStatistic(StatisticTime time) const { return statistic[time]; };
     unsigned GetStatisticCurrentValue(StatisticType idx) const { return statisticCurrentData[idx]; }
 
+    // remove all wares that are already scheduled but ignoring emergency protocol
```
Flamefire: Not sure what "ignoring ..." means here. Maybe:
```suggestion
    // Stop wares restricted in emergency mode that are waiting in warehouse to be transported already
```

tests/s25Main/integration/testEmergencyProtocol.cpp
```diff
@@ -0,0 +1,87 @@
...
+        // wait for some more ticks to give time if not working to deliver more boards
+        RTTR_SKIP_GFS(200);
+        // check boards are still fine and protocol working
+        BOOST_TEST_CHECK(world.GetPlayer(0).GetHQ()->GetInventory()[GoodType::Boards] == 10);
+    }
+};
+
+BOOST_FIXTURE_TEST_CASE(EmergencyProtoclActiveWoodcutterAndSawmillCanBuild, EmergencyFixture)
+{
+    initGameRNG();
```
Flamefire: move to fixture

tests/s25Main/integration/testEmergencyProtocol.cpp
```diff
@@ -0,0 +1,87 @@
...
+#include "worldFixtures/WorldWithGCExecution.h"
+#include "worldFixtures/initGameRNG.hpp"
+#include "gameTypes/GO_Type.h"
+#include <boost/test/unit_test.hpp>
+#include <buildings/nobHQ.h>
+#include <buildings/nobUsual.h>
+
+struct EmergencyFixture : public WorldFixture<CreateEmptyWorld, 1>
+{
+    nobHQ* HQ = world.GetPlayer(0).GetHQ();
```
Flamefire:
```suggestion
    nobHQ* hq = world.GetPlayer(0).GetHQ();
```

tests/s25Main/integration/testEmergencyProtocol.cpp
```diff
@@ -0,0 +1,87 @@
...
+    EmergencyFixture()
+    {
+        HQ->AddToInventory(HQ->getStartInventory(StartWares::VLow), true);
+        MapPoint pos;
+
+        pos = world.GetPlayer(0).GetHQPos() + MapPoint(3, 0);
+        world.SetBuildingSite(BuildingType::Farm, pos, 0);
+        world.BuildRoad(0, false, world.GetNeighbour(pos, Direction::SouthEast),
```
Flamefire: Maybe derive from `WorldWithGCExecution` and use `BuildRoadForBlds` which checks the the roads are built. Especially in the tests it is otherwise not fully clear that the building site is actually connected. Here it is reasonably clear, but for `GetHQPos() + MapPoint(-1, 2)` it is a bit trickier

tests/s25Main/integration/testEmergencyProtocol.cpp
```diff
@@ -0,0 +1,87 @@
...
+        pos = world.GetPlayer(0).GetHQPos() + MapPoint(-3, 0);
+        world.SetBuildingSite(BuildingType::Farm, pos, 0);
+        world.BuildRoad(0, false, world.GetNeighbour(pos, Direction::SouthEast),
+                        std::vector<Direction>(3, Direction::East));
+
+        // wait until emergency protocol should be activated
+        RTTR_EXEC_TILL(500, HQ->GetInventory()[GoodType::Boards] == 10);
+
+        // activate program (with 10 boards it should trigger)
+        world.GetPlayer(0).TestForEmergencyProgramm();
```
Flamefire: Can't we test that this is done automatically? I.e. shouldn't the above activate it? Or at least in the next GF

tests/s25Main/integration/testEmergencyProtocol.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+        // wait for some more ticks to give time if not working to deliver more boards
+        RTTR_SKIP_GFS(200);
+        // check boards are still fine and protocol working
+        BOOST_TEST_CHECK(world.GetPlayer(0).GetHQ()->GetInventory()[GoodType::Boards] == 10);
```
Flamefire:
```suggestion
        // No more boards are carried out to the farms due to emergency protocol
        RTTR_SKIP_GFS(200);
        BOOST_TEST_CHECK(hq->GetInventory()[GoodType::Boards] == 10);
```

tests/s25Main/integration/testEmergencyProtocol.cpp
```diff
@@ -0,0 +1,87 @@
...
+    }
+};
+
+BOOST_FIXTURE_TEST_CASE(EmergencyProtoclActiveWoodcutterAndSawmillCanBuild, EmergencyFixture)
```
Flamefire: Typo here and below
```suggestion
BOOST_FIXTURE_TEST_CASE(EmergencyProtocolActiveWoodcutterAndSawmillCanBuild, EmergencyFixture)
```

tests/s25Main/integration/testEmergencyProtocol.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+    MapPoint posSawmill = world.GetPlayer(0).GetHQPos() + MapPoint(-2, 4);
+    world.SetBuildingSite(BuildingType::Sawmill, posSawmill, 0);
+    world.BuildRoad(0, false, world.GetNeighbour(posSawmill, Direction::SouthEast),
+                    std::vector<Direction>(2, Direction::NorthEast));
+
+    // check if inventory boards are given out
+    RTTR_EXEC_TILL(200, HQ->GetInventory()[GoodType::Boards] < 10);
+
+    // check if building where found
```
Flamefire:
```suggestion
    // check that buildings are built
```

tests/s25Main/integration/testEmergencyProtocol.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+    MapPoint pos = world.GetPlayer(0).GetHQPos() + MapPoint(-1, 2);
+    world.SetBuildingSite(BuildingType::Watchtower, pos, 0);
+    world.BuildRoad(0, false, world.GetNeighbour(pos, Direction::SouthEast),
+                    std::vector<Direction>(2, Direction::NorthEast));
+
+    // wait for some more ticks to give time if not working to deliver more boards
+    RTTR_SKIP_GFS(500);
+    // check boards are still fine and protocol working
+    BOOST_TEST_CHECK(world.GetPlayer(0).GetHQ()->GetInventory()[GoodType::Boards] == 10);
```
Flamefire:
```suggestion
    // No boards are carried out to the farms or watchtower due to emergency protocol
    RTTR_SKIP_GFS(500);
    BOOST_TEST_CHECK(hq->GetInventory()[GoodType::Boards] == 10);
```

Flamefire (changes requested):
I verified with the original too in various situations and you are right: They are immediately stopped in warehouses.

We do have a difference though: S2 activates at 4 boards, RTTR at 10.

Anyway, some minor polishing here and it's good to go

tests/s25Main/integration/testEmergencyProtocol.cpp
tost11: I think so too, but without it, the emergency protocol is not triggering, but I can't find out why

Flamefire (dismissed):
Looks good!
I updated the branch and added a few small fixes to the new test: Lot's of unused includes and use of a fixture-suite and fixing the typo (moved parts to suite name)

The change I added now does the emergency protocol check in the tests similar to the real game. So that explicit check could be removed.
If I haven't missed anything and CI passes I'll merge, unless you notice anything

Flamefire (dismissed):

Flamefire (approved):

tost11:
nice :D, i also think the issue: #953 can be closed then

PR #1967, opened 2026-07-29
Fix double free for geologist sign

Xellzul:
A geologist replacing an old sign with a new one destroys and frees it immediately, leaving the kill list holding a dangling pointer and double-freeing it moments later.

Not sure if it impacts meaningfully replays.


Flamefire:
Can you explain a bit what happens here and why?

From quick inspection the existing code looked correct:
- When Destroy is called the object is immediately removed. The event is removed so the code in HandleEvent doesn't trigger
- When the event triggers the object is added to the kill list where destroy is called removing the object

So what exactly happens to lead to this bug?

PR #1968, opened 2026-08-06
Draw overlay images with an offset.

### LINKED ISSUE #1966: [BUG] Incorrect Y offsets for some figure sprites
Some settler sprites have incorrect y offsets between the body and the overlay images. This can be noticed most easily when the offset is too big, because then the head "floats" in the air. Here are some screenshots

Fat carrier, walking south west.
Image

Thin carrier with a roman shield, walking north east.
Image

Hunter walking west.
Image

It is usually just one of the eight animation steps.

To verify, I exported spritesheets for each job and each good and also saw this. (Using pink as background to see it easily)

Notice the south west and north west rows, fifth column.
Image

I found that when parsing the BOB files, two bytes are considered unknown data.

[uint16_t unknown;](https://github.com/Return-To-The-Roots/libsiedler2/blob/7ae27b5973970872bd34b9ef759bc098bb44decd/src/ArchivItem_Bob.cpp#L147-L148
)

I printed the unknown values and they were either 0, 0x0001 or 0xFFFF. As signed int16_t this would be 0, +1 and -1. Then I changed the exported spritesheets in a way that all +1 sprites would be drawn in a green rect, and all -1 sprites are drawn in a blue rect. Et voilà:

Image

Image

Image

For the green ones it's rather obvious, the blue ones often look like they are kind of bent forward. So the unknown data seems to be some kind of ny-correction. Here are all spritesheets to verify:

[export.zip](https://github.com/user-attachments/files/30521997/export.zip)

1. Which version is affected?
v0.9.5 - 397f2b2 and 20260730 - f519458
2. What did you do/shall be done to reproduce the bug?
Observe a fat carrier walking souh west or north west without good. Press pause until the incorrect animation step is drawn. 
3. What is expected?
No floating heads or squashed necks.
4. What happens instead?
See above.

Lknadfodr:
Fixes #1966
Requires [https://github.com/Return-To-The-Roots/libsiedler2/pull/29](https://github.com/Return-To-The-Roots/libsiedler2/pull/29)

### DIFF
--- a/libs/s25main/ogl/glArchivItem_Bob.cpp
+++ b/libs/s25main/ogl/glArchivItem_Bob.cpp
@@ -20,8 +20,8 @@ void glArchivItem_Bob::Draw(unsigned item, libsiedler2::ImgDir direction, bool f
     auto* glOverlay = dynamic_cast<glArchivItem_Bitmap_Player*>(overlay.first);
     if(glOverlay)
     {
-        DrawPoint overlayDrawPt = DrawPoint(drawPt.x, drawPt.y - overlay.second);
-        glOverlay->DrawFull(overlayDrawPt, COLOR_WHITE, color);
+        drawPt.y -= overlay.second;
+        glOverlay->DrawFull(drawPt, COLOR_WHITE, color);
     }
 }
 
--- a/libs/s25main/ogl/glArchivItem_Bob.cpp
+++ b/libs/s25main/ogl/glArchivItem_Bob.cpp
@@ -20,8 +20,8 @@ void glArchivItem_Bob::Draw(unsigned item, libsiedler2::ImgDir direction, bool f
     auto* glOverlay = dynamic_cast<glArchivItem_Bitmap_Player*>(overlay.first);
     if(glOverlay)
     {
-        DrawPoint drawOverlayPt = DrawPoint(drawPt.x, drawPt.y - overlay.second);
-        glOverlay->DrawFull(drawOverlayPt, COLOR_WHITE, color);
+        DrawPoint overlayDrawPt = DrawPoint(drawPt.x, drawPt.y - overlay.second);
+        glOverlay->DrawFull(overlayDrawPt, COLOR_WHITE, color);
     }
 }
 
--- a/libs/s25main/ogl/glArchivItem_Bob.cpp
+++ b/libs/s25main/ogl/glArchivItem_Bob.cpp
@@ -3,6 +3,7 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 
 #include "glArchivItem_Bob.h"
+#include "DrawPoint.h"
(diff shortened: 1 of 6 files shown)


libs/s25main/ogl/glBitmapItem.cpp
```diff
@@ -13,9 +13,9 @@ glBitmapItem::glBitmapItem(libsiedler2::baseArchivItem_Bitmap* b, bool shadow, b
     b->getVisibleArea(pos.x, pos.y, size.x, size.y);
     origin = Position(b->getNx(), b->getNy()) - pos;
 }
-glBitmapItem::glBitmapItem(libsiedler2::ArchivItem_Bitmap_Player* b, bool isOwning)
+glBitmapItem::glBitmapItem(libsiedler2::ArchivItem_Bitmap_Player* b, int16_t nyOffset, bool isOwning)
```
Flamefire: Use `int` in the interface if not requiring the specific width. And do we need "ny"? At this point it could just be "yOffset", can't it? Especially as you subtract it from Ny similar to `pos`. Is that subtraction correct?

libs/s25main/ogl/glArchivItem_Bob.cpp
```diff
@@ -15,9 +16,13 @@ void glArchivItem_Bob::Draw(unsigned item, libsiedler2::ImgDir direction, bool f
...
     if(body)
         body->DrawFull(drawPt, COLOR_WHITE, color);
-    auto* overlay = dynamic_cast<glArchivItem_Bitmap_Player*>(getOverlay(item, fat, direction, animationstep));
-    if(overlay)
-        overlay->DrawFull(drawPt, COLOR_WHITE, color);
+    auto overlay = getOverlay(item, fat, direction, animationstep);
+    auto* glOverlay = dynamic_cast<glArchivItem_Bitmap_Player*>(overlay.first);
+    if(glOverlay)
+    {
+        DrawPoint overlayDrawPt = DrawPoint(drawPt.x, drawPt.y - overlay.second);
```
Flamefire: Maybe just: `drawPt.y -= overlay.second`? I want to avoid `.x`/`.y` pairs where possible as they are prone to C&P mistakes and the suggestion is more clear that the y-coord is modified.

libs/s25main/ogl/glBitmapItem.cpp
Lknadfodr: Changed it.

Yes, the subtraction is correct, I tested both. But I can't explain why, I gess the S2 developers used different coordinate systems.

libs/s25main/ogl/glArchivItem_Bob.cpp
Lknadfodr: Aye.

Flamefire (approved):

PR #1969, opened 2026-08-17
Fix underflow when trading

ottml:
Steps to reproduce:

- Build two storehouses
- Both storehouses have figures with one type, for example woodcutter
- Both storehouses have 2 woodcutters
- Send 1 woodcutter to ally storehouse
- Then 3 woodcutters are sent to the ally instead of 1

It sends the requested number from the first WH and then basically ALL from the other warehouses due to count - available underflowing

### DIFF
--- a/libs/s25main/buildings/nobBaseWarehouse.cpp
+++ b/libs/s25main/buildings/nobBaseWarehouse.cpp
@@ -1451,7 +1451,7 @@ unsigned nobBaseWarehouse::GetAvailableFiguresForTrading(const Job job) const
     if(job == Job::Helper)
         return (inventory[Job::Helper] - 1) / 2; // need one as leader
     else
-        return std::min(inventory[job], inventory[Job::Helper] - 1);
+        return inventory[job];
 }
 
 void nobBaseWarehouse::StartTradeCaravane(const boost_variant2<GoodType, Job>& what, const unsigned count,
--- a/tests/s25Main/integration/testTrading.cpp
+++ b/tests/s25Main/integration/testTrading.cpp
@@ -211,7 +211,6 @@ BOOST_AUTO_TEST_CASE(TradeLessFiguresThenInStoreHouse)
     initGameRNG();
 
     auto* const hqPlayer1 = world.GetSpecObj<nobBaseWarehouse>(players[1]->GetHQPos());
-    BOOST_TEST_REQUIRE(hqPlayer1->GetNumRealFigures(Job::Woodcutter) == 8);
 
     // Add second warehouse
     auto* wh1 = static_cast<nobBaseWarehouse*>(BuildingFactory::CreateBuilding(
--- a/tests/s25Main/integration/testTrading.cpp
+++ b/tests/s25Main/integration/testTrading.cpp
@@ -206,7 +206,7 @@ BOOST_AUTO_TEST_CASE(TradeToMuch)
     testAfterLeaving(20);
 }
 
-BOOST_AUTO_TEST_CASE(NoOverflowWhenTradingLessFiguresThenInStoreHouse)
+BOOST_AUTO_TEST_CASE(TradeLessFiguresThenInStoreHouse)
 {
     initGameRNG();
 
--- a/tests/s25Main/integration/testTrading.cpp
+++ b/tests/s25Main/integration/testTrading.cpp
@@ -14,6 +14,7 @@
 #include "gameData/JobConsts.h"
 #include <rttr/test/LogAccessor.hpp>
 #include <boost/test/unit_test.hpp>
+#include <factories/BuildingFactory.h>
 #include <variant.h>
 
 struct TradeFixture : public WorldWithGCExecution3P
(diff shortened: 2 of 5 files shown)


ottml:
@Flamefire @Flow86 Could it be that some permissions are missing for pullrequests from a fork? Because the ci above is not starting because of 

> 2 workflows awaiting approval

libs/s25main/buildings/nobBaseWarehouse.cpp
```diff
@@ -1451,7 +1451,7 @@ unsigned nobBaseWarehouse::GetAvailableFiguresForTrading(const Job job) const
     if(job == Job::Helper)
         return (inventory[Job::Helper] - 1) / 2; // need one as leader
     else
-        return std::min(inventory[job], inventory[Job::Helper] - 1);
```
Flamefire, line 1454: This looks like a 2nd bug: It seems like this intended to include recruiting but trading does not recruit so the number here was wrong. Might be worth a 2nd test or part of the test: Trade more or less than the number of available figures

tests/s25Main/integration/testTrading.cpp
```diff
@@ -205,6 +206,38 @@ BOOST_AUTO_TEST_CASE(TradeToMuch)
     testAfterLeaving(20);
 }
 
+BOOST_AUTO_TEST_CASE(NoOverflowWhenTradingLessFiguresThenInStoreHouse)
+{
+    auto* const hqPlayer0 = world.GetSpecObj<nobBaseWarehouse>(players[0]->GetHQPos());
+    auto* const hqPlayer1 = world.GetSpecObj<nobBaseWarehouse>(players[1]->GetHQPos());
+    BOOST_TEST_REQUIRE(hqPlayer0->GetNumRealFigures(Job::Woodcutter) == 8);
+    BOOST_TEST_REQUIRE(hqPlayer1->GetNumRealFigures(Job::Woodcutter) == 8);
```
Flamefire: That can be removed:
`hqPlayer0` doesn't seem to matter in the test and the other is tested again belo

tests/s25Main/integration/testTrading.cpp
```diff
@@ -205,6 +206,38 @@ BOOST_AUTO_TEST_CASE(TradeToMuch)
     testAfterLeaving(20);
 }
 
+BOOST_AUTO_TEST_CASE(NoOverflowWhenTradingLessFiguresThenInStoreHouse)
```
Flamefire: I don't see a test for an "overflow" here. I guess the test is simply:

```suggestion
BOOST_AUTO_TEST_CASE(TradeLessFiguresThenInStoreHouse)
```

Flamefire (changes requested):
> Steps to reproduce:
> 
> -     Build two storehouses
> -     Both storehouses have figures with one type, for example woodcutter
> -     Both storehouses have 2 woodcutters
> -     Send 1 woodcutter to ally storehouse
> -     Then 3 woodcutters are sent to the ally instead of 1


I don't see how 2 WHs are required for this bug. It seems the issue is that all woodcutters are sent instead of only the requested one. Can you explain?

> @Flamefire @Flow86 Could it be that some permissions are missing for pullrequests from a fork? Because the ci above is not starting because of
> 
> > 2 workflows awaiting approval

That's intentional: Especially with AI PRs this avoids people flooding CI. Not sure if (still) required here though

ottml:
> > Steps to reproduce:
> > 
> > * ```
> >   Build two storehouses
> >   ```
> > * ```
> >   Both storehouses have figures with one type, for example woodcutter
> >   ```
> > * ```
> >   Both storehouses have 2 woodcutters
> >   ```
> > * ```
> >   Send 1 woodcutter to ally storehouse
> >   ```
> > * ```
> >   Then 3 woodcutters are sent to the ally instead of 1
> >   ```
> 
> I don't see how 2 WHs are required for this bug. It seems the issue is that all woodcutters are sent instead of only the requested one. Can you explain?
> 
> > @Flamefire @Flow86 Could it be that some permissions are missing for pullrequests from a fork? Because the ci above is not starting because of
> > > 2 workflows awaiting approval
> 
> That's intentional: Especially with AI PRs this avoids people flooding CI. Not sure if (still) required here though

If you have only one warehouse we get a overflow but this has now effect because the warehouse loop is left after the one woodcutter is sent from this warehouse. Only if there are more then one warehouse with the specific job the overflow has an effect. Because the count variable has an infite value and the count==0 check does never match.

Ah ok this is because of the AI. Must the approval be done after each commit or only for a newly created pull request? If this must be approved every time it could be difficult to fix unit-test and clang-tidy.

tests/s25Main/integration/testTrading.cpp
```diff
@@ -205,6 +206,38 @@ BOOST_AUTO_TEST_CASE(TradeToMuch)
...
+    BOOST_TEST_REQUIRE(hqPlayer1->GetNumRealFigures(Job::Woodcutter) == 8);
+    BOOST_TEST_REQUIRE(wh1->GetNumRealFigures(Job::Woodcutter) == 2);
+
+    initGameRNG();
+
+    // Trade less figures then in nearest warehouse wh1
+    this->TradeOverLand(players[0]->GetHQPos(), Job::Woodcutter, 1);
+
+    // Run enough GFs so all trade caravans are out (~numTradeItems + 1 people need to leave taking 30GFs max each)
+    RTTR_EXEC_TILL(30 * (1 + 1), wh1->GetLeavingFigures().empty());
```
Flamefire: Maybe this should also consider the HQ. Otherwise there could be carriers leaving from the HQ while the the trade caravan from wh1 is leaving and you'll miss the wrongly leaving woodcutters from the HQ which was the issue before.
Best to enhance the comment:
> Trade less figures then in nearest warehouse wh1 should use only figures from that WH

tests/s25Main/integration/testTrading.cpp
```diff
@@ -205,6 +206,38 @@ BOOST_AUTO_TEST_CASE(TradeToMuch)
...
+
+    PeopleCounts inv;
+    inv[Job::Woodcutter] = 2;
+    inv[Job::Helper] = 1;
+    wh1->AddToInventory(inv, true);
+
+    BOOST_TEST_REQUIRE(hqPlayer1->GetNumRealFigures(Job::Woodcutter) == 8);
+    BOOST_TEST_REQUIRE(wh1->GetNumRealFigures(Job::Woodcutter) == 2);
+
+    initGameRNG();
```
Flamefire: This seems odd to have here and moves the first and 2nd check for num figures further apart.
Best to move this to the start and/or all of those `initGameRNG` to the fixture instead

Flamefire:
> If you have only one warehouse we get a overflow but this has now effect because the warehouse loop is left after the one woodcutter is sent from this warehouse. Only if there are more then one warehouse with the specific job the overflow has an effect. Because the count variable has an infite value and the count==0 check does never match.

Ok so the issue is that it sends the requested number from the first WH and then basically ALL from the other warehouses due to `count - available` **under**flowing. Can you add that to the PR description?

> Ah ok this is because of the AI. Must the approval be done after each commit or only for a newly created pull request? If this must be approved every time it could be difficult to fix unit-test and clang-tidy.

It's done for all PRs just because they could be AI.
Not fully sure. At least reviewers get a notification on pushes and can re-approve timely

libs/s25main/buildings/nobBaseWarehouse.cpp
ottml: I think we do not need an additonal test because the less part is already tested in in the new test `TradeLessFiguresThenInStoreHouse` and the more case in the test case above `TradeToMuch`

tests/s25Main/integration/testTrading.cpp
```diff
@@ -205,6 +206,38 @@ BOOST_AUTO_TEST_CASE(TradeToMuch)
     testAfterLeaving(20);
 }
 
+BOOST_AUTO_TEST_CASE(TradeLessFiguresThenInStoreHouse)
+{
+    initGameRNG();
+
+    auto* const hqPlayer0 = world.GetSpecObj<nobBaseWarehouse>(players[0]->GetHQPos());
+    auto* const hqPlayer1 = world.GetSpecObj<nobBaseWarehouse>(players[1]->GetHQPos());
+    BOOST_TEST_REQUIRE(hqPlayer1->GetNumRealFigures(Job::Woodcutter) == 8);
```
Flamefire: This can be removed too as it is checked below (line 227
```suggestion
```

tests/s25Main/integration/testTrading.cpp
```diff
@@ -205,6 +206,38 @@ BOOST_AUTO_TEST_CASE(TradeToMuch)
     testAfterLeaving(20);
 }
 
+BOOST_AUTO_TEST_CASE(TradeLessFiguresThenInStoreHouse)
+{
+    initGameRNG();
+
+    auto* const hqPlayer0 = world.GetSpecObj<nobBaseWarehouse>(players[0]->GetHQPos());
```
Flamefire: unused
```suggestion
```

Flamefire (approved):

PR #1971, opened 2026-08-26
GHA CI adjustments

Flamefire:
- `softprops/action-gh-release` doesn't support/need the token input
- Updated markdown-lint option requires the included files, not the excluded ones
- Use `$GITHUB_OUTPUT` to avoid deprecation
- Continue full matrix for tests - Useful when running into compiler-(version) specific issues to have other results available too to see if it is a general issue

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -31,6 +31,7 @@ env:
 jobs:
   Windows:
     strategy:
+      fail-fast: false
       matrix:
         include:
           - { os: windows-2022, generator: Visual Studio 17 2022, type: Debug,   platform: Win32}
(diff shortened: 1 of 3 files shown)


.github/workflows/unit-tests.yml
```diff
@@ -31,6 +31,7 @@ env:
 jobs:
   Windows:
     strategy:
+      fail-fast: false
```
Flow86, line 34: this may be an issue if a lot of PRs are running that we get into throttling due too many jobs running,

maybe leave the fail-fast on for now?

Flamefire: It is only an issue if there are a lot of PRs starting CI at the same time and failing in at least some jobs before the others are finished.
This currently isn't an issue, is it?

Also these CI jobs take around 30min so I'd say fast enough that we won't run into those wait times, at least not often.

So IMO the value of seeing if the failure is platform specific is higher than the cost of some longer wait times in rare situations.

PR #1973, opened 2026-08-26
Fix ambiguous clamp and switch to `std::chrono::milliseconds`

Flamefire:
C++17 introduced `std::clamp` which makes passing a type from std namespace (such as chrono-types) ambiguous in the semi-recursive call. Use constexpr-ifs and `std::clamp` in the final path.

### DIFF
--- a/libs/s25main/network/GameClient.cpp
+++ b/libs/s25main/network/GameClient.cpp
@@ -1199,7 +1199,7 @@ void GameClient::DecreaseSpeed()
         SetNewSpeed(framesinfo.gfLengthReq / SPEED_STEP * SPEED_STEP + SPEED_STEP);
 }
 
-void GameClient::SetNewSpeed(FramesInfo::milliseconds32_t gfLength)
+void GameClient::SetNewSpeed(std::chrono::milliseconds gfLength)
 {
     static_assert(MIN_SPEED >= SPEED_GF_LENGTHS[GameSpeed::VeryFast], "Not all speeds reachable");
     static_assert(MAX_SPEED <= SPEED_GF_LENGTHS[GameSpeed::VerySlow], "Not all speeds reachable");
@@ -1221,7 +1221,7 @@ bool GameClient::OnGameMessage(const GameMessage_Server_NWFDone& msg)
     if(!nwfInfo)
         return true;
 
-    if(!nwfInfo->addServerInfo(NWFServerInfo(msg.gf, msg.gf_length, msg.nextNWF)))
+    if(!nwfInfo->addServerInfo(NWFServerInfo(msg.gf, std::chrono::milliseconds(msg.gf_length), msg.nextNWF)))
     {
         RTTR_Assert(false);
         LOG.write("Failed to add server info. Invalid server?\n");
@@ -1293,7 +1293,7 @@ void GameClient::ExecuteGameFrame()
     if(framesinfo.forcePauseLen.count())
     {
         if(currentTime - framesinfo.forcePauseStart > framesinfo.forcePauseLen)
-            framesinfo.forcePauseLen = FramesInfo::milliseconds32_t::zero();
+            framesinfo.forcePauseLen = framesinfo.forcePauseLen.zero();
         else
             return; // Pause
     }
@@ -1342,7 +1342,7 @@ void GameClient::ExecuteGameFrame()
 
                     ExecuteNWF();
 
-                    FramesInfo::milliseconds32_t oldGFLen = framesinfo.gf_length;
+                    const auto oldGFLen = framesinfo.gf_length;
                     nwfInfo->execute(framesinfo);
                     if(oldGFLen != framesinfo.gf_length)
                     {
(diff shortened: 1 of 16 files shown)


Flow86:
pipeline fails?

Flamefire:
Hm, seems like this didn't work standalone due to mixing duration types. Switching to use the chrono types consistently now. There was one place where we used our own 32 bit ms type but that's pretty much irrelevant now in 64bit days.
Using the standard types allows using the chrono literals too

libs/s25main/network/GameClient.cpp
```diff
@@ -1604,34 +1605,33 @@ unsigned GameClient::GetGlobalAnimation(const unsigned short max, const unsigned
...
     RTTR_Assert(ev);
     // TODO: Move to some animation system that is part of game
-    FramesInfo::milliseconds32_t elapsedTime;
+    std::chrono::milliseconds elapsedTime;
     if(state == ClientState::Game)
         elapsedTime = (GetGFNumber() - ev->startGF) * framesinfo.gf_length + framesinfo.frameTime;
     else
-        elapsedTime = FramesInfo::milliseconds32_t::zero();
-    FramesInfo::milliseconds32_t duration = ev->length * framesinfo.gf_length;
+        elapsedTime = 0ms;
```
Flow86: isnt this automatically 0 when using std::chrono::milliseconds above? otherwise init it  with 0 and we can skip the else completely

libs/s25main/network/GameClient.cpp
```diff
@@ -1604,34 +1605,33 @@ unsigned GameClient::GetGlobalAnimation(const unsigned short max, const unsigned
...
 {
     RTTR_Assert(ev);
-    FramesInfo::milliseconds32_t elapsedTime;
+    std::chrono::milliseconds elapsedTime;
     if(state == ClientState::Game)
         elapsedTime = (GetGFNumber() - ev->startGF) * framesinfo.gf_length + framesinfo.frameTime;
     else
-        elapsedTime = FramesInfo::milliseconds32_t::zero();
-    FramesInfo::milliseconds32_t duration = ev->length * framesinfo.gf_length;
+        elapsedTime = 0ms;
```
Flow86: same here (code is duplicated ? maybe time to have another  helper?)

libs/s25main/network/GameServer.cpp
```diff
@@ -558,7 +559,7 @@ bool GameServer::StartGame()
     framesinfo.gfLengthReq = framesinfo.gf_length = SPEED_GF_LENGTHS[ggs_.speed];
 
     // NetworkFrame-Länge bestimmen, je schlechter (also höher) die Pings, desto länger auch die Framelänge
-    framesinfo.nwf_length = CalcNWFLenght(FramesInfo::milliseconds32_t(highest_ping));
+    framesinfo.nwf_length = CalcNWFLenght(std::chrono::milliseconds(highest_ping));
```
Flow86: thought we already fixed all these typos `CalcNWFLenght` vs ` CalcNWFLength`

libs/s25main/network/GameServer.cpp
```diff
@@ -595,7 +595,7 @@ unsigned GameServer::CalcNWFLenght(FramesInfo::milliseconds32_t minDuration) con
 void GameServer::SendNWFDone(const NWFServerInfo& info)
 {
     nwfInfo.addServerInfo(info);
-    SendToAll(GameMessage_Server_NWFDone(info.gf, info.newGFLen, info.nextNWF));
+    SendToAll(GameMessage_Server_NWFDone(info.gf, static_cast<unsigned>(info.newGFLen / 1ms), info.nextNWF));
```
Flow86, line 598: instead of `/ 1ms` can't we  use ` .count()`? (and we also should not need the  `static_cast` then?)

tests/testHelpers/rttr/test/TmpFolder.hpp
```diff
@@ -24,6 +24,8 @@ class TmpFolder
     ~TmpFolder() { boost::filesystem::remove_all(folder); }
     const boost::filesystem::path& get() const { return folder; }
     boost::filesystem::path operator/(const boost::filesystem::path& rhs) const { return folder / rhs; }
-    operator const boost::filesystem::path &() const { return folder; }
+    // clang-format off
+    operator const boost::filesystem::path&() const { return folder; }
+    // clang-format on
```
Flow86, line 29: totally unrelated to this branch

Flamefire: Kind of: Everytime I formatted the changes of this branch it would change this which would fail CI.
And it's so small of a change that it's not worth opening a separate PR for that, is it?

libs/s25main/network/GameServer.cpp
Flamefire: The `x / duration` basically means: Give me `x` in units of `duration`. That's what we want here. The alternative is to cast to milliseconds (no-op in this case) and then use `.count()` because at this point we need to be explicit about having the count in ms
The cast is required either way: `std::milliseconds` is defined as a signed integer (all std durations are) but in the message negative values don't make sense

libs/s25main/network/GameClient.cpp
Flamefire: Same semantics as `int`, so no.
Initialized now

libs/s25main/network/GameClient.cpp
Flamefire: Introduced helper

libs/s25main/network/GameServer.cpp
Flamefire: Seemingly missed that one. One more in external/s25edit/SGE/sge_collision.cpp but that's a semi-external library. So IMO not worth changing the one variable

Flow86 (approved):

PR #1979, opened 2026-09-09
Store the addon in AddonGui and iterate the GUIs directly

MichalLabuda:
Follow-up to [73aafa2db](https://github.com/Return-To-The-Roots/s25client/commit/73aafa2db35babed5da0d791b3b4d17c29ccba81)

Five loops in `iwAddons` walked `addonGuis_` and `ggs.getAddon(i)` by the same
index, with nothing enforcing that the two stay aligned, and dereferenced the
returned `const Addon*` unchecked - while the constructor guards that same call
with `assertNonNull`. `AddonGui` already receives the addon to build its label
and description button, so it now keeps it and exposes `getAddon()` alongside
`getWindow()`.

Drops two German comments on rewritten lines: one naming `ADDONMANAGER`,
which no longer exists, and one redundant.

### DIFF
--- a/libs/s25main/addons/Addon.cpp
+++ b/libs/s25main/addons/Addon.cpp
@@ -7,7 +7,7 @@
 #include "Window.h"
 #include "s25util/colors.h"
 
-AddonGui::AddonGui(const Addon& addon, Window& window, bool /*readonly*/) : window_(window)
+AddonGui::AddonGui(const Addon& addon, Window& window, bool /*readonly*/) : addon_(addon), window_(window)
 {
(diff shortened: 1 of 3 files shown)


Flamefire (approved):

MichalLabuda:
Merge is blocked by the codecov/patch check. The refactored lines were already uncovered on master, so no coverage was lost — codecov/project is +0.01%. Could you merge it with the admin bypass?

Flamefire:
Coverage is informational only. I just forgot to enable auto-merge

PR #1980, opened 2026-09-09
Act on the selected addon preset, not a typed name

MichalLabuda:
Follow-up to [#1951](https://github.com/Return-To-The-Roots/s25client/pull/1951)

Removed **Delete** button from **Save Addon Preset** window
Removed name field from **Load Addon Preset** window

Load and Delete resolved their target from the name field, with the list
selection only prefilling it. So editing the name after picking a preset
silently retargeted the action, an unmatched name raised "Preset Not Found",
and an empty one did nothing.

Both now act on the list selection and are disabled until a preset is picked.
That makes those states unrepresentable, so `GetTargetFilePath()`,
`GetTargetFileOrNotify()` and the "Preset Not Found" strings are gone.

A typed name is only needed for saving, and deleting belongs with browsing
presets rather than with saving one.

image
image

### DIFF
--- a/libs/s25main/ingameWindows/iwAddonPresets.cpp
+++ b/libs/s25main/ingameWindows/iwAddonPresets.cpp
@@ -7,6 +7,8 @@
 #include "Loader.h"
 #include "RttrConfig.h"
 #include "WindowManager.h"
+#include "commonDefines.h"
+#include "controls/ctrlButton.h"
 #include "controls/ctrlEdit.h"
 #include "controls/ctrlTable.h"
 #include "controls/ctrlText.h"
@@ -65,8 +67,8 @@ static std::optional<std::map<unsigned, unsigned>> LoadPresetsFromFile(const bfs
 }
 
 // iwAddonPresetsBase
-iwAddonPresetsBase::iwAddonPresetsBase(const std::string& title, const std::string& actionLabel)
-    : IngameWindow(CGI_ADDON_PRESETS, IngameWindow::posLastOrCenter, Extent(440, 330), title,
+iwAddonPresetsBase::iwAddonPresetsBase(const std::string& title, const unsigned height)
+    : IngameWindow(CGI_ADDON_PRESETS, IngameWindow::posLastOrCenter, Extent(440, height), title,
                    LOADER.GetImageN("resource", 41), true)
 {
     const bfs::path presetsDir = GetPresetsDir();
@@ -91,20 +93,12 @@ iwAddonPresetsBase::iwAddonPresetsBase(const std::string& title, const std::stri
     AddText(ID_txtFolder, DrawPoint(20, 236), presetsDir.string(), COLOR_YELLOW, FontStyle::TOP, SmallFont)
       ->setMaxWidth(400);
 
-    // maxLength 251 = 255 filename limit - 4 chars for ".ini"; just discourages absurdly long
-    // input, isValidFileName() may still reject it since it counts bytes, not codepoints.
-    AddEdit(ID_edtName, DrawPoint(20, 254), Extent(400, 22), TextureColor::Green2, NormalFont, 251);
-    GetCtrl<ctrlEdit>(ID_edtName)->SetType(EditType::Filename);
-
-    AddTextButton(ID_btAction, DrawPoint(20, 284), Extent(185, 22), TextureColor::Green2, actionLabel, NormalFont);
-    AddTextButton(ID_btDelete, DrawPoint(235, 284), Extent(185, 22), TextureColor::Red1, _("Delete"), NormalFont);
-
     RefreshTable();
 }
 
 void iwAddonPresetsBase::RefreshTable()
 {
-    auto& table = *GetCtrl<ctrlTable>(ID_tblPresets);
+    auto& table = assertNonNull(GetCtrl<ctrlTable>(ID_tblPresets));
     table.DeleteAllItems();
 
     for(const auto& file : ListDir(GetPresetsDir(), "ini"))
@@ -113,56 +107,10 @@ void iwAddonPresetsBase::RefreshTable()
     table.SortRows(0, TableSortDir::Ascending);
 }
 
-bfs::path iwAddonPresetsBase::GetTargetFilePath() const
-{
-    const std::string name = GetCtrl<ctrlEdit>(ID_edtName)->GetText();
-    if(name.empty())
-        return {};
-
-    const auto& table = *GetCtrl<ctrlTable>(ID_tblPresets);
-    for(unsigned short i = 0; i < table.GetNumRows(); ++i)
-    {
-        if(table.GetItemText(i, 0) == name)
-            return table.GetItemText(i, 1);
-    }
-    return {};
-}
-
-bfs::path iwAddonPresetsBase::GetTargetFileOrNotify() const
-{
-    bfs::path path = GetTargetFilePath();
-    if(!path.empty())
-        return path;
-
-    const std::string name = GetCtrl<ctrlEdit>(ID_edtName)->GetText();
-    if(name.empty())
-        return {};
-
-    WINDOWMANAGER.Show(std::make_unique<iwMsgbox>(_("Preset Not Found"),
-                                                  helpers::format(_("Preset '%1%' was not found."), name), nullptr,
-                                                  MsgboxButton::Ok, MsgboxIcon::ExclamationRed));
-    return {};
-}
-
-void iwAddonPresetsBase::Msg_EditEnter(const unsigned /*ctrl_id*/)
-{
-    DoAction();
-}
-
 void iwAddonPresetsBase::Msg_ButtonClick(const unsigned ctrl_id)
 {
-    switch(ctrl_id)
-    {
-        case ID_btAction: DoAction(); break;
-        case ID_btDelete: ConfirmDelete(); break;
-        default: break;
-    }
-}
-
-void iwAddonPresetsBase::Msg_TableSelectItem(const unsigned /*ctrl_id*/, const std::optional<unsigned>& selection)
-{
-    const auto& table = *GetCtrl<ctrlTable>(ID_tblPresets);
-    GetCtrl<ctrlEdit>(ID_edtName)->SetText(selection ? table.GetItemText(*selection, 0) : "");
+    RTTR_Assert(ctrl_id == ID_btAction);
+    DoAction();
 }
 
 void iwAddonPresetsBase::Msg_TableChooseItem(const unsigned /*ctrl_id*/, const unsigned /*selection*/)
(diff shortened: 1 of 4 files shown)


libs/s25main/ingameWindows/iwAddonPresets.cpp
```diff
@@ -113,101 +107,43 @@ void iwAddonPresetsBase::RefreshTable()
...
+    : iwAddonPresetsBase(_("Save Addon Preset"), 330), states_(std::move(states))
 {
-    switch(ctrl_id)
-    {
-        case ID_btAction: DoAction(); break;
-        case ID_btDelete: ConfirmDelete(); break;
-        default: break;
-    }
-}
+    if(ShouldBeClosed())
```
Flamefire: What is this for in the ctor? How can that be true?

libs/s25main/ingameWindows/iwAddonPresets.cpp
```diff
@@ -113,101 +107,43 @@ void iwAddonPresetsBase::RefreshTable()
...
 }
 
-void iwAddonPresetsBase::Msg_EditEnter(const unsigned /*ctrl_id*/)
+void iwAddonPresetsBase::Msg_TableChooseItem(const unsigned /*ctrl_id*/, const unsigned /*selection*/)
 {
     DoAction();
 }
 
-void iwAddonPresetsBase::Msg_ButtonClick(const unsigned ctrl_id)
+// iwSaveAddonPreset
```
Flamefire: useless comment

libs/s25main/ingameWindows/iwAddonPresets.cpp
```diff
@@ -262,26 +198,54 @@ void iwSaveAddonPreset::SaveToPath(const bfs::path& filePath)
...
+        SaveToPath(GetPresetsDir() / fileNameResult.name);
 }
 
 // iwLoadAddonPreset
 iwLoadAddonPreset::iwLoadAddonPreset(std::function<void(const std::map<unsigned, unsigned>&)> onLoad)
-    : iwAddonPresetsBase(_("Load Addon Preset"), _("Load")), onLoad_(std::move(onLoad))
-{}
+    : iwAddonPresetsBase(_("Load Addon Preset"), 300), onLoad_(std::move(onLoad))
+{
+    if(ShouldBeClosed())
```
Flamefire: ?

Flamefire (changes requested):
Not sure about removing the delete button from saving. I can imagine a use case where a user wants to store one but delete an existing one first to avoid confusing it with another one.

How about another layout:
```
<list>
<delete button>
<space>
<edit>
<save button>
```

So "delete" only works on the selected item

MichalLabuda:
On the **Delete** button: in the **Save Addon Preset** window it had two possible targets - the row selected in the list and the name typed in the edit box - and those differ as soon as the user types after selecting. In the **Load Addon Preset** window the selection is the only target for both buttons. Remoing **Delete** button from save window, is the only solution I could think of that makes the UI most intuitive and eliminates edit field vs. selected preset confusion (saving as whatever is in the edit field (minus leading spaces) is usually expected, not what's selected. Quite the opposite is true for delete action). This also makes the code a bit more readable and predictable.

As for your layout, it could resolve the ambiguity somewhat but then load window would need buttons rearranged too as to keep both windows visually consistent (same Delete button placement). Also it doesn't solve problems like inconsistency when you add spaces at the beginning of the existing preset name in the edit box. Before this PR, if you select a preset (it fills edit field) and then add spaces at the beginning of this existing preset name, **Save** will ask if you want to overwrite, because spaces at the beginning are trimmed on save and it points back to the preset you selected. At the same time clicking **Delete** will say "Preset not found" because it looks for exact match (with spaces).
Yet another solution to this problem could be selecting/deselecting a preset on the list depending on the edit field content. Matching preset would be selected on the list and we could disable **Delete** button if nothing is selected. But auto-select will face the same problem with trimming spaces. To do it right some additional real-time input processing would have to be added or auto-selection abandoned. In general, what this PR introduces is the simplest solution that mostly addresses all the problems except maybe for manual OS-side filename manipulations but there has to be limit of responsibility somewhere...

I wanted to fix it before merging the original addon presets PR but was late with pushing my work. I had it done but not fully verified/thought through yet so I lingered. That's why this follow-up came in such a short time after original PR was merged - work was already done.

libs/s25main/ingameWindows/iwAddonPresets.cpp
MichalLabuda: It could only be true in a single situation: when the ctor couldn't ensure the presets folder exists. In that case the base ctor showed an error and called Close() without building the table, so this skipped building the rest of a window that was about to close anyway.

That was a bit clunky, so I rewrote it: there's now iwAddonPresetsBase::EnsurePresetsFolder(), which iwAddons calls before opening either preset window. If the folder can't be created the user gets the error and the window is never constructed at all, so the ctor can just assume the folder is there and both guards are gone. Much more readable.

Flamefire: > so this skipped building the rest of a window that was about to close anyway.

IMO that solution had a big advantage: No non-local assumptions. What happens now when the folder does not exist when the window is opened?

It looks like the guards are not even required: Yes it will do some useless work, but that is never noticeable, is it? Or would it crash?

Flamefire:
> Yet another solution to this problem could be selecting/deselecting a preset on the list depending on the edit field content. Matching preset would be selected on the list and we could disable Delete button if nothing is selected. But auto-select will face the same problem with trimming spaces

So the only issue is that the save-window has an edit which may confuse users. Your idea there can actually be used: `Msg_EditChange` is called when the edit is modified. In this case the delete button can simply be disabled. It is only ever enabled when selecting a preset which fills the edit and enables the button. If the user modifies the edit it is disabled even if that happens to match an entry.
I'd say deleting a preset by typing its name and then clicking a button is unusual enough to not need to support it. But deleting a file before saving a new one seems more usual to me. And disabling the delete button when nothing is selected seems consistent too
And it seems easy enough to implement.

Or what do you think?

As for placement of controls: save window only needs to add edit and save button to the bottom. Could even be done semi-automatically by having a constant for the additional height and then using getIwHeight to get the intended position for the edit.
So the delete button and logic can be in the base

libs/s25main/ingameWindows/iwAddonPresets.cpp
MichalLabuda: I've now made it consistent with the rest of the code, added folder creation to the startup sequence (`InitDirectories`) and removed any special guards.

Worth pointing out that wherever in the code folders are read, the app crashes if read permission is missing from the folder in question, while the startup sequence will happily pass. So if we were to fix it, it would have to be a global fix, probably in `ListDir` - though `ArchiveLocator` iterates a directory on its own, so that alone wouldn't cover everything. Write operations fail gracefully (at least those I checked do). We could also implement a write-read-delete round-trip for each folder at startup but this seems excessive. Your take on this?

Side note: the spare slot in `std::array<std::string, 10> dirs` has been empty since 55cd323bb (2020) folded `lstsUser` and `gameLstsUser` into `assetsUserOverrides` - harmless, since the empty path expands to the prefix dir which always exists, and `addonPresets` now simply fills it again.

MichalLabuda:
I've restored the `Delete` button to the save preset window. It is now disabled when no preset is selected or user types in the edit box. No auto-reselection.

Before, when a save failed, the edit field was cleared so the user had to write the entire preset name again. Now it is preserved - the same goes for deleting a preset.

libs/s25main/ingameWindows/iwAddonPresets.h
```diff
@@ -29,15 +27,24 @@ class iwAddonPresetsBase : public IngameWindow
...
+    static constexpr unsigned contentWidth = 400;
+    static constexpr unsigned tableY = 30;
+    static constexpr unsigned tableHeight = 200;
+    static constexpr unsigned rowGap = 8;
+    static constexpr unsigned rowHeight = 22;
+    static constexpr unsigned bottomMargin = 24;
+    /// Subclasses add their own controls here; the bottom row is reserved for the buttons
+    static constexpr unsigned contentStartY = tableY + tableHeight + rowGap + rowHeight;
+    static constexpr unsigned halfWidth = (contentWidth - rowGap) / 2;
+    static constexpr unsigned rightColumnX = contentX + halfWidth + rowGap;
```
Flamefire, line 40: Can you move them to the cpp file please?
a) to not having to recompile dependents when we change the internal layout and b) to not switch between 2 files when reading the control placement.
And a comment for rowGap/rowHeight pair would be good. I think in other places we used `ctrlHeight`, so maybe that and `ctrlSpacingY` or similar.

libs/s25main/ingameWindows/iwAddonPresets.cpp
```diff
@@ -200,13 +160,60 @@ void iwAddonPresetsBase::Msg_MsgBoxResult(const unsigned msgbox_id, const Msgbox
...
+
+void iwSaveAddonPreset::Msg_EditEnter(const unsigned /*ctrl_id*/)
+{
+    DoAction();
+}
+
+void iwSaveAddonPreset::Msg_EditChange(const unsigned /*ctrl_id*/)
+{
+    auto& table = assertNonNull(GetCtrl<ctrlTable>(ID_tblPresets));
+    if(!table.GetSelection())
```
Flamefire, line 196: Is the check required? Can't we just do `assertNonNull(GetCtrl<ctrlTable>(ID_tblPresets)).SetSelection(std::nullopt);`?

libs/s25main/ingameWindows/iwAddonPresets.cpp
```diff
@@ -200,13 +160,60 @@ void iwAddonPresetsBase::Msg_MsgBoxResult(const unsigned msgbox_id, const Msgbox
...
+}
+
+void iwSaveAddonPreset::Msg_TableSelectItem(const unsigned ctrl_id, const std::optional<unsigned>& selection)
+{
+    iwAddonPresetsBase::Msg_TableSelectItem(ctrl_id, selection);
+    if(deselectingFromEdit_)
+        return;
+    const auto& table = assertNonNull(GetCtrl<ctrlTable>(ID_tblPresets));
+    auto& edit = assertNonNull(GetCtrl<ctrlEdit>(ID_edtName));
+    // Suppress the change event so Msg_EditChange won't deselect the row just selected
```
Flamefire, line 212: Or maybe reuse `deselectingFromEdit_`? Just handle it fully in `EditChange`: Ignore the call if `deselectingFromEdit_` is True, else set it and change the table selection. Then you don't need it here at all.
Or actually: Isn't the `if(!table.GetSelection()) return` enough already? I.e. you don't need `deselectingFromEdit_` and `setNotify` at all

libs/s25main/ingameWindows/iwAddonPresets.cpp
```diff
@@ -247,41 +254,51 @@ void iwSaveAddonPreset::SaveToPath(const bfs::path& filePath)
...
 }
 
-// iwLoadAddonPreset
 iwLoadAddonPreset::iwLoadAddonPreset(std::function<void(const std::map<unsigned, unsigned>&)> onLoad)
-    : iwAddonPresetsBase(_("Load Addon Preset"), _("Load")), onLoad_(std::move(onLoad))
-{}
+    : iwAddonPresetsBase(_("Load Addon Preset"), _("Load"), contentStartY + rowHeight + bottomMargin),
+      onLoad_(std::move(onLoad))
+{
+    GetCtrl<ctrlButton>(ID_btAction)->SetEnabled(false);
```
Flamefire, line 290: I guess this could be done in the base class: You can not load if nothing is selected and not save if nothing is selected/entered (so I guess edtChange should also enable/disable this button)

libs/s25main/ingameWindows/iwAddonPresets.h
```diff
@@ -29,15 +27,24 @@ class iwAddonPresetsBase : public IngameWindow
...
+
+    iwAddonPresetsBase(const std::string& title, const std::string& actionLabel, unsigned height);
```
Flamefire, line 42: Maybe `additionalHeight` You have a common `contentStartY + rowHeight + bottomMargin` part you always need because you add 1 button in the base class

tests/s25Main/UI/testWindows.cpp
```diff
@@ -200,44 +232,108 @@ BOOST_FIXTURE_TEST_CASE(AddonPresetExtensionInNameIsDistinct, AddonPresetFixture
...
+    base.Msg_ButtonClick(iwAddonPresetsBase::ID_btAction);
+    base.Msg_MsgBoxResult(iwSaveAddonPreset::ID_mbOverwrite, MsgboxResult::Yes);
+
+    const auto* msgbox = topMsgbox();
+    BOOST_TEST_REQUIRE(msgbox);
+    BOOST_TEST(msgbox->GetTitle() == _("Save Failed"));
+    BOOST_TEST(!wnd.ShouldBeClosed());
+    BOOST_TEST(base.GetCtrls<ctrlEdit>().at(0)->GetText() == "blocked");
+    closeTopMsgbox();
+    closeTopMsgbox();
```
Flamefire, line 296: I'd say it is worth adding this as a method to the fixture: Delivering  a result and closing the window. Makes it clearer what you do and why. Here I was wondering which one you close.
And maybe close the 2nd one using `msgbox`?

tests/s25Main/UI/testWindows.cpp
```diff
@@ -200,44 +232,108 @@ BOOST_FIXTURE_TEST_CASE(AddonPresetExtensionInNameIsDistinct, AddonPresetFixture
...
+
+    const std::map<unsigned, unsigned> states{{1, 2}};
+    iwSaveAddonPreset wnd(states);
+    Window& base = wnd;
+    base.GetCtrls<ctrlEdit>().at(0)->SetText("blocked");
+    base.Msg_ButtonClick(iwAddonPresetsBase::ID_btAction);
+    base.Msg_MsgBoxResult(iwSaveAddonPreset::ID_mbOverwrite, MsgboxResult::Yes);
+
+    const auto* msgbox = topMsgbox();
+    BOOST_TEST_REQUIRE(msgbox);
```
Flamefire, line 291:
```suggestion
    const auto& msgbox = ensureNonNull(topMsgbox());
```

Similar below for e.g. the button. I.e. check pointers always using this helper

Flamefire (changes requested):

